PR #4, opened 2016-02-03
finish part 1 of expanded mode

robyww:


src/firefly/js/visualize/iv/ImageViewerLayout.jsx
```diff
@@ -0,0 +1,245 @@
+/*
```
xiuqin: A new file?

PR #12, opened 2016-02-12
DM-4580 Charts area toolbar

tgoldina:
Reorganized charts area to be able to switch between histogram and scatter plot views of the table
Added toolbar with selection buttons (zoom, selectrows, filter), only zoom is implemented at the moment.
Option default values now come from the store (if previous values for this form were saved)


robyww:
It looks good.  I do think you should start exporting you dispatchers directly.  They have long enough names anyway.

PR #17, opened 2016-02-19
Dm 4468 mouse readout

lznakano:


src/firefly/js/visualize/ImagePlotCntlr.js
```diff
@@ -390,12 +409,30 @@ export function dispatchChangeExpandedMode(expandedMode) {
...
+
+export function dispatchChangeMouseReadoutReadout1( newRadioValue) {
+
+    flux.process({type: CHANGE_MOUSE_READOUT_READOUT1, payload: {newRadioValue }});
+}
+
+
+export function dispatchChangeMouseReadoutReadout2(newRadioValue) {
+    flux.process({ type: CHANGE_MOUSE_READOUT_READOUT2, payload: {newRadioValue} });
+
```
robyww: Why don't you make these two actions into 1.  Just make a parameter determine which readout it is.

src/firefly/js/visualize/ui/MouseReadout.jsx
```diff
@@ -30,75 +33,249 @@ var rS= {
...
 
-function showReadout(plot, mouseState, coordinate){
-	if (!plot) return'';
-	if (isBlankImage(plot)) return'';
+MouseReadout.propTypes= {
+	visRoot:React.PropTypes.object.isRequired,
+	plotView: React.PropTypes.object,
+	mouseState:React.PropTypes.object.isRequired,
+};
+
```
robyww: You can import PropTypes directly so you don't have to use React.PropTypes.
`import React, {PropTypes} from 'react;`

src/firefly/js/visualize/ui/MouseReadout.jsx
```diff
@@ -30,75 +33,249 @@ var rS= {
...
+			break;
+		case 'pixelSize':
+			gLabel='Pixel Size:';
+			break;
+		case 'sPixelSize':
+			gLabel='Screen Pixel Size:';
+			break;
+	}
+    return gLabel;
+}
```
robyww: Why don't you make an const object that has this mapping.  I think it would be cleaner than a switch.

src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
```diff
@@ -46,27 +70,89 @@ function getDialogBuilder(fieldKey) {
...
+		case 'sPixelSize':
+			coordinate = CoordinateSys.SCREEN_PIXEL;
+			break;
+		default:
+			coordinate=CoordinateSys.UNDEFINED;
+			break;
+
+
+
+	}
```
robyww: You could make this an object as well.  Shorter and cleaner.

src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
```diff
@@ -101,10 +185,18 @@ class MouseReadoutOptionDialog extends React.Component {
...
+					radioValue={this.props.radioValue}
+			/>;
 		}
 		else {
-			form= <CoordinateOptionDialogForm  groupKey={this.props.groupKey} fieldKey={this.props.fieldKey}/>;
+			form= <CoordinateOptionDialogForm
+				groupKey={this.props.groupKey}
+				fieldKey={this.props.fieldKey}
+				radioValue={this.props.radioValue}
+			/>;
```
robyww: Better to start his methods with
`const {groupKey,fieldKey,radioValue}= this.props;`

src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
robyww: I notice that is function is only use in MouseReadout, not MouseReadoutOptionPopups. You should probably move it to there so you don't have to export it.

robyww:
There are several issues with this block:
```
switch (coordinate) {                  
            case CoordinateSys.EQ_J2000:
				if (type === 'hms') {
					result = ' ' +  hmsLon + ',  ' + hmsLat;
		        }
				else {
					//convert to decimal representation
					var dmsLon = CoordUtil.convertStringToLon(hmsLon,coordinate);//wpt.getCoordSys());
					var dmsLat = CoordUtil.convertStringToLat( hmsLat,coordinate);//wpt.getCoordSys());
					var dmsLonShort =dmsLon.toString().substring(0, 10);
					var dmsLatShort = dmsLat.toString().substring(0, 10);
					result = ' ' + dmsLonShort + ',  ' +dmsLatShort;
				}
				break;
			case CoordinateSys.GALACTIC:
			case CoordinateSys.SUPERGALACTIC:

				var lonShort = lon.toString().substring(0, 10);
				var latShort = lat.toString().substring(0, 10);
				result= ' ' + lonShort + ',  '+latShort;
				break;
			case CoordinateSys.EQ_B1950:
				result = ' ' +  hmsLon + ',  ' + hmsLat;
				break;

			case CoordinateSys.PIXEL:
				var pt =plot.projection.getPixelScaleArcSec();
				var ptShort = pt.toString().substring(0, 5);
				result = ' '+ ptShort+'"';
				break;
			case CoordinateSys.SCREEN_PIXEL:
				var pt =plot.projection.getPixelScaleArcSec()/plot.zoomFactor;
				var ptShort = pt.toString().substring(0, 5);
				result = ' '+ ptShort+'"';
				break;

```

1.  You should use the number and format them to the correct number of decimal places. (Xiuqin says we should do 7 for lat/lon).  3 for image/screen pixel. To formate we are currently using numeral.js.
http://numeraljs.com/ You can look at: `drawingLayers/DistanceTool.js` for examples.
1. You redeclaing pt and ptShort.  In javascript `var` does not work like that it can only do it once. it is hoisted to the top.  In this case you could use `const` or `let` but  in a switch statement I don't think that would work.
1. Consider using template strings:
```
var size =plot.projection.getPixelScaleArcSec()/plot.zoomFactor;
result= `  ${numeral(size).format('0.000')}"`
```

src/firefly/js/visualize/ui/MouseReadout.jsx
```diff
@@ -18,88 +18,255 @@ import {showMouseReadoutOptionDialog} from './MouseReadoutOptionPopups.jsx';
...
+	var spt= mouseState.screenPt;
+	if (!spt) return EMPTY;//hideReadout;
+
+	var column1 = {width: 90, paddingRight: 5, textAlign:'right',textDecoration: 'underline', color: 'DarkGray', fontStyle:'italic' ,  display: 'inline-block'};
+    var column2 = {width: 60, display: 'inline-block'};
+	var column3 = {width: 75, paddingRight: 5, textAlign:'right',textDecoration: 'underline', color: 'DarkGray', fontStyle:'italic' ,  display: 'inline-block'};
+	var column4 = {width: 170,display: 'inline-block'};
+	var column5 = {width: 90, paddingLeft:8, display: 'inline-block'};
+	var column5_1 = {width: 90, paddingLeft:5, display: 'inline-block'};
+
```
robyww: Since all these column variables are not computed each time. you can move them outside the function and make them const.  That would be slightly more efficient.

src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
```diff
@@ -155,35 +178,73 @@ function CoordinateOptionDialogForm({groupKey,fieldKey}) {
...
+function PixelSizeOptionDialogForm( {groupKey,fieldKey, radioValue} ) {
 
 
-	var leftColumn = { display: 'inline-block', paddingLeft:125, verticalAlign:'middle', paddingBottom:15};
+	var leftColumn = { display: 'inline-block', paddingLeft:50, verticalAlign:'middle', paddingBottom:15};
 
 	var rightColumn = {display: 'inline-block',  paddingLeft:18};
 
-	var dialogStyle = { minWidth : 300, minHeight: 100 , padding:5};
+	var dialogStyle = { minWidth : 300, minHeight:54 , padding:5};
```
robyww: You also can move then out of the function and make them const.

src/firefly/js/visualize/ImagePlotCntlr.js
```diff
@@ -84,7 +84,22 @@ const CHANGE_EXPANDED_MODE= 'ImagePlotCntlr.changeExpandedMode';
...
+//} ;
+//const CHANGE_MOUSE_READOUT_READOUT2= ()=>{
+//    return {
+//        type: CoordinateSys,
+//        text: 'ImagePlotCntlr.changeMouseReadoutModeReadout2'
+//    };
+//};
+const CHANGE_MOUSE_READOUT_READOUT1='ImagePlotCntlr.changeMouseReadoutModeReadout1';
+const CHANGE_MOUSE_READOUT_READOUT2='ImagePlotCntlr.changeMouseReadoutModeReadout2';
+const CHANGE_MOUSE_READOUT_PIXEL= 'ImagePlotCntlr.changeMouseReadoutModeReadoutPixel';
```
robyww: I don't think I clearly communicated about the actions.  I saw you changed the 3 dispatcher into one.  However, I think you should make the action ID into one.  Let the parameters decide which of the three are changed. In the dispatcher you pass a type. Just keeping sending this type all the way through into the reducer.  Then let the reducer do the correct update.

We can talk about this some more if you want.

PR #20, opened 2016-02-23
dm-4500-crop: Implemented crop, did some clean up, some refactoring

robyww:


loitly:
- maybe it's a problem with the selection tool, but i cannot re-draw at a different location after i've created a box.
- minor bug: 
    crop once, then reselect another box at the very edge of the image.  a box is not drawn, but crop tool icon appears.  clicking crop does nothing.
- center the cropped image would be a nice touch.
Everything else goods great.

robyww:
It is a selection tool problem, that I have seen.  I will go ahead and fix it as part of this ticket.

PR #24, opened 2016-02-27
DM-4574: Table (JS): text view

loitly:
Added text view.  Also, added the ability to save the table.


robyww:
The code all looks fine. First pull request that I don't have anything to say about any line of code!

I am getting an exception:

TableStore:
```
    onUnmount() {
        const {tbl_id} = this.tableModel;   //<---- here, line 26
        this.handleAction(TblCntlr.TABLE_REMOVE, {tbl_id});
    }
```
The exception happens when I show the table in the tri-view and try to switch to another view, such as side-by-side.

PR #26, opened 2016-02-29
DM 4788

lznakano:
@robyww : please review it.  Thanks!


robyww:
Small UI Stuff:

- don't clear the label, only clear the value - it looks less flashy
- don't make the label look like a link.

robyww:
I think you have the right concept. I think it is working correctly.  Most of the comments are just small stuff. The only major thing is, I don't think you need `componentDidMount` or `componentWillUnmount`.  Make sure you have eslint running in IntelliJ.

Otherwise, Good Job!

PR #30, opened 2016-03-03
DM-5163 support for expanded view.

loitly:


robyww:
Looks good. 

The objects you use with `generateLayout` seem a little complex for just declaring without some sort of factory method.  However, I don't really have a good solution on what such a method would look like or if it would help.

PR #32, opened 2016-03-05
DM-5198-statspoint: 1. add draw layer to show highlighted point from the stats popup. 2. update stats popup to work for both 3 color and no band cases.

cwang2016:


robyww:
I made a second pass today. I know I made a lot of comments, however, overall the code is very clean.  

Good job.

PR #35, opened 2016-03-15
Dm 5361 imgselpan

cwang2016:
More updates will be added after tab is made as one member in the field group.

### DIFF
--- a/src/firefly/js/visualize/ui/ImageSelectPanel.jsx
+++ b/src/firefly/js/visualize/ui/ImageSelectPanel.jsx
@@ -36,15 +36,15 @@ const leftpadding = 'leftpadding';
 const unitSign = { 'arcsec':'"', 'arcmin':'\'', 'deg':' Deg' };
 
 const keyMap = {
-    'selTab':   'SELECTIMAGEPANEL_SelectedCatalog',
+    'selTab':    'SELECTIMAGEPANEL_SelectedCatalog',
     'targettry': 'SELECTIMAGEPANEL_targettry',
     'irsatypes': 'SELECTIMAGEPANEL_IRSASelectLists_types',
-    '2masstypes': 'SELECTIMAGEPANEL_2MASSSelectLists_types',
+    '2masstypes':'SELECTIMAGEPANEL_2MASSSelectLists_types',
     'wisetypes': 'SELECTIMAGEPANEL_WISESelectLists_types',
     'wisebands': 'SELECTIMAGEPANEL_WISESelectLists_bands',
     'sizefield': 'SELECTIMAGEPANEL_ImgFeature_size',
     'unitfield': 'SELECTIMAGEPANEL_ImgFeature_unit',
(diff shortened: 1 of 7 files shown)


robyww:
UI Notes:

- Take out 3 color checkbox, for now, we are going to do 3 color different. It will probably be put back in a different location.  You, I, Xiuqin will have to talk about it.
- The Size field needs to change it values when you change the units.  See the GWT version.
- We will have to break the size (text input, units, text feedback) our into it own component. Other parts of the application will need it.  I forgot to tell you earlier.  I don't know it that must be done right away, but in the next 2 or 3 weeks.
- The NED or Simbad option is really part of the target panel.  It needs to be put in there.  I was not planning on you doing that.  Just leave it there for now but don't do anything with it.
- If there is an error, show a message to you user and don't hide the dialog.  You can use `PopupUtil.showInfoPopup`

src/firefly/js/visualize/ui/ImageSelectPanel.jsx
```diff
@@ -0,0 +1,533 @@
...
+                                </div>
+                             </td>
+                        </tr>
+                    </tbody>
+                </table>
+            </FieldGroup>
+            </div>
+        );
+     }
+}
```
robyww: We are trying to get away from tables for Layout.  Using table to layout a table is fine but we are trying to use divs for other layout.  Since I am out today, Loi could probably give you an overview of the best techniques we have been using. 

- display:'inline-block'
- display: 'flex'

It will be a little more work at first but cleaner in the end.

src/firefly/js/visualize/ui/ImageSelectPanelProp.js
```diff
@@ -0,0 +1,68 @@
...
+                {'item':'2', 'id': 1, 'name': 'Band 2'},
+                {'item':'3', 'id': 0, 'name': 'Band 3'},
+                {'item':'4', 'id': 1, 'name': 'Band 4'}],
+            'Default': '1',
+        },
+        'range': {'min': 36, 'max': 10800, 'unit': 'arcsec'},
+        'size': 500
+    }
+];
+
```
robyww: Nice use a breaking out the data from the UI. Very readable.

src/firefly/js/visualize/ui/ImageSelectPanelProp.js
```diff
@@ -0,0 +1,68 @@
...
+export default panelCatalogs;
```
robyww: We are trying to get away from `export default`. Just export `panelCatalogs` directly. It works better with IntelliJ.

robyww:
`ImageSelectPanel` is a big file and will probably get bigger before you are done with all three tickets.  You might want to consider breaking it up.  You have to break out the size field anyway.  However, there might  be other parts as well.

robyww:
Overall it is a good start. The code is clear and show a good understanding of the task. Good Job!

PR #38, opened 2016-03-17
DM-5552: Add renderer option to js table

loitly:
- Fixed scrollToRow bug.
- Added renderers option to table
- Created several common renderers

### DIFF
--- a/src/firefly/js/tables/ui/BasicTable.jsx
+++ b/src/firefly/js/tables/ui/BasicTable.jsx
@@ -23,8 +23,8 @@ export class BasicTable extends React.Component {
         super(props);
         this.state = {
             showMask: false,
-            widthPx: 0,
-            heightPx: 0,
+            widthPx: 75,
+            heightPx: 75,
             columnWidths: makeColWidth(props.columns, props.data, props.showUnits)
         };
 
--- a/src/firefly/js/tables/ui/BasicTable.jsx
+++ b/src/firefly/js/tables/ui/BasicTable.jsx
@@ -9,16 +9,14 @@ import Resizable from 'react-component-resizable';
 import {debounce, get, isEmpty, pick} from 'lodash';
 
 import {SelectInfo} from '../SelectInfo.js';
-import {FilterInfo, FILTER_TTIPS} from '../FilterInfo.js';
-import {InputField} from '../../ui/InputField.jsx';
-import {SORT_ASC, UNSORTED, SortInfo} from '../SortInfo';
+import {FilterInfo} from '../FilterInfo.js';
+import {SortInfo} from '../SortInfo';
 import {tableToText} from '../TableUtil.js';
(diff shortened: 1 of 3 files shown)


src/firefly/js/tables/ui/BasicTable.jsx
```diff
@@ -30,21 +28,18 @@ export class BasicTable extends React.Component {
...
         };
 
-        this.onResize = this.onResize.bind(this);
+        this.onResize =  debounce((size) => {
+                if (size) {
+                    var widthPx = size.width;
+                    var heightPx = size.height;
+                    this.setState({widthPx, heightPx});
+                }
+            }, 100);
```
ejoliet: I don't know if it matters but just noticed that you used half waiting time in the debounce function from the previous settings.

ejoliet: If i use a different URL table than the default, such as http://web.ipac.caltech.edu/staff/roby/demo/test-table4.tbl, the table appears with column header but without rows, all background white. This happen also in 'dev' currently.

src/firefly/js/tables/ui/BasicTable.jsx
```diff
@@ -92,15 +87,15 @@ export class BasicTable extends React.Component {
                     }
                 };
 
-        const colProps = pick(this.props, ['columns', 'data', 'selectable', 'selectInfoCls', 'tableStore']);
+        const colProps = pick(this.props, ['columns', 'data', 'selectable', 'selectInfoCls', 'tableStore', 'renderers']);
```
ejoliet: I didn't know about 'pick' function and i think is very nice. Actually i went to loadash docs and find out couple of others like 'omitBy' which is also interesting for oposite case when you want to pick from an object leaving out (omitting) attributes from it.

src/firefly/js/tables/ui/BasicTable.jsx
ejoliet: The console logs gives more insight of what could happen:
TableStatsCntlr.js:178 Failed to fetch table statistics: TypeError: Cannot read property 'getNativeFromReactElement' of undefined
Please, have a look.

ejoliet: Another comment i want to make is related to the design flow. When i open a different table, the tab currently shown is always the first tab. I don't know if this is the expected behavior but i was more in favor of getting automatically the tab switch to the recent table loaded. What do you think?

loitly: Yep, the debounce was on purpose.  I wanted it to be more responsive.

The small table not rendering is related to the 'scrollToRow bug' I mentioned as fixed.  For some reason, the fix was not in.  I added it now.

I agree with the selected tab comment.  The current behavior is temporary.  It will mimic that of the old application once converted.

PR #40, opened 2016-03-23
DM-5488: Field group updates

robyww:
- changed mixin to high order component: FieldGroupConnector
- simplified field group reducer
- now value change can take any object not just required fields, this was good for TargetPanel
- cleanup, added dispatchers
- support for function in value of InputField
- tabs now can be wrapped with InputField
- FieldGroup no longer export default
- hidden values in FieldGroup, Field group class, use initValue prop
- tab page now classes
- removed mixin
- added field group support unmounted fields
- made new DialogCntlr.js


src/firefly/js/core/DialogCntlr.js
```diff
@@ -0,0 +1,112 @@
...
+//============ private functions =================================
+//============ private functions =================================
+
+const showDialogChange= function(state,action) {
+    var {dialogId,ownerId}= action.payload;
+
+    if (!state[dialogId] || !state[dialogId].visible || ownerId!==state[dialogId].ownerId) {
+        state= Object.assign({},state, {[dialogId]: {visible: true, ownerId}});
+    }
+    return state
```
tgoldina: Could you, please, add semicolon here for consistency? (My ESLint is complaining.)

src/firefly/js/core/ReduxFlux.js
```diff
@@ -30,6 +30,7 @@ import DrawLayer, {DRAWING_LAYER_KEY} from '../visualize/DrawLayerCntlr.js';
 import DrawLayerFactory from '../visualize/draw/DrawLayerFactory.js';
 import DrawLayerCntlr, {makeDetachLayerActionCreator} from '../visualize/DrawLayerCntlr.js';
 import MultiViewCntlr, {IMAGE_MULTI_VIEW_KEY} from '../visualize/MultiViewCntlr.js';
+import DialogCntlr, {DIALOG_KEY} from '../core/DialogCntlr.js'
```
tgoldina: Missing semicolon. Are you using firefly/.eslintrc? (It has "semi": [2, "always"] - should we change the rule?)

tgoldina: And – since I am looking at this class – addListener for multiple types does not seem to be correct.

tgoldina:
For RadioGroupInputField, when no value is supplied, the first option value must be the group value. 
It's broken. In XYPlot options, try supplying 2 as X/Y ratio.  (I had to switch the radio button below (Fit/Fill) to Fill and back to Fit to make it work. (It's working in the previous version - check http://irsawebdev1.ipac.caltech.edu/fftools/fftools.html

src/firefly/js/core/ReduxFlux.js
robyww: I just missed the semicolon.  Remember I edited 42 files. I was cutting and pasting a lot.

robyww:
Now that you mention it. I think I saw this. I don't quite understand how the value was getting in there before.  I will have to go back and look at the original.

tgoldina:
In example dialog: 
-if the initial value is not valid the validate state is true. 
-if I click OK first time the dialog is displayed, only the fields in the selected tab will have value. If I then switch the tab, both selected tab field and previous tab fields values will be listed. Should there be only the selected tab value?
-looks like something is wrong with validators for the tab fields - I can not change the value to a valid one

src/firefly/js/ui/RadioGroupInputField.jsx
```diff
@@ -1,76 +1,40 @@
...
-            groupKey: React.PropTypes.string
-        },
+    var {value,options}= params;
+    value= convertValue(value,options);
 
-        componentWillMount() {
-            // if no default value is specified, select the first option
-            if (typeof this.state.fieldState.value === 'undefined' || this.state.fieldState.value==='') {
-                this.state.fieldState.value = this.props.options[0].value;
-            }
```
tgoldina: This part was handling the undefined value situation for RadioButtonGroup.

src/firefly/js/fieldGroup/FieldGroupUtils.js
```diff
@@ -5,84 +5,96 @@
...
+            if (typeof newValue=== 'object' &&       // check to see if return is an object with {value:string,valid:boolean}
+                Object.keys(newValue).length===2 &&
+                newValue.hasOwnProperty('valid') &&
+                newValue.hasOwnProperty('value') ) {
+                dispatchValueChange({fieldKey:key,groupKey,valid:newValue.valid,value:newValue.value});
+            }
+            else {
+                dispatchValueChange({fieldKey:key,groupKey,valid:true,value:newValue});
+            }
+        }
```
tgoldina: I would like to understand this piece of code. When would the field value be a function?

src/firefly/js/fieldGroup/FieldGroupUtils.js
```diff
@@ -5,84 +5,96 @@
...
+    });
+
+
+    //===============
+
+    fields= getGroupFields(groupKey); // need a new copy
```
tgoldina: getGroupFields(groupKey) is not creating a copy, is it?

src/firefly/js/fieldGroup/FieldGroupUtils.js
robyww: The field value would be a function on the file upload case. Therefore the upload does not activate until validation.  In the upload case the function would return a promise.  However, It could return a value or an object with a value and a valid status,

robyww: Now the value key of a field can contain a promise or function or primitive. The function can return a primitive, a promise, or an object with primitive and status.

src/firefly/js/fieldGroup/FieldGroupUtils.js
robyww: The store might have updated because of the dispatch updated.  I am not copying it. The previous fields object is out of date.

src/firefly/js/ui/ExampleDialog.jsx
```diff
@@ -179,7 +178,7 @@ class FieldGroupTest extends Component {
 
     constructor(props)  {
         super(props);
-        FieldGroupUtils.initFieldGroup('DEMO_FORM',exDialogReducer);
+        dispatchInitFieldGroup('DEMO_FORM',false,null,exDialogReducer);
```
tgoldina: This dispatch triggers warnings every time ExampleDialog comes up:
"Warning: setState(...): Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state."  

Another warning comes up as I click OK in ExampleDialog:

Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
warning @ warning.js:45
getInternalInstanceReadyForUpdate @ ReactUpdateQueue.js:34
ReactUpdateQueue.enqueueSetState @ ReactUpdateQueue.js:181
ReactComponent.setState @ ReactComponent.js:65
(anonymous function) @ ExampleDialog.jsx:193
(anonymous function) @ FieldGroupUtils.js:136
(anonymous function) @ createStore.js:132
dispatch @ createStore.js:131
(anonymous function) @ fftools.js:92535
(anonymous function) @ index.js:88
process @ ReduxFlux.js:209
dispatchHideDialog @ DialogCntlr.js:34
validUpdate @ CompleteButton.jsx:28
(anonymous function) @ CompleteButton.jsx:34
validateResolvedSingle @ FieldGroupUtils.js:17
(anonymous function) @ FieldGroupUtils.js:56

One item of the ticket was understanding what to do about these warnings. What are your current thoughts?

robyww: I studied it a lot, fixed it a little and I am very frustrated with the way react-redux works.  I think I want to take react-redux out completely.  The problem is not with the field group stuff. The problem is the react-redux and some of out code indiscriminately calling setState when its state has not changed. When the field group does its mount or init dispatch (which does not affect anything else) other parts of the application is calling setState without any compare. 

Any call to setState if it all the objects are equals make this message.  If the code did a shallow equals instead of relying on react there would not be a problem.

Any work around I tried to do the with field group stuff has not worked, though I am still experimenting.

robyww: If fftools.js lines 102-158 you can see my experimenting with taking out the connector. It works fine and does eliminate one of the warning messages.

tgoldina:
I have another issue about the initial value. (It is still present in this version.)

I have two input fields - one is mounted, second is mounted on a radio button switch (numBins and falsePositiveRate in Histogram Options: one or the other shows up when user switches between Byesian Blocks and Fixed size). Both fields define the default value the same way in initial state:

                        value: FieldGroupUtils.getFldValue(fields, 'falsePositiveRate', 0.05),
or
                        value: FieldGroupUtils.getFldValue(fields, 'numBins', 10),

The initial value of the originally unmounted field is not set. How can I get the default falsePositiveRate to be 0.05? (It's HistogramOptions.jsx - am I doing something wrong?)

tgoldina:
I am glad you are doing all this work. This ticket is high impact, because it affects all the code, which handles input. Would it be possible to highlight some of the changes (and reasons behind them) at the meeting today? I think the whole group would benefit.

robyww:
I think we should get it finished a stabilized first and I would like to wait until we are all back. (Cindy is out this week). The ticket is high impact but the changes are all pretty subtle.

robyww:
About your concern about unmounted fields showing in validation.  That is because the CompleteButton now has an option 'includeUnmounted'.  I set it to true for testing.  If it is false the it behaves as you would expect.

robyww:
>> I have another issue about the initial value. (It is still present in this version.)

I think there were a couple of problems.  I just cleaned up the component mount reducer more.  I think I finally have it right.  Even then, FieldGroupUtils.getFldValue did not return the default value when the fields object was defined  but the field.value was not.  It now returns the default in both cases.

PR #41, opened 2016-03-23
DM-4595: first pass at history and routing.

loitly:
an early implementation of history and routing.  it's based on the old framework with placeholder for additional features.

### DIFF
--- a/src/firefly/js/core/History.js
+++ b/src/firefly/js/core/History.js
@@ -2,48 +2,79 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
+/**
+ * url pattern
+ * /{module_name}/{launch_page};{path_params}?{query_params}
+ *
+ *  module_name: synonymous to context name, or application deployed path. i.e.  fftools, application/wise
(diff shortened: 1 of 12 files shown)


src/firefly/js/core/History.js
```diff
@@ -2,48 +2,79 @@
...
 
-const history = {
-    HISTORY_PATH,
-    getList,
-    getCurrent,
-    add
-};
+function genericHandler(baseUrl='') {
+    return {
+        actionToUrl: (action) => {
```
tgoldina: You might want to add a comment here, just in case someone like me is unaware of URL path parameter syntax.
(http://doriantaylor.com/policy/http-url-path-parameter-syntax)

tgoldina:
I see the URL changes as I move between dropdown and search, and I can use back and forward buttons. Looks like a good start.

If I paste a URL after table search to another tab, the URL becomes:
http://localhost:8080/fftools/fftools.html;a=table_space.tableSearch?request=[object%20Object]&resultId=tbl_ui_gid-5&tbl_ui_id=tbl_ui_id-8

There are other things like tab switch/remove, getting related plots, etc. I guess those issues will be addressed later.

PR #42, opened 2016-03-25
Dm 4581 densityplot

tgoldina:
First cut of density plot and large data set support. 
- Support for large data sets:
  -the data are decimated on server (currently 10000 points, 1x1 aspect ratio for all plots)
  -highlight  is passed with point values
- Performance improvements on highlight, zoom, resize, etc. 
- Density plot support
  -2d histogram display with weights
  -scalable symbol (supposed to change in resize)
  
-It is still SVG (the number of points is limited to 1000). Highcharts has an experimental boost.js module, which automatically switches to canvas for large number of points. This needs to be explored.
-Handling zoom is in a separate ticket.
-More work needs to be done on aspect ratio choice, using canvas, better look, separating density plot from scatter plot, handling density plot options, etc.


src/firefly/java/edu/caltech/ipac/firefly/server/util/QueryUtil.java
```diff
@@ -534,10 +534,10 @@ public static DataAccessException createEndUserException(String msg, String deta
     }
 
 
-    private static final int DECI_DEF_MAX_POINTS = AppProperties.getIntProperty("decimation.def.max.points", 100000);
+    private static final int DECI_DEF_MAX_POINTS = AppProperties.getIntProperty("decimation.def.max.points", 10000);
```
cwang2016: As I see the number is modified, would that be better to use a variable to represent the number instead of hardcode?

src/firefly/java/edu/caltech/ipac/firefly/server/util/QueryUtil.java
```diff
@@ -559,20 +559,25 @@ public static DataGroup doDecimation(DataGroup dg, DecimateInfo decimateInfo) {
 
         boolean doDecimation = dg.size() >= DECI_ENABLE_SIZE;
 
-        DataType[] columns = new DataType[doDecimation ? 6 : 3];
+        DataType[] columns = new DataType[doDecimation ? 5 : 3];
```
cwang2016: same as previous comment, use a variable instead of hardcode number?

src/firefly/js/tables/DecimateInfo.js
```diff
@@ -0,0 +1,49 @@
...
+            dinfo = new DecimateInfo(...kv[1].split(','));
+            return dinfo.isValid() ? dinfo : null;
+        }
+    }
+
+    serialize() {
+        const {xColumnName, yColumnName, maxPoints, xyRatio, xMin, xMax, yMin, yMax} = this;
+        const orderedNumericProps = [maxPoints, xyRatio, xMin, xMax, yMin, yMax];
+        const deciStrStart = DECIMATE_TAG + '=' + xColumnName + ',' + yColumnName;
+        const deciStr = orderedNumericProps.reduce((sres, val)=>{
```
cwang2016: in ES6, deciStrStart = `${DECIMATE_TAG}=${xColumnName},${yColumnName}`

src/firefly/js/tables/DecimateInfo.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    static parse(str) {
+        if (!str) return;
+
+        const kv = str.split('=', 2);
+        let dinfo = null;
+        if (kv && kv.length === 2 && kv[0].equals(DECIMATE_TAG)) {
+            dinfo = new DecimateInfo(...kv[1].split(','));
+            return dinfo.isValid() ? dinfo : null;
```
cwang2016: Does dinfo.inValid() call the function starting at line 25?  there is no argument passed in isValid(),  that means isValid returning 'false'?

src/firefly/js/tables/DecimateInfo.js
cwang2016: deciStrStart = `${DECIMATE_TAG}=${xColumnName},${yColumnName}`

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -357,13 +362,13 @@ var ChartsPanel = React.createClass({
...
+                         title='Select enclosed points'
                          src={SELECT_ROWS}
                          onClick={() => this.addSelection()}
-                    />
+                    />}
                     <img style={selectionBtnStyle}
-                         title='Filter in the selected points - ToDo'
+                         title='Filter in the selected points'
                          src={FILTER_IN}
                          onClick={() => this.addFilter()}
```
cwang2016: how about change the cursor while hovering the img in order to give the user a hint to click the img?

src/firefly/js/visualize/XYPlot.jsx
```diff
@@ -34,288 +35,610 @@ const plotParamsShape = PropTypes.shape({
...
+    if (weight == 1) return 1; // 1pt;
+    else {
+        if (logShading) {
+            //use log scale for shade assignment
+            let max; //min=2, max;
+            const base = Math.pow(maxWeight+0.5, 0.2);
+            for (var e = 1; e <=5; e++) {
+                max = Math.round(Math.pow(base, e));
+                if (weight <= max) {
+                    if (max > maxWeight) { max = maxWeight; }
```
cwang2016: I may be wrong. Just wondering what  this assignment (max = maxweight) would impact?

src/firefly/js/visualize/XYPlotCntlr.js
```diff
@@ -227,24 +236,47 @@ function updatePlotData(data) {
...
 function fetchPlotData(dispatch, activeTableServerRequest, xyPlotParams) {
 
+    if (!xyPlotParams) { return; }
+
+    const decimateInfoCls = new DecimateInfo(xyPlotParams.x.columnOrExpr, xyPlotParams.y.columnOrExpr, 10000);
+
     // todo support expressions
     const req = Object.assign({}, activeTableServerRequest, {'startIdx' : 0, 'pageSize' : 1000000,
-        'inclCols' : xyPlotParams.x.columnOrExpr+','+xyPlotParams.y.columnOrExpr
+        'inclCols' : xyPlotParams.x.columnOrExpr+','+xyPlotParams.y.columnOrExpr,
```
cwang2016: some minor thing,
'inclCols' : `${xyPlotParams.x.columnOrExpr},${xyPlotParams.y.columnOrExpr}`

src/firefly/java/edu/caltech/ipac/firefly/server/util/QueryUtil.java
robyww: I think this is the way we want it.  The number is a fallback to the property.  Setting the property will override it.

robyww:
I noticed some changes on the server side.  They are all backward compatible, right?

src/firefly/js/tables/DecimateInfo.js
```diff
@@ -0,0 +1,49 @@
...
+    serialize() {
+        const {xColumnName, yColumnName, maxPoints, xyRatio, xMin, xMax, yMin, yMax} = this;
+        const orderedNumericProps = [maxPoints, xyRatio, xMin, xMax, yMin, yMax];
+        const deciStrStart = DECIMATE_TAG + '=' + xColumnName + ',' + yColumnName;
+        const deciStr = orderedNumericProps.reduce((sres, val)=>{
+            return sres + ',' + (Number.isFinite(val) ? val : '');
+        }, deciStrStart);
+        return deciStr;
+    }
+}
```
robyww: I question if this needs to be a class. Why not have a object literal and a couple of functions . `parse` is static so it does not need to be in the class so to be for is valid and `serialize`. I have been trying to get a way from `class` unless I absolutely and mostly work with pure functions.

src/firefly/js/tables/TableUtil.js
```diff
@@ -142,6 +142,18 @@ export function isFullyLoaded(tbl_id) {
...
 
+export function getCellValue(tableModel, rowIdx, colName) {
+    if (tableModel.tableData && tableModel.tableData.data) {
+        const cols = tableModel.tableData.columns;
+        const colIdx = cols.findIndex((col) => {
+            return col.name === colName;
+        });
+        // might be undefined if row is not loaded
+        return get(tableModel.tableData.data, [rowIdx, colIdx]);
+    }
```
robyww: good use of `get`

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -52,11 +56,11 @@ var ChartsPanel = React.createClass({
...
     getInitialState() {
         return {
-            chartType : SCATTER,
-            optionsShown : true,
+            chartType: SCATTER,
+            optionsShown: true,
             widthPx: 700,
             heightPx: 300,
-            debouncedResize: debounce(this.onResize, 500, {'leading':false, 'trailing':true})};
+            debouncedResize: debounce(this.onResize, 200, {'leading':false, 'trailing':true, 'maxWait':10000})};
```
robyww: `'leading':false, 'trailing':true` are the defaults so I don't think you need them

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -79,13 +83,16 @@ var ChartsPanel = React.createClass({
 
     onResize(size) {
         if (size) {
-            this.setState({ widthPx: (size.width-10), heightPx: (size.height-30), debouncedResize: this.state.debouncedResize });
+            this.state.debouncedResize.cancel;
+            this.setState({
+                widthPx: (size.width - 10),
+                heightPx: (size.height - 30),
+                debouncedResize: debounce(this.onResize, 200, {'leading':false, 'trailing':true, 'maxWait':10000})
+            });
```
robyww: same think with `leading` and `trailing`

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -79,13 +83,16 @@ var ChartsPanel = React.createClass({
 
     onResize(size) {
         if (size) {
-            this.setState({ widthPx: (size.width-10), heightPx: (size.height-30), debouncedResize: this.state.debouncedResize });
+            this.state.debouncedResize.cancel;
```
robyww: i think you mean:
```
this.state.debouncedResize.cancel();
```

src/firefly/js/visualize/ChartsTableViewPanel.jsx
robyww: Also- after thinking about it I don't think you need a max wait.  The use case would be someone holding the mouse a resizing the component for 10 seconds with out letting up.  It that is what they want to do then just let me.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
robyww: This is a more serious concern: Why do you have `debounceResize` in the state? It could be a method on the class object. Also, you are creating a new function with each setState which means that if `widthPx` and `heightPx` are the same the component will still re-render.

robyww:
This might be outside the scope of your most recent changes but I think it is worth looking at.  `ChartsPanel` in `ChartsTableViewPanel.jsx` is a very large class. It calls a lot of methods starting at line 231 that could be pure functions that just two or three parameters. I am beginning to agree with the functional people that claim it is just easier to follow code when the parameters are passed. We want to minimize to use of `this`.
For Example - `addFilter` could become 
```
function addFilter(chartType, tblPlotData, tableModel) {...}
```
or `addZoom`
```
function addZoom(chartType, tblId, tblPlotData) {...}
```
I also think that it could be broken up into some smaller react components such as the toolbar and the options

src/firefly/js/visualize/XYPlot.jsx
```diff
@@ -34,288 +35,610 @@ const plotParamsShape = PropTypes.shape({
...
+
+    onSelectionEvent(event) {
+        const xAxis = event.xAxis[0];
+        const yAxis = event.yAxis[0];
+
+        this.props.onSelection({xMin: xAxis.min, xMax: xAxis.max, yMin: yAxis.min, yMax: yAxis.max});
+    }
+
+    makeSeries(chart) {
+        //const chart = this.refs.chart && this.refs.chart.getChart();
```
robyww: I think this function could be broken out into a pure function and maybe even put into another file (`XYPlotHelper.js`???).  It is over 100 lines and is key to the charts.

src/firefly/js/visualize/XYPlot.jsx
```diff
@@ -34,288 +35,610 @@ const plotParamsShape = PropTypes.shape({
...
+                marker = {symbol: 'rectangle', radius: xUnitPx/2, hD: (xUnitPx-yUnitPx)/2};
+
+                const {xMin, xUnit, yMin, yUnit} = parseDecimateKey(decimateKey);
+                const getCenter = (xval,yval) => {
+                    return {
+                        // bitwise operators convert operands to 32-bit integer
+                        // hence they can be used as a fast way to truncate a float to an integer
+                        x: xMin+(~~((xval-xMin)/xUnit)+0.5)*xUnit,
+                        y: yMin+(~~((yval-yMin)/yUnit)+0.5)*yUnit
+                    };
```
robyww: I am glad you documented this! I have seen it before but would not have remembered. If you are worried about the performance it might be better to use a for loop instead of reduce on line 430. I would normally say reduce is better but I have heard in performance critical javascript that a for loop is faster.

src/firefly/js/tables/DecimateInfo.js
tgoldina: Good catch. I forgot to test the converted code and since we don't use parse on the client, I got away with it.

src/firefly/js/tables/DecimateInfo.js
tgoldina: Got it. Looks like Github is eating backquotes.

src/firefly/js/tables/DecimateInfo.js
tgoldina: You are right, we probably do not need a class. Moreover, client code will never need to parse back decimate info (only to serialize), because it has more informative decimation key in the result.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
tgoldina: I think you are right. I was trying to figure out why I get at least two resizes. (I wanted to have just one, because it is expensive.) One of the guesses was that the default maxWait is too short. I think it's canceling the previous debounce that fixed the problem.

src/firefly/js/visualize/XYPlot.jsx
tgoldina: You are right. Both min and max can be removed here. They were used to calculate the label in the original version.

PR #43, opened 2016-03-25
DM-5584: create reusable file_upload component

loitly:
A FileUpload field is added to the SearchPanel.
To test, select a file to upload.  Click search, then monitor this directory for uploaded file:
/hydra/workarea/fftools/temp_files/
File should start with 'upload_'.

### DIFF
--- a/src/firefly/js/core/History.js
+++ b/src/firefly/js/core/History.js
@@ -17,7 +17,7 @@ import {get, pick} from 'lodash';
 
 import {flux} from '../Firefly.js';
 import {TABLE_SEARCH} from '../tables/TablesCntlr.js';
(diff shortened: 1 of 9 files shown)


robyww:
The upload component seems to give some sort of security exception when I upload a file and then return to the search panel. I think it is trying to set the value of the file.

DOMPropertyOperations.js:168 Uncaught InvalidStateError: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.

robyww:
There is no feedback when uploading.  This will have to be done, but probably not in this ticket.  I am going to make a second ticket.

PR #44, opened 2016-03-29
DM-4484: Catalogs Overlay

robyww:
Catalogs Overlay

Also the following:
  - Introduction of Sagas
  - fixed TargetPanel bug
  - added forceReinit property to FieldGroup components
  - added z-index for dialogs


src/firefly/js/visualize/saga/CatalogWatcher.js
```diff
@@ -0,0 +1,151 @@
...
+        lonCol: s[0],
+        latCol: s[1],
+        csys : CoordinateSys.parse(s[2])
+    };
+    
+    if (!tableData.columns.find( isCName(columns.lonCol)) && !tableData.columns.find(isCName(columns.latCol))) {
+        return;
+    }
+
+    const params= {
```
tgoldina: Here you might want to specify "decimate" parameter to handle large catalogs:
decimate: serializeDecimateInfo(columns.lonCol, columns.lonCol.latCol, numPoints [,limits]) 

The helper function is in Decimate.js

tgoldina:
OK, I've got basic idea of how you use sagas. Though I should say generators are mind twisting. (Hopefully, we'll get used to them.)

One thing I noticed: When I bring another image and another catalog, the highlighted row is not reflected in the catalog, even though clicking on a point seems to change it.

robyww:
>> OK, I've got basic idea of how you use sagas. Though I should say generators are mind twisting. 

Once you do get your head around generators these saga really simplify the flow. Everything become really straight forward. I am really excited about them.

robyww:
I am already working on part two of the ticket (DM-4503). I will fix the bug and add the decimate in that one.

src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -0,0 +1,182 @@
...
+            return isEmpty(lastDataRet) ? computeDrawLayer(tableData, columns) : lastDataRet;
+            break;
+        case DataTypes.HIGHLIGHT_DATA:
+            return isEmpty(lastDataRet) ? 
+                          computeHighlightLayer(tableData, drawLayer.highlightedRow, columns) : lastDataRet;
+            break;
+        case DataTypes.SELECTED_IDX_ARY:
+            return null;
+            break;
+    }
```
lznakano: I think the "break" is not needed here.

robyww: your right, in this case none of them are needed.

ejoliet: Do we want to return null or throw an exception so it can be re-thrown and trace upfront? Or do we handle the 'null' on the component view after dispatch?

src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -0,0 +1,182 @@
...
+        hasPerPlotData:false,
+        isPointData:true,
+        canUserDelete: true,
+        canUseMouse:true,
+        canHighlight: true,
+        canSelect: true,
+        canFilter: true,
+        helpLine : helpText,
+        canUserChangeColor: ColorChangeType.STATIC
+    };
```
ejoliet: One improvement we should add here (or to a different ticket) is that the user should be able to change the shape and size of the overlay symbol

src/firefly/js/visualize/saga/CatalogWatcher.js
```diff
@@ -0,0 +1,151 @@
...
+        startIdx : 0,
+        pageSize : 1000000,
+        inclCols : `${columns.lonCol},${columns.latCol}`
+    };
```
ejoliet: That's really cool, i didn't know i could use templates inside js object directly. I mean, it make sense but it would have never occurred to me to use it that way. Cool!

src/firefly/js/drawingLayers/Catalog.js
robyww: That would be pretty easy to do. More work in the UI than anything else.

src/firefly/js/drawingLayers/Catalog.js
robyww: The way the `getDrawData` is made it supports nulls.  However, the next ticket (the one I am working on now) we will support selected.

PR #45, opened 2016-03-30
DM-4794

wmiipac:
Zoom Option Popup

- the zoom option panel stays on while user could either use zoom button or select from level list

- reached max zoom level info popup


robyww:
Wei and I have already gone of this branch some already.  It looks good, since she is on vacation I am not going to list any picky issues.  There is some unused code that I will remove and push. The I will merge.

I like the way this popup works better than the gwt one.

PR #47, opened 2016-03-31
DM-5362-imgalltabs, add all tabs into the select image panel, create reusable radius input field, integrate field group and upload file implementation.

cwang2016:
- This version replaces the image on the selected plot, if no plot is selected, then add a new plot into the available viewer.

### DIFF
--- a/src/firefly/js/ui/RadiusInputFields.jsx
+++ b/src/firefly/js/ui/RadiusInputFields.jsx
@@ -1,0 +1,205 @@
+import React, {Component, PropTypes}  from 'react';
+import {get} from 'lodash';
+
+import {fieldGroupConnector} from './FieldGroupConnector.jsx';
+import {convertAngle} from '../visualize/VisUtil.js';
+import {InputFieldView} from './InputFieldView.jsx';
+import {ListBoxInputFieldView} from './ListBoxInputField.jsx';
+import Validate from '../util/Validate.js';
+
+
+const invalidSizeMsg = 'size is out of range';
+const DECDIGIT = 6;
+const unitSign = { 'arcsec':'"', 'arcmin':'\'', 'deg':' Deg' };
+
+/*
+ * remove trailing zero from toFixed result
+ */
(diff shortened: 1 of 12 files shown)


src/firefly/js/ui/RadiusInputFields.jsx
```diff
@@ -0,0 +1,205 @@
...
+        var valid = isSizeValid(props.value, props.min, props.max);
+        var value = (valid&&props.value) ? props.value  : '';
+        var displayValue = props.displayValue ? props.displayValue : sizeFromDeg(value, unit);
+
+        this.state = {
+            value,
+            unit,
+            displayValue,
+            valid
+        };
```
robyww: This is not the right way to update state in react.  I am surprised it even works.  in the constructor `this.state= {...}` is correct.  However anywhere else you should call `this.setState({...})`

src/firefly/js/ui/RadiusInputFields.jsx
```diff
@@ -0,0 +1,205 @@
...
+            this.props.onChange(Object.assign({}, ev, {sizeInfo: stateUpdate}));
+        }
+    }
+
+    render() {
+        var minInUnit = sizeFromDeg(this.props.min, this.state.unit);
+        var maxInUnit = sizeFromDeg(this.props.max, this.state.unit);
+        var sign = unitSign[this.state.unit];
+        var message = `Valid range between: ${minInUnit}${sign} and ${maxInUnit}${sign}`;
+        var {displayValue, valid, unit} = this.state;
```
robyww: _Just being picky_ - you could move this line up and not have a `this.state.unit` in lines 147,148, and 149

robyww:
As part of this ticket, I would like you to go ahead and put the image select panel into the Images drop down and not just a dialog.  This is so we can get the tri-view working later today.

robyww:
When you are calling `makeWiseRequest` sometime a `undefined` is being pass to the `survey` parameter,

PR #48, opened 2016-03-31
DM-4584 Zoom support for density plot

tgoldina:
When zooming decimated data, a request should go to the server to fetch zoomed data. If the result is not decimated, further zoom does not require server call, but restoring the original does. Hence we need to remember if the unzoomed data were decimated or not.

### DIFF
--- a/src/firefly/js/visualize/XYPlotCntlr.js
+++ b/src/firefly/js/visualize/XYPlotCntlr.js
@@ -1,7 +1,7 @@
 import {flux} from '../Firefly.js';
 
 import update from 'react-addons-update';
-import {has, omitBy, isUndefined, isString} from 'lodash';
+import {debounce, get, has, omitBy, isUndefined, isString} from 'lodash';
 
 
 import {doFetchTable, isTableLoaded, findTblById} from '../tables/TableUtil.js';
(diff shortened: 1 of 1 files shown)


src/firefly/js/visualize/XYPlotCntlr.js
```diff
@@ -283,6 +320,7 @@ function fetchPlotData(dispatch, activeTableServerRequest, xyPlotParams) {
             console.error(`Failed to fetch XY plot data: ${reason}`);
         }
     );
+
 }
```
lznakano: 1. getInitState() is not needed since it returns {}.  export function reducer(state={}, action={}) should be fine.
2. console.error should change to logError instead.
3. Is any reason that some places the function expressions are used and other places function declarations are used?

robyww: The reason for logError is that is shows stack traces by default and give us a place to upgrade our error logging when we want.  It is a very simple function but all error logging should go through it.

tgoldina: Thank you for the comments. I followed your suggestions.

PR #49, opened 2016-03-31
DM-4503: ImageViewer: Select, Unselect, Filter

robyww:
Includes:
 - Select, Unselect, Filter
 - Drawer support the selected layer
 - Handling decimation
 - VisUtil function to get all points in a selected area
 - Highlighted point fix

### DIFF
--- a/src/firefly/js/drawingLayers/Catalog.js
+++ b/src/firefly/js/drawingLayers/Catalog.js
@@ -3,8 +3,8 @@
  */
 
 
-import {isEmpty} from 'lodash';
-import {primePlot} from '../visualize/PlotViewUtil.js';
+import {isEmpty,get} from 'lodash';
+import {primePlot,getAllDrawLayersForPlot} from '../visualize/PlotViewUtil.js';
 import {visRoot} from '../visualize/ImagePlotCntlr.js';
 import PointDataObj, {DrawSymbol} from '../visualize/draw/PointDataObj.js';
 import {makeDrawingDef, COLOR_PT_1, COLOR_PT_2, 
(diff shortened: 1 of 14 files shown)


src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -142,11 +153,29 @@ function getLayerChanges(drawLayer, action) {
...
+    const{tableData, columns}= drawLayer;
+    switch (dataType) {
+        case DataTypes.DATA:
+            return isEmpty(lastDataRet) ? computeDrawLayer(tableData, columns) : lastDataRet;
+        case DataTypes.HIGHLIGHT_DATA:
+            return isEmpty(lastDataRet) ? 
+                          computeHighlightLayer(drawLayer, columns) : lastDataRet;
+        case DataTypes.SELECTED_IDXS:
+            return isEmpty(lastDataRet) ?
+                computeSelectedIdxAry(drawLayer.selectInfo) : lastDataRet;
```
tgoldina: Looks like computeSelectedIdxAry returns a function, while other layers return arrays.

src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -162,16 +191,16 @@ function computeDrawLayer(tableData, columns) {
     return ptAry;
```
tgoldina: Very minor: you assign a value to variable just to return. If you want to document the code this way, it would be better to use JSDoc.

src/firefly/js/drawingLayers/Catalog.js
robyww: yes, it can now return a function or an array of indexes.  I think if make the whole think more optimal.  It is documented in DrawLayer.js.

src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -180,3 +209,112 @@ function computeHighlightLayer(tableData, highlightedRow, columns) {
...
+
+export function selectCatalog(pv,dlAry) {
+    var p= primePlot(pv);
+    var sel= p.attributes[PlotAttribute.SELECTION];
+    if (!sel) return;
+    const catDlAry= getLayers(pv,dlAry);
+    if (catDlAry.length) {
+        const tooBig= catDlAry.some( (dl) => dl.dataTooBigForSelection);
+        if (tooBig) {
+            showInfoPopup('Your data set is too large to select. You must filter it down first.',
```
tgoldina: An optimization would be not to show select option when the data are too big.

robyww: I thought about that but I am afraid it would be more confusing.  The user might wonder why the select does not show up sometimes.

src/firefly/js/drawingLayers/Catalog.js
```diff
@@ -142,11 +153,29 @@ function getLayerChanges(drawLayer, action) {
...
 
-const findColIdx= (columns,colId) => columns.findIndex( (c) => c.name===colId);
+
+function getDrawData(dataType, plotId, drawLayer, action, lastDataRet) {
+
+    const{tableData, columns}= drawLayer;
+    switch (dataType) {
+        case DataTypes.DATA:
+            return isEmpty(lastDataRet) ? computeDrawLayer(tableData, columns) : lastDataRet;
+        case DataTypes.HIGHLIGHT_DATA:
```
tgoldina: If I have two catalogs, and I click on a point, the nearest points on both catalogs are "highlighted". I think there should be only one highlighted point, which corresponds to the active table.

robyww: What happens if you want both tables to go to that spot? I am going to go ahead and keep it as it is for this ticket and we can talk about the right way to do it.  I have mixed feelings so I made it affect all the visible layers for all the functions: highlight, select, filter.  Table that are uncheck are not affected.

PR #50, opened 2016-04-05
Dm-4494 recenter & DM-3022

robyww:
I combined two tickets into one pull request (probably a mistake). I keep in as two commit, one for each tickets.

- DM-4494: restore defaults button and recenter button, both on toolbar
- DM-3022: color dialog.  This has been half done for months.  I finished it.


src/firefly/js/visualize/ui/ColorDialog.jsx
```diff
@@ -2,248 +2,226 @@
...
+            return renderStandardView(plot,fields);
+        }
+    }
+}
+
+function renderThreeColorView(plot,rFields,gFields,bFields) {
+    const {plotState}= plot;
+    return (
+        <div style={{paddingTop:4}}>
+            <FieldGroup groupKey={'colorDialogTabs'} keepState={false}>
```
cwang2016: why is 'keepState' false here?

src/firefly/js/visualize/ui/ColorDialog.jsx
```diff
@@ -2,248 +2,226 @@
...
+        request.lowerRange,
+        useZ ? ZSCALE :request.upperWhich,
+        request.upperRange,
+        request.DR, request.BP, request.WP,
+        request.gamma, request.algorithm, request.zscaleContrast,
+        request.zscaleSamples, request.zscaleSamplesPerLine);
 
-export default ColorDialog;
+    return RangeValues.serializeRV(rv);
+}
```
cwang2016: got what you made on the 'layered fieldGroup, very clear and very good for tab case. I will try to apply the way to ImageSelectPanel for 3 color plots. Thanks.

src/firefly/js/visualize/ui/ColorDialog.jsx
```diff
@@ -2,248 +2,226 @@
...
 
-    onClick() {
-        this.doClose();
-    },
+function replot(request) {
+    if (request.colorDialogTabs) {
+        replot3Color(
+            request.redPanel,request.greenPanel,
+            request.bluePanel,
+            request.colorDialogTabs.colorTabs);
```
cwang2016: so the fieldgroup for red, green, and blue are in the parallel position in the request object as request.colorDialogTabs? I thought they represent fieldgroup inside another fieldgroup?

src/firefly/js/visualize/ui/ColorDialog.jsx
robyww: maybe it should be true, I should play with it.

src/firefly/js/visualize/ui/ColorDialog.jsx
robyww: yes, all the fields group are made parallel.  I think that is cleaner than trying to put them inside others.

PR #51, opened 2016-04-05
DM-5700-ImgSelDropdown: add ImageSelect panel to be ImageSelect dropdown for the main menu

cwang2016:


src/firefly/js/visualize/ui/ImageSelectPanel.jsx
```diff
@@ -82,7 +81,11 @@ export function computeCurrentCatalogId( fields, catalogId = IRSA ) {
 // this version is made to add a new plot to an existing viewer if no plot is found
 // or replace the plot from the active viewer
 
-const PlotSelectMode = new Enum(['AddNewPlot', 'ReplacePlot', 'ReplaceOrCreate']);
+const [REPLACE, CREATE, CREATE3COLOR] = [0x100, 0x010, 0x001];
```
robyww: I have never thought of using this syntax like this before.  It is a great way to defining related constants!

src/firefly/js/ui/ImageSelectDropdown.jsx
```diff
@@ -0,0 +1,93 @@
...
+        this.iAmMounted = true;
+        this.removeListener = flux.addListener(() => this.stateUpdate());
+
+    }
+
+    stateUpdate() {
+        var fields = FieldGroupUtils.getGroupFields(panelKey);
+
+        if (fields && this.iAmMounted ) {
+             this.setState({addPlot: isCreatePlot(this.plotInfo.plotMode, fields)});
```
robyww: since you are is using fields in this component, I don't think  you need it in the state.  Am I missing something?

robyww:
It all looks good!

src/firefly/js/ui/ImageSelectDropdown.jsx
cwang2016: 'fields' is used to calculate the value of 'addPlot' for the state. Not very sure the meaning of  'need it in the state'? If I have any misunderstanding, please let me know again. Thanks.

PR #52, opened 2016-04-05
DM-5638: file upload should display feedback while it's busy uploading.

loitly:
While file is being uploaded, a spinning icon will appears on the right end of the file to indicate it is busy uploading.  This should happen right after you submit the form.

This pull request also fixes the problem where fits file are different once uploaded onto the server.
The upload files will be stored at: /hydra/workarea/fftools/visualize/fits-cache/upload_*


cwang2016:
the fits in the cache looks good as being viewed by 'fv'.

PR #53, opened 2016-04-07
DM-5723: table fail to resize in some cases

loitly:
Just minor tweak to table container components to make it play well with other container types.
Not much to test.  Table should still appears correctly after changes.


cwang2016:
Done.

PR #54, opened 2016-04-07
DM-4571-suggestbox

tgoldina:
Created suggest box input field.
Modified ExampleDialog to show an example and XYPlotOptions to use it for column selection. 
(In future it should assist creating column expressions.)

### DIFF
--- a/src/firefly/js/ui/SuggestBoxInputField.css
+++ b/src/firefly/js/ui/SuggestBoxInputField.css
@@ -1,5 +1,5 @@
 .SuggestBoxInputField {
-  position: relative;
+    position: relative;
 }
 
 .SuggestBoxPopup {
@@ -20,7 +20,7 @@
 
 .SuggestBox__Suggestion {
   display: block;
-  padding: 2px 20px;
+  padding: 2px 20px 2px 10px;
   cursor: default;
 }
 
--- a/src/firefly/js/ui/SuggestBoxInputField.css
+++ b/src/firefly/js/ui/SuggestBoxInputField.css
@@ -1,0 +1,31 @@
+.SuggestBoxInputField {
+  position: relative;
+}
+
+.SuggestBoxPopup {
+    position: absolute;
+    overflow: auto;
+    top: 100%;
+    background: #fff;
+    z-index: 1;
+    max-height: 200px;
+    border: 2px solid #7eadd9;
+}
+
+.SuggestBox {
+  list-style-type: none;
+  padding: 0;
+  margin: 0;
+}
+
+.SuggestBox__Suggestion {
+  display: block;
+  padding: 2px 20px;
+  cursor: default;
+}
+
+.SuggestBox__Suggestion--highlighted {
+  outline: 0;
+  color: white;
+  background: #7eadd9;
+}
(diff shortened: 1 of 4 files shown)


src/firefly/js/ui/SuggestBoxInputField.css
```diff
@@ -0,0 +1,31 @@
...
+  display: block;
+  padding: 2px 20px;
+  cursor: default;
+}
+
+.SuggestBox__Suggestion--highlighted {
+  outline: 0;
+  color: white;
+  background: #7eadd9;
+}
```
loitly: Good use of BEM naming convention.  I've been converting over whenever I get a chance as well.

loitly:
Code looks fine.  I found a couple of possible bugs.
- suggestion box dropdown on top of validation message, is that intentional?  
- suggestion box mouse scroll to advance options works fine.
  however keypress down arrow does not advance options.  try scrollIntoView to see if that helps.

src/firefly/js/ui/SuggestBoxInputField.jsx
```diff
@@ -0,0 +1,175 @@
...
+        this.props.fireValueChange({ value : displayValue, message, valid});
+    }
+
+
+    render() {
+
+        const {displayValue, valid, message, highlightedIdx, isOpen } = this.state;
+        var {validator, label, labelWidth, tooltip, inline, fireValueChange, getSuggestions, renderSuggestion, valueOnSuggestion} = this.props;
+        const currentSuggestions = getSuggestions(displayValue);
+
```
robyww: I thought `getSuggestions` was going to return a promise? did you decide not to make it async?

src/firefly/js/ui/SuggestBoxInputField.jsx
```diff
@@ -0,0 +1,175 @@
...
+                    isOpen && changeHighlighted(isUndefined(highlightedIdx) ? currentSuggestions.length - 1 : Math.max(0, highlightedIdx - 1));
+                    break;
+                case 40: // arrow down
+                    isOpen && changeHighlighted(isUndefined(highlightedIdx) ? 0 : Math.min(highlightedIdx + 1, currentSuggestions.length - 1));
+                    break;
+                default:
+                    break;
+            }
+        };
+
```
robyww: Why are you defining `handleKeyPress` and `changeValue` inside the render function? since they both need to call `setState` why not just make them methods?

robyww:
UINotes:
The only real UI issue I have is that I think the drop down should be as wide as the text input field. For example, I compared to the google search box, or the Safari URL input. The suggestions should probably left justified as well.

robyww:
>> suggestion box dropdown on top of validation message, is that intentional?

You might not need z-index at all for the suggestion box popup

tgoldina:
I actually do. If I have two suggest fields one under another, the bottom
one will be on top of suggest box if I don't set z-index.

2016-04-08 9:43 GMT-07:00 Trey Roby <notifications@github.com>:

robyww:
Then we might have a much bigger problem, but it might be mine to solve.  DM-5659 is about how multiple popups interact. I don't think this is a problem unless suggestion box is in a dialog.

tgoldina:
Warning below suggestion box shows only in dialogs. In XY Plot Options it
shows on top.

2016-04-08 11:55 GMT-07:00 Trey Roby <notifications@github.com>:

PR #55, opened 2016-04-09
DM-5722: Add table client-side sorting

loitly:
You can see this in TablePanel options and fits header dialog.

### DIFF
--- a/src/firefly/js/tables/TableStore.js
+++ b/src/firefly/js/tables/TableStore.js
@@ -11,6 +11,12 @@ import * as TblUtil from './TableUtil.js';
 import {SelectInfo} from './SelectInfo.js';
 
 export class TableStore {
+
+    /**
+     * 
+     * @param props props from Table
+     * @param changeListener function called when data changed
+     */
     constructor(props, changeListener) {
         this.init(props);
         this.changeListener = changeListener;
@@ -39,15 +45,17 @@ export class TableStore {
         }
     }
 
+    onUnmount() {}
+
     onSort(sortInfoString) {
-        const {request} = this.cState.tableModel;
-        request.sortInfo = sortInfoString;
+        var {request} = this.cState.tableModel;
+        request = Object.assign({}, request, {sortInfo: sortInfoString});
         this.handleAction(TblCntlr.TABLE_FETCH, {request});
     }
 
     onFilter(filterIntoString) {
-        const {request} = this.cState.tableModel;
-        request.filters = filterIntoString;
+        var {request} = this.cState.tableModel;
+        request = Object.assign({}, request, {filters: filterIntoString});
         this.handleAction(TblCntlr.TABLE_FETCH, {request});
     }
 
@@ -93,15 +101,14 @@ export class TableStore {
         this.handleAction(TblCntlr.TABLE_SELECT, {tbl_id, selectInfo});
     }
 
-    onOptionUpdate({pageSize, columns, showUnits, showFilters}) {
+    onOptionUpdate({pageSize, columns, showUnits, showFilters, colSortDir}) {
         if (pageSize) {
             this.onPageSizeChange(pageSize);
-        } else {
-            const changes = omitBy({columns, showUnits, showFilters}, isUndefined);
-            if ( !isEmpty(changes) ) {
-                changes.tbl_ui_id = this.cState.tbl_ui_id;
-                this.handleAction(TblUiCntlr.TBL_UI_STATE_UPDATE, changes);
-            }
+        }
+        const changes = omitBy({columns, showUnits, showFilters, colSortDir}, isUndefined);
+        if ( !isEmpty(changes) ) {
+            changes.tbl_ui_id = this.cState.tbl_ui_id;
+            this.handleAction(TblUiCntlr.TBL_UI_STATE_UPDATE, changes);
         }
     }
 
@@ -116,17 +123,25 @@ export class TableStore {
     }
 
     handleAction(type, payload) {
+        var {tableModel, origTableModel} = this.cState;
+        if (! origTableModel && tableModel) {
+            // for localStore
+            // need to save orig tableModel when user wants to reset.
+            this.cState.origTableModel = cloneDeep(tableModel);
+            origTableModel = this.cState.origTableModel;
+        }
         switch (type) {
-            case (TblCntlr.TABLE_FETCH_UPDATE)  :
-                throw new Error('sorting and filtering is not implemented for localstore, yet.');
+            case (TblCntlr.TABLE_FETCH)  :
+                if (get(payload, ['request', 'sortInfo']) !== get(tableModel, ['request', 'sortInfo'])) {
+                    tableModel = TblUtil.sortTable(origTableModel, payload.request.sortInfo);
+                }
+                this.updateState({tableModel});
                 break;
 
             default:
-                var tableModel = Object.assign({}, this.cState.tableModel, payload);
+                tableModel = Object.assign({}, this.cState.tableModel, payload);
                 this.updateState({tableModel});
         }
-
-
     }
 
     /**
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -2,10 +2,11 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {get, isEmpty, uniqueId, padEnd, cloneDeep} from 'lodash';
+import {get, set, isEmpty, uniqueId, padEnd, cloneDeep} from 'lodash';
 import * as TblCntlr from './TablesCntlr.js';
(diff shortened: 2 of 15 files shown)


src/firefly/js/visualize/ui/FitsHeaderView.jsx
```diff
@@ -156,56 +155,18 @@ function getFileSizeAndPixelSize(plot, band, fitsHeaderInfo) {
...
-            columns={columns}
-            data={data}
+            key={tableModel.tbl_id}
+            tableModel={tableModel}
             height='calc(100% - 42px)'
-            sortInfo={sortInfo}
-            //tableStore={tableStore}
         />
     );
 }
```
lznakano: Adding the sorting functionality simplified FitsHeaderViewer a lot.

src/firefly/js/visualize/ui/FitsHeaderView.jsx
```diff
@@ -15,7 +15,6 @@ import CompleteButton from '../../ui/CompleteButton.jsx';
 import {getSizeAsString} from '../../util/WebUtil.js';
 import HelpIcon from '../../ui/HelpIcon.jsx';
 import {SortInfo} from '../../tables/SortInfo';
-import {TableStore} from  '../../tables/TableStore.js';
```
lznakano: This import is no longer needed.  I will take it out.

src/firefly/js/tables/TableStore.js
```diff
@@ -39,15 +45,17 @@ export class TableStore {
         }
     }
 
+    onUnmount() {}
```
robyww: Is `onUnmount` a place holder?

src/firefly/js/tables/TableUtil.js
```diff
@@ -228,6 +233,41 @@ export function smartMerge(target, source) {
...
+
+    var tableModel = cloneDeep(origTableModel);
+    set(tableModel, 'request.sortInfo', sortInfoStr);
+
+    var comparator;
+    if (!col.type || ['char', 'c'].includes(col.type) ) {
+        comparator = (r1, r2) => {
+                const [s1, s2] = [r1[colIdx], r2[colIdx]];
+                return multiplier * (s1 > s2 ? 1 : -1);
+            };
```
robyww: what happens when this comparator returns a NaN? such as when a column is undefined?

src/firefly/js/tables/TableUtil.js
```diff
@@ -228,6 +233,41 @@ export function smartMerge(target, source) {
...
+    } else {
+        comparator = (r1, r2) => {
+                const [v1, v2] = [r1[colIdx], r2[colIdx]];
+                return multiplier * (Number(v1) - Number(v2));
+            };
```
robyww: what happens when this comparator returns a NaN? such as when a column is undefined?

src/firefly/js/tables/TableUtil.js
```diff
@@ -262,6 +302,19 @@ export function tableToText(columns, dataAry, showUnits=false) {
...
+    const {startIdx, endIdx, hlRowIdx, currentPage, pageSize,totalPages} = gatherTableState(tableModel);
+    var data = tableModel.tableData.data.slice(startIdx, endIdx);
+    var tableRowCount = data.length;
+    const filterInfo = get(tableModel, 'request.filters');
+    const filterCount = filterInfo ? filterInfo.split(';').length : 0;
+    const sortInfo = get(tableModel, 'request.sortInfo');
+
+    return {startIdx, hlRowIdx, currentPage, pageSize,totalPages, tableRowCount, sortInfo, selectInfo, filterInfo, filterCount, data};
+}
+
```
robyww: I am impressed how effectively you use `get`

src/firefly/js/tables/ui/BasicTableView.jsx
```diff
@@ -0,0 +1,233 @@
...
+        })
+    ),
+    callbacks: PropTypes.shape({
+        onRowHighlight: PropTypes.func,
+        onRowSelect: PropTypes.func,
+        onSelectAll: PropTypes.func,
+        onSort: PropTypes.func,
+        onFilter: PropTypes.func
+    })
+};
```
robyww: Good example of `PropTypes.shape`. I have never used it before. I probably should start.

src/firefly/js/tables/ui/BasicTableView.jsx
```diff
@@ -0,0 +1,233 @@
...
+
+function makeColWidth(columns, data, showUnits) {
+    return !columns ? {} : columns.reduce((widths, col, cidx) => {
+        const label = col.name;
+        var nchar = col.prefWidth;
+        const unitLength = showUnits ? get(col, 'units.length', 0) : 0;
+        if (!nchar) {
+            nchar = Math.max(label.length, unitLength, get(data, `0.${cidx}.length`, 0));
+        }
```
robyww: why not `get(data, [0,cidx,'length'],0)`?

src/firefly/js/visualize/ui/FitsHeaderView.jsx
lznakano: Just test FitsHeader, the sorting works.  Nice job. I have a lot to learn, especially, the lodash, I never though of using it. Look at Loi's code, I found that get is very handy for searching something.

src/firefly/js/tables/ui/BasicTableView.jsx
```diff
@@ -0,0 +1,233 @@
...
+    shouldComponentUpdate(nProps, nState) {
+        return sCompare(this, nProps, nState);
+    }
+
+    render() {
+        const {columns, data, hlRowIdx, showUnits, showFilters, filterInfo,
+                    sortInfo, callbacks, textView, rowHeight} = this.props;
+        const {widthPx, heightPx, columnWidths, showMask} = this.state;
+
+        if (isEmpty(columns)) return (<div style={{top: 0}} className='loading-mask'/>);
```
tgoldina: It's probably a general question. What is our policy on self-closing tags? Should we adhere to HTML5 or XHTML syntax?

robyww: Well we use XHTML all the time with react components.  Are you talking only about normal html components like the `div` above?

src/firefly/js/tables/ui/TablePanelOptions.jsx
```diff
@@ -3,63 +3,23 @@
...
-    } );
-    var tableRowCount = data.length;
-
-    return {cols, data, tableRowCount, selectInfoCls };
-}
-
 export const TablePanelOptions = (props) => {
-    const {columns, pageSize, showUnits, showFilters, onChange} = props;
+    const {columns, pageSize, showUnits, showFilters, onChange, colSortDir} = props;
     if (isEmpty(columns)) return false;
```
tgoldina: Is it OK to return a primitive from a functional component?
I saw this: https://github.com/facebook/react/issues/5355

src/firefly/js/visualize/ui/FitsHeaderView.jsx
tgoldina: In which situations do we sort on the client? Does it have implications for XYPlot or image, connected to a table (because they are sharing highlightedIdx)?

src/firefly/js/tables/ui/BasicTableView.jsx
loitly: You are right.  We should adhere to HTML5 which is closing any non-void elements tags.  I'll fix it.

src/firefly/js/visualize/ui/FitsHeaderView.jsx
tgoldina: Unrelated to sorting: I noticed that some comments in Fits header show as "null". It might be better to show no value instead.

src/firefly/js/tables/TableStore.js
loitly: It's actually being called.  But in this case, it does nothing.

src/firefly/js/tables/TableUtil.js
loitly: it will still work okay since comparator only check for greater/less than zero.  i've tried it, and it did not crash.

src/firefly/js/tables/ui/BasicTableView.jsx
loitly: that'll work, too.  this is older code when i am trying to see which way is easier to read.. i.e.  'a.b.c.0'  or ['a', 'b', 'c', 0].

src/firefly/js/tables/ui/BasicTableView.jsx
tgoldina: Actually, it looks like all tags should be closed in JSX (
https://facebook.github.io/react/tips/self-closing-tag.html) – no matter
custom or not.

2016-04-11 9:47 GMT-07:00 loitly <notifications@github.com>:

PR #56, opened 2016-04-12
DM-5757, DM-5775:make the resize work in 3-color bands

lznakano:
The table sorting and popup's resizing are both working in this branch.  Please take a look and let me know your comments. 

Thanks!

### DIFF
--- a/src/firefly/js/ui/panel/TabPanel.css
+++ b/src/firefly/js/ui/panel/TabPanel.css
@@ -39,7 +39,6 @@
     \-webkit-border-top-right-radius: 4px;
     border-top-left-radius: 4px;
     border-top-right-radius: 4px;
-    height:100%;
     overflow: hidden;
     clear: both;
     flex-grow: 1;
--- a/src/firefly/js/ui/panel/TabPanel.css
+++ b/src/firefly/js/ui/panel/TabPanel.css
@@ -39,7 +39,7 @@
     \-webkit-border-top-right-radius: 4px;
     border-top-left-radius: 4px;
     border-top-right-radius: 4px;
-
+    height:100%;
     overflow: hidden;
     clear: both;
     flex-grow: 1;
--- a/src/firefly/js/visualize/ui/FitsHeaderView.jsx
+++ b/src/firefly/js/visualize/ui/FitsHeaderView.jsx
@@ -42,11 +42,12 @@ const titleStyle = {width: '100%', height: 20};
 
 //define the first column in the textStyle div
 const textColumn1 = {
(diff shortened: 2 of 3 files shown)


src/firefly/js/ui/panel/TabPanel.css
```diff
@@ -39,7 +39,7 @@
     \-webkit-border-top-right-radius: 4px;
     border-top-left-radius: 4px;
     border-top-right-radius: 4px;
-
+    height:100%;
```
loitly: height of 100% will ruin the flex-grow properties.

loitly:
Code looks fine.  However, it is breaking Tab's layout at other places, like tables.
Also, there is an obvious scrollbar on the right side of the dialog which means sizing is not right.
It also made it possible for the title to scroll off the top.

Please correct before first.

src/firefly/js/visualize/ui/FitsHeaderView.jsx
```diff
@@ -24,17 +24,18 @@ const popupPanelResizableStyle = {
...
 };
 
 
 //define the display style for the file size and pixel information and the table in the same div
-const tableAndTitleInfoStyle = {width: '100%', height: 'calc(100% - 40px)'};
+const tableAndTitleInfoStyle = {width: '100%', height: 'calc(100% - 40px)', display: 'flex'};
 
 //define the table style only in the table div
-const tableStyle = {width: '100%', height: 'calc(100% - 30px)',  overflow: 'auto', flexGrow: 1, display: 'flex'};
+const tableStyle = {paddingLeft:5,paddingRight:5, width: '100%', height: 'calc(100% - 60px)',  overflow: 'hidden', flexGrow: 1, display: 'flex'};
```
loitly: FYI:  not a big deal, since it's not very visible.  But, by adding padding while keeping 100% width just pushes the whole table to the right.  So, if you look closely, you will see that the table is chopped off by 5px to the right.  Had parent's overflow was not 'hidden', a scrollbar would appears.
You can fix this by using calc to subtract the paddings.  But, a better way to do this is to use box-sizing: border-box.  This takes paddings and borders into the '100%' calculation.

loitly:
Beside the one comment I had, everything else looks fine.  I'll let you decide whether or not you want to fix it.  Okay to merge when ready.

robyww:
>>  FYI: not a big deal, since it's not very visible. But, by adding padding while keeping 100% width just >> pushes the whole table to the right. 

You should fix this. It is little but it is there.

src/firefly/js/visualize/ui/FitsHeaderView.jsx
lznakano: Yes, you are right.  I will fix it.

PR #58, opened 2016-04-13
DM-5760 Charts need to be expandable

tgoldina:
Support expanded view for charts (histogram and plot)
(expand button in the right upper corner of the chart area)

### DIFF
--- a/src/firefly/js/visualize/ChartsTableViewPanel.jsx
+++ b/src/firefly/js/visualize/ChartsTableViewPanel.jsx
@@ -55,8 +55,8 @@ class ChartsPanel extends React.Component {
         this.state = {
             chartType: localStorage.getItem(PREF_CHART_TYPE) || SCATTER,
             optionsShown: !(localStorage.getItem(PREF_OPTIONS_SHOWN)==='false'),
-            widthPx: 50,
-            heightPx: 300
+            widthPx: 20,
+            heightPx: 20
         };
 
         this.onResize = debounce((size) => {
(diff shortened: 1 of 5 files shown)


src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -402,14 +413,24 @@ class ChartsPanel extends React.Component {
     }
 
     renderToolbar() {
+        const {expandable, expandedMode} = this.props;
```
loitly: IMO and without full knowledge of this component, I think breaking the many render functions into private functional components will make the main/public component looks cleaner.  Also from a performance perspective, it may not have to re-render.
i.e.  
function Toolbar({expandable, expandedMode}) {...}
then use <Toolbar> in place of renderToolbar

loitly:
In Safari, xy-options panel is acting strangely.  Maybe it's the suggestion box growing too tall pushing everything else off the 'hidden' container.  This is problematic.  I think you should fix it before merging.  If you can't duplicate, call me and we can try to make it happen.

Also, I noticed here: visualize/ChartsTableViewPanel.jsx:511, where autoflow: 'auto' is causing the scrollbar to flash(show/hide) in expanded mode when resizing the whole browser.  Flashing probably means the component had to render twice.

tgoldina:
I have fixed suggest box display and expanded view in Safari. I can not change visualize/ChartsTableViewPanel.jsx:511 autoflow: 'auto' to 'hidden', because it will prevent scrolling down the plot when stretch is 'fill'.

PR #60, opened 2016-04-18
DM-5364-ImgSelAddPlot replace or modify the plot based on if there is  highlighted plot or if the highlighted plot is 3-colors.

cwang2016:
The image select panel provides the options to replace or modify the plot in either 3 colors or not. The option is  based on if  any plot  is selected before opening the dropdown  or if the viewer of the selected plot can receive new plot.

### DIFF
--- a/src/firefly/js/visualize/ui/ImageSelectPanelResult.js
+++ b/src/firefly/js/visualize/ui/ImageSelectPanelResult.js
@@ -334,10 +334,5 @@ export function resultSuccess(plotInfo, hideDropdown = false) {
             dispatchPlotImage(nPlotId, wpSet[0]);
         }
 
-/*
-        if (hideDropdown) {
-            dispatchHideDropDownUi();
-        }
-*/
     };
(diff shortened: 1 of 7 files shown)


src/firefly/js/visualize/ui/ImageSelectPanelResult.js
```diff
@@ -323,15 +323,21 @@ export function resultSuccess(plotInfo, hideDropdown = false) {
...
+        }
+
         if (plotInfo.isThreeColor) {
             dispatchPlotImage(nPlotId, wpSet, true);
         } else {
             dispatchPlotImage(nPlotId, wpSet[0]);
         }
 
-
+/*
```
robyww: nit picky- If you think you need to leave this code in you should probably use the // syntax.

robyww:
I looks good. Nice job on all the of the tickets.  It works really well.  The three color is a great improvement to how we were doing it before.

PR #63, opened 2016-04-20
Tabs and CollapsiblePanel should be able to preserve state...

tgoldina:
...when not part of the field group.

- Added FieldGroupCollapsible wrapper, which manages the state of collapsible panel when it's a part of field group
- Added componentKey property to Tabs and CollapsiblePanel. If it is defined, it will be used as a key to save state through component unmounts and subsequent mounts. 
- Renamed DialogCntrl into ComponentCntrl and added dispatcher for component's state changes 

I am not sure the two last items are implemented the the right way or the change is a good idea. Perhaps we should keep the components 'dumb' and manage their state externally. For the limited number of components we are talking about, saving the state in the local storage (as a preference) might be a better idea than going through controller/dispatcher. (My main concern is that this implementation creates yet another way of managing tab/collapsible state in addition to managing state with field group connector or externally via property update.)


src/firefly/js/core/ComponentCntlr.js
```diff
@@ -0,0 +1,177 @@
...
+   Structure of the store:
+   dialogOrComponent:
+     dialog:
+        [dialogId]: {
+            visible
+            ownerId
+        }
+     component:
+        [componentId]: savedComponentState
+ */
```
robyww: I like this `structure of the store` section.  I think we all have been doing this different ways.  We might think about standardizing it.

src/firefly/js/core/ComponentCntlr.js
```diff
@@ -0,0 +1,177 @@
...
+    if (!dialogsState[dialogId] || !dialogsState[dialogId].visible || ownerId!==dialogsState[dialogId].ownerId) {
+        state = update(state,
+                        {[DIALOG_KEY] :
+                            {$merge: {
+                                [dialogId]: {
+                                    visible: true,
+                                    ownerId
+                                }
+                            }}});
+    }
```
robyww: This code is correct and clean. However, it reminds me that (after using it some) I really don't like this `update` function.  I find it very hard to read.  Personally, I am using it as the last resort now.  This comment is not about this code. It is to stimulate more discussion about the clearest way to update our stores.

robyww: An Alternative:
`const clone = (obj,params={}) => Object.assign({},obj,params);`
Once defined:
```
     state = clone(state,
                       {[DIALOG_KEY] :
                           clone(state[DIALOG_KEY],
                               { [dialogId]: {
                                   visible: true,
                                   ownerId
                               } })});
```

It is more clear what a function is and what an object is.

robyww:
Code all looks good. Tatiana and I talked about componentWillReceiveProps.

src/firefly/js/core/ComponentCntlr.js
tgoldina: For deeper updates, the advantages of update
<https://facebook.github.io/react/docs/update.html> are more prominent.

Writing something like this:

state = update(state,
    {
        [COMPONENT_KEY]: {[componentId]: {$merge: componentState}}
    });

would require 2 clone calls. The deeper you go, the more levels you have to
clone.



2016-04-20 9:06 GMT-07:00 Trey Roby <notifications@github.com>:

loitly: I am in favor of using update.  I think the syntax it's just a matter of getting used to.  You do get a consistent way of handling other command beside 'set', like merge.  Also, it's well documented.

PR #64, opened 2016-04-21
DM-5803: fetchUrl is not handling post requests correctly.

loitly:
- Change 'post' to use 'form-urlencoded' instead of 'multipart/form-data'.
- Add 'file-upload'.  It should use 'multipart/form-data'.
- Switch plot and table request method to 'post'.


robyww:
Looks good.  At some point, we should remember to test that this all works with CORS.

PR #66, opened 2016-04-22
DM-5283:fixed 2-color flux readout  layout issue and FitsHeader's issues

lznakano:
Please check and let me know if both FitsHeader and MouseReadout look OK.  Thanks!

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/CommandService.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/CommandService.java
@@ -70,6 +70,7 @@ public class CommandService extends BaseHttpServlet {
 
 //            res.setContentType("text/plain");
         } catch (Exception e) {
+            e.printStackTrace();
             StringBuilder sb = new StringBuilder(500);
             sb.append("[{");
             sb.append("\"success\" :  \"").append(false).append("\",");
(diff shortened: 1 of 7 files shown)


xiuqin:
The ticket number is wrong in the branch name. It should be DM-5820 if you want to read the ticket.

xiuqin:
issue: 
- The FITS header popup window does not show when I made a two color or one color image.
- The position readout need to be a little longer to be able to display 12h12m12.12s +12d12m12.12s
   It overlaps with the image name sometimes. Try WISE m31, with mouse around center of the target.

robyww:
- I saw the same as Xiuqin, two color does not come up with FITS header
- On safari, the resize indicator is better, however, there is still an extra one.
- I did not see the text wrapping issue, @loitly did, so I am not sure which test to use.  Maybe he can try it.

lznakano:
I emailed the error messages on Friday night.  There is a java server side issue which caused the two-color FitsHeader does not show up.  The issue is not from the client (javascript side).  FYI

lznakano:
Trey,

I changed the text font type and the background color.  Please take a look and let me know if you spot anything else.  

Thanks!

lznakano:
Xiuqin and Trey,

Please verify it again to see if everything looks OK.  Thanks!

robyww:
put this in hipchat but repeating it here:
I have looked.  You still don't have the  background set to look like the standard header dialog, it is whiteish
goto line 61 of FitsHeaderView and set the background to #e3e3e3; then remove any color style that override them in the tab case

robyww:
Looks good.  Go ahead and merge.

PR #67, opened 2016-04-22
DM-5689: Added TABLE_NEW_LOADED action, fired when table is completel…

loitly:
- added new action TABLE_NEW_LOADED to table; fired when table is completely loaded.
- added table error handling.
- fix active table not updating after an active tab is removed.

### DIFF
--- a/src/firefly/js/tables/reducer/TableDataReduer.js
+++ b/src/firefly/js/tables/reducer/TableDataReduer.js
@@ -2,10 +2,12 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 import update from 'react-addons-update';
+import {get} from 'lodash';
 
 import * as TblUtil from '../TableUtil.js';
 import * as Cntlr from '../TablesCntlr.js';
 import {SelectInfo} from '../SelectInfo.js';
+
 
 
 /*---------------------------- REDUCERS -----------------------------*/
@@ -18,15 +20,18 @@ export function dataReducer(state={data:{}}, action={}) {
                 return update(root, { [tbl_id] : {selectInfo: {$set: selectInfo}}});
             } else return root;
 
+        case (Cntlr.TABLE_NEW_LOADED)  :
+            if (get(root, [tbl_id, 'tableMeta', 'Loading-Status'], 'COMPLETED') !== 'COMPLETED') {
+                return update(root, { [tbl_id] : {tableMeta: {['Loading-Status']:  {$set: 'COMPLETED'}}}});
+            } else return root;
+
         case (Cntlr.TABLE_HIGHLIGHT)  :
-        case (Cntlr.TABLE_LOAD_STATUS)  :
         case (Cntlr.TABLE_UPDATE)  :
             return TblUtil.smartMerge(root, {[tbl_id] : action.payload});
 
         case (Cntlr.TABLE_NEW)  :
         case (Cntlr.TABLE_REPLACE)  :
             const nTable = Object.assign({selectInfo: SelectInfo.newInstance({}).data},action.payload);
-            
             return update(root, {$merge: {[tbl_id] : nTable}});
 
         case (Cntlr.TABLE_REMOVE)  :
--- a/src/firefly/js/tables/reducer/TableUiReducer.js
+++ b/src/firefly/js/tables/reducer/TableUiReducer.js
@@ -3,9 +3,8 @@
  */
 
 import update from 'react-addons-update';
-import {set, has, omit, omitBy, get, isEmpty, isUndefined, cloneDeep} from 'lodash';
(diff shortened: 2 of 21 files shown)


src/firefly/js/tables/reducer/TableDataReduer.js
```diff
@@ -18,15 +20,18 @@ export function dataReducer(state={data:{}}, action={}) {
                 return update(root, { [tbl_id] : {selectInfo: {$set: selectInfo}}});
             } else return root;
 
+        case (Cntlr.TABLE_NEW_LOADED)  :
+            if (get(root, [tbl_id, 'tableMeta', 'Loading-Status'], 'COMPLETED') !== 'COMPLETED') {
+                return update(root, { [tbl_id] : {tableMeta: {['Loading-Status']:  {$set: 'COMPLETED'}}}});
```
robyww: could this be?
 `return update(root, { [tbl_id] : {tableMeta: {Loading-Status:  {$set: 'COMPLETED'}}}});`

As I said earlier,  I am begging to really not like update because it is so hard to read. I though you wrote something else to stick an object into a deep object?

src/firefly/js/tables/reducer/TableUiReducer.js
```diff
@@ -60,15 +53,16 @@ function removeTable(root, action) {
...
+    // if (!get(tableModel, 'tableData')) return ui;
     const {startIdx, endIdx, ...others} = TblUtil .getTblInfo(tableModel);
-    var data = tableModel.tableData.data.slice(startIdx, endIdx);
-    var tableRowCount = data.length;
     const filterInfo = get(tableModel, 'request.filters');
     const filterCount = filterInfo ? filterInfo.split(';').length : 0;
     const sortInfo = get(tableModel, 'request.sortInfo');
     const showLoading = !TblUtil.isTableLoaded(tableModel);
 
+    var data = has(tableModel, 'tableData.data') ? tableModel.tableData.data.slice(startIdx, endIdx) : [];
```
robyww: never seen `has` before, cool!

src/firefly/js/tables/reducer/TableUiReducer.js
```diff
@@ -79,7 +73,21 @@ function uiStateReducer(ui, tableModel) {
         ui = update(ui, {$merge: {[tbl_ui_id]: uiData}});
     });
     return ui;
+}
 
+function updateAllUi(ui, tbl_id, tbl_ui_id, payload) {
+    if (tbl_ui_id) {
+        const changes = set({}, [tbl_ui_id], payload);
+        return TblUtil.smartMerge(ui, changes);
```
robyww: concerning my previous comment about update would a `set`/`smartMerge` be a substitute for using `update` with a `$set`?

robyww:
Test and appears to work. Code looks good.

src/firefly/js/tables/reducer/TableDataReduer.js
ejoliet: Minor: file name misses 'c' for reduCer.js ;-)

--
Emmanuel Joliet
Office: 170
Phone:  (626) 395-1489

loitly: Regarding 'update', I agree the syntax is cryptic.  But, it's also a good idea to use already written code instead of writing it ourselves.  Why don't we wrap lodash 'set' with 'update' to create update funtions with lodash 'set' syntax..  ie.. using the above example...
        udSet(root, [tbl_id, 'tableMeta', 'Loading-Status'], 'COMPLETED');
   other use cases of update..
        udMerge((root, [tbl_id, 'tableData', 'data', rowIdx], newRowOfData);
        udSet((root, 'data.currrentTable.error', errorMsg);
            in this case, path is a string.

robyww: That is a good idea, we still us it but put in into some simple wrappers with lodash type syntax that we all like and seems clear.

tgoldina:
Sort on a table column produces TypeError: Cannot read property 'toLocaleString' of undefined in TablePanel.jsx:206

tgoldina:
Another bug: when I go to the last page (could be highlight-driven) and filter from plot, the table comes with "No Data Found".

tgoldina:
I all looks good to me. I think the bugs I mentioned are long standing and should not prevent you from merging this one in.

tgoldina:
I have committed the changes to chart-related controllers to this branch to remove the dependency on search request and use tbl_id directly. (This was done to remove tbl_id from the serialized search request, which we use as a handle to the table on the server. The processing of this request on the server was triggering the erroneous updates to the original table.)

loitly:
- added wrapper funtions for update.
- renamed TableDataReducer.js
- fixed 'No Data Found' after filtered when at end of table.

PR #68, opened 2016-04-22
DM-5800 Tab titles need to shrink to accommodate long names

tgoldina:
Tabs are now resizable.
If not enough space is available ellipsis character will be added at the end.
The tooltip with full name will come up when pointing to the tab.

I have experimented with making Mac style mid string truncation, as discussed [here](http://stackoverflow.com/questions/831552/ellipsis-in-the-middle-of-a-text-mac-style). Unfortunately, the most promising [solution](http://jsfiddle.net/r96vB/1/)  breaks in Firefox (resulting in overlap). It looks like purely CSS solution will ether have partial characters or uneven sizing due to truncated characters. To look good, the original string needs to be manipulated, which requires iterating on it because we can not tell in advance many pixels a string it will take.

### DIFF
--- a/src/firefly/js/ui/panel/TabPanel.css
+++ b/src/firefly/js/ui/panel/TabPanel.css
@@ -20,6 +20,8 @@
     \-moz-border-radius-topleft: 5px;
     \-webkit-border-top-left-radius: 5px;
     border-top-left-radius: 5px;
+    white-space: nowrap;
+    overflow: hidden;
 }
 
 .TabPanel__Tab--selected {
@@ -45,3 +47,9 @@
     display: flex;
 }
 
+.ellipsis {
+    display: inline-block;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+}
--- a/src/firefly/js/ui/panel/TabPanel.css
+++ b/src/firefly/js/ui/panel/TabPanel.css
@@ -47,9 +47,3 @@
     display: flex;
 }
 
-.ellipsis {
-    display: inline-block;
-    white-space: nowrap;
-    text-overflow: ellipsis;
-    overflow: hidden;
-}
--- a/src/firefly/js/ui/panel/TabPanel.jsx
+++ b/src/firefly/js/ui/panel/TabPanel.jsx
@@ -100,7 +100,7 @@ export class Tabs extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        this.state= Object.assign(tabsStateFromProps(nextProps));
+        this.state= tabsStateFromProps(nextProps);
     }
     
     shouldComponentUpdate(np, ns) {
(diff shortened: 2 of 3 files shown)


xiuqin:
Some initial calculation is wrong. The image search panel came up with  '...' in all the tabs.

tgoldina:
Indeed, I always include 12px for remove button, which is not always
present.

2016-04-22 16:18 GMT-07:00 Xiuqin Wu <notifications@github.com>:

tgoldina:
I have reworked the calculation to preserve relative size of tabs. (This should have fixed the bug Xiuqin has found.)

src/firefly/js/ui/panel/TabPanel.css
```diff
@@ -45,3 +47,9 @@
     display: flex;
 }
 
+.ellipsis {
+    display: inline-block;
```
loitly: shouldn't this be .TabPanel__Tab--ellipsis?

src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -34,14 +100,13 @@ export class Tabs extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        this.state= tabsStateFromProps(nextProps);
+        this.state= Object.assign(tabsStateFromProps(nextProps));
```
loitly: what does this 'assign' statement do?  for assign, don't you need at least 2 parameters?

loitly:
I can't switch tab on safari.  Will you double check?
Otherwise, it looks good.

tgoldina:
I have tested in Safari 9.0.3 on Mac OS 10.10.5, it works fine.

I am not sure we should struggle to support old browser versions, for which
the support was dropped or is about to drop.

For example, I have Safari 6.2.8, In which table does not load at all,
producing the following error:

WARNING: Unable to parse json message into ServerEvent: [object
Object]com.google.gwt.json.client.JSONException: Error parsing JSON:
SyntaxError: JSON Parse error: Unexpected identifier "object"
at Unknown.fillInStackTrace(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@27:4898
)
at Unknown.Throwable_0(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@24:4859
)
at Unknown.Exception_0(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@22:4979
)
at Unknown.RuntimeException_0(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@22:5019
)
at Unknown.JSONException(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@29:29531
)
at Unknown.throwJSONException(
http://localhost:8080/fftools/28537EC390FCED68FAC6C8C469A388BF.cache.js@32:30026
)

2016-04-25 9:45 GMT-07:00 loitly <notifications@github.com>:

src/firefly/js/ui/panel/TabPanel.css
tgoldina: I don't think so. It's not a modifier. It is quite generic style for text: if there is not enough room, the text is clipped and the clipped part is represented with '...' 

I will rename it to text-ellipsis and move it to global.css

PR #71, opened 2016-04-26
build parser to parse DS9 region JSON resulted by region file upload and region objects to contain data and properties of all types of regions

cwang2016:
- create Region class to contain various types of region data
- build parser to parse region JSON generated by region file upload.

- some accessory update: change the label of 3 color image to be '3-Color Image'.

### DIFF
--- a/src/firefly/js/visualize/region/Region.js
+++ b/src/firefly/js/visualize/region/Region.js
@@ -7,20 +7,31 @@
  */
 
 import Enum from 'enum';
+import {has, isNil} from 'lodash';
 
 export const RegionType = new Enum(['circle', 'annulus', 'ellipse', 'ellipseannulus', 'box',
-                                    'boxannulus', 'line', 'point', 'polygon', 'text', 'undefined']);
+                                    'boxannulus', 'line', 'point', 'polygon', 'text', 'message',
+                                    'undefined'], {ignoreCase: true});
 
 export const RegionCsys = new Enum(['PHYSICAL', 'FK4', 'B1950', 'FK5', 'J2000',
                                     'IMAGE', 'ECLIPTIC', 'GALACTIC',
-                                    'ICRS', 'AMPLIFIER', 'LINEAR', 'DETECTOR', 'UNDEFINED']);
+                                    'ICRS', 'AMPLIFIER', 'LINEAR', 'DETECTOR', 'UNDEFINED'], {ignoreCase: true});
 
 export const RegionValueUnit = new Enum(['CONTEXT', 'DEGREE', 'RADIUS', 'ARCMIN', 'ARCSEC',
-                                         'SCREEN_PIXEL', 'IMAGE_PIXEL']);
+                                         'SCREEN_PIXEL', 'IMAGE_PIXEL'], {ignoreCase: true});
 
 export const RegionPointType = new Enum(['circle', 'box', 'cross', 'diamond', 'x',
-                                         'arrow', 'boxcircle', 'undefined']);
+                                         'arrow', 'boxcircle', 'undefined'], {ignoreCase: true});
 
+
+
+var cloneArg = (arg) => Object.keys(arg).reduce((prev, key) =>
+                        {
+                            if (!isNil(arg[key])) {
+                                prev[key] = arg[key];
+                            }
+                            return prev;
+                        }, {});
 
 /**
  *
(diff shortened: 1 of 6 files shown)


src/firefly/js/visualize/region/Region.js
```diff
@@ -0,0 +1,291 @@
...
+ */
+export var Region =
+    ( {wpAry = null,
+      radiusAry = null,
+      dimensionAry = null,
+      angle = null,
+      type = RegionType.undefined,
+      options = null,
+      highlighted = false}
+    ) => ({type, wpAry, radiusAry, dimensionAry, angle, options, highlighted});
```
robyww: Let a a head and make the Factory functions lower case such a `makeRegion` instead of `Region` and `makeRegionOptions` instread of `RegionOptions`, etc

src/firefly/js/visualize/region/Region.js
```diff
@@ -0,0 +1,291 @@
...
+export class RegParseException extends Error {
+    constructor(message) {
+        super(message);
+        this.message = message;
+        this.name = 'RegionParseError';
+    }
+}
+
+
+var findEnumKey = (enumObj, val) => ( enumObj.enums.find((item) => item.key.toLowerCase() === val.toLowerCase()));
```
robyww: You don't need this.  
```
 RegionType = new Enum([ 'circle', 'annulus', 'ellipse', 'ellipseannulus', 'box',
                                    'boxannulus', 'line', 'point', 'polygon',
                                     'text', 'undefined'], 
                      { ignoreCase: true });
```
then
```
RegionType.get('AnnLus'); // => RegionType.annulus
```
see https://www.npmjs.com/package/enum
about half way down.

src/firefly/js/visualize/region/Region.js
```diff
@@ -0,0 +1,291 @@
...
+ * @param highlightable bool
+ * @param deletable     bool
+ * @param fixedSize     bool
+ * @param include       bool
+ * @param lineWidth     number
+ * @param offsetX       number
+ * @param offsetY       number
+ * @param message       string, containing parsing message
+ * @constructor
+ */
```
robyww: I like the documetation

src/firefly/js/visualize/region/Region.js
robyww: I don't think we need to carry unnecessary null or false info in Region. I like the way you take an object but you don't need to default everything.  Just take the object and do an object assign.  Of course you have to force a type key like you are doing now. There could be a lot and this not having unrelated keys would make the object lighter.

src/firefly/js/visualize/region/Region.js
robyww: I have similar feeling about unnecessary keys here, but since they mostly default to true it is harder.

src/firefly/js/visualize/region/Region.js
```diff
@@ -0,0 +1,291 @@
...
+        deletable = true,
+        fixedSize = false,
+        include   = true,
+        lineWidth = 0,
+        offsetX   = 0,
+        offsetY   = 0,
+        message  = null}
+    ) => ({color, text, font, pointType, pointSize, editable, movable, rotatable, highlightable,
+        deletable, fixedSize, include, lineWidth, offsetX, offsetY, message});
+
```
robyww: I have similar feeling about unnecessary keys here, but since they mostly default to true it is harder.
 Add a line note

src/firefly/js/visualize/region/RegionFactory.js
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    /**
+     * parsePart parses the region data of JSON result (one item from RegionData array)
+     * @param regionStr
+     * @param index
+     * @returns {Region} or throw RegParseException
+     */
+    static parsePart(regionStr, index = -1) {
+
```
robyww: I am not convinced we need to throw an exception anymore, since we are only supporting parsePart.  You could just `return false`, return an object with a RegionType of some sort of error and a message key. I get more nervous about throwing exceptions in JavaScript since catching is not enforced.

robyww: All that said, an exception still might be the best choice when you are deep in parsing code.  Just stick with the exception.

robyww:
The parsing all looks good. The key issues are the enum case insensitive, carrying all the keys, and naming factory methods starting with lower case.  I like the way you have Region.js organized, it is very easy to follow.

PR #73, opened 2016-04-28
DM-4573 expression support (converted expression parsing library to JS)

tgoldina:
Converted GWT expression library to JS.
Wrote a unit test for its functions and operators.
Added expression support to XY plot column selection.

### DIFF
--- a/src/firefly/js/util/expr/Parser.js
+++ b/src/firefly/js/util/expr/Parser.js
@@ -37,40 +37,41 @@ const rators2 = [
         Expr.ATAN2, Expr.MAX, Expr.MIN
 ];
 
- //Parses strings representing mathematical formulas with variables.
- //The following operators, in descending order of precedence, are
- //defined:
- //
- //<UL>
- //<LI>^ (raise to a power)
- //<LI>* /
- //<LI>Unary minus (-x)
- //<LI>+ -
- //<LI>&lt; &lt;= = &lt;&gt; &gt;= &gt;
- //<LI>and
- //<LI>or
- //</UL>
- //
- //^ associates right-to-left; other operators associate left-to-right.
- //
- //<P>These unary functions are defined:
- //abs, acos, asin, atan,
- //ceil, cos, exp, floor,
- //log, round, sin, sqrt,
- //tan.  Each requires one argument enclosed in parentheses.
- //
- //<P>There are also binary functions: atan2, min, max; and a ternary
- //conditional function: if(test, then, else).
- //
- //<P>Whitespace outside identifiers is ignored.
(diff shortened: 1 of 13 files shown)


src/firefly/js/util/expr/Parser.js
```diff
@@ -0,0 +1,287 @@
...
+ //
+ //<P>Whitespace outside identifiers is ignored.
+ //
+ //<P>Examples:
+ //<UL>
+ //<LI>42
+ //<LI>2-3
+ //<LI>cos(x^2) + sin(x^2)
+ //<UL>
+
```
loitly: shouldn't these comments be in a jsdoc block?

src/firefly/js/visualize/XYPlotOptions.jsx
```diff
@@ -11,12 +11,37 @@ import {FieldGroup} from '../ui/FieldGroup.jsx';
...
+    for (i = text.length-1; i>=0; i--) {
+        c = text.charAt(i);
+        if (!/[A-Za-z\d_]/.test(c)) {
+            priorIdx = i;
+            break;
+        }
+    }
+    if (priorIdx > 0) priorContent = text.substring(0, priorIdx+1);
+    if (priorIdx < text.length) token = text.substring(priorIdx+1);
+
```
loitly: i am sure this works fine.
just sharing info.. have you tried using string.match?  works great for taking chunks out of matching strings..
i.e.  
> 'cos(x^2)'.match(/(.*)([^A-Za-z\d_])/)
[ 'cos(x^2)', 'cos(x^2', ')', index: 0, input: 'cos(x^2)' ]

loitly:
A couple of things I noticed:

1.  It performs badly in Firefly. There is too much lags between keystrokes.  Not sure what you can do about that.
2.  Selecting points in xy-plot will highlight table, but not vise-versa.  Is that intentional?

Otherwise, it is fine.  I will put this ticket into reviewed.  If you made changes and wanted me to review again, feel free to re-assign.

tgoldina:
Thank you for testing. I've fixed highlight from table. (I forgot to add
support for expressions when calculating highlight.)

Also thank you for suggesting [match](
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match),
it simplifies things.

Regarding Firefox performance, I noticed I still get unexpected table
reloads. (For example, when I enter or remove a key in a suggest box.) When
only plot is on the page (as in expanded view), it seems to perform much
better.


2016-04-28 18:08 GMT-07:00 loitly <notifications@github.com>:

PR #74, opened 2016-04-28
DM-4920: Multi image fits support and react optimization work

robyww:
- When you load a multi image fits the UI will control which image is visible.
 - Zoom and rotate will work on all the images.
 - Then... a lot of react optimization. Hard to tell if it helps but the FPS is higher


cwang2016:
Looks good to me. 
A note: for  "load only one" for multi-image  case doesn't work. Some error message from 'PlotState.js' is shown.

robyww:
>> A note: for "load only one" for multi-image case doesn't work. Some error message from 'PlotState.js' is shown.

I forget to check that.  I will look at it.

PR #75, opened 2016-04-30
DM-5959: multiple dialog are not working well together

robyww:


src/firefly/js/ui/DialogRootContainer.jsx
```diff
@@ -93,22 +133,59 @@ function reRender(dialogs,tmpPopups) {
...
+    return retval;
+}
+
+
+
 /**
  * @param popup {object}
  */
 function showTmpPopup(popup) {
     if (!divElement) init();
```
lznakano: Can init() function be ES6 format:
const init = ()=>{
...
 }? 
 I am just wondering?

src/firefly/js/ui/PopupPanel.jsx
```diff
@@ -26,8 +26,11 @@ export var PopupPanel= React.createClass(
         title : PropTypes.string,
```
lznakano: There are a few unused import:
Component
 getAbsoluteLeft
 getAbsoluteTop

src/firefly/js/ui/DropDownMenu.jsx
```diff
@@ -81,5 +81,6 @@ DropDownMenuWrapper.propTypes= {
     visible : PropTypes.bool,
     x : PropTypes.number.isRequired,
     y : PropTypes.number.isRequired,
-    content : PropTypes.object.isRequired
+    content : PropTypes.object.isRequired,
+    zIndex : PropTypes.number
 };
```
lznakano: The argument "e" in computePosition is no longer needed.

lznakano: Everything looks good to me. The comments I made are all small things.  I also learned how to use reduce from reviewing Trey's code.  I have a GAIN.!!!

src/firefly/js/ui/DialogRootContainer.jsx
robyww: it could be but it really just need to be a simple function. I wrote it like that before we had standardized on how to do it.  I will fix it.

src/firefly/js/ui/PopupPanel.jsx
robyww: good catch, I will fix the lodash import as well, it is using the old style

PR #78, opened 2016-05-05
Dm 4485 region drawing

cwang2016:
region drawing for the following regions, 
circle, ellipse, box, polygon, point (circle, box, diamond, cross, x, arrow, boxcircle), line, text, annulus, box annulus, ellipse annulus.  (annulus is made for the case with at least two repeated regions of the same type). 

add the following functions to file ShapeDataObj.js
- makeRectangleByCenter, makeEllipse
- drawEllipse, 
- makeTextLocationRectangle, makeTextLocationEllipse
- update drawRectangle by adding the case which is given the rotating angle and rectangle center.

### DIFF
--- a/src/firefly/js/core/ReduxFlux.js
+++ b/src/firefly/js/core/ReduxFlux.js
@@ -44,6 +44,7 @@ import NorthUpCompass from '../drawingLayers/NorthUpCompass.js';
 import Catalog from '../drawingLayers/Catalog.js';
 import WebGrid from '../drawingLayers/WebGrid.js';
 
+import RegionPlot from '../drawingLayers/RegionPlot.js';
 
 import {showExampleDialog} from '../ui/ExampleDialog.jsx';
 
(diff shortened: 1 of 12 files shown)


robyww:
On the drawing Layer dialog: You should put the file name in the title.  When we start using the api, there might not be a file name, then you should use the drawing layer id.  If is also possible the api will provide a title. For now, put the file name in the title.

cwang2016:
ok, the title will be updated. One more thing, when should the dialog do when the draw button is clicked? 
for now, some error message is shown below the upload button if upload or drawing fails.

cwang2016:
And the dialog stays open for now.

robyww:
>> One more thing, when should the dialog do when the draw button is clicked? 
Close the dialog after the upload completes successfully complete,  If it fail leave it open with a message.

robyww:
You did a great job on this ticket! Adding more region point type support, now supporting ellipse, better annulus, and great test files. This is a big step forward in our region support. 

_I am impressed._

PR #79, opened 2016-05-06
DM-5986 Usa sagas in place of side-effects

tgoldina:
Removed or converted side-effects in chart related controllers


src/firefly/js/visualize/XYPlotCntlr.js
```diff
@@ -72,25 +74,55 @@ export function dispatchLoadPlotData(xyPlotParams, searchRequest) {
...
 }
 
-export function dispatchSetZoom(tblId, selection) {
+/*
+ * Zoom XY plot to a given selection or reset zoom if no selection is given
+ * @param {String} tblId - table id
+ * @param {Object} selection - {xMin, xMax, yMin, yMax}
+ */
+export function dispatchZoom(tblId, selection) {
+    const {xyPlotData, xyPlotParams, decimatedUnzoomed} = flux.getState()[XYPLOT_DATA_KEY][tblId];
```
loitly: this is a good place to use lodash 'get'; too many ways to get undefined error:
get(flux.getState(), [XYPLOT_DATA_KEY,tblId], {});

src/firefly/js/visualize/XYPlotCntlr.js
```diff
@@ -154,103 +170,56 @@ export function reducer(state={}, action={}) {
...
-                newState[tblId] = {plotDataReady: false};
-                const request = findTblById(tblId).request;
-                action.sideEffect((dispatch) => fetchPlotData(dispatch, request, nextPlotParams));
-            }
-            return newState;
-
+            const newParams = Object.assign({}, state[tblId].xyPlotParams);
+            Reflect.deleteProperty(newParams, 'selection');
+            Reflect.deleteProperty(newParams, 'zoom');
+            return updateSet(state, [tblId,'xyPlotParams'], newParams);
```
loitly: lodash omit is a good option for this.
const newParams = omit(state[tblId].xyPlotParams, ['selection', 'zoom']);

src/firefly/js/core/MasterSaga.js
```diff
@@ -30,6 +31,7 @@ export function* masterSaga() {
     // This section starts any predefined Sagas
     yield fork( watchCatalogs, null, dispatch, getState);
     yield fork( imagePlotter, null, dispatch, getState);
+    yield fork( syncCharts, null, dispatch, getState);
```
loitly: Question for you and @robyww, is this where we are suppose to initialize these sagas?
What if it's not needed?

src/firefly/js/visualize/saga/ChartsSync.js
```diff
@@ -0,0 +1,54 @@
...
+                }
+                histogramState = flux.getState()[HistogramCntlr.HISTOGRAM_DATA_KEY];
+                if (has(histogramState, tbl_id)) {
+                    const histogramParams = xyPlotState[tbl_id].histogramParams;
+                    HistogramCntlr.dispatchLoadColData(histogramParams, request);
+                }
+                break;
+        }
+    }
+}
```
loitly: Same question; if there is no histogram or xyplot, will it still dispatch the load?

loitly:
I see js/side-effect is still there.  And, it's still in redux's middleware.
I believe this is the last of the side-effects.  If so, you should remove side-effects as well.

loitly:
After table sort, histogram reset or disappear.  Is that normal?  Do table need to send out a different action to differentiate a sort vs a filter?

src/firefly/js/visualize/saga/ChartsSync.js
tgoldina: 'has' checks if there is statistics/xyplot/histogram connected to the
table. The load will be dispatched only if there is something connected.
But it does not take into account if the chart is displayed or not. I
should probably have some logic for lazy (delayed) load. If a chart is not
on display, the data should not load. The load should be dispatched only
when the chart becomes visible. I'll create a ticket for it.

2016-05-06 16:39 GMT-07:00 loitly <notifications@github.com>:

src/firefly/js/core/MasterSaga.js
robyww: This is for predefined sagas,  we may want to move the list to ReduxFlux where all the other list of things that are initialized are. 

>> where are we suppose to initialize these sagas? What if it's not needed?

I don't understand the questions.

We current have  many more sagas than these three they are more short lived and dynamically started in certain context using `dispatchAddSaga`.

src/firefly/js/visualize/XYPlotCntlr.js
```diff
@@ -72,25 +74,55 @@ export function dispatchLoadPlotData(xyPlotParams, searchRequest) {
...
+            } else {
+                dispatchSetZoom(tblId, selection);
+            }
+        } else {
+            // reset zoom
+            if (decimatedUnzoomed || isUndefined(decimatedUnzoomed)) {
+                const tableModel = findTblById(tblId);
+                if (tableModel) {
+                    const paramsWithoutZoom = Object.assign({}, xyPlotParams);
+                    Reflect.deleteProperty(paramsWithoutZoom, 'zoom');
```
robyww: picky: you could use lodash `omit` here.

src/firefly/js/visualize/saga/ChartsSync.js
robyww: if you look at `CatalogWatcher.js`, you see the first line is:
```
yield take(ImagePlotCntlr.PLOT_IMAGE);
```
This keeps the sega from doing anything until there is a image loaded.  You might be able to so something like that.

tgoldina: Then we need an action for setting/destroying chart area. Or we have one?

2016-05-07 10:08 GMT-07:00 Trey Roby <notifications@github.com>:

robyww: I think you are going to have to have an action at a lower level than a result area type layout.  The problem is that with the api you might put a chart in a div and a table in another div.  Then you will not have a result area laying them out.

tgoldina:
Yes, it would be beneficial to have in the TABLE_NEW_LOADED payload a
trigger field, which would differentiate actions triggered by sort (where
data do not change, only their order) or filter from other loads. We don't
need to reload table statistics or histogram on sort. But we do need to to
reload them on filter, as you noticed.

2016-05-06 16:55 GMT-07:00 loitly <notifications@github.com>:

PR #81, opened 2016-05-10
DM-6050 table caching optimizations

tgoldina:
We've been trying to fix it for a while. Still in the latest build, loading table, showing XY plot options, and getting XY data result in 3 base catalog requests. This is an attempt to cure it - only one base catalog request should be made.
 - Same key is returned even if the order of the parameters is different
 - DECIMATE_INFO is included into SYS_PARAMS
 - Also cleaned warnings in the classes I have touched


loitly:
code looks good.  tested it, and confirmed that caching is working as described.
good to merge.

PR #82, opened 2016-05-11
DM-5188: firefly API initial work

robyww:
- Defined new directory js/api
- lowlevel api
- Starting of highlevel api, hooks for more work
- remove VisMouseCntlr
- Added VisMousesSync so mouse events don't go into redux
- Redid the way the do mouse Readout with the MouseReadoutWatch saga
- New Mouse Readout controller
- Example mouse readout (VerySimpleMouseReadout.jsx)
- beginning of example html files
- Adjust ui to work better in API mode
- _also_ fixes a popup bug: DM-6029


src/firefly/html/demo/ffapi-highlevel-test.html
```diff
@@ -0,0 +1,116 @@
...
+            var v2= firefly.makeImageViewer('old2');
+            v1.plot(req);
+            v2.plotURL('http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band2.fits');
+        }
+   }
+   
+</script>
+
+
+<script src="fftools.js"></script>
```
tgoldina: Why does it depend on fftools? Where is this file coming from?

src/firefly/js/api/ApiBuild.js
```diff
@@ -0,0 +1,155 @@
...
+        findActionType(XYPlotCntlr, XYPlotCntlr.XYPLOT_DATA_KEY ),
+        findActionType(TablesCntlr, TablesCntlr.DATA_PREFIX),
+        findActionType(TablesCntlr, TablesCntlr.RESULTS_PREFIX),
+        findActionType(TablesCntlr, TablesCntlr.UI_PREFIX),
+        findActionType(ReadoutCntlr, ReadoutCntlr.READOUT_PREFIX),
+        findActionType(MultiViewCntlr, MultiViewCntlr.IMAGE_MULTI_VIEW_PREFIX),
+        findActionType(ImPlotCntlr.default, ImPlotCntlr.PLOTS_PREFIX),
+        findActionType(AppDataCntlr, AppDataCntlr.APP_DATA_PATH),
+    );
+
```
tgoldina: extra comma after the last parameter

src/firefly/js/api/ApiBuild.js
```diff
@@ -0,0 +1,155 @@
...
+        MultiViewStandardToolbar,
+        ApiExpandedDisplay,
+        ExpandedModeDisplay,
+        ImageMetaDataToolbar,
+        TablesContainer,
+        TablePanel
+    };
+    
+    const util= Object.assign({}, ApiUtil, {image:ApiUtilImage}, {xyplot:{}}, {table:{}}, {data:{}} )
+
```
tgoldina: missing semicolon

src/firefly/js/api/ApiUtil.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+
+
+/**
+ *
+ * @param {string|{}} div a div element or a string id of the div element
+ * @param {{}} Component a react component
+ * @param {{}} [props] props for the react component
+ * @param {{}} [style] css style object, style of this root element
```
tgoldina: We might want to mention at least camel case here, the inline styles are somewhat React-specific: https://facebook.github.io/react/tips/inline-styles.html For example, you can specify pixel values as numbers.

src/firefly/js/api/ApiUtil.js
```diff
@@ -0,0 +1,97 @@
...
+            <Component {...props} />
+        </div>
+    );
+
+    ReactDOM.render(renderStuff,divElement);
+}
+
+
+/**
+ * removed the redenered element
```
tgoldina: typos here

src/firefly/html/demo/ffapi-highlevel-test.html
robyww: that is a mistake, I will remove it.

src/firefly/js/api/ApiUtil.js
robyww: They don't have to be react specific, react accepts `{width:'123px'}` or `{width:123}`.  Your right, they are JavaScript api style specific. I will mention it, it will not hurt.  It is standard JavaScript.

Let me look up the correct term. I don't think it is called 'JavaScript api style'

src/firefly/js/visualize/VisMouseSync.js
```diff
@@ -0,0 +1,81 @@
...
+    return new Promise((resolve) => addOneTimeMouseCall( (s) => resolve(s)) );
+}
+
+export function fireMouseCtxChange(mouseCtx) {
+    lastCtx= mouseCtx;
+    listenerList.forEach((l) => l(mouseCtx));
+    oneTimeCallList.forEach((l) => l(mouseCtx));
+    oneTimeCallList=[];
+}
+
```
tgoldina: Nice. No more mouse-related actions.

src/fftools/js/fftools.js
```diff
@@ -47,5 +48,8 @@ if (app) {
     ReactDOM.render(React.createElement(FireflyViewer, props),
         document.getElementById(props.div));
 }
+else {
+    initApi();
+}
```
loitly: fftools.js is no longer build.  Should we remove this file?

tgoldina:
Nice. Some things I have noticed about firefly/demo/ffapi-highlevel-test.html
- Expanded mode has a scrollbar and you can scroll to see empty placeholders.
- The size is not quite right: mouse readout is cut.
- Layers/Tools/Expand buttons appear on mouse enter, but do not disappear on mouse leave (unless you it enters another plot).

The next one is a question: would it make sense to use the same plot group id by default for the images shown in the same div?

src/firefly/html/demo/ffapi-highlevel-test.html
```diff
@@ -0,0 +1,116 @@
...
+                    plotGroupId : 'myGroup',
+                    Service   : 'TWOMASS',
+                    Title     : '3 color',
+                    SurveyKey  : 'j',
+                    WorldPt    : '10.68479;41.26906;EQ_J2000',
+                    SizeInDeg  : '.12'
+                },
+                {
+                    Type      : 'SERVICE',
+                    PLOTgroupId : 'myGroup',
```
tgoldina: are the props case insensitive?

src/firefly/html/demo/ffapi-highlevel-test.html
```diff
@@ -0,0 +1,116 @@
...
+                    Type      : 'SERVICE',
+                    plotGroupId : 'myGroup',
+                    Service   : 'TWOMASS',
+                    title     : '3 color',
+                    SurveyKey  : 'k',
+                    WorldPt    : '10.68479;41.26906;EQ_J2000',
+                    SizeInDeg  : '.12'
+                }
+            ];
+
```
loitly: A util/helper function to create plot request, if not already exposed?
It'll be a good place to document what the options are.

src/firefly/js/api/ApiHighlevelBuild.js
```diff
@@ -0,0 +1,227 @@
...
+
+    const hApi= {};
+    const {RequestType}= llApi.util.image;
+
+    /**
+     * The general plotting function to plot a FITS image.
+     * @param targetDiv to put the image in.
+     * @param request the request object literal with the plotting parameters
+     */
+    hApi.showImage= (targetDiv, request)  => showImageInMultiViewer(llApi, targetDiv, request);
```
loitly: these functions are nicely declared and exposed here.  but, it may be hard to generate jsdoc from it.  may need to revisit when we start looking at generating jsdoc.

src/firefly/js/api/ApiUtil.js
```diff
@@ -0,0 +1,97 @@
...
+    if (!isElement(divElement)) debugMsg(`the div element ${isString(div)?div:''} is not defined in the html` );
+    if (!Component) debugMsg('Component must be defined');
+
+
+    const renderStuff= (
+        <div style={style}
+             className='rootStyle' >
+            <Component {...props} />
+        </div>
+    );
```
loitly: I wonder if we can simply add rootStyle directly into the divElement instead of creating a new one.
It's more natural for the api user to style their div directly instead of passing it into this function.

src/firefly/js/util/BrowserUtil.js
```diff
@@ -133,9 +139,11 @@ export const modifyURLToFull= function(url, rootPath) {
 };
 
 
+const hPref= ['http', 'https', '/', 'file'];
+
 const isFull= function (url) {
     url = url.toLowerCase();
-    return ['http', 'https', '/', 'file'].some( (s) => url.startsWith(s));
+    return hPref.some( (s) => url.startsWith(s));
```
loitly: I've not seen 'some' used before.  Good to know.

src/firefly/js/visualize/ui/ImageMetaDataToolbarView.jsx
```diff
@@ -93,20 +101,40 @@ export function ImageMetaDataToolbarView({activePlotId, viewerId, viewerPlotIds,
...
+            <VisInlineToolbarView
+                plotId={pv.plotId} dlAry={dlAry}
+                showLayer={lVis}
+                showExpand={true}
+                showToolbarButton={tb}
+                showDelete ={false}
+            />
+        </div>
+    );
+}
```
loitly: This should be a functional component.. You never know.  May get that 'performance' gain some day.

loitly:
vistoolbar is not displaying description correctly in api mode.
api code looks great.

src/fftools/js/fftools.js
robyww: I can. Do you want me to?

loitly: Yes.  I think you should.

src/firefly/html/demo/ffapi-highlevel-test.html
robyww: yes, I have a function that goes through them and finds the right key.  
BTW - they are not props, the are parameters to the request.

src/firefly/html/demo/ffapi-highlevel-test.html
robyww: I could expose WebPlotRequest which has many helper functions.  then it would be either a WebPlotRequest object or a object literal.  It would work, maybe more confusing.

src/firefly/js/api/ApiHighlevelBuild.js
robyww: It may. It has to be in the closure.

src/firefly/js/api/ApiUtil.js
robyww: The user could just do than.  The point of this is if they do something like flex.

src/firefly/js/visualize/VisMouseSync.js
robyww: That is true sort of.  They is a new controller that holds the readout state.  It could get updated a lot.  The VisMosueState turned out to be more a way to listen to events instead of state.

robyww:
>> Layers/Tools/Expand buttons appear on mouse enter, but do not disappear on mouse leave (unless you it enters another plot).

Anything the user specifically request, I don't autohide.  I think this is the right answer but I am open to other alternatives.


>> The next one is a question: would it make sense to use the same plot group id by default for the images shown in the same div?

I think you are right, I will do that.  It might need to be unlocked.

PR #86, opened 2016-05-15
DM-6042: Make the viewer API

robyww:
- viewer api framework and working with images
   - change image create so more happening in action creator
   - changes to take object in dispatchers and better deal with json
   - clean out a little more of the gwt code
   - Server side not can review a action and the dispatch it to a channel


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/PushCommands.java
```diff
@@ -311,14 +312,52 @@ public String doCommand(Map<String, String[]> paramMap) throws Exception {
 
         public String doCommand(Map<String, String[]> paramMap) throws Exception {
             SrvParam sp= new SrvParam(paramMap);
-            String ip= sp.getRequired(ServerParams.IP_ADDRESS);
-            boolean active= PushJob.isBrowserClientActive(ip);
+            String channel= sp.getOptional(ServerParams.CHANNEL_ID);
+            boolean active= PushJob.getBrowserClientActiveCount(channel,0)>1;
```
tgoldina: What happens if channel is null? Is eventChannel in RequestOwner is always non-null?

src/firefly/js/api/ApiHighlevelBuild.js
```diff
@@ -163,22 +163,27 @@ function buildDeprecated(llApi) {
...
+    const {findInvalidWPRKeys,confirmPlotRequest}= llApi.util.image;
+    const {renderDOM,debug}= llApi.util;
     const {MultiImageViewer, MultiViewStandardToolbar}= llApi.ui;
 
     highlevelImageInit(llApi);
-    request= confirmPlotRequest(request,globalImageViewDefParams,targetDiv);
+    const testR= Array.isArray(request) ? request : [request];
+    testR.forEach( (r) => {
+        const badList= findInvalidWPRKeys(r);
+        if (badList.length) debug(`plot request as the following bad keys: ${badList}`);
```
tgoldina: typo 'as' -> 'has'

src/firefly/js/api/ApiViewer.js
```diff
@@ -0,0 +1,187 @@
...
+//================================================================
+
+
+//================================================================
+//---------- Private Image functions
+//================================================================
+
+
+
+function plotRemoteImage(request, channel, dispatch) {
```
tgoldina: Is channel parameter a placeholder for future?

src/firefly/js/core/JsonUtils.js
```diff
@@ -84,12 +84,11 @@ export const jsonRequest= function(baseUrl, cmd, paramList, doPost) {
                 return;
             }
             response.json().then( (result) => {
-                if (result && result[0]) {
-                    if (result[0] && result[0].success && result[0].success !== 'false' && result[0].data) {
-                        resolve(result[0].data);
-
+                if (has(result,'0')) {
+                    if (Boolean(get(result,'0.success'))) {
```
tgoldina: Interesting, I would not think using index in path like this would work.
Since you check that result[0] exists, why using get here instead of 'if (result[0].success)'? Boolean(undefined) returns false anyway.

src/firefly/js/core/JsonUtils.js
```diff
@@ -84,12 +84,11 @@ export const jsonRequest= function(baseUrl, cmd, paramList, doPost) {
...
+                        resolve(result[0].data ? result[0].data : result[0]);
                     }
-                    else if (result[0] && get(result,'0.error')){//result[0].error) {
+                    else if (has(result,'0.error')){
```
tgoldina: Same here. Could be result[0].error.

src/firefly/java/edu/caltech/ipac/firefly/data/ServerParams.java
```diff
@@ -58,6 +58,7 @@
     public static final String ID = "id";
     public static final String BID = "bid";
     public static final String CHANNEL_ID = "channelID";
+    public static final String TRY_MS = "tryMS";
```
tgoldina: My guess is MS stands for milliseconds, but I had to look how it's used. We might want to document the meaning of some parameters.

src/firefly/js/api/ApiViewer.js
robyww: Channel works.  However, it should not be passed here. It is in the dispatcher.

src/firefly/java/edu/caltech/ipac/firefly/server/vispush/PushJob.java
```diff
@@ -152,9 +152,20 @@ public static boolean pushRemoveMask(String maskId) {
...
-        String channel= ServerContext.getRequestOwner().getEventChannel();
-        return ServerEventManager.getActiveQueueChannelCnt(channel)>1;
+    public static int getBrowserClientActiveCount(String channel, int tryTime) {
+        if (channel==null) channel= ServerContext.getRequestOwner().getEventChannel();
+        int cnt=  ServerEventManager.getActiveQueueChannelCnt(channel);
+        long endTry= System.currentTimeMillis()+tryTime;
+
+        try {
+            while (tryTime>0 && cnt==0 && System.currentTimeMillis()<endTry) {
+                Thread.sleep(200);
```
tgoldina: did you mean to use tryTime here?

src/firefly/js/ui/TestQueriesPanel.jsx
```diff
@@ -393,12 +393,12 @@ function doImages(request) {
 
 
     //wpr1.setAnnotationOps(AnnotationOps.TITLE_BAR);
-    dispatchPlotImage('TestImage1', wpr1);
-    dispatchPlotImage('TestImage2', wpr2);
-    dispatchPlotImage('TestImage3', wpr3);
-    dispatchPlotImage('TestImage4', wpr4);
+    dispatchPlotImage({plotId:'TestImage1', wpRequest:wpr1});
```
tgoldina: I like that you added parameter names. I think it makes it more clear.

src/firefly/html/demo/ffapi-highlevel-test.html
```diff
@@ -23,6 +23,14 @@
     <br>
 </div>
 
+<div>
+    <!--<a href='javascript:firefly.getViewer("testChannel").showImage({url:"http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band2.fits"})'>test open</a>-->
+    <a href='javascript:firefly.getViewer().plotURL("http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band2.fits")'>test open</a>
```
tgoldina: opens  new app with the image in expanded view but the search drop-down on top of it

tgoldina: if I click on the first link again, another viewer shows next to the first, but with a different zoom level

src/firefly/java/edu/caltech/ipac/firefly/server/vispush/PushJob.java
robyww: No, it is right.  I guess I do need to document it more.  `tryTime` is the about of time that is will check for.  It polls every 200 ms until success or tryTime runs out.  This way it will return quite fast.  `tryTime` is really the time until it gives up. The polling is only looking at an array that is updated in another thread.

src/firefly/html/demo/ffapi-highlevel-test.html
robyww: It is suppose to show a new one.  the other does only replaces.  I will look at the zoom level.

robyww: >> opens new app with the image in expanded view but the search drop-down on top of it

I made a ticket, DM-6108.  There are several more issues the need to be worked in controlling the tri-view layout.

loitly: re: 'test open'
Safari:  failed with msg:  Firefly: Operation fail: Could not launch a new browser tab.
Chrome:  blocked by popup blocker
Firefox: works fine.
I know the popup blocker is not a popular complains, but someone will.  If there's an easy way to move the window.open code closer to the clicked event, it should solve that problem.

src/firefly/java/edu/caltech/ipac/firefly/data/ServerParams.java
loitly: maybe naming it something like CONN_TIMEOUT or WAIT_TIME will have some implicit meaning.

src/firefly/js/api/ApiUtil.js
```diff
@@ -35,8 +35,8 @@ export const isDebug = () => get(window, 'firefly.debug', false);
  * @param {String|Error} msg any number of messages
  */
 export function debug(...msg) {
-    if (!isEmpty(msg) && isDebug()) {
-        logError(...msg);
+    if (isDebug() && !isEmpty(msg)) {
+        logErrorWithPrefix('Firefly:', ...msg);
```
loitly: this is more of a question.  is this how msg suppose to be passed?
you cannot just do msg instead of ...msg?

robyww: you can pass a single string. That is what I am doing.   However, you can also pass multiple parameters. Such as
```
debug('message 1', new Error('this is the error', {a: 'error obj', b: 'more' });
```
When it comes out it will start with 'Firefly: '

src/firefly/js/api/ApiViewer.js
```diff
@@ -0,0 +1,187 @@
...
+    var root= getCookie('usrkey');
+    if (root) {
+        const idx= root.indexOf('/');
+        if (idx>-1) root= root.substring(0,idx);
+    }
+    else {
+        root= 'remote';
+    }
+    return `${root}-viewer`;
+}
```
loitly: not sure if you are aware, WebSocketClient capture the channel information and stored the connection id and channel in a 'seinfo' cookie.  it would be trivial to expose those information through WebSocketClient.

src/firefly/java/edu/caltech/ipac/visualize/plot/WorldPt.java
```diff
@@ -56,16 +55,16 @@ public String serialize() {
     protected static WorldPt stringAryToWorldPt(String wpParts[]) {
         WorldPt retval= null;
         if (wpParts.length>=3) {
-            double lon= StringUtils.parseDouble(wpParts[0]);
-            double lat= StringUtils.parseDouble(wpParts[1]);
+            double lon= parseDouble(wpParts[0]);
+            double lat= parseDouble(wpParts[1]);
```
ejoliet: I think we still need a check on the input parameters of parseDouble. StringUtils was basically checking 'NaN' and returning Double.NaN. Why removing it?

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/PushCommands.java
robyww: I think it defaults to self. I need to confirm with loi.

src/firefly/js/core/JsonUtils.js
robyww: this does need has.  I should only do this if error is defined.

src/firefly/js/core/JsonUtils.js
robyww: Your right, I will fix it.

src/firefly/java/edu/caltech/ipac/visualize/plot/WorldPt.java
robyww: It is doing the same as before.  I move the parse double to the base class Pt.java.  I am trying to remove the dependency on StringUtils to make the new GWT cross compile smaller.

ejoliet: oh, ok! great then!

On 5/16/16 6:57 PM, Trey Roby wrote:
>
> In src/firefly/java/edu/caltech/ipac/visualize/plot/WorldPt.java 
> <https://github.com/Caltech-IPAC/firefly/pull/86#discussion_r63455053>:
>
> > @@ -56,16 +55,16 @@ public String serialize() {
> >      protected static WorldPt stringAryToWorldPt(String wpParts[]) {
> >          WorldPt retval= null;
> >          if (wpParts.length>=3) {
> > -            double lon= StringUtils.parseDouble(wpParts[0]);
> > -            double lat= StringUtils.parseDouble(wpParts[1]);
> > +            double lon= parseDouble(wpParts[0]);
> > +            double lat= parseDouble(wpParts[1]);
>
> It is doing the same as before. I move the parse double to the base 
> class Pt.java. I am trying to remove the dependency on StringUtils to 
> make the new GWT cross compile smaller.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub 
> <https://github.com/Caltech-IPAC/firefly/pull/86/files/3dddf13ead9a8b65a93254dce2ab5fcd6f9f4283#r63455053>
>

-- 
Emmanuel Joliet
Office: 170
Phone:  (626) 395-1489

PR #87, opened 2016-05-18
DM-5799: changed both the server side and client side to implement th…

lznakano:
The correct asinh algorithm is used in this implementation.  Please view this branch and let me know if any changes are needed. 

Please check 
1.  whether displaying dataMin/dataMax and beta looks OK
2.  I don't think that we need dataMin and dataMax in the list . Data alone is enough.  When users choose   data, the dataMin is the default for the lower range and the dataMax is the default for the upper range.  dataMin and dataMax really do nothing. 

Thanks!


src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1107,25 +1144,129 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
+     *    m_o=2.5log(flux_0), flux_0 is the flux of an object with magnitude 0.0.
+     *    b is an arbitrary "softening" which determines the flux level as which the liner behavior is set in, and x is the flux,
+     *
+     *
+     *   Since  mu_0 and a are constant, we can use just:
+     *     mu =  asinh( flux/(2.0*b)) ) = asin(x/beta); where beta=2*b;
+     * @param beta
+     * @return
+     */
+    private double  getASinhStretchedPixelValue(double flux, double maxFlux, double minFlux, double beta)  {
```
robyww: can this be static?

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1107,25 +1144,129 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
 
     }
-    private static double getASinhStretchedPixelValue(double x, double dr, double zp, double mp, double bp, double wp)  {
 
+    /**
+     * Calculate the mean flux value
+     * @param data
+     * @return
+     */
+     private  double getMean(double [] data ) {
```
robyww: can this be static?

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1107,25 +1144,129 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
 
-        double rd = dr * (x - zp) / mp; //in the range of 0 - dr (when zp=0.0)
-        double nsd = asinh(rd) / asinh(mp - zp); // in the range 0-1 when zp=0.0
-        double pixValue = 255 * (nsd - bp) / wp;
+    /**
+     * Calculate variance and then standard deviation
+     * @param data
+     * @return
+     */
+      private double getStdDev(double[] data) {
```
robyww: same here, static?

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1107,25 +1144,129 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
 
+          return Math.sqrt(temp/size);
+      }
 
+    /**
+     * Process the image fluxes to exclude the 0.0 and NaN and infinity values
+     * @return
+     */
+      private double[] getNoneZeroValidReadoutArray(){
+        ArrayList<Double> list= new ArrayList<>();
```
robyww: static here as well, might need to pass something

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1107,25 +1144,129 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
+     * This sigma value is calculated using the whole image data.
+     * sigma = SQRT [  ( sum (xi-x_average)^2 )/n
+     *   where x_average is the mean value of the x array
+     *   n is the total number of the element of x array
+     * @return
+     */
+    private double computeSigma() {
+
+        //get none zero and finite flux values
+        double [] validData = getNoneZeroValidReadoutArray();
```
robyww: static here as well, if you passed something

robyww:
This looks good.  There is a couple of things I want to talk to you about.

- How we can keep FitsRead as static as possible.  I see continue refactorings of FitsRead, static functions will make it easier.
- Your work made me aware that there are about 5 JS files we can delete.  I want to do that as part of this ticket.

robyww:
Little thing: "Beta" needs to be capitalized in the UI.

lznakano:
All of the methods mentioned can be static.   I need to do a minor refactoring first.  The reason I did not use static is to keep the same pattern as it is now.  With a minor refactoring, I can change all those method to static.   

I will change the UI using Beta. 

Thanks!

robyww:
Another UI thing.  The decimal places need to be cut down in Beta.  I don't know what is the right amount but it seems like no more than 4 maybe 2.

robyww:
Sorry, I keep finding little ui stuff.  The validation range is wrong.  It is set from 0 to 100.  I don't know what is correct but on my images the default is more than 100.

robyww:
The validation and 'Beta' look good but you still need to get the decimal place number right.  I don't know how many but it must just be a couple (maybe no decimal places)

lznakano:
Trey,

For beta, I cut it down to 2 as you suggested.  Should we do the same for dataMin and dataMax?  Do you agree to remove Data Min and Data Max from the list?   

Lijun

robyww:
This is the problem.
14d

PR #90, opened 2016-05-24
DM-6139: server call for region on images

robyww:
This is the new server call for making a hard copy for image with the drawing layers


cwang2016:
Sounds good for making PNG hard copy.

PR #91, opened 2016-05-25
create PNG for regions from drawing layers on image.

cwang2016:


cwang2016:
The png file is created for regions on images from all selected drawing layers based on the server call done by DM-6142.

PR #92, opened 2016-05-25
DM 5190 & DM 5189: chart and table api

loitly:
- Table API including external viewer.
- Chart API.  No external API, yet.
- lots refactoring
Test url:  http://localhost:8080/firefly/demo/ffapi-highlevel-test.html

Require server-side build.


src/firefly/js/api/ApiUtil.js
```diff
@@ -73,6 +72,19 @@ export function renderDOM(div, Component, props, style={width:'100%', height:'10
...
+    if (!isElement(divElement)) debug(`the div element ${isString(div)?div:''} is not defined in the html` );
+    if (!Component) debug('Component must be defined');
+
+
+    const renderStuff= (
+            <Component {...props} />
+    );
+
+    ReactDOM.render(renderStuff,divElement);
+}
```
loitly: Is there something different you want to do here that's not in renderDOM?  If it can be applied to renderDOM, it should be combine into it.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -527,7 +589,7 @@ export class ChartsTableViewPanel extends Component {
     }
 
     getNextState() {
-        const tblId = TblUtil.getActiveTableId();
+        const tblId = this.props.tblId || TblUtil.getActiveTableId();
```
loitly: this may be an issue.  TblUtil.getActiveTableId() returns the active table for 'main' group.  if this chart is connected to a different group, it may switches to an unintended table.

src/firefly/js/visualize/Histogram.jsx
```diff
@@ -7,13 +7,15 @@ import ReactHighcharts from 'react-highcharts/bundle/highcharts';
 import numeral from 'numeral';
 import {getFormatString} from '../util/MathUtil.js';
 
+
 var Histogram = React.createClass(
```
loitly: should convert to es6 style.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -149,7 +149,7 @@ class ChartsPanel extends React.Component {
                 let popupTitle = 'Chart Options';
 
                 const reqTitle = get(tableModel, 'tableMeta.title');
-                if (reqTitle) { popupTitle += ': '+reqTitle; }
+                if (reqTitle) { popupTitle += ': '+decodeURIComponent(reqTitle); }
```
loitly: decoding here is not necessary.  it should be fixed already.

src/firefly/js/api/ApiUtilChart.jsx
```diff
@@ -0,0 +1,35 @@
...
+//import * as HistogramCntlr from '../visualize/HistogramCntlr.js';
+//import * as TableStatsCntlr from '../visualize/TableStatsCntlr.js';
+
+export {uniqueChartId} from '../visualize/ChartUtil.js';
+
+export function loadPlotDataForTbl(tblId, chartId, xyPlotParams) {
+    dispatchAddSaga(getChart, {tblId, chartId, xyPlotParams});
+}
+
+export function* getChart({tblId, chartId, xyPlotParams}) {
```
loitly: there is already a saga for watching when a table is loaded.  i'll see if it can be used in place of this.  any objection?

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -130,33 +139,28 @@ class ChartsPanel extends React.Component {
...
-            const {tableModel, tblStatsData} = this.props;
             if (optionsShown) {
+                const {tableModel, tblStatsData, tblId, chartId} = this.props;
                 // show options popup
-                const popupId = getChartOptionsPopupId(tblId);
                 let popupTitle = 'Chart Options';
 
                 const reqTitle = get(tableModel, 'tableMeta.title');
-                if (reqTitle) { popupTitle += ': '+decodeURIComponent(reqTitle); }
+                if (reqTitle) { popupTitle += ': '+chartId+' '+decodeURIComponent(reqTitle); }
```
loitly: same decoding comment applies here.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -130,33 +139,28 @@ class ChartsPanel extends React.Component {
...
 
                 var popup = (
                     <PopupPanel title={popupTitle} closeCallback={()=>{this.toggleOptions();}}>
                         <div
                             style={{overflow:'auto',width:OPTIONS_WIDTH,height:300,paddingTop:10,paddingLeft:10,verticalAlign:'top'}}>
-                            {this.renderChartSelection()}
-                            <OptionsWrapper {...{tblId, tableModel, tblStatsData, chartType}}/>
+                            {chartId===tblId ? this.renderChartSelection() : false}
```
loitly: chartId is same as tblId?  I thought chartId has an added prefix?

src/firefly/js/visualize/ChartsTableViewPanel.jsx
tgoldina: This is not the latest code. Now I have

    getNextState() {
        var {tblId, chartId} = this.props;
        tblId = tblId || chartId ? getTblIdForChartId(chartId) : TblUtil.getActiveTableId();
        chartId = this.props.chartId || tblId;
        const tableModel = TblUtil.getTblById(tblId);

It covers several situations: 
1. tblId is passed as a property: the chart is only connected to a particular table
2. chartId is passed as a property: tblId for a given chart is set in the store (this is the api case when chart is connected to the active table of a given group).
3. both tblId and chartId are empty: current firefly app scenario, chart is connected to the active table of the main group.

src/firefly/js/visualize/Histogram.jsx
tgoldina: Yes, I have created a DM-6287 for refactoring. I will add converting to es6 classes to it.

src/firefly/js/api/ApiHighlevelBuild.js
```diff
@@ -37,20 +37,130 @@ function build(llApi) {
...
+     * @prop {string}  xUnit        unit for x axis
+     * @prop {string}  yUnit        unit for y axis
+     * @prop {string}  xOptions     comma separated list of x axis options: grid,flip,log
+     * @prop {string}  yOptions     comma separated list of y axis options: grid,flip,log
+     */
+
+    /**
+     * The general plotting function to plot an XY Plot.
+     * @param {string|HTMLDivElement} targetDiv to put the chart in.
+     * @param {XYPlotOptions} parameters the request object literal with the chart parameters
```
loitly: thanks for using @typedef in consistent with table api.  if this does not work out, at least we can address them in the same manner.

src/firefly/js/api/ApiUtilChart.jsx
tgoldina: I think you are right. Just need XYPlotCntlr.LOAD_PLOT_DATA action creator to check if the table is fully loaded. If not, just update the store and  do nothing: syncCharts saga will take care of the server call.

src/firefly/js/api/ApiUtilImage.jsx
```diff
@@ -47,40 +47,40 @@ export {ExpandType} from '../visualize/ImagePlotCntlr.js';
...
+    //     if (vr.expandedMode!==ExpandType.COLLAPSE && !isExpanded) {
+    //         isExpanded= true;
+    //         expandedDivEl.style.visibility= 'visible';
+    //         renderDOM(expandedDivEl, Component, Object.assign({closeFunc:unExpand}, props),
+    //             { left: 1, right: 1, top: 1, bottom: 1, position: 'absolute' });
+    //     }
+    // });
+    //
     
     dispatchApiToolsView(true);//todo move this somewhere, not sure where
```
loitly: @robyww , this is the line I mentioned to you before.  it should be moved to somewhere else so we can remove this function.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
tgoldina: It's only for the current firefly app scenario: one chart per table. In future this scenario might go away, but I was not sure I did not want to break it now. It's special, because in this case we display a choice between Scatter Plot and Histogram, which we don't give if the chart type is known (api case).

src/firefly/js/visualize/ChartsContainer.jsx
```diff
@@ -2,24 +2,41 @@
...
-        const expandedMode = getExpandedMode() === LO_EXPANDED.xyPlots.view;
-        this.state = {
-            expandedMode: props.expandedMode,
-            closeable: props.closeable
-        };
+        this.state = nextState(this.props);
+    }
+
+    componentWillReceiveProps(np) {
+        if (!this.isUnmounted && !shallowequal(this.props, np)) {
```
loitly: this.isUnmounted is used here and a few lines below.  however, it was never set.  should be set in componentWillUnmount.

src/firefly/html/demo/ffapi-highlevel-test.html
```diff
@@ -121,6 +141,53 @@
...
             v2.plotURL('http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band2.fits');
+
+
+            //-----------------  TABLE DEMO ------------------------//
+            tblReq1 = firefly.util.table.makeIrsaCatalogRequest('wise_allwise_p3as_psd', 'WISE', 'wise_allwise_p3as_psd', null,
+                    {   position: '10.68479;41.26906;EQ_J2000;m31;ned',
+                        SearchMethod: 'Cone',
+                        radius: 300
+                    });
+            tblReq2 = firefly.util.table.makeIrsaCatalogRequest('another wise catalog', 'WISE', 'wise_allwise_p3as_psd', null,
```
tgoldina: I wanted to ask long ago: when do we use alt_source?

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -42,35 +47,45 @@ import LOADING from 'html/images/gxt/loading.gif';
...
 
         const normal = (size) => {
             if (size) {
                 var widthPx = size.width - 10;
-                var heightPx = size.height - 30;
+                var heightPx = size.height;
                 //console.log('width: '+widthPx+', height: '+heightPx);
-                this.setState({widthPx, heightPx});
+                if (widthPx !== this.state.widthPx || heightPx != this.state.heightPx) {
+                    this.setState({widthPx, heightPx, immediateResize: false});
```
loitly: In general, when debounced is used, you should check for isUnmounted condition before calling setState().  If not, there's a chance setState will be called after component unmounted.

src/firefly/java/edu/caltech/ipac/firefly/server/util/QueryUtil.java
```diff
@@ -95,6 +96,38 @@ public static TableServerRequest convertToServerRequest(HttpServletRequest req)
         return retval;
     }
 
+    public static TableServerRequest convertToServerRequest(String str) {
```
tgoldina: What is the difference between QueryUtil.convertToServerRequest and TableServerRequest.parse? When should we use each?

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -438,58 +510,61 @@ class ChartsPanel extends React.Component {
...
         if (optionsShown) {
-            return (
-                <div style={{display:'inline-block',overflow:'auto',width:(OPTIONS_WIDTH-20),height:heightPx,paddingLeft:10,verticalAlign:'top'}}>
-                    {this.renderChartSelection()}
-                    <OptionsWrapper  {...{tblId, tableModel, tblStatsData, chartType}}/>
-                </div>
-            );
-        } else {
-            return undefined;
+            if (!optionsPopup) {
```
loitly: nitpicking.. this if condition can be easily combined with previous if statement.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -501,6 +576,11 @@ ChartsPanel.defaultProps = {
     expandable: true
 };
 
+function getTblIdForChartId(chartId) {
+    return  get(flux.getState()[XYPlotCntlr.XYPLOT_DATA_KEY], [chartId, 'tblId']) ||
+            get(flux.getState()[HistogramCntlr.HISTOGRAM_DATA_KEY], [chartId, 'tblId']);
+}
```
loitly: this function should be in ChartUtil.  if there is not a ChartUtil, it may be a good idea to create one.  i am sure there will be more of these functions in the future when more chart related features are added.

src/firefly/js/visualize/Histogram.jsx
```diff
@@ -27,16 +29,55 @@ var Histogram = React.createClass(
...
-                source: undefined,
-                height: 300,
-                logs: undefined,
-                reversed: undefined,
                 desc: 'Sample Distribution',
                 binColor: '#d1d1d1'
             };
         },
 
+        shouldComponentUpdate(nextProps) {
```
loitly: this is purely my opinion without studying the code.
there is a lot of logic in this check.  it would be helpful if there is a short comments outlining when this component should update.

src/firefly/js/visualize/HistogramOptions.jsx
```diff
@@ -75,7 +75,7 @@ var HistogramOptions = React.createClass({
                 <ValidationField
                     style={{width: 30}}
                     initialState= {{
-                        value: 0.05,
+                        value: get(histogramParams, 'falsePositiveRate')||'0.05',
```
loitly: fyi, get takes a default value as a 3rd parameter.  
i.e.  get(histogramParams, 'falsePositiveRate', '0.05')
just sharing info.  no change is needed.  it's just as good the way you have it.

src/firefly/js/visualize/TableStatsCntlr.js
```diff
@@ -76,14 +77,17 @@ export function reducer(state=getInitState(), action={}) {
...
         }
         case (TablesCntlr.TABLE_REMOVE)  :
         {
-            const {tbl_id} = action.payload.tbl_id;
+            const {tbl_id} = action.payload;
             if (has(state, tbl_id)) {
                 const newState = Object.assign({}, state);
-                Reflect.deleteProperty(newState, [tbl_id]);
+                Reflect.deleteProperty(newState, tbl_id);
                 return newState;
```
loitly: the above 3 lines can be replaced with
return omit(state, [tbl_id]);

src/firefly/html/demo/ffapi-highlevel-test.html
loitly: alt_source will be used if source fail to load.  one use-case is to put a local path to a file in source, and a url to the same file in alt_source.  so, if file is accessible by firefly, it saves the overhead of retrieving the file from a url.

src/firefly/java/edu/caltech/ipac/firefly/server/util/QueryUtil.java
loitly: I am glad you ask.  QueryUtil uses URLDecoder.decode to decode nested parameter values.  TableServerRequest is used on both the client by gwt and the server. gwt cannot use URLDecoder.  Therefore, a simple decoder function is used there.  When we no longer need to support gwt, I plan to move the parsing logic back into TableServerRequest.

src/firefly/js/visualize/Histogram.jsx
tgoldina: What I am trying to do here is to update the existing chart rather then
re-render:  true is returned by a single line in the beginning. But I have
added a comment.

2016-05-26 10:30 GMT-07:00 loitly <notifications@github.com>:

src/firefly/js/tables/TableUtil.js
```diff
@@ -101,54 +200,64 @@ export function isTblDataAvail(startIdx, endIdx, tableModel) {
...
-export function findTblById(id) {
+export function getTblById(id) {
     return get(flux.getState(),[TblCntlr.TABLE_SPACE_PATH, 'data', id]);
 }
 
 /**
- * return the table results area
+ * return the group information
+ * @param {string} tbl_group    the group to look for
  * @returns {*}
```
tgoldina: Is it useful to say the function returns whatever you want?

loitly: you are right.  but, it's the default intelliJ produces.  so, it's everywhere.  i will slowly update the docs as i come across them.

robyww:
I think the expanded might look at little better if you have it 3-5 pixels on the top and left

robyww:
I think the charts need a simple border in api mode

src/firefly/js/api/ApiUtilImage.jsx
robyww: we need a highLevelApiInit which does the following:
```
dispatchApiToolsView(true)
call image init
call table init (probably a noop)
call chart init (probably a noop)
```

robyww:
OK. I give up! It is too much to review.  You guys win the prize.  As far as I can tell is all looks good. The code is where I expected it. The only design issue we have it how to init (which I punted on when I did the initial work). 

You guys did a ton of work.

tgoldina:
Regarding chart border, do you mean the thinner border around chart or the border around chart and toolbar? (I am planning to update options, so that they go over the chart, not at the side.)

robyww:
>> Regarding chart border, do you mean the thinner border around chart or the border around chart and toolbar? (I am planning to update options, so that they go over the chart, not at the side.)

my suggestions: 
 - border around it all: `border: 1px solid rgba(0,0,0, .2);`
 - with of chart match width off container, not left padding (it is on the div under the resizer)
 
Don't try to do that before you leave.

PR #95, opened 2016-06-02
DM-6437: Convert GWT java algorithmic to JavaScript

robyww:
- Conversion of the coordinate system routines
 - convert the GWT projection to JavaScript
 - Server now sends projection info in JSON as well, old way not yet removed
 - VisUtil.USE_GWT will control if the new is used or the old.


lznakano:
The conversions look OK to me since majority codes are just pure algorithm.  Whether the conversion works the same as the Java version is unknown until the test harness is in place.

PR #96, opened 2016-06-03
implement functions for region controls from API to create region layer, delete region layer, add region entry and remove region entry

cwang2016:
4 major region control functions are implemented:
- create region layer based on  region description strings or a uploaded region file.  
- delete region layer
- add region(s) to a selected region layer
- remove region(s) from a  selected region layer

You may try those region control functions from the page of "ffapi-highlevel-test.html".

### DIFF
--- a/src/firefly/js/visualize/DrawLayerCntlr.js
+++ b/src/firefly/js/visualize/DrawLayerCntlr.js
@@ -8,35 +8,42 @@ import ImagePlotCntlr, {visRoot}  from './ImagePlotCntlr.js';
 import DrawLayerReducer from './reducer/DrawLayerReducer.js';
 import {without,union,omit,isEmpty} from 'lodash';
 
+
 export {selectAreaEndActionCreator} from '../drawingLayers/SelectArea.js';
 export {distanceToolEndActionCreator} from '../drawingLayers/DistanceTool.js';
 
-const RETRIEVE_DATA= 'DrawLayerCntlr.retrieveData';
-const CREATE_DRAWING_LAYER= 'DrawLayerCntlr.createDrawLayer';
-const DESTROY_DRAWING_LAYER= 'DrawLayerCntlr.destroyDrawLayer';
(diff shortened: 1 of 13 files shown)


src/firefly/js/visualize/DrawLayerCntlr.js
```diff
@@ -217,6 +226,48 @@ export function dispatchDetachLayerFromPlot(id,plotId, detachPlotGroup=false,
 }
 
 
+export function dispatchCreateRegionLayer(regionId, layerTitle, fileOnServer, rgAry, dispatcher = flux.process) {
+    if (!fileOnServer) {
+        fileOnServer = '';
+    }
+    if (!rgAry) {
+        rgAry = [];
+    }
```
robyww: you can use default parameters for `fileOnServer` and `rgAry`.
```
export function dispatchCreateRegionLayer(regionId, layerTitle, fileOnServer='', rgAry=[], dispatcher = flux.process)
```

src/firefly/js/visualize/DrawLayerCntlr.js
```diff
@@ -217,6 +226,48 @@ export function dispatchDetachLayerFromPlot(id,plotId, detachPlotGroup=false,
...
+
+    var regionAry =  rgAry && (Array.isArray(rgAry) ? rgAry : [rgAry]);
+
```
robyww: lodash: isArray

robyww:
I am concerned you might be doing too much in the dispatch functions that should be done in the action creators.  This may make the python code have to work too hard later.  We should talk about it.

robyww:
I don't see in the demo page where you are adding region locally, just on the remote tab.  Am I missing it?

robyww:
RegionCntlr.js should be call something else. We only call things *Cntlr.js when they define reducers for the store.  This is really part of the drawing layer reducer. Could be cal RegionTask.js, we can talk about it more on Monday.

robyww:
It is looking good.  There are a few little think I would like us to talk about and do before you merge.  Let's talk on Monday morning.

PR #99, opened 2016-06-08
DM-6026: distinguish table sort from filter.

loitly:
created TABLE_SORT action to distinguish sorting from filtering.  sorting should not reload xyplot nor catalog overlay.

Also:
- disable history when in api mode.
- ensure tableMeta.source reflects the file on the server.
- fix TablePanelOptions not resetting columns selection.
- remove 'Fits Data' tab when no images available.
- fix 'Coverage' appearing when it should.


cwang2016:
Done with review. Looks good to me.

tgoldina:
I am sorry to raise it again, but how do I add a side-effect on TABLE_SORT action?
(I wanted to add it to my syncCharts saga to make sure xy plot data are updated on sort.)

PR #103, opened 2016-06-16
DM-6542: Prevent external viewer from popup blockers.

loitly:
ticket: https://jira.lsstcorp.org/browse/DM-6542

to test:
goto http://localhost:8080/firefly/demo/ffapi-highlevel-test.html
at top of page, click "test open".  a new tab(external viewer) should popup.
cmd+t(open new tab), goto http://localhost:8080/firefly/demo/ffapi-highlevel-test.html
click "test again", external viewer's title will flash "Updated" to indicate content of that page has been updated.
switch over to external viewer tab to see another image plotted.

at no point in this test should you be prompted by the browser's popup blocker.


cwang2016:
Both review and test are done. Looks good.

PR #105, opened 2016-06-16
DM-5763: XYPlot decimation options

tgoldina:
- Added decimation options to XYPlot options panel when numpoints exceed decimation limit of 5000
- Changed selected series to be on top of data series to speed up selection change in Firefox
- Added legend to decimated plot
- Updated syncCharts saga to ignore sort action when refetching table statistics and histogram data

### DIFF
--- a/src/firefly/js/visualize/XYPlot.jsx
+++ b/src/firefly/js/visualize/XYPlot.jsx
@@ -1,7 +1,7 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import {isUndefined, debounce} from 'lodash';
+import {isUndefined, debounce, get} from 'lodash';
 import shallowequal from 'shallowequal';
 import React, {PropTypes} from 'react';
 import ReactHighcharts from 'react-highcharts/bundle/highcharts';
@@ -15,10 +15,7 @@ export const axisParamsShape = PropTypes.shape({
     columnOrExpr : PropTypes.string,
     label : PropTypes.string,
     unit : PropTypes.string,
-    options : PropTypes.string, // ex. 'grid,log,flip'
-    nbins : PropTypes.number,
-    min : PropTypes.number,
-    max : PropTypes.number
+    options : PropTypes.string // ex. 'grid,log,flip'
 });
 
 export const selectionShape = PropTypes.shape({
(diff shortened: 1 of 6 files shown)


loitly:
- clicking ‘ok’ on the options should show validation errors on required fields, i.e. X, Y.  instead of doing nothing.

- just an idea.. when it’s decimated, you cannot ‘select’ multiple points.  Yet, there is no indication that is normal.  so, instead of show/hide the icons, gray it out when it’s not applicable.  and in the decimation case, you can specifically say that it’s not supported.

- some code is modifying the table’s results after filtering.  i’ve seen this when tbl_id is not removed from request when doing a new search.  may not be xyplot problem.  could be image. 

- xyPlot selected points is off.  i plot ra/dec.  sorted table by descending ra.  zoom in plot on ra < 19.5.  highlight a row where ra > 150.. nothing show(good).  then selected the same row.. a green dot shows(bad).

- when X/Y ratio is large.. plot overflow.. but, have no way to scroll.
mouse drag to select an area.  no way to unselect it.  i assume a click will unselect it.  but, it does not.

- set x-bin and y-bin to 500, legend no longer show the color, only label.

src/firefly/js/visualize/XYPlot.jsx
```diff
@@ -216,6 +235,8 @@ export class XYPlot extends React.Component {
                     chart.get(HIGHLIGHTED).setData(highlightedData);
                 }
 
+
+                // plot parameters change
```
loitly: adding simple comments like these in this function really helps.  it was hard to read without it.

src/firefly/js/visualize/saga/ChartsSync.js
```diff
@@ -34,9 +34,21 @@ export function* syncCharts() {
...
             case TablesCntlr.TABLE_NEW_LOADED:
                 const {tbl_id} = action.payload;
-                tableStatsState = flux.getState()[TableStatsCntlr.TBLSTATS_DATA_KEY];
-                if (has(tableStatsState, tbl_id)) {
-                    TableStatsCntlr.dispatchLoadTblStats(request);
+
+                // table statistics and histogram data do not change on table sort
+                if (action.type !== TablesCntlr.TABLE_SORT) {
+                    tableStatsState = flux.getState()[TableStatsCntlr.TBLSTATS_DATA_KEY];
+                    if (has(tableStatsState, tbl_id)) {
```
loitly: just a comment on lodash's 'has'.  it will return true as long as tbl_id property is defined.  its value can be anything, including null, undefined, or ''.  in this case, i think a 'get' would be more precise.

tgoldina: I'd like to leave it as is: presence of tbl_id as an attribute indicates that we'd like to watch this table. tbl_id is omitted cleared when table is removed. If leftovers are present it indicates a problem.

loitly:
Looks good.

PR #107, opened 2016-06-20
firefly api related issues due to irsa integration.

loitly:
- default table cell to render html content as html. 
 - fix firefly_loader.js to take into consideration the location of from where it is loaded.
 - paging bar show non-ascii characters in irsa's html

### DIFF
--- a/src/firefly/js/tables/ui/TableRenderer.js
+++ b/src/firefly/js/tables/ui/TableRenderer.js
@@ -170,9 +170,10 @@ export class TextCell extends React.Component {
     //
     render() {
         var val = getValue(this.props);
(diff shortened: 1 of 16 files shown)


src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -168,10 +169,10 @@ export class TextCell extends React.Component {
     // }
     //
     render() {
+        var val = getValue(this.props);
+        val = (val.search(html_regex) >= 0) ? <div dangerouslySetInnerHTML={{__html: val}} /> : val;
```
wmiipac: haha, react actually call this prop "dangerous"

loitly:
made another commit based on updated ticket and pushed.  ready for review, again.

PR #111, opened 2016-06-28
DM-6135: vo search simplified panel

ejoliet:
Please, review this branch and let me know. VO search panel will be found in the last right-hand tab on the 'Catalog' search dropdown panel and can be tested by using VO URL value as:  
http://vizier.u-strasbg.fr/viz-bin/votable/-A?-source=J/A+A/402/549
and 'ngc3293' as target.

### DIFF
--- a/src/firefly/js/ui/VoSearchPanel.jsx
+++ b/src/firefly/js/ui/VoSearchPanel.jsx
@@ -2,14 +2,16 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import React, {Component, PropTypes} from 'react';
+import React from 'react';
 import {TargetPanel} from '../ui/TargetPanel.jsx';
 import {SizeInputFields, sizeFromDeg} from './SizeInputField.jsx';
+import {ValidationField} from './ValidationField.jsx';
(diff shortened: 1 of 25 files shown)


src/firefly/js/ui/VoSearchPanel.jsx
```diff
@@ -79,8 +85,27 @@ var sizeArea = () => {
...
 };
 
-function voSearchArea() {
+var voSearchArea = () => {
     return (
-        <div>Search registry box, table result and cone search field here!</div>
+        <ValidationField
+            fieldKey={'vourl'}
+            /*validator={urlValidator}*/
+            tooltip='Enter the VO cone search URL directly (or use the link below to open external NVO search and find the VO cone search URL)'
```
cwang2016: please check if this tooltip is shown?

ejoliet: You are right, it doesn't. I will fix that. Need to be in a 'initialState' object to be set properly.

cwang2016:
review is done. Looks good.

tgoldina:
Just one more comment: VO URL field should show validation errors. When search is clicked and the VO URL field is invalid or empty, nothing happens. There should be a red validation error mark just like for empty target. (You can set nullAllowed=false and validator in initial state.)

PR #112, opened 2016-06-28
DM-6726 Default chart and other optimizations

tgoldina:
- Removed chart selection from chart area
- Dropdown for chart selection (can be omitted if single chart type is used)
- Populate current values in chart options
- Support Clear and Reset in chart options
- For tables, connected to charts, if no default parameters are specified, default chart id XY plot with center columns for catalogs or two first numeric columns for other tables.
- Label, matching column expression, and unit, matching table model, are default parameters for both app and api now.
- Cleaned chart API a bit

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-test.html
+++ b/src/firefly/html/demo/ffapi-highlevel-test.html
@@ -196,8 +196,8 @@
             firefly.showTable('tables-2', tblReq2, {removable: false, showUnits: true, showFilters: true});
             //-----------------  CHART DEMO ------------------------//
             //--- using QUERY_ID: 'upload' does not work getActiveTableId('upload') returns undefined ---//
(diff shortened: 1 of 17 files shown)


cwang2016:
Review is done. Looks good. 
Just wondering, for Unit entry (on X Label/Unit/Options), is there an allowed list for the unit? if there is, would that be better to let the user to pick the unit from a list? if there is some design concern, just ignore my sudden thought.

tgoldina:
If left empty, the unit should be set to the unit of the corresponding
column. But we want to give user flexibility to enter whatever he likes,
especially when expressions are involved.

2016-06-28 14:31 GMT-07:00 Cindy Wang <notifications@github.com>:

ejoliet:
Expanded mode doesn't work (Chrome and Firefox tested). It shows histogram option instead (?). Should be XY plot in full size i guess. Please check.

PR #113, opened 2016-07-06
DM-6022 Lazy load related chart data on table data update

tgoldina:
- Introduced CHART_MOUNTED/CHART_UNMOUNTED actions, deleted SETUP_TBL_TRACKING.
- Reworked the logic for chart related controllers. Chart data are updated only if a chart is mounted, no updates on unmounted charts.
- Now checking if connected table or server parameters have changed before placing server request for data
- Fixed setState calls on unmounted component in ChartContainer.
- Fixed delete button being shown in expanded mode.

### DIFF
--- a/src/firefly/js/visualize/ChartsTableViewPanel.jsx
+++ b/src/firefly/js/visualize/ChartsTableViewPanel.jsx
@@ -67,7 +67,7 @@ class ChartsPanel extends React.Component {
                 var widthPx = size.width;
                 var heightPx = size.height;
                 //console.log('width: '+widthPx+', height: '+heightPx);
-                if (widthPx !== this.state.widthPx || heightPx != this.state.heightPx) {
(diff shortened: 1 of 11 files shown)


src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -147,21 +134,28 @@ class ChartsPanel extends React.Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        this.setState({chartType: getChartType(nextProps)});
+        const {tblId, chartId, chartType} = nextProps;
+        if (!tblId || !chartId) { return; }
+        if (chartId != this.props.chartId || chartType != this.props.chartType || !this.props.tblId) {
```
cwang2016: just picky, !==  instead of '!='

cwang2016:
review is done. looks good!

PR #114, opened 2016-07-08
DM-5761_XYPlot should support selecting columns from a table

wmiipac:
Add XYPlot column select popup
 Added description and type to table statistics


src/firefly/js/ui/TextButton.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.ff-MenuItem-linkButton .menuItemText {
+    font-size: 10pt;
+    line-height: 14pt;
+}
+
+.ff-MenuItem-linkButton .menuItemText:HOVER {
+    background-color: rgba(255,255,255,.97);
+}
```
tgoldina: I think, this css file has noting to do with the TextButton and needs to be deleted.

wmiipac: deleted

tgoldina:
I think, it's ready to merge.

PR #115, opened 2016-07-11
DM-6288 Chart Options Display

tgoldina:
This ticket is about making chart options (and chart) display more consistent with that of the table. Chart Options will be displayed over the right side of the table. Overall look of the chart is more similar to the table.

### DIFF
--- a/src/firefly/js/visualize/ChartsTableViewPanel.jsx
+++ b/src/firefly/js/visualize/ChartsTableViewPanel.jsx
@@ -30,10 +30,6 @@ import {XYPlot} from '../visualize/XYPlot.jsx';
 import HistogramOptions from '../visualize/HistogramOptions.jsx';
 import Histogram from '../visualize/Histogram.jsx';
 
-import {PopupPanel} from '../ui/PopupPanel.jsx';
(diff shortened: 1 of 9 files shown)


ejoliet:
I didn't see any problem whatsoever. Looks good.

src/firefly/js/visualize/ChartsTableViewPanel.jsx
```diff
@@ -47,19 +43,34 @@ import FILTER_IN from 'html/images/icons-2014/24x24_FilterAdd.png';
...
+    padding: '0 5px',
+    top: 1,
+    right: 1,
+    bottom: 2,
+    backgroundColor: 'beige',
+    borderLeft: '1px solid #CCC',
+    zIndex: 100,
+    flex: '0 0 auto',
+    overflow:'auto',
+    boxShadow: '-2px 0 2px 0 rgba(0,0,0,0.50)'
```
loitly: i see lots of inline styling in this file.  having them in a css file may look cleaner.

loitly:
Only in Safari when scrollbar is visible, it cuts into the Reset button.
Otherwise, it looks great!

PR #117, opened 2016-07-14
DM-6516-footprint footprint tool implementation. Integrating with the  implementation of marker tool. Both are made in consistent style.

cwang2016:
Implementation: (both footprint tool and marker tool are made in consistent style)   
- footprint description file - in DS9 region format, footprint description is located at the server side.
-  update the service to load the footprint description from the server. 
-  create footprint factory to produce regions and drawing objects based on footprint description. 
-  create  drawing layer for footprint.
-  create functions to handle the operation on footprint (or marker) on the drawing layer - rotate, resize, or relocate or translation.
-  handle the cases that the footprint coverage is bigger than the image size or the footprint's position is out of image covered area.          
- UI components for drawing layer panel

Notes for operation: 
Rotate handle, resize handle and outline box are provided for the user to rotate, resize and relocate the footprint or marker on the U if any of the operations is applied.
- Outline box is initially an upright rectangle which
    a) encloses all footprint regions or
    b) is around the center of the footprint (marked by a cross symbol on the UI) or the marker if a) is not visible or 
    c) is around the plot center if none of a) and  b) is visible.  
    The outline box becomes slanted as the footprint is rotated. 
- Resize handles appear at the corners of the outline box. (marker is resizable and footprint is not resizable).
- Rotate handle appears only at one side of the outline box, starting from the right, the top, the left, or the bottom side  according to its visibility on the plot.
  
Testing example,
   1. get image from image search panel (Images), target: m51, select WISE and click search button
   2. click icon for "overlay marker and instrument footprints"
      and select "add marker" or any item for adding footprint from the dropdown list
      => a marker or the footprint will appear in the middle of the plot
   3. click the mouse at the place where you want to locate the footprint           

   When the mouse hovers over the footprint, the cursor will change to be 'pointer'.
   Once 'pointer' is clicked down, an outline box with either resize or rotate handle will appear for 3 seconds after the mouse is up.
  
   4. Click the resize handle and drag to resize the marker,  or 
       click the rotate handle and drag to rotate the footprint or
       click on the pointer and drag to move the marker or footprint. 

   If the outline box is moved out of the plot area  while operating the footprint or marker, the outline box of other kind will continue to appear.


cwang2016:
Note for compiling: there are some changes made on Java codes. Please make full compilation before running it.

ejoliet:
First comment i have: WOW!!! Impressive! This is way better from before migration! It is now very useful! There are a couple of issues:

-  Angle input field in layer window seems to be in radians, should be in degree and i couldn't enter any value either.
- For the case of a full footprint, sometime the box is out side the viewport but no handle to rotate is visible? Can we make the handle always visible and not necessarily on the right? Maybe 4 handles on each side of the box?
- I think the idea was to remove the link "Add" from the layer window as we can keep adding any overlay from the menu directly?
- PNG image saved shows the overlay with different color than the one displayed 
- Handle appears always on the right hand side but it should be visible in the viewport always

cwang2016:
Additional feature
- the footprint is displayed and operated on multiple images simultaneously.

cwang2016:
Notes:

- Angle entry bug in the drawing layer panel is corrected. 
- An outline box around the center of the plot will continue to appear once the footprint is outside the viewport. 
- The rotate handle is shown on one side which is the right, the top, the left or the bottom side, depending the visibility on any of the sides.

ejoliet:
A second review after last commit revealed the following problem:
- title in layer dialog not same as before, format should be: 'Show: Footprint #N: TELESCOPE APERTURE/INSTRUMENT' instead of the one currently in this branch: 'Show: TELESCOPE_APERTURE_#N'
- JWST MIRI footprint region is actually the sum of JWST MIRI and HST WFPC2. Please, check that footprint regions migrated contain and match the shapes defined in FootprintFactory class.
- When more than 1 footprint is displayed, ones cannot select the footprint unless is visually separate (i think is taking into account the box surrounding the footprint as surface to be selected)
- When rotate 'NICMOS' footprint, the shape lines seems to be moving, maybe a precision problem in using screen to world point and viceversa?

src/firefly/java/edu/caltech/ipac/visualize/resources/Footprint_JWST.reg
```diff
@@ -0,0 +1,67 @@
...
+POLYGON   -0.13937751  -0.08964411  -0.14094694  -0.08941355  -0.14077194  -0.08770800  -0.13922195  -0.08790245  -0.13937751  -0.08964411 # color=yellow tag={MIRI}
+POLYGON   -0.13938306  -0.08969967  -0.14096360  -0.08946632  -0.14076916  -0.08776356  -0.13922195  -0.08795800  -0.13938306  -0.08969967 # color=yellow tag={MIRI}
+POLYGON   -0.13938028  -0.08969967  -0.14092749  -0.08946632  -0.14073861  -0.08776356  -0.13921362  -0.08795800  -0.13938028  -0.08969967 # color=yellow tag={MIRI}
+POLYGON   -0.13901084  -0.09011910  -0.14103860  -0.08987188  -0.14059694  -0.08768578  -0.13863306  -0.08800800  -0.13901084  -0.09011910 # color=yellow tag={MIRI}
+POLYGON   -0.13901917  -0.09009966  -0.14104138  -0.08985798  -0.14057749  -0.08767189  -0.13861639  -0.08798856  -0.13901917  -0.09009966 # color=yellow tag={MIRI}
+POLYGON   -0.13902472  -0.09008299  -0.14099972  -0.08982187  -0.14058027  -0.08765244  -0.13862751  -0.08797467  -0.13902472  -0.09008299 # color=yellow tag={MIRI}
+POLYGON  360.00148056  -0.00349167 359.98601112  -0.01929166 359.97019723  -0.00381111 359.98566389   0.01199167 360.00148056  -0.00349167 # color=yellow tag={MIRI}
```
ejoliet: those 3 polygon belongs to WFPC2. remove them and check that MIRI and WFPC2 is correctly defined.

src/firefly/js/drawingLayers/FootprintTool.js
```diff
@@ -0,0 +1,538 @@
...
+ * @param footprintDrawObj
+ * @param plotId
+ * @returns {*}
+ */
+function updateFootprintAngle(angleDegStr, footprintDrawObj, plotId) {
+    const plot = primePlot(visRoot(), plotId);
+    var cc = CsysConverter.make(plot);
+    var angleDeg;
+
+    angleDeg = angleDegStr ? parseFloat(angleDegStr) : 0.0;
```
ejoliet: float could be a loss of precision, maybe this is why we see lines moving when rotating? Ex. rotating NICMOS footprint.

ejoliet: after reading javascript docs, it seems that parseFloat is the highest precision available, the loss of precision must be somewhere else where we transform angles and distances.

robyww:
Fire Impressions. I agree with Immanuel.  It looks _really_ good.  

UI Issues:

- Label does not update as the uses types. You have to click on the footprint for it to update. Changing corners of the label also has this problem. The marker work like it should, this is just a footprint isssue.
- There issue with the active footprint changing.  The first one does not seem to want to let go unless it is hidden.  My guess is this might be my problem in `ImageViewerLayout.`
- I think we only need 1 decimal place for the angle and it should be in degrees. The input box can be smaller.
- I think the center field should take  the full width so that it does not wrap.  It will look cleaner.
- I really like the way you control the size the rotate control.
- We need @ejoliet input, but the label could default to the name of the footprint.
- In response to @ejoliet on add new footprints: I still like having a an 'Add Marker' for markers but the add for the footprint is probably not be necessary.

src/firefly/js/drawingLayers/FootprintTool.js
```diff
@@ -0,0 +1,538 @@
...
+const TYPE_ID= 'OVERLAY_FOOTPRINT_TYPE';
+const factoryDef= makeFactoryDef(TYPE_ID,creator,null,getLayerChanges,null, getFootprintToolUIComponent);
+const FootprintStatus = new Enum(['attached', 'select', 'attached_relocate', 'relocate', 'rotate']);
+
+export default {factoryDef, TYPE_ID}; // every draw layer must default export with factoryDef and TYPE_ID
+
+var idCnt=0;
+
+export function footprintCreateLayerActionCreator(rawAction) {
+    return (dispatcher) => {
```
robyww: You could break this function up a little.  It is pretty big.

src/firefly/js/drawingLayers/FootprintTool.js
```diff
@@ -0,0 +1,538 @@
...
+
+/**
+ * action create for FOOTPRINT_MOVE
+ * @param rawAction
+ * @returns {Function}
+ */
+export function footprintMoveActionCreator(rawAction) {
+    var angleBetween = (origin, v1, v2) => {
+        var [v1_x, v1_y] = [v1.x - origin.x, v1.y - origin.y];
+        var [v2_x, v2_y] = [v2.x - origin.x, v2.y - origin.y];
```
robyww: I have not seen this done before.  A nice way to make x, y operations clear.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
```diff
@@ -1434,9 +1437,66 @@ public static WebPlotResult getDS9Region(String fileKey) {
             retval = createError("on getDSRegion", null, e);
         }
         return retval;
+    }
 
+    private static final Map<String, String> footprintMap;
+    static {
```
robyww: you and I and @loitly should talk about where the right place to store these files are.  I don't know if we want put it in 'visualize/reousrces'

robyww:
The code looks good. However, it is so much it is hard to evaluate. This is a problem with code reviews. 

I have notice in the last couple of months.  In May I, @loitly, @tgoldina and @ejoliet all had really big commits. How do we look at large commits?  I think we should bring it up for ideas in our Wednesday meeting.

ejoliet:
You are right. And i would add that i am also concerned that we are changing the server too. Java classes are changing in dev but we are not testing for IFE app as well in those cases.

--
Emmanuel Joliet
Office: 170
Phone:  (626) 395-1489

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
cwang2016: besides the issue regrading where to put the footprint files, I also have some concern regrading if it is ok to hardcode the footprint map in the java file? or put the map data into a text file and create the map by loading the text file?

ejoliet: Yes, good point. I would decouple the data content and the loading. The same way we load properties. Actually you could think a future use case of a user-defined footprint to be loaded and added to the layers as part of all the predefined footprint available. 

--
Emmanuel Joliet
Office: 170
Phone:  (626) 395-1489

cwang2016:
Notes: 

- Build firefly_data.jar to include fits (previously contained in fits.jar) and footprint description file.

- Create methods to add and get footprint entry (footprint file path and the associated key) into a map at server side.   

- footprint and marker text: the text is displayed as the drawing layer title initially and the text content is editable in the drawing layer panel.

cwang2016:
Please Note:
Please do full compile (gradle :firefly:war :firefly:deploy ) as the server side are modified.

PR #120, opened 2016-07-15
DM-6909: Filtering from expanded mode cancels expanded mode

loitly:
https://jira.lsstcorp.org/browse/DM-6909

- fixed above ticket
- added close ('x') button to table options and xyplot options
- fixed table's fixed column's transparency issue

### DIFF
--- a/src/firefly/js/charts/ui/ChartsTableViewPanel.jsx
+++ b/src/firefly/js/charts/ui/ChartsTableViewPanel.jsx
@@ -452,7 +452,8 @@ class ChartsPanel extends React.Component {
         if (optionsShown) {
             return (
                 <div className='ChartPanelOptions'>
-                    <OptionsWrapper  {...{chartId, tableModel, tblStatsData, tblPlotData, tblHistogramData, chartType}}/>
+                    <OptionsWrapper toggleOptions={this.toggleOptions}
+                        {...{chartId, tableModel, tblStatsData, tblPlotData, tblHistogramData, chartType}}/>
                 </div>
             );
         }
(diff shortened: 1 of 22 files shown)


tgoldina:
The layout issues described in the ticket are fixed. 

However, after some filtering and sorting, XYPlot, which relies on tableMeta.source to detect table changes, does not follow the table.

It seems that tableMeta.source is not always set to the sorted or filtered table location in ${temp_files}. Sometimes it erroneously points to the original table in ${perm_files}

Even if the request passed to the server does not have META_INFO.source, it is being set [incorrectly] somewhere in server code. Then JsonTableUtil.java lines 40-44 copies all meta from request. If source attribute happens to be in the request, it overwrites the correct source set in TableDef.

src/firefly/js/charts/ui/ChartsTableViewPanel.jsx
```diff
@@ -452,7 +452,8 @@ class ChartsPanel extends React.Component {
         if (optionsShown) {
             return (
                 <div className='ChartPanelOptions'>
-                    <OptionsWrapper  {...{chartId, tableModel, tblStatsData, tblPlotData, tblHistogramData, chartType}}/>
+                    <OptionsWrapper toggleOptions={this.toggleOptions}
+                        {...{chartId, tableModel, tblStatsData, tblPlotData, tblHistogramData, chartType}}/>
```
tgoldina: I am wondering if toggleOptions can be handled here. 
Otherwise we need to propagate this prop to all options components (for now it included HistogramOptions) and handle situations when toggleOptions do not make sense, like in Charts dropdown.

tgoldina:
I am done with the review. The original issues for which this ticket was created are fixed.

There are two new bugs I noticed: 
1) x in Charts dropdown, when scatter plot is selected. It should not be there.
2) table-chart sync problems, because source meta attribute does not always reflect the location of the source IPAC table in our workarea correctly.

The second issue was not introduced in this ticket. Please, feel free to create a separate ticket for it.

tgoldina:
I have tested the latest commit, and everything works as expected. Thank you. Ready to merge.

PR #121, opened 2016-07-21
DM-6952: Table problems

loitly:
https://jira.lsstcorp.org/browse/DM-6952

Table component has couple of problems:
  - Scrambled table values after column selection and saving the table, then reset mess up the table. Saving the table and reset makes the table comes back.
  - table display no longer redefines column names in the table based on the column label, e.g. "Field Size" instead of "s_fov".
  - Downloaded file is not a valid IPAC table.
  - Filtering table does not change image overlay or plot until table is saved. At that point, the filtering works, but the plot symbol changes (happens when result is decimated, datapoints > 5000?)
  - In Edit Table Options, it's unclear what the reset button resets to.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/DataGroupFilter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/DataGroupFilter.java
@@ -94,7 +94,7 @@ public class DataGroupFilter {
         StopWatch.getInstance().start("DataGroupFilter");
 
         TableDef tableMeta = IpacTableUtil.getMetaInfo(source);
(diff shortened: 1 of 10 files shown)


loitly:
In regard to the 5th bullet, reset button will:
 - unchecked show units and filters
 - clear all filters
 - set page size to 100
 - reset column selection to when table was first loaded.

wmiipac:
Can't really test the fix with Atlas,shall  I point firefly to you instead of irsawebdev1?

ejoliet:
I've checked that the issues are fixed apart from the second bullet that ill be tested separately. Will be confirmed.
Regarding the ipac table format, table header is not kept after sorting/filtering.

PR #122, opened 2016-07-21
DM-6908 Add filter editor to the chart toolbar

tgoldina:
- Added filter editor to the charts toolbar
- When filtering returns empty data set, server side will return header. Before, it was returning  \Loading-Status = COMPLETED, and TablesCntlr:265 would never activate a new table callback.
- Charts are cleared when table data are being fetched. 
- Label and unit in XYPlotOptions will now get cleared whenever x or y changes
- removed tbl_id from CatalogWatcher and CoverageWatcher's server requests to prevent interference with the table
- changed the order of lat,lon->lon,lat in CoverageWatcher's server request (just for consistency: it was dec, ra before)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
@@ -134,51 +134,51 @@ public class HistogramProcessor extends IpacTablePartProcessor {
      */
     public DataGroup createHistogramTable(double[] columnData) throws DataAccessException {
 
-
-        /*if (!scale.equalsIgnoreCase(LINEAR_SCALE)){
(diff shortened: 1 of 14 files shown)


loitly:
clear filter button disappears after a region is selected.
- apply a filter, clear filter button appears on top-left.
- select a region in xyplot, clear button disappears.

loitly:
clear filter text did not clear filter.
- apply a filter. ie.  ra > 0
- select the text 'ra >0' then press 'delete'.  tab or enter to apply the change.  nothing happen.

loitly:
table in filter editor is not sortable.
- click on 'Column' column, nothing happen.

loitly:
Done with review.  Code is good.  There are several things I pointed out above that may need fixing.

loitly:
It looks good.  Good to merge.

PR #123, opened 2016-07-21
DM-6953: fixed mutliple image issues

robyww:
- There is no panner for the image (is it missing from the API or is it a bug in calling it?)
 - No readout value from thumbnail image
 - The top bar readout doesn't include units for the pixel flux
 - Clicking on the expand icon deletes the image (in API only)
 - Clicking on the expand icon diabled the expand mode of table and xy plot (in API only)
 - Image button ask you for a position and displays that, wiping out whatever image brought you to IRSA Viewer in the first place. It should give another tile of the same field, drawn from the selected data set. (IRSAVIewer only)
 - grid with label off
 - Made loading images from remote consistent
 - Started conversion of python API
 - fixed range values serialization issue

### DIFF
--- a/src/firefly/js/visualize/ui/ApiExpandedDisplay.jsx
+++ b/src/firefly/js/visualize/ui/ApiExpandedDisplay.jsx
@@ -10,13 +10,14 @@ import {VisHeaderView} from './VisHeaderView.jsx';
 import {ExpandedModeDisplay} from '../iv/ExpandedModeDisplay.jsx';
 // import {currMouseState} from '../VisMouseCntlr.js';
 import {addMouseListener, lastMouseCtx} from '../VisMouseSync.js';
+import {readoutRoot} from '../../visualize/MouseReadoutCntlr.js';
 
 
 
 export class ApiExpandedDisplay extends Component {
     constructor(props) {
         super(props);
-        this.state= {visRoot:visRoot(), currMouseState:lastMouseCtx()};
+        this.state= {visRoot:visRoot(), currMouseState:lastMouseCtx(), readout:readoutRoot()};
     }
 
     shouldComponentUpdate(np,ns) { return sCompare(this,np,ns); }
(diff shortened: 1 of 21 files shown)


src/firefly/js/visualize/ui/ApiExpandedDisplay.jsx
```diff
@@ -33,8 +34,10 @@ export class ApiExpandedDisplay extends Component {
     }
 
     storeUpdate() {
-        if (visRoot()!==this.state.visRoot || lastMouseCtx() !==this.state.currMouseState) {
-            this.setState({visRoot:visRoot(), currMouseState:lastMouseCtx()});
+        if (visRoot()!==this.state.visRoot || 
+            lastMouseCtx() !==this.state.currMouseState || 
+            readoutRoot()!==this.state.readout) {
+            this.setState({visRoot:visRoot(), currMouseState:lastMouseCtx(), readout:readoutRoot()});
```
cwang2016: Do functions (visRoot(), lastMouseCtx(), readoutRoot()) get called twice possibly if the code is written like this?  Sometime I was puzzled by  similar situation like this, not sure if this is a concern as thinking of execution time.

cwang2016:
Review is done. Looks good to me.

ejoliet:
From what i could check, everything has been addressed and solved! Works like a charm!

The only (minor) difference i've noticed from testing external viewer call from demo page in chrome, safari and firefox is that in firefox, the first time you call the external viewer, the tab will get active but the subsequent calls won't. In Safari and Chrome, once you click to open the viewer, it will automatically focus and make the tab active where the image is opened independently of the first one called already or not.

ejoliet:
"fixed range values serialization issue" needs to be confirmed and tested by @wmiipac btw.

src/firefly/js/visualize/ui/ApiExpandedDisplay.jsx
robyww: Ultimately I don't think it matters for performance. Maybe it would if it were called in a loop.  Even if you think about performance, most of the time the if would be `false`.  Therefore putting it is 3 variables might take more resources.

robyww:
>> The only (minor) difference i've noticed from testing external viewer call from demo page in chrome, safari and firefox is that in firefox, the first time you call the external viewer, the tab will get active but the subsequent calls won't. In Safari and Chrome, once you click to open the viewer, it will automatically focus and make the tab active where the image is opened independently of the first one called already or not.

We should make a new ticket for that.  Our solution with the old system it is blink the title on the tab until it become active.

PR #124, opened 2016-07-25
DM-6439: Remove GWT from build

loitly:
https://jira.lsstcorp.org/browse/DM-6439

  - remove GWT build from firefly
  - AnyFileUpload no longer pre-load fits images
  - fix firefly showing blank page when no results are present.  should show drop-drown.


robyww:
Looks good.

PR #126, opened 2016-07-28
DM-7029: fixed some api bugs

robyww:
- Image is coming with one draw layer. (I can delete this draw layer and nothing changes on the image)
 - When draw layer is deleted, and no more layers are present, the layers dialog should be closed. (It stays with nothing to
  display, you have to click x to close it.)
 - After selecting an area in one viewer, I select an area in another viewer, then move the mouse to the first one: 147
  ImageViewerLayout.jsx:314 Uncaught TypeError: Cannot read property 'x' of null at ImageViewerLayout.jsx:314 Nothing works after
that. I have to reload.
 - Selection appears with an offset, if the page, which contains the viewer is scrolled. (Load the attached script, press
  'Start selection tracking' click to select a point, then scroll page a bit down, then click to select another point - it shows
down from where it should be.)
 - I have 2 image viewers in separate divs. Selected line in one, then selected line in the other. The line from the first one
  disappeared, but its label is still there. (See attached image.)
 - Selection is working differently from distance. To select in another plot, I need to press selection again. I don't need to press ruler again to select new distance in another plot.
 - The payload.attValue of CHANGE_PLOT_ATTRIBUTE action is using WorldPt for area and point selections (when payload.attKey is  'SELECTION' or 'ACTIVE_POINT'), but ImagePt for line selection (when payload.attKey is 'ACTIVE_DISTANCE') How can I make them
all use image coordinates? Now exporting the coordinage system conversion code.

`var imagePt= firefly.util.image.CCUtil.getImageCoords(plot,pt);`

### DIFF
--- a/src/firefly/js/api/ApiUtilImage.jsx
+++ b/src/firefly/js/api/ApiUtilImage.jsx
@@ -33,6 +33,8 @@ export {RequestType} from '../visualize/RequestType';
 export {ExpandType, dispatchApiToolsView} from '../visualize/ImagePlotCntlr.js';
 export {CsysConverter} from '../visualize/CsysConverter.js';
 export {CCUtil} from '../visualize/CsysConverter.js';
+export {primePlot} from '../visualize/PlotViewUtil.js';
+export {visRoot} from '../visualize/ImagePlotCntlr.js';
 export {watchCoverage} from '../visualize/saga/CoverageWatcher.js';
 import {watchImageMetaData} from '../visualize/saga/ImageMetaDataWatcher.js';
 
--- a/src/firefly/js/api/ApiUtilImage.jsx
+++ b/src/firefly/js/api/ApiUtilImage.jsx
@@ -31,6 +31,8 @@ export {RangeValues} from '../visualize/RangeValues.js';
 export {WPConst, WebPlotRequest, findInvalidWPRKeys, confirmPlotRequest} from '../visualize/WebPlotRequest.js';
 export {RequestType} from '../visualize/RequestType';
 export {ExpandType, dispatchApiToolsView} from '../visualize/ImagePlotCntlr.js';
+export {CsysConverter} from '../visualize/CsysConverter.js';
+export {CCUtil} from '../visualize/CsysConverter.js';
 export {watchCoverage} from '../visualize/saga/CoverageWatcher.js';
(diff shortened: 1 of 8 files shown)


src/firefly/js/api/ApiUtilImage.jsx
```diff
@@ -31,6 +31,8 @@ export {RangeValues} from '../visualize/RangeValues.js';
 export {WPConst, WebPlotRequest, findInvalidWPRKeys, confirmPlotRequest} from '../visualize/WebPlotRequest.js';
 export {RequestType} from '../visualize/RequestType';
 export {ExpandType, dispatchApiToolsView} from '../visualize/ImagePlotCntlr.js';
+export {CsysConverter} from '../visualize/CsysConverter.js';
+export {CCUtil} from '../visualize/CsysConverter.js';
```
tgoldina: To use CCUtil, we need to add the following exports:
export {primePlot} from '../visualize/PlotViewUtil.js';
export {visRoot} from '../visualize/ImagePlotCntlr.js';

tgoldina:
Thank you for quick turnaround. Most of the issues are fixed. 

Here is what I have noticed:

1. After using distance tool in one plot, then the other, clicking again in the first plot does not make it active. (You have to click on the border of the plot to make it active, clicking inside does not help.)

This seems to cause strange behavior, when selections do not work as expected. For example:  select ruler, select some lines alternating first and second plot. Unselect ruler, select area icon. Selecting in the first plot will still show distance. I had to delete distance tool drawing layer or click on the border for things to start showing area selection. 

In general, there is some confusion with active plot, when I have two viewers. Should a plot become active as soon as mouse enters is? Otherwise in readout, compass thumbnail will still show active plot, while readout thumbnail could use another plot.

2. In Selection Tool drawing layer the help text is so long that 'Delete' link is not visible. The text is also cut.

src/firefly/js/api/ApiUtilImage.jsx
robyww: ok, I will do that too.

robyww:
With selection, there is no delete. The drawing layers make that stuff optional. I can turn it on.

robyww:
#1 is a little beyond the scope of this ticket.  I will make new ticket from it.

tgoldina:
Thank you. Then it's ready to merge.

2016-07-29 15:13 GMT-07:00 Trey Roby <notifications@github.com>:

PR #128, opened 2016-08-01
DM-7016: Image loading feedback

robyww:
- includes a fix to Target Panel, the automatic active target was not working correctly


wmiipac:
review is done. All looks good.
It also fixed the two issues in DM-7068, listed under Atlas. BTW Atlas code is updated accordingly.

PR #129, opened 2016-08-01
DM-7055: fix miscellaneous table issues

loitly:
- disable sorting when content is html
 - table options: auto-adjust all column width based on content
 - table refractoring..
    - renamed a few actions to better reflect what it's doing.
    - added TABLE_FILTER
    - document sequence of actions where applicable.
 - update build script to skip buildClient when possible.

### DIFF
--- a/src/firefly/js/tables/reducer/TableDataReducer.js
+++ b/src/firefly/js/tables/reducer/TableDataReducer.js
@@ -20,7 +20,7 @@ export function dataReducer(state={data:{}}, action={}) {
                 return updateSet(root, [tbl_id, 'selectInfo'], selectInfo);
             } else return root;
 
-        case (Cntlr.TABLE_NEW_LOADED)  :
+        case (Cntlr.TABLE_LOADED)  :
             const statusPath = [tbl_id, 'tableMeta', 'Loading-Status'];
             if (get(root, statusPath, 'COMPLETED') !== 'COMPLETED') {
                 return updateSet(root, statusPath, 'COMPLETED');
@@ -28,13 +28,14 @@ export function dataReducer(state={data:{}}, action={}) {
 
         case (Cntlr.TABLE_HIGHLIGHT)  :
         case (Cntlr.TABLE_UPDATE)  :
-            return TblUtil.smartMerge(root, {[tbl_id] : action.payload});
+            return TblUtil.smartMerge(root, {[tbl_id] : {isFetching:false, ...action.payload}});
 
-        case (Cntlr.TABLE_NEW)      :
+        case (Cntlr.TABLE_FETCH)      :
+        case (Cntlr.TABLE_FILTER)      :
         case (Cntlr.TABLE_SORT)     :
         case (Cntlr.TABLE_REPLACE)  :
             const rowCount = action.payload.totalRows || get(action, 'payload.tableData.data.length', 0);
-            const nTable = Object.assign({isFetching:false, selectInfo: SelectInfo.newInstance({rowCount}).data},action.payload);
+            const nTable = Object.assign({isFetching:false, selectInfo: SelectInfo.newInstance({rowCount}).data}, action.payload);
             return updateSet(root, [tbl_id], nTable);
 
         case (Cntlr.TABLE_REMOVE)  :
--- a/src/firefly/js/visualize/saga/CatalogWatcher.js
+++ b/src/firefly/js/visualize/saga/CatalogWatcher.js
@@ -3,11 +3,11 @@
  */
 
 import {take} from 'redux-saga/effects';
-import {isEmpty, get, omit} from 'lodash';
+import {isEmpty, get} from 'lodash';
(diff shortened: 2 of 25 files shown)


tgoldina:
I think 5th issue (Catalog overlay should not use table id in drawing layer description. (See the attached screenshot.) It should be using MetaConst.CATALOG_OVERLAY_TYPE attribute value) was added after your last commit, and you did not have chance to fix it.

src/firefly/js/visualize/saga/CoverageWatcher.js
```diff
@@ -111,7 +111,7 @@ export function* watchCoverage({viewerId, options= {}}) {
 
         switch (action.type) {
 
-            case TABLE_NEW:
+            case TABLE_LOADED:
                 if (!getTableInGroup(tbl_id)) continue;
                 decimatedTables[tbl_id]= null;
                 displayedTableId = updateCoverage(tbl_id, viewerId, decimatedTables, options);
```
tgoldina: Should we update coverage on invoked_by=TABLE_SORT? (Since all it is doing is collecting points to calculate coverage image.)

src/firefly/js/tables/reducer/TableDataReducer.js
```diff
@@ -20,21 +20,22 @@ export function dataReducer(state={data:{}}, action={}) {
...
+        case (Cntlr.TABLE_LOADED)  :
             const statusPath = [tbl_id, 'tableMeta', 'Loading-Status'];
             if (get(root, statusPath, 'COMPLETED') !== 'COMPLETED') {
                 return updateSet(root, statusPath, 'COMPLETED');
             } else return root;
 
         case (Cntlr.TABLE_HIGHLIGHT)  :
         case (Cntlr.TABLE_UPDATE)  :
-            return TblUtil.smartMerge(root, {[tbl_id] : action.payload});
+            return TblUtil.smartMerge(root, {[tbl_id] : {isFetching:false, ...action.payload}});
```
tgoldina: Just curious, why isFetching set to false here? I though on TABLE_UPDATE the data are still coming. Or isFetching marks the state before first TABLE_UPDATE?

src/firefly/js/visualize/saga/CatalogWatcher.js
```diff
@@ -42,13 +42,15 @@ export function* watchCatalogs() {
...
+        const action= yield take([TABLE_LOADED, TABLE_SELECT,TABLE_HIGHLIGHT, TABLE_UPDATE,
                                   TABLE_REMOVE, ImagePlotCntlr.PLOT_IMAGE]);
         const {tbl_id}= action.payload;
         switch (action.type) {
-            case TABLE_SORT:
-            case TABLE_NEW_LOADED:
-                handleCatalogUpdate(tbl_id);
+            case TABLE_LOADED:
+                const {invokedBy} = action.payload;
+                if (invokedBy !== TABLE_SORT) {
```
tgoldina: I think this is a bug. Catalog needs to be updated on TABLE_SORT. Otherwise highlights won't match table or plot.

src/firefly/js/visualize/saga/ImageMetaDataWatcher.js
```diff
@@ -37,7 +37,7 @@ export function* watchImageMetaData({viewerId}) {
     var tbl_id;
     var paused= true;
     while (true) {
-        const action= yield take([TABLE_NEW,TABLE_SELECT,TABLE_HIGHLIGHT, TABLE_UPDATE, TABLE_REMOVE,
+        const action= yield take([TABLE_SELECT,TABLE_HIGHLIGHT, TABLE_UPDATE, TABLE_REMOVE,
```
tgoldina: TABLE_HIGHLIGHT will take care of the first row, hence no need to take TABLE_LOADED?

src/firefly/js/visualize/saga/CatalogWatcher.js
ejoliet: before, when sorting, the table highlight was the first row. So it might not be a bug.

tgoldina: I mean that we do need to refetch catalog data on TABLE_SORT. We did it
before.

2016-08-02 15:34 GMT-07:00 Emmanuel Joliet <notifications@github.com>:

src/firefly/js/visualize/saga/ImageMetaDataWatcher.js
loitly: if you are watching TABLE_UPDATE, then TABLE_LOADED is not needed, since UPDATE is also fired on first load as well.

src/firefly/js/visualize/saga/CatalogWatcher.js
loitly: @tgoldina , you are right.

src/firefly/js/tables/reducer/TableDataReducer.js
loitly: Yes, it's only true before any data returned.

src/firefly/js/visualize/saga/CoverageWatcher.js
loitly: yep..  added logic to skip updateCoverage when table is reloaded due to sorting.

PR #130, opened 2016-08-02
DM-5192 Implement Python API to support Firefly actions.

cwang2016:
- move 'python' and 'test/python' from fftools to be  under firefly and firefly/test.
- implement functions to dispatch actions for showing fits, showing table, and adding extension, operating the image (zoom, stretch, and pan), creating/deleting region layer,  adding/removing region entries. 
- More functions for adding/removing mask and showing XYPlot will be implemented further.  

Fixed some region/footprint related issues in Javascript implementation,
- per DS9 document, the default system is implicitly set to be 'PHYSICAL' if no coordinate system is explicitly specified in the region file or region data. 
- fixed the footprint/marker display in case the target is represented in various world coordinate systems.


cwang2016:
- some footprint data is updated, please build and deploy firefly first.
  (gradle :firefly:war :firefly:deploy, and gradle :wise:war)

cwang2016:
The test file, firefly/test/python/demo-notebook.ipynb, can be run by setting PYTHONPATH = <path> where firefly/python/display/FireflyClient.py is located.

robyww:
Clean up `firefly/src/fftools/python/display/FireflyClient.pyc` needs to be removed from git.

robyww:
in the test files you should use relative paths to the data. The put come documentation at the top saying which directly to expect to start notebook in.

robyww:
Review Complete. Code looks good. Good Job.

stargaser:
Copied from DM-5192 (I couldn't put all of these in comments on code diffs since many are legacy issues):

Here are some style comments on the FireflyClient.py module, drawn in part from the [DM Python Style Guide](https://developer.lsst.io/coding/python_style_guide.html). I'm still thinking about the method signatures in detail and will comment a little later on those.

* The methods you don't intend users to use should be prefixed with a single underscore (`handleEvent, sendURLAsGet, sendURLAsPost, isPageConnected, createRV, makePidParam`)
* Static methods which use other attributes or methods of the class should be class methods instead. See [this discussion on Stack Overflow](http://stackoverflow.com/questions/136097/what-is-the-difference-between-staticmethod-and-classmethod-in-python). The methods in this case include `createRV, createRangeValuesStandard, createRangeValuesZScale, genItemId, createImageURL`.
* Instead of `if type(var) is list` , should use `if isinstance(var, list)`

stargaser:
This is another comment not on the specific diffs in this pull request:

As far as the API goes: I know that the Javascript API is organized around Actions. Python users will want to get information back from Firefly. So a simple `zoom` or `pan` does not indicate by itself whether the user is setting the zoom in Firefly or getting the currently set value back. If a `getZoom` or `getPan` is doable with the Javascript API, then I'd want to have `setZoom` and `setPan`. Astronomers will also want to be able to use world coordinates (ra, dec) for specifying the panning, in addition to pixel units.

stargaser:
I like the naming of `showFits`, `showTable`, and `showXYPlot` as they clearly indicate that something is going to happen in the Firefly window, and they are in keeping with e.g. the APLpy methods.

Similarly, `addRegionData`, `deleteRegion`, `removeRegionData` indicate actions though I would expect `deleteRegion` to delete a single region marker and not an entire layer.

src/firefly/test/python/demo-notebook.ipynb
```diff
@@ -0,0 +1,423 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from FireflyClient import *\n",
```
stargaser: Please use `from FireflyClient import FireflyClient`. I also wonder if the module should be named `fireflyClient.py` -- I couldn't tell for sure from reading the DM Python Style guide.

src/firefly/python/display/FireflyClient.py
```diff
@@ -533,11 +621,41 @@ def createRangeValuesZScale(algorithm, zscaleContrast=25, zscaleSamples=600, zsc
...
+        :return: an ID string
+        """
 
+        if item in FireflyClient.itemID:
+            FireflyClient.itemID[item] += 1
+            return item + '-' + str(FireflyClient.itemID[item])
+        else:
+            return None
+
+    @staticmethod
```
stargaser: make a `@classmethod` since the method references the class

robyww:
> As far as the API goes: I know that the Javascript API is organized around Actions. Python users will want to get information back from Firefly. So a simple zoom or pan does not indicate by itself whether the user is setting the zoom in Firefly or getting the currently set value back. If a getZoom or getPan is doable with the Javascript API, then I'd want to have setZoom and setPan. Astronomers will also want to be able to use world coordinates (ra, dec) for specifying the panning, in addition to pixel units.

I think you are right about the naming.  However, the names where made to be similar to the afw.display api.  We should probably make better name and be aware they they will be different.  Maybe down the road we can improve the afw.display names.

src/firefly/python/display/FireflyClient.py
```diff
@@ -533,11 +621,41 @@ def createRangeValuesZScale(algorithm, zscaleContrast=25, zscaleSamples=600, zsc
...
         """
-        retval= FireflyClient.createRV('zscale',1,2,algorithm,zscaleContrast, zscaleSamples, zscaleSamplesPerLine)
+        retval= FireflyClient.createRV('zscale',1,1,algorithm,zscaleContrast, zscaleSamples, zscaleSamplesPerLine)
         if not retval:
             a= algorithm if algorithm.lower() in FireflyClient.stretchAlgorithmDict else 'linear'
             retval= FireflyClient.createRV('zscale',1,2,a,25,600,120)
         return retval
 
 
+    @staticmethod
```
stargaser: make this a `@classmethod` since it references the class

src/firefly/python/display/FireflyClient.py
```diff
@@ -332,119 +387,151 @@ def showXYPlot(self, fileOnServer, additionalParams=None):
...
+        extension = {'id': extensionId, 'plotId': plotId, 'imageUrl': imageUrl,
+                     'title': title, 'extType': extType, 'toolTip': toolTip}
+        payload={'extension': extension}
+        return self.dispatchRemoteActionByPost(self.channel, FireflyClient.actionDict['AddExtension'], payload)
+
+    #----------------------------
+    # actions on image
+    #----------------------------
+
+    def zoom(self, plotId, factor = 1.0):
```
stargaser: recommend to call this `setZoom`

src/firefly/python/display/FireflyClient.py
```diff
@@ -332,119 +387,151 @@ def showXYPlot(self, fileOnServer, additionalParams=None):
...
+            return  self.dispatchRemoteAction(self.channel, FireflyClient.actionDict['ZoomImage'], payload)
+
+        if type(plotId) in (tuple, list):
+            return [zoomOnePlot(x, factor) for x in plotId]
+        else:
+            return zoomOnePlot(plotId, factor)
 
 
-    def pan(self, plotId, x, y):
+    def pan(self, plotId, x=None, y=None):
```
stargaser: recommend to call this `setPan`

src/firefly/python/display/FireflyClient.py
```diff
@@ -332,119 +387,151 @@ def showXYPlot(self, fileOnServer, additionalParams=None):
...
-        return self.sendURLAsGet(url)
+
+        payload = {'plotId': plotId}
+        if x and y :
+            payload.update({'centerPt': {'x': x, 'y': y, 'type': 'ImagePt'}})
+
+        return self.dispatchRemoteAction(self.channel, FireflyClient.actionDict['PanImage'], payload)
+
 
     def stretch(self, plotId, serializedRV):
```
stargaser: recommend to call this `setStretch`

src/firefly/python/display/FireflyClient.py
```diff
@@ -533,11 +621,41 @@ def createRangeValuesZScale(algorithm, zscaleContrast=25, zscaleSamples=600, zsc
...
-        retval= FireflyClient.createRV('zscale',1,2,algorithm,zscaleContrast, zscaleSamples, zscaleSamplesPerLine)
+        retval= FireflyClient.createRV('zscale',1,1,algorithm,zscaleContrast, zscaleSamples, zscaleSamplesPerLine)
         if not retval:
             a= algorithm if algorithm.lower() in FireflyClient.stretchAlgorithmDict else 'linear'
             retval= FireflyClient.createRV('zscale',1,2,a,25,600,120)
         return retval
 
 
+    @staticmethod
+    def genItemId(item):
```
stargaser: Is this for internal use only? If yes, please preface with a single underscore.

src/firefly/test/python/demo-notebook.ipynb
robyww: Also they may be trying to get away from camel casing.  At one time they decided they wanted to do it like they but I remember some discussion. Might be worth asking Tim Jenness

src/firefly/python/display/FireflyClient.py
```diff
@@ -533,11 +621,41 @@ def createRangeValuesZScale(algorithm, zscaleContrast=25, zscaleSamples=600, zsc
...
+        """
 
+        if item in FireflyClient.itemID:
+            FireflyClient.itemID[item] += 1
+            return item + '-' + str(FireflyClient.itemID[item])
+        else:
+            return None
+
+    @staticmethod
+    def createImageURL(imageSource):
```
stargaser: I'm wondering if this should be called (or made into) `downloadPng` because originally I thought it meant "give me a link to the FITS file".

stargaser:
Another comment on legacy code:

The methods for specifying the stretch of the image look very forbidding to the astronomer, especially `createRV`, `createRangeValuesStandard` and `createRangeValuesZScale`. It would be better to hide a lot of this under the hood in a `setStretch` method that would allow for either Zscale, MinMax, Percentile, or absolute levels for the stretch.

src/firefly/test/python/demo-notebook.ipynb
```diff
@@ -0,0 +1,423 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "#file= fc.uploadFile('./data/c.fits')\n",
+    "file= fc.uploadFile('/hydra/cm/firefly/src/firefly/test/python/data/wise-1b-1.fits')\n",
```
stargaser: The user might be using standalone `firefly-exec.war` so please use a relative path: `.data/wise-1b-1.fits`

src/firefly/python/display/FireflyClient.py
```diff
@@ -242,35 +277,53 @@ def uploadTextData(self, stream):
...
+
+    #-------------------------
+    # dispatch actions
+    #-------------------------
+
+    #----------------------------------------------------------------
+    # action on showing fits, tables, XYPlot and adding extension
+    #----------------------------------------------------------------
+
     def showFits(self, fileOnServer=None, plotId=None, additionalParams=None):
```
stargaser: Looking at [this Pythonic style guide](http://docs.python-guide.org/en/latest/writing/style/) under Function Arguments, it is more Pythonic to pass additional keyword arguments as `**kwargs` which will be a dictionary of additional named parameters, rather than to have the user construct a dictionary and pass it in.

src/firefly/test/python/demo-notebook.ipynb
```diff
@@ -0,0 +1,423 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "pParams= { 'URL' : 'http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band4.fits','ColorTable' : '9', 'title': 'hello'}\n",
+    "status= fc.showFits(fileOnServer=None, plotId='p2', additionalParams=pParams)"
```
stargaser: As noted in the diff for `showFits` in `FireflyClient.py`, it is more Pythonic to use `**kwargs` in the function definition, then make this call `status = fc.showFits(fileOnServer=None, plotId='p2', URL='http://web.ipac.caltech.edu/staff/roby/demo/wise-m51-band4.fits', ColorTable='9', title='hello')`

stargaser:
Review complete. The region functionality is very nice!
Apologies for so many comments that were on legacy code predating this pull request.

stargaser:
An extra note. Tim Jenness confirmed that LSST will move towards more Pythonic lowercase_and_underscores instead of camelCase. The DM Python style guide needs updating. So we should rename `removeRegionData` to `remove_region_data`, etc.

PR #134, opened 2016-08-02
DM-7090: fix IrsaViewer catalog panel style so labels and fields do n…

loitly:
Irsa Catalog Search panel:

 - Remove white spaces between label, input, and drop-down in the target panel.
 - The position of those element remain fixed during user input.
 - Do the same for target panel in VO Catalog tab.
 - Align the fields so they look less clutter.


ejoliet:
Looks good to me, better than before! Thanks! Nothing wrong to say about the code. Review done ok.

PR #135, opened 2016-08-03
DM-7088 make the image select work with coverage correctly

cwang2016:
- update the plot container to be able to create a new plot, replace an existing plot or neither while doing image select.


robyww:
OK, I found another condition.  If there is a coverage but no other images then there is not way to add images in the DEFAULT_FITS_VIEWER_ID container. So I guess we need to make it so that you _do_ have an add plot on the drop down but it the current container can't handle it then it goes to DEFAULT_FITS_VIEWER_ID.

I think you should first make sure the container exist.  There might be use case in other application that the remove it so that there is not way to add a plot.

robyww:
Pushed a bug fix I just found in Point.js

cwang2016:
I will make the change.

robyww:
I don't see any other issues. Review Complete.

cwang2016:
Made some update which allows 'adding' a plot in addition to replace the plot  in case the container of the active plot can only replace the plot  by finding an available plot container which allows adding new plot.

robyww:
Tested. It looks fine.

PR #136, opened 2016-08-04
DM-6473: Fixes mostly related to image view of the meta data via table

robyww:
- images cannot be remove, but in expanded mode, it can.
 - plot group fails are not failing correctly, no message shown
 - When a non-meta table is selected, images are shown, but not the toolbar.
 - After table is removed, images are still there and other problem with removing tables.
 - Firefly viewer uses 'triViewImages’ viewer_id while api has no viewer_id. as a result, images loaded by api will be lost once table or other searched data are returned. (I think this is fixed, but I may not understand it)
 - Catalog or Grid overlay are drawn outside of the images.


loitly:
Review completed without any issues.
I noticed a 1 pixel gap between the image viewer's tab and its content.  
It also appears in dev, so I am sure it's not caused by this pull request.

PR #137, opened 2016-08-08
DM-6954 XY plot problems found

tgoldina:
- Added min/max options for plot boundaries
- Label changes for decimation: X-Bins and Y-Bins: Number of X-Bins, Number of Y-Bins
- Reduced marker radius to 3px, highlight to 4px
- The units on the plot are in braces: ex. "dec (deg)" 
- Updated PointerPopup so that error message on min/max fields appears within windows boundaries.

(The precision issue is not yet fixed, it's the subject of DM-6920)


ejoliet:
I have reviewed the PR. Everything has been fixed as written here but i have a couple of comment about getting density plot for decimated data:
1/ When i search for 'm16' around 10' on 2mass, i get the shaded plot without legend. The legend appears only in expanded mode, unless i change 'linear' to 'log' shade in 'binning options', the legend appears. To be consistent we probably don't want to show the legend or we show the legend always, what do you think?
2/ When switching from 'linear' to 'log' is reflecting the change without the need to click on 'apply'. Again, for consistency, i wasn't expected that behavior (which i found cool!) but maybe for consistency, we could let the user click apply instead. Again, what do you think?

ejoliet:
When i select X-bin =10, Y-bin=10, no legend is displayed nor in expanded mode. Probably because no space for that? Take a look please.

ejoliet:
What is the reset button suppose to do. If i change X-bins, Y-bins, then click on reset, i expect to get back the default value (100, 100) from where i started but nothing change. Any idea?

tgoldina:
Currently, in Highcharts it's not possible to adjust legend symbol size for custom symbols. For this reason, I do not display legend if symbol size it too small or too big. Hopefully, we can overcome this restriction in the future.

ejoliet:
I see, we should raise a separate ticket so we track it and hopefully fix it.
Everything on this ticket has been fixed, all good to me, review completed.

tgoldina:
Reset is there to be able to reset back to the currently used parameters.  Otherwise, if you change them without Apply, there is no way to go back. Clear is used to enter new parameters for the plot. (So the parameters are set to some reasonable values.)

PR #138, opened 2016-08-09
DM-7149: Add more support for FITS table display

loitly:
- Add tbl_index to table request indicate which table to display if it's a multi-table file.
 - Add default behavior for column with array type to display the array's data-type and length.

https://jira.lsstcorp.org/browse/DM-7149

To test:
Follow above link to download table.fits file.
Using Firefly's "Data Sets..." sample dialog, upload the sample file to view fits table data.
Use the Table Index field to select which table to view.  There should be 4 tables in this file.


cwang2016:
Review and test are done. Looks good.

PR #139, opened 2016-08-09
DM-7155: Filtering from image is wrong on sorted tables.

loitly:
https://jira.lsstcorp.org/browse/DM-7155

from ticket:
> Sort a table first. Then using image selection tool, select an area, then filter: the filter fails. Points are outside the selected area.

PS.. Title of ticket mentioned that is was wrong for XyPlot as well.  That is NOT true.  XyPlot works fine.


tgoldina:
Review complete. Works as expected.

PR #140, opened 2016-08-09
DM-6920 Yellow dot highlight precision

tgoldina:
Use the original data whenever available.
For expressions, calculate precision based on min-max difference.


ejoliet:
Review done, all good and fixed.

tgoldina:
In the latest commit:
- Turned off log scale options for decimated plot
- Use reversed ra axis for catalogs

Bug fixes:
 - disappearing axis, when log option is selected and log can not be applied 
   (when data have 0 or negative numbers, the plot falls back to linear scale)
 - aspect ration not being passed to server when decimating
 - marker size not being updated on resize
 - extra resize when plot is loading

PR #142, opened 2016-08-11
DM-7038-JSDoc

lznakano:
To look at the doc, please open jsdoc/index.html file.  I have edited the source files (.js and .jsx) to add document tags, add parameter' types and names etc to the best of my knowledge.  Please examine the them and let me know  of any of them are wrong.  I am sure some of them are not right. 

NOTE: 
Module:  it is for documentation purpose only, lowLevelApi and highLevelApi are just for listing the method/function etc in two category.  

Known issue:
1.  The sorting in the template uses longname (path+methodName).  
2.  The README.md file can be accepted by jsdoc, however, it does not translate the links inside    README.md file correctly.  Thus, I used a fireflyOverView.md instead.  There is an option to copy the static files but apparently, it does not work as it was described. 
3. I put the example markdown file into tutorial directory

Please read it through and let me know your comments and suggestions.  

Thanks!

LZ


docs/fireflyOverView.md
```diff
@@ -0,0 +1,219 @@
...
+ -  [Gradle 2.3+] (https://gradle.org/downloads)
+    Gradle is an open source build automation system.
+
+
+ -  [Tomcat 7+] (http://tomcat.apache.org/download-70.cgi)
+    Apache Tomcat is an open source software implementation of the Java Servlet and JavaServer Pages technologies.
+
+ -  [Node 0.12.0.+] (https://nodejs.org/download/)
+    Javascript interpreter for command line environment, used for development tools
+
```
tgoldina: The 4 items above do not show as links. Perhaps, we should put them in the form `[text](link)` with no spaces. Is this file replacing README.md?

src/firefly/js/api/ApiBuild.js
```diff
@@ -49,16 +49,27 @@ import {buildViewerApi} from './ApiViewer.js';
...
     const lowlevelApi= buildLowlevelAPI();
     const viewInterface= buildViewerApi();
+
     const highLevelApi= buildHighLevelApi(lowlevelApi);
     if (!window.firefly) window.firefly= {};
     window.firefly.ignoreHistory = true;
     Object.assign(window.firefly, lowlevelApi, highLevelApi, viewInterface);
+    /**
+     * @namespace firefly
+     * @desc Eefine the firefly object that includes highLevelApi and lowLevelApi and the viewInterface
```
tgoldina: Eefine -> Define

loitly:
My initial reaction is that the js docs should be generated as part of a build, and not commit as source.  Can we at least talk about this before this pull request is merged?

ejoliet:
BTW, jsdocs html result shouldn't need to be in the commit for 'dev', we only need a version for 'ops' when tagging. please remove result htlm that is not needed to produce a jsdoc from gradle.
 
Anyway, following up Loi's comment, as we said previously, in general, the build should compile, test (+reports), generate docs and/or deploy. If one step fails, the build fails even if it's jsdocs!  User should know, for example warnings, error by looking at reports. 
So for JSDOC, we should have a build (gradle docs should take care of java and js) to generate jsdocs from a js file or the entire js folder (maybe excluding *test*, although test should be a good guide on how use or not use functions).

jsdoc/api_ApiBuild.js.html
```diff
@@ -0,0 +1,278 @@
+<!DOCTYPE html>
```
ejoliet: this is a result from generating jsdoc, no need to be in git, only for version tagged or ops in general. jsdocs reflect stable ops version not ongoing unstable 'dev'.

tgoldina: Couple more suggestions:

1. About structure of JSDoc. I think, it would be preferable to display API
jsdoc hierarchically around the namespaces:

- firefly
- firefly.action
- firefly.action.type
- firefly.ui
- firefly.util
- firefly.util.image
- firefly.util.table
- firefly.util.chart

Ideally, we should be able to place functions and components under
namespaces by using  @memberof <namespace> tag in the *original* function
doc. I think, we should not rely on single or double star in comment to
decide what goes into API docs and what does not.

2. We might want to use jsdoc.json instead of conf.json, then it would be
more clear why this config file is for.


2016-08-12 10:48 GMT-07:00 Emmanuel Joliet <notifications@github.com>:

tgoldina: As an alternative to using single quote to mark private functions, we can
mark everything we want to go into the API as @public and use command-line
'access' property to generate docs just for public members.

2016-08-12 13:20 GMT-07:00 Tatiana Goldina <tgoldina@gmail.com>:

PR #143, opened 2016-08-11
DM-6910 Tab panel issues

tgoldina:
Made resizable tabs an option (most cases, we don't need tabs to resize)
When tabs are resizable, multiple tabs will have the same size.


loitly:
Works great..  Review completed.

PR #144, opened 2016-08-11
DM-5793: Mask overlay support

robyww:
Notes:
 - zoom works, rotate works.
 - files can't be rotate before reading in mask (no check for this yet)
 - drawing layer controls work
 - flip might work. I think it works on  the client. Currently there is a bug fliping certain types of data on the server side.
 - crop does not work. I think the client is fine but there is a server bug as well
 - python support for mask: update add_mask and remove_mask for mask layer (this work was done by Cindy)
 - bugs fixes with loading image and mask in same python cell

### DIFF
--- a/src/firefly/js/visualize/PlotChangeTask.js
+++ b/src/firefly/js/visualize/PlotChangeTask.js
@@ -322,7 +322,8 @@ function processPlotReplace(dispatcher, result, pv, makeSuccessAction, makeFailA
             var overlayPlotViews = [];
             resultAry.forEach((r, i) => {
                 if (i === 0) return;
-                var plot = WebPlot.makeWebPlotData(pv.plotId, r.data[WebPlotResult.PLOT_CREATE], true);
+                const {imageOverlayId}= pv.overlayPlotViews[i-1];
+                var plot = WebPlot.makeWebPlotData(imageOverlayId, r.data[WebPlotResult.PLOT_CREATE][0], {}, true);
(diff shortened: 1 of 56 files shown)


src/firefly/js/visualize/PlotChangeTask.js
```diff
@@ -322,7 +322,8 @@ function processPlotReplace(dispatcher, result, pv, makeSuccessAction, makeFailA
             var overlayPlotViews = [];
             resultAry.forEach((r, i) => {
                 if (i === 0) return;
-                var plot = WebPlot.makeWebPlotData(pv.plotId, r.data[WebPlotResult.PLOT_CREATE], true);
+                const {imageOverlayId}= pv.overlayPlotViews[i-1];
+                var plot = WebPlot.makeWebPlotData(imageOverlayId, r.data[WebPlotResult.PLOT_CREATE][0], {}, true);
                 overlayPlotViews[i - 1] = {plot};
             });
```
cwang2016: just would like to confirm that that resultant length of overPlotViews is one less that of resultAry.

cwang2016:
Review completed. What a big effort work!
one note: 
- dispatchPlotMask has no 'fileKey' parameter passed.

src/firefly/js/visualize/PlotChangeTask.js
robyww: Yes, the main image is 0 and the others are overlays

PR #145, opened 2016-08-12
DM-6950: improve overlay labels

ejoliet:
I've removed white background on the text labels and and made color the same as the overlay itself.
Please check that it is the case for the overlays in general (grid, markers, distance tool, compass).
Thanks!

### DIFF
--- a/src/firefly/js/visualize/draw/DirectionArrowDrawObj.js
+++ b/src/firefly/js/visualize/draw/DirectionArrowDrawObj.js
@@ -127,7 +127,8 @@ function drawDirectionArrow(ctx,drawTextAry,startPt,endPt,drawParams,renderOptio
 
     //FIXME:color text black on white background - yellow on white background is not readable
     //TODO: better solution would be to adapt text color with background
-    DrawUtil.drawText(drawTextAry, text, ret.textX, ret.textY,'black', renderOptions);
+    // DM-6950
+    DrawUtil.drawText(drawTextAry, text, ret.textX, ret.textY, color, renderOptions);
 }
 
 function toRegion(startPt,endPt,plot,drawParams,renderOptions) {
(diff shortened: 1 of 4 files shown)


src/firefly/js/visualize/draw/DirectionArrowDrawObj.js
```diff
@@ -127,7 +127,8 @@ function drawDirectionArrow(ctx,drawTextAry,startPt,endPt,drawParams,renderOptio
 
     //FIXME:color text black on white background - yellow on white background is not readable
     //TODO: better solution would be to adapt text color with background
-    DrawUtil.drawText(drawTextAry, text, ret.textX, ret.textY,'black', renderOptions);
+    // DM-6950
+    DrawUtil.drawText(drawTextAry, text, ret.textX, ret.textY, color, renderOptions);
```
robyww: you should take out the todo comments here since it is done.

src/firefly/js/visualize/draw/DrawUtil.js
```diff
@@ -116,14 +116,15 @@ function drawText(drawTextAry,text, x,y,color,
...
         fontWeight,
         fontStyle,
-        'backgroundColor': 'white',
-        'MozBorderRadius': '5px',
-        'borderRadius': '5px',
-        'WebkitBorderRadius': '5px'
+        //'backgroundColor': 'white',
+        'MozBorderRadius': '1px',
+        'borderRadius': '1px',
+        'WebkitBorderRadius': '1px'
```
robyww: I would still like to support background at the DrawUtil level.  I think you should add a extra parameter at the end (`backgroundColor`, default it to `null`  or `undefined`).  When a background color is not passed (the normal case), don't set a background color and don't set the padding.

src/firefly/js/visualize/draw/ShapeDataObj.js
```diff
@@ -633,7 +633,8 @@ export function drawText(drawObj, drawTextAry, plot, inPt, drawParams) {
         //TODO: better solution would be to adapt text color with background
 
         // in case shape 'text' defines the color of its own
-        var color = has(drawObj, 'color') ? drawObj.color : 'black';
+        // DM-6950: remove background white and update color with same used for the object itself (grid, distance tool, marker label,etc)
```
robyww: take out the DM-6950 comment. the ticket give us the history

robyww:
This all looks good. I think it is better than it was. Looks cleaner. My only issue is just keeping the support at the low level so that if we case a case we want it.

**Review Complete.**

PR #146, opened 2016-08-13
Dm 7038 js doc testing

lznakano:
Tatiana,

Please check out this one and then run under filefly:
   jsoc -c conf.json  

You should see the document under jsdoc.  Open jsdoc/index.html, you will see the new look of the low level api's. 

Cheers,
Lijun


docs/fireflyOverView.md
```diff
@@ -0,0 +1,219 @@
+# Firefly
```
loitly: This file is very similar to firefly-api-overview.md.  We can rearrange the content so that it can be used by both, instead of keeping 2 separate copies.

jsconf.json
```diff
@@ -0,0 +1,71 @@
...
+        "dictionaries": ["jsdoc","closure"]
+    },
+    "source": {
+        "include": [
+           "src/firefly/js/api/",
+           "src/firefly/js/tables/TableUtil.js",
+           "src/firefly/js/charts/XYPlotCntlr.js",
+           "src/firefly/js//visualize/CsysConverter.js",
+           "src/firefly/js/visualize/DrawLayerCntlr.js"
+          // ,"src/firefly/js/charts/HistogramCntlr.js"
```
loitly: json does not support comments.  jsdoc's parser allow for it?

package.json
```diff
@@ -84,7 +84,8 @@
     "style-loader": "^0.12.3",
     "url-loader": "^0.5.7",
     "webpack": "^1.8.2",
-    "react-addons-perf" : "0.14.7"
```
loitly: Is there a reason you removed react-addons-perf?

jsconf.json
ejoliet: true, a way to do it is to put the '//' inside the double quote, that looks more as a comment.

src/firefly/js/api/ApiBuild.js
```diff
@@ -49,16 +49,28 @@ import {buildViewerApi} from './ApiViewer.js';
 import './ApiStyle.css';
 
 
-/**
+/*
```
loitly: I am in favor of @tgoldina's suggestion; using @public instead of /** to show/hide functions from jsdoc.

src/firefly/js/api/ApiHighlevelBuild.js
```diff
@@ -123,10 +127,12 @@ function buildTablePart(llApi) {
      */
 
     /**
-     * The general plotting function to plot a table.
+     *@function showTable
      * @param {string|HTMLDivElement} targetDiv to put the table in.
      * @param {Object} request         request object created from
      * @param {TblOptions} options     table options.
+     * @memeberof firefly
+     * @example fierfly.showTable
```
loitly: This is nice.  It works great for the these API functions.  Too bad it does not recognize TblOptions @typedef which is declared just above it.  We'll have to figure out how to pull these typefefs in as well.

src/firefly/js/api/ApiUtil.js
```diff
@@ -42,15 +49,11 @@ export function debug(...msg) {
...
-
-
-
 /**
  *
  * @param {string|Object} div a div element or a string id of the div element
  * @param {Object} Component a react component
  * @param {Object} [props] props for the react component
- * @namespace firefly
  * @Function renderDOM
```
loitly: I hope these @function declaration are not necessary in these cases where it's clearly a function.

loitly:
In regard to jsconfig.json, I prefer it named jsdoc_config.json if you want to preserve the default 'config.json' name or just jsdoc.json.

It's looking very good so far.  Can't wait to see the final product.

src/firefly/js/api/ApiHighlevelBuild.js
tgoldina: Actually, it does see the type definition and generates Type Definitions
section correctly, but it does not recognize the namepath and does not
create link to the type definition. (See
https://github.com/jsdoc3/jsdoc/issues/969)

There are 3 solutions:

1. @param {module:firefly~TblOptions} table options

or

2. mark typedef @static and use
    @param {module:firefly.TblOptions} table options


3. inline object definition:

  @param {Object} TblOptions - table options
  @param {string}  TblOptions.tbl_group - the group this table belongs to.
Defaults to 'main'.
  @param {boolean} TblOptions.removable - true if this table can be removed
from view.  Defaults to true.
  @param {boolean} TblOptions.showUnits - defaults to false
  @param {boolean} TblOptions.showFilters - defaults to false
  @param {boolean} TblOptions.selectable - defaults to true
  @param {boolean} TblOptions.expandable - defaults to true
  @param {boolean} TblOptions.showToolbar - defaults to true
  @param {boolean} TblOptions.border - defaults to true

2016-08-15 17:33 GMT-07:00 loitly <notifications@github.com>:

tgoldina:
One thing I noticed, 'firefly' module methods are marked 'inner' (meaning module private) unless you explicitly mark them as '@memberof module:firefly'. Then they appear as static, which we want. It probably happens because JSDoc can not figure out how the methods are exported, since our exports are built dynamically. 

http://usejsdoc.org/about-namepaths.html#namepaths-in-jsdoc-3

src/firefly/js/api/ApiViewer.js
```diff
@@ -24,6 +25,11 @@ var viewerWindow;
  * Build the interface to remotely communicate to the firefly viewer
  * @return {{getViewer: getViewer, getExternalViewer: getExternalViewer}}
  */
+/**
+ * @desc  build  viewer api
+ * @module highLevelApi
```
tgoldina: remove this module declaration

src/firefly/js/api/ApiViewer.js
```diff
@@ -32,7 +38,9 @@ export function buildViewerApi() {
  *
  * @param {string} [channel] the channel id string, if not specified then one will be generated
  * @param file the html of the viewer to launch. In time there will be several
+ * @memeberof highLevel
```
tgoldina: should be @memberof module:firefly

src/firefly/js/api/ApiBuild.js
lznakano: jsdoc has a command line option (or can be added in the configuration file) to specify the levels access of the documents.  In the source code, we need to use @access tag to specify the levels.   After this ticket is merged, I will send out the list of frequent used tags so each of us can edit the source codes to put proper tags.

src/firefly/js/api/ApiUtil.js
lznakano: @function tag is only needed when it is an inner function, such as an inner arrow function etc.

lznakano: Yes, it is not necessary.  There ware still try/error things left in the source codes.  Usually the @function is not needed.

package.json
lznakano: I did not remover it.  It may be an accident when I cleaned up the package.json file.

jsconf.json
lznakano: Yes, it take // as a comment.

src/firefly/js/api/ApiViewer.js
lznakano: I will rename it as jsdoc_config.json.

lznakano:
If everyone agrees with layout/look &feel and the appearance,  I will rename the configuration file, add the deleted package.json entry back and then do a push again.   I will prepare a list of tags and examples.  Then we can decide what we want to expose, for example, we want to show a method as static or inner, how we want the object type to be listed etc.   Then each of us can edit a portion of the source codes and then have a configuration file to run one's own test.  For example, I can edit MouseReadout.jsx, ApiBuild.js..., then I create a a configuration file "jsdoc_config_lz.json" to include only those files, then I run jsdoc -c jsdoc_config_lz.json to see how the docs look like.   Once we are all done our editing, we can generate the jsdoc as a whole.

tgoldina:
I don't think we ever want to expose "inner" methods. Our exposed methods
should be static. We should mark them the right way to overcome JSDoc
confusion.

2016-08-16 9:56 GMT-07:00 Lijun Zhang <notifications@github.com>:

PR #147, opened 2016-08-15
Client-side filtering and table related cleanup

loitly:
https://jira.lsstcorp.org/browse/DM-7136
 - removes BasicTable
 - client-side filtering
 - added more jsdoc to table related code
 - fix filter input field auto-correct issue.

Added table options to basic table as a property.  Currently, options is available to fits headers as well as XYplot columns selections dialog.


src/firefly/js/tables/TablesCntlr.js
```diff
@@ -102,6 +102,99 @@ export const TABLE_UPDATE = `${DATA_PREFIX}.update`;
...
+ * @typedef {Object} TableGroupItem
+ * @prop {string}   tbl_group  table group name
+ * @prop {string}   tbl_id     unique id of the table data
+ * @prop {string}   tbl_ui_id  unique id of the table's UI data
+ * @prop {string}   title      title or label of the table
+ * @prop {boolean}  removable  true if this item can be removed from group.
+ * @prop {Object.<string, *>}   options   table options, ie.  selectable, expandable
+ */
+
+
```
ejoliet: Although the documentation above is dense and good, i don't think it will be written out in JSDocs html. What do you think?

ejoliet:
All ok. Removed deprecated component, client filtering added, basic table options nicely used in the FITS headers dialog. Love the jsdocs added!
Review done.

src/firefly/js/tables/TablesCntlr.js
loitly: JSDoc should be able to generate docs with `@typedef` and `@callback` tags.  They are standard tags.

PR #148, opened 2016-08-19
DM-7271: table enhancements

loitly:
DM-7271: table enhancements

 - do not apply default data format
 - introduce optional display format
 - add "AUTO" and "NONE" as format keywords.
 - for fits table, set display format to "%.9g" for double type.
 - add support for TDISPn fits table header
 - remove condition so temp files will not get written into source directory

When reading fits table, set display format to "%.9g" for double if TDISPn is not given.


src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
```diff
@@ -104,11 +104,38 @@ public static JSONObject toJsonTableRequest(TableServerRequest req) {
...
+        DataType[] columns = data.getDataDefinitions();
+        for (int colIdx = 0; colIdx < columns.length; colIdx++) {
+            DataType dt = columns[colIdx];
+            String fkey = DataSetParser.makeAttribKey(DataSetParser.FORMAT_DISP_TAG, dt.getKeyName());
+            if (tableDef.contains(fkey)) {
+                dt.getFormatInfo().setDataFormat(tableDef.getAttribute(fkey).getValue());
+                dt.getFormatInfo().setWidth(Arrays.stream(new int[]{
+                                dt.getKeyName().length(),
+                                dt.getDataUnit().length(),
+                                dt.getTypeDesc().length()}).max().getAsInt());
```
tgoldina: Can data unit or type desc be null?

tgoldina:
Review complete. Looks good. 

As we talked, it would be nice to save display format in decimated data too. Currently, when catalog is loaded from a FITS table, XYPlot points have higher precision than table data, which can result in a gap between the displayed point and the highlight when the resolution is high or there is only one point.  

Later we might want to carry server-side precision to the client and use display format only to format.

tgoldina:
Review complete. Works as expected for the 3 table types I've tested: IRAS catalog, FITS binary, uploaded IPAC table without type info.

PR #149, opened 2016-08-20
DM-7190: fix the issues related to add region, highlight box while zooming the image, and update the dispatch functions related to add/remove region entry.…

cwang2016:
region bugs fixing and enhancement:
- the size of highlight box is updated accordingly while zooming the image
- add any number of regions to a drawing layer. 
- In order to meet the need for adding the regions to a plot layer, the region related actions are updated as below:
# dispatchCreateRegionLayer: create the plot layer containing regions per region file or region description, no empty region layer is not supported. 
# dispatchAddRegionEntry: add regions to an existing plot layer or a new plot layer will be created automatically to contain the added regions in case the specified layer does not exist.
# dispatchRemoveRegionLayer: removes the specified regions from a plot layer and the plot layer is destroyed automatically in case all regions are removed from that plot layer.

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-test.html
+++ b/src/firefly/html/demo/ffapi-highlevel-test.html
@@ -77,7 +77,7 @@
 </div>
 <div>
     <a href='javascript:addRegionEntry()'>Add Region Entry </a>
-    <select id='regionEntry_add'><option selected disabled>Regiion Layers:</option></select>
+    <select id='regionEntry_add'><option selected disabled>Region Layers:</option></select>
     <a href='javascript:removeRegionEntry()'>Remove Region Entry</a>
     <select id='regionEntry_remove'><option selected disabled>Regiion Layers:</option></select>
     <div>
@@ -266,6 +266,7 @@
                 'PHYSICAL;box (360, 220, 60, 60, 0) #color=orange text={ph3}'
             ];
 
+            window.emptyRegion = '';
             window.regionAry2 = [
                 'J2000',
                 'boxcircle point    202.45    47.262 # color=red text="pt boxcircle 6" delete=0',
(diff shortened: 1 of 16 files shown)


cwang2016:
Test method: 
- start localhost:8080/firefly/demo/ffapi-highlevel-test.html
- click 'test open'
- click 'Create Region Layer'  on 'Region_Plot_1' (near the bottom of page) 
- click 'Create Region Layer' on 'Region_Plot_2'
   ==> two drawing layers (Region_Plot_1 & Region_Plot_2) are rendered
- selecting 'Region_Plot_2' from the dropdown list by 'Remove Region Entry', then click 'Remove Region Entry'
   ==> drawing layer 'Region_Plot_2' is gone (all regions are removed from the layer)
- selecting 'Region_Plot_2' from the dropdown list by 'Add Region Entry', then click 'Add Region Entry'
    ==>  drawing layer 'Region_Plot_2' is created again. (the plot layer is created to contain the regions to be added)

tgoldina:
Works as expected. A few minor things I noticed:

1. There is a small gap between highlighted outline and a non-rotated box region. It seems to get bigger when I zoom image, but disappears when I select another region, and select this one again.

2. We might want to update docs (docs/firefly-api-overview and JSDoc) with new dispatchAddRegionLayer signature and correct statements for missing plotId in dispatchCreateRegionLayer and dispatchAddRegionLayer:  "if no plotId is given, the region layer is created on all plots" -> "if no plot Id is given, the region layer is created on all plots of the active plot group". (If plots are in different plot groups the region will only appear on an active plot.)

tgoldina:
One more thing I noticed on the test page is that when I expand the image with region layers, they originally appear, but then disappear. In fact, they seem to disappear every time 'Zoom to fit' is clicked.

cwang2016:
Yes, Just discovered that the content of all kinds of drawing layer got disappeared when the 'zoom to fit' is clicked. Will dig into this issue. Thanks.

cwang2016:
the content of some layers got disappeared when the 'zoom to fit' button is clicked.

cwang2016:
More fixings: 
- highlight box consistently wraps around the region while the image is zoomed. 
- the drawing layer content consistently appears  in various image zooming mode including 'zoom to fit'.

PR #150, opened 2016-08-20
DM-7275: fixed expanded modes display issues

robyww:
- always does single mode when the is only one plot
  - better use of expanded mode toolbar space using flex layout
  - fixed: color change failing when mask is plotted

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-test.html
+++ b/src/firefly/html/demo/ffapi-highlevel-test.html
@@ -108,7 +108,7 @@
             var req= {
                 plotId: 'xxq',
                 Type      : 'SERVICE',
-//                plotGroupId : 'myGroup',
+                plotGroupId : 'myGroup',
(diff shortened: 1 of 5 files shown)


ejoliet:
Thanks for looking at it!
I've checked that calling external viewer gives correctly single view mode on one image and tiled view on next images.
But when trying the expand mode on the 2x2 grid from here:

I can only see one image (single view?) and options to switch to tiled view or available list are no longer displayed. Could you check?

robyww:
You are right.  The ffapi-highlevel-test.html has all the images in different groups. The options button should have been there to choose the images.  However, there is something still not right in how we are doing grouping, I am going to look at it a little more.

robyww:
@ejoliet OK, i think i have fixed the bugs. You are welcome to look at it again or I can go ahead and merge.

ejoliet:
Looks good. Review completed.

PR #151, opened 2016-08-25
DM-7239: Table reset does not reset to original settings.

loitly:
- reset options based on original settings.
 - do not reload document unless it is needed.
 - do not redraw coverage with table is sorted

Reset should work like it did before.
Show units, show filters, column selections, and filters will be reset to what it was originally when table was loaded.  On table sort, coverage image will not reset.  However, it will be reset on filters changed.  This is how it is in production as well.


ejoliet:
Thanks!. Works as before migration. Review done.

PR #153, opened 2016-08-25
DM-7317 Log scale doesn't work with sample table attached

tgoldina:
- Fixed log scale issue (when minimum value is close to 0
- Added more support for error conditions (still no way to catch errors in render)
- More validation for log scale in XYOptions
- Trey's changes for MULTI_VALUE_CHANGE - field reducer is now called once on form submit

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -120,7 +120,7 @@ function colWithName(cols, name) {
 function getNumericCols(cols) {
     const ncols = [];
     cols.forEach((c) => {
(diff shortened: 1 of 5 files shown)


src/firefly/js/charts/ChartUtil.js
```diff
@@ -153,7 +153,7 @@ export function getDefaultXYPlotParams(tbl_id) {
     // otherwise use the first one-two numeric columns
     if (!isCatalog) {
         const numericCols = getNumericCols(tableData.columns);
-        if (numericCols.length > 2) {
+        if (numericCols.length >= 2) {
```
ejoliet: That's a good fix too! Thanks!

ejoliet:
Works ok to me. Review done.

PR #154, opened 2016-08-25
DM-7280: display wrong data when 'hidden' columns are in front of visible ones

loitly:
- fix hidden column problem
  - TabPanel
     - fix annoying gap between tabs and content
     - fix cursor shows click-able at edges, but nothing happens when clicked


ejoliet:
Column problem fixed. Also the UI edge annoying problem is solved. All good. Review done.

PR #155, opened 2016-08-26
DM-7277: complete commit and fix

ejoliet:
Please, review and let me know.

I've hidden the description column in xy-plot selecting column dialog (same behavior as in table component) when no description available.
This is the case in Gator/Atlas, which both use upload table mechanism in the API. In OPS, before migration, no description is shown in the table/xy-plot component.

To test, upload any ipac table catalog, this is not getting any description from the ipac table header, so nothing should be shown.


wmiipac:
works as expected. good job. review completed.

PR #156, opened 2016-08-26
DM-6289 Chart options - reset to default

tgoldina:
- Fixed truncated columns in table statistics processor (when col name is long)
- Fixed NaN related issues in histogram processor
- Reset in xyplot or histogram options resets options to the default.

How do we know the options are the default?

- Chart is created from the menu drop down, options selected there are "the default"
- Default chart, created with the default parameters
- Chart is created via API, options passed via the API function are "the default"

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
@@ -37,6 +37,13 @@ public class HistogramProcessor extends IpacTablePartProcessor {
         new DataType("binMin", Double.class),
         new DataType("binMax", Double.class)
     };
+    static {
+        // set default precision to 9 significant digits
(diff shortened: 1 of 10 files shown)


ejoliet:
I've tested the reset feature. It restore the default values that appeared the first time it displays the plot.

One caveat though: i did change values randomly and then reset, all is restored and plot is displayed ok. But labels are not replaced in the option panel when you reset, only after you apply and open the option panel back.

tgoldina:
When label and unit are set by the user (for example in chart drop down), they will be preserved on Reset.

When label and unit are not specified (empty), they are not part of the default parameters. When the chart is created, and no label or unit are present, the values from the table header will be used. If showing them in options creates confusion, we could probably remove them from table options.

Making unknown label and unit part of the default parameters is problematic, because the table is not always loaded at the time the plot is requested. (Like in API case.)

By the way, label and unit are automatically cleared, whenever the columns are changing.

PR #157, opened 2016-08-27
DM-7406: Create Light curve skeleton app

loitly:
- introduce template into firefly entry point.
 - create light curve as a template.
 - create viewer, controller, and results component for light curve.
   - these components are place holders and are not fully implemented.
 - create upload form for use during development.

To view light curver viewer use lc.html:  ie. http://localhost:8080/firefly/lc.html
Upload a table to see the results.

### DIFF
--- a/src/firefly/html/lc.html
+++ b/src/firefly/html/lc.html
@@ -1,0 +1,36 @@
+<!doctype html>
+<html>
+
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <link rel="icon" type="image/x-icon" href="images/fftools-logo-16x16.png">
+    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <title>Light Curve Viewer</title>
+
+    <script>
+        window.firefly = {
+            app: {
+                template: 'LightCurveViewer',
+                menu: [
+                    {label:'Upload', action:'LCUpload'},
+                    {label:'Catalogs', action:'IrsaCatalogDropDown'},
+                    {label:'Help', action:'app_data.helpLoad', type:'COMMAND'},
+                    ]
+            },
+        };
+    </script>
+    <script  type="text/javascript" src="firefly_loader.js"></script>
+</head>
+
+<body style="margin: 0; background-color: rgb(200,200,200)">
+<!-- attached loaction for firefly app -->
+<div id='app'/>
+
+
+
+
+</body>
+
+</html>
+
+
--- a/src/firefly/js/FFEntryPoint.js
+++ b/src/firefly/js/FFEntryPoint.js
@@ -8,7 +8,7 @@ import ReactDOM from 'react-dom';
 import {get} from 'lodash';
 
 import {firefly} from './Firefly.js';
-import {FireflyViewer} from './core/FireflyViewer.js';
+import {FireflyViewer} from './templates/fireflyviewer/FireflyViewer.js';
 import {LcViewer} from './templates/lightcurve/LcViewer.jsx';
 import {initApi} from './api/ApiBuild.js';
 import {HELP_LOAD} from './core/AppDataCntlr.js';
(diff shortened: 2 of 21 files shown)


ejoliet:
Looks great!!! Nice starter page! I've uploaded a file and checked that everything was ok.
LC layout and skeleton is good. Review done.

src/firefly/html/lc.html
```diff
@@ -0,0 +1,36 @@
...
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <link rel="icon" type="image/x-icon" href="images/fftools-logo-16x16.png">
+    <meta name="apple-mobile-web-app-capable" content="yes" />
+    <title>Light Curve Viewer</title>
+
+    <script>
+        window.firefly = {
+            app: {
+                template: 'LightCurveViewer',
```
robyww: The triview is the default template?  I think we should probably specify it.

src/firefly/js/FFEntryPoint.js
```diff
@@ -51,9 +66,10 @@ if (app) {
                 {label:'Example Js Dialog', action:'exampleDialog', type:'COMMAND'}
         ]
     };
-    const props = Object.assign({}, defProps, app);
+    const props = Object.assign(defaults, app);
+    const viewer = app.template === Template.LCV ? LcViewer : FireflyViewer;
```
robyww: should this line and the constants above be in a some sort of factory?

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -0,0 +1,117 @@
...
+        var {expanded, standard, closeable} = mode || {};
+        const content = {};
+        var visToolbar;
+        if (showImages) {
+            visToolbar = <VisToolbar key='res-vis-tb'/>;
+            content.imagePlot = (<TriViewImageSection key='res-tri-img'
+                                        closeable={closeable}
+                                        imageExpandedMode={expanded===LO_VIEW.images}
+                                        {...images}  />);
+        }
```
robyww: I don't think you want a TriViewImageSection here.  Must a MultiImageViewer.  If we want coverage and user added image in the LCViewer the we probably need to generalize TriViewImageSection to another component.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -0,0 +1,209 @@
+/*
```
robyww: LcManager is mostly a placeholder, right?

robyww:
Is LayoutCntlr more general than just the tri-view? I know it is required for expanded but it is not clear if is for others.  Maybe I don't understand it well enough.  I am having trouble which what goes in LayoutCntlr.

robyww:
should the triview stuff be moved to templates/tri-view?

robyww:
Conclusions:

- The structure looks good.  It provides both flexibility and simplicity.
- I think we might want to have everything about a templates together: constant and code. Maybe we import a template object that has the constant and the React Element?
- I think I need to further separate the saga from the TriViewImageSection and rename it.
- I need to understand LayoutCntrl better, I have used with out enough understanding.

I think the only thing you should change for this ticket is the template organization (tri-view in templates directory), other stuff we can iterate on).

This is really good path forward.  
Review Complete.

src/firefly/html/lc.html
loitly: Yes, FireflyViewer is the default.  It's documented via jsdoc in FFEntryPoint.app.

src/firefly/js/FFEntryPoint.js
loitly: I was thinking just a private map.  I don't think anyone else need to know about this.

src/firefly/js/templates/lightcurve/LcResult.jsx
loitly: I am sure you are right.  TriViewImageSection is just a placeholder.  I did not load the saga.  Beside the raw table, all of the other components are placeholders as well.

src/firefly/js/templates/lightcurve/LcManager.js
loitly: Yes, it will be a complete rewrite.  There is a jira ticket for this.

PR #159, opened 2016-08-31
DM-7448: Table reset unset option 'Show unit' initially set

loitly:
https://jira.lsstcorp.org/browse/DM-7448

Reset button does not work when options are set explicitly.

This the demo page below, the first table has showUnits and showFilters set to true.  It used to work reset properly.  


ejoliet:
The option 'show unit' and 'filters' are kept after reset if they were selected initially. But if the page size is '15', after reset it returned to '100'. The reset should return '15' if this was the inital value.
Example: http://localhost:8080/firefly/demo/ffapi-highlevel-test.html,
'open table' will open external viewer with table page size set to 15. 'Reset' button reset the value to 100 instead of keeping the initial value in that case '15'.
I think is still a bug if we think about how reset should restore initial values. What do you think?

ejoliet:
Has been fixed now. Thanks! Review completed.

PR #161, opened 2016-08-31
DM-7162: adding first skeleton and test for lightcurve handler

ejoliet:
I've added two processors (maybe it can be combined). They are both under firefly/server/query.

One is taking care of returning periodogram and peaks tables. The other is in charge of phase folding an input IPAC table (raw LC table).

I've added a handler to deal with the Nexsci API in the future. Ans an IRSA handler to deal with the extraction/parsing of both tables from the API result VOTable xml.
There is also a test to demonstrate how to get the tables out of a request.

For now, i'm faking the return API with a sample VOTable.

The request object (PeriodogramAPIRequest) expects input parameters needed for computing the periodogram (calling the API) and phase folding a raw light curve. (Note: Maybe we need to add the URL as part of the request so we can always change the API url directly from the client instead of doing it from the server?)

The phase folding is under development and need to be integrated in the handler. For now i'm returning the same raw table that i've passed in as input in the request.

I've added the test buttons in firefly-dev.html, they appear in the tab 'Compute periodogram' in the TestQueriesPanel (.jsx) component.

More offline discussion might be needed though.

Please, let me know if that make sense regarding the feature requirements and if it covers the future need.


src/firefly/java/edu/caltech/ipac/firefly/server/query/AlgorithmDefinition.java
```diff
@@ -0,0 +1,29 @@
+package edu.caltech.ipac.firefly.server.query;
```
loitly: Please include on top of file the caltech/ipac license's information.  See other source file for example.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AlgorithmDefinition.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Created by ejoliet on 8/22/16.
+ */
+public enum AlgorithmDefinition {
```
loitly: Class name is too generic.  Could be inner static of Periodogram.

src/firefly/java/edu/caltech/ipac/firefly/server/query/IrsaLightCurveHandler.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+package edu.caltech.ipac.firefly.server.query;
```
loitly: We need to find a better place to put these classes.

src/firefly/java/edu/caltech/ipac/firefly/server/query/IrsaLightCurveHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+/**
+ * .
+ * Should handle the LC transformations to get files out of the API result VOtable xml
+ *
+ * @author ejoliet
+ * @see PeriodogramAPIRequest
+ */
+public class IrsaLightCurveHandler extends NexsciLcApiHandler {
```
loitly: Is there a reason for subclassing?  I don't see multiple versions of nexsci handlers.

src/firefly/java/edu/caltech/ipac/firefly/server/query/LightCurveAPIHandler.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.io.File;
+
+/**
+ * Created by ejoliet on 8/22/16.
+ * API handler to deal with result from calling LC API.
+ * Result is a file or several that contain the power vs period and peaks table
+ * The prefered output is votable containing those 2 tables
+ */
+public abstract class LightCurveAPIHandler implements LightCurveHandler {
```
loitly: This is not used.  Not sure what the usage of this would be.

loitly:
Beside the placement of the added classes, I would also like to see how they can be simplified.
The multiple level of abstractions may not be necessary.  Can we talk about it?

ejoliet:
Cleanup and move to .lc package. Please, have a look.

loitly:
It looks good.  Review is complete.

PR #162, opened 2016-09-01
DM-7460 No help link/anchor option for XY-Plot

tgoldina:
Provide help_id option to define help anchor string.
If firefly is built locally, an example can be found in the first chart at


ejoliet:
Works as expected. Review done.

PR #166, opened 2016-09-07
Dm 7508 eslint rules

lznakano:
Please take a look at .eslintrc file.  It lists a set of rules.  Only parameter is required, the rest rules are set to warning (0:off, 1:warning, 2:true).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,14 +1,28 @@
 {
   "plugins": [
-     "react"
+     "react",
+    "jsdoc"
+
    ],
+  "settings": {
+    "jsdoc": {
+      "tagNamePreference": {
+        "param": "param",
+        "returns": "returns"
+      }
+    }
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ],
   "env": {
     "es6": true,
     "browser": true
   },
   "parserOptions": {
-    "ecmaVersion": 7,
-    "ecmaFeatures": {
+     "ecmaVersion": 7,
+     "ecmaFeatures": {
       "arrowFunctions" : true,
       "binaryLiterals" : true,
       "blockBindings": true,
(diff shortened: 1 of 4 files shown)


.eslintrc
```diff
@@ -64,7 +75,20 @@
...
+    "jsdoc/check-tag-names": 1,
+    "jsdoc/check-types": 1,
+    "jsdoc/newline-after-description": 1,
+    "jsdoc/require-description-complete-sentence": 1,
+    "jsdoc/require-hyphen-before-param-description": 1,
+    "jsdoc/require-param": 2,
+    "jsdoc/require-param-description": 1,
+    "jsdoc/require-param-type": 1,
+    "jsdoc/require-returns-description": 1,
+    "jsdoc/require-returns-type": 1
```
tgoldina: We might want to relax the rules a bit, otherwise we'll have too many warnings, and people will not pay attention to real problems.

I suggest the following defaults:
    "jsdoc/check-param-names": 2,
    "jsdoc/check-tag-names": 1,
    "jsdoc/check-types": 1,
    "jsdoc/newline-after-description": 0,
    "jsdoc/require-description-complete-sentence": 0,
    "jsdoc/require-hyphen-before-param-description": 0,
    "jsdoc/require-param": 2,
    "jsdoc/require-param-description": 0,
    "jsdoc/require-param-type": 0,
    "jsdoc/require-returns-description": 0,
    "jsdoc/require-returns-type": 1

We would still want to preserve those rules in the file, so that people who work on API documentation can turn them on to make docs consistent.

Reference link: https://github.com/gajus/eslint-plugin-jsdoc

tgoldina:
Other than the defaults everything looks great. 

I noticed that LayoutCntlr.js is failing because the following function signature format is unsupported by eslint-plugin-jsdoc: 
  dispatchShowDropDown({view}={}). 

Unfortunately, the rules that require function signature parsing are the most useful:
  "jsdoc/check-param-names": 2,
  "jsdoc/require-param": 2,

tgoldina:
Review complete. 

As Loi said, we can fix LayoutCntlr.js
  dispatchShowDropDown({view}={}) -> dispatchShowDropDown({view})
and move on.

PR #168, opened 2016-09-09
DM-7543: color backgrnd image viewer inline tool change

ejoliet:
Please, have a look to this change. The ticket suppose to solve the consistency with the rest of the color to be more consistent and readable. Blue is not consistent with the rest (grey) and the zoom level is not readable. Sometimes the inline tool icons (top right-hand side) are not apparent enough.

This is step 1 change: making the background to be the same grey color on image viewer left and right, and plot title black, zoom level label red.

Another proposed solution is step 2:  black font on white background if step 1 doesn't solve the issue.

Thanks.


xiuqin:
Build the branch and image title "black on gray" looks OK. I don't remember we want to change the zoom level to color red. It is not changed anyway. let's leave it since it is the same green color in IRSA operation site.  

Review done.

PR #171, opened 2016-09-09
DM-4498: WCS match

robyww:
This commit implements **WCS match** in firefly and make some tweeks to the code:
- updates JSDocs in many, many places
- includes some code cleanup so we can eventually remove underscore.string
- deletes an unused file
- moves a couple of files

Reviewers should only have to concentrate on changes to `ImageViewerLayout.js`, `HandlePlotChange.js`, `PlotView.js`, and the one new file `WcsMatchTask.js`

WCS match seems to have many edge cases so I probably have missed some.  Testing it out is important.


ejoliet:
I just started but WCS seems to give an error in the console and image disappear.
TypeError: Cannot read property 'x' of null
    at Object.findScrollPtForImagePt (PlotView.js:620)
    at installTiles (HandlePlotChange.js:220)
    at reducer (HandlePlotChange.js:76)
    at reducer (ImagePlotCntlr.js:854)
    at combination (combineReducers.js:117)
    at dispatch (createStore.js:179)
    at middleware.js:52
    at index.js:14
    at dispatch (applyMiddleware.js:45)
    at processZoomSuccess (ZoomUtil.js:228)

Step to reproduce: 
- image search on band 1 and band 4 around m43. Then click on 'wcs match'. The band 1 image disappear and console throw log error reported above.

Will try to check other things in the meantime.

xiuqin:
- When switching to single image mode, should the image retain its original zoom? 
- The arrows to go through the images are not on the same line. screenshot attached.
dm-4498
- Sometimes a semi-transparent window shows up with message "success" and an OK button. Is it for debug purpose?

robyww:
> When switching to single image mode, should the image retain its original zoom?

I found and issue: when switching between single and grid the WCS match gets out of sync.

> Sometimes a semi-transparent window shows up with message "success" and an OK button. Is it for debug purpose?

This should not happen, how do I reproduce it?

ejoliet:
Overall, i did check that rotation and scale is preserved correctly. 
I didn't see the message 'success' so i can't reproduce that. But i could see the misalignment of the arrows.

Images are still out of sync in single image mode when i rotated the image and went to the next one. I had to unclick and click back again the 'WCS' checkbox to get the images in sync. Could you check?

robyww:
There are certain cases where I automatically turn off wcs sync (i.e. image add or remove). I should do the same thing with rotate and crop.

xiuqin:
I could not make the message "Success" appear any more.
A new issue: get 2MASS J. H, K images at different size, say 0.11, 0.12, 0.13 deg.
rotate the first one in grid mode, then click on 1X for zoom, the image disappears. It will be back when zoom up (+).  I tried twice with sequence K,H,J, the same thing happened to the first image.

robyww:
@xiuqin - I think both the bugs you were found were fixed in this last commit.

PR #172, opened 2016-09-10
Dm 7160 lc input form

wmiipac:
The light curve input forms are not connected to the processors. They are incomplete at this point, many parameters need to be add when adding the processors and backend APIs.


ejoliet:
I've switch to your branch but don't see anything, seems an error loading:

Error: Cannot find module "./LCPanels.css"

Could you check?

wmiipac:
Oh, I will add in. It was not used in the current panels, but will 
probably be used when we add more fields.

src/firefly/js/templates/lightcurve/LCPanels.css
```diff
@@ -0,0 +1,26 @@
...
+    min-width: 400px;
+    height: 300px;
+    min-height: 300px;
+    overflow: auto;
+    background-color: #b3edff;
+    padding: 2px;
+    position: relative
+}
+
+.Header {
```
loitly: CSS style name should follow BEM guildline: http://getbem.com/introduction/
Generic name like this may clash with others.
We use '__' for blocks separator and '--' for modifiers.
ie.  LCPanels__Header[--if-needed]

wmiipac: Good point, noted.

ejoliet:
Couple of changes we might need to do to be more consistent with the LC workflow and wordings:

- 'Period finding' -> 'Periodogram'
- in this tab 'Period finding': 'search' -> 'compute periodogram' 
- 'LC param' -> 'Phase folding'
- Remove tab 'Plotting options'

Thanks!

ejoliet:
Yes agree. Review done.

PR #174, opened 2016-09-13
DM-7593: use phase folded code in processor and update test

ejoliet:
Please review this PR which make use of the phase folded code, updating phase folded processor and tests.
To see in action, go to 

http://localhost:8080/firefly/firefly-dev.html

And click 'compute periodogram' tab.
Change period and phase fold.

The phase folded request built can be seen here: js/ui/TestQueriesPanel.jsx:142

The phase column displays '0.0' which is not enough (see ticket [DM-7593])

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/lc/IrsaLightCurveHandler.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/lc/IrsaLightCurveHandler.java
@@ -3,8 +3,10 @@
  */
 package edu.caltech.ipac.firefly.server.query.lc;
 
+import edu.caltech.ipac.astro.IpacTableException;
 import edu.caltech.ipac.astro.IpacTableWriter;
 import edu.caltech.ipac.firefly.server.ServerContext;
+import edu.caltech.ipac.firefly.server.util.ipactable.DataGroupReader;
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/lc/IrsaLightCurveHandler.java
```diff
@@ -51,15 +53,33 @@ public File getPeaksTable(PeriodogramAPIRequest request) {
...
+        try {
+            DataGroup dg = DataGroupReader.readAnyFormat(tbl);
+            PhaseFoldedLightCurve plc = new PhaseFoldedLightCurve();
+            plc.addPhaseCol(dg, period, timeColName);
+            tempFile = createPhaseFoldedTempFile();
+            IpacTableWriter.save(tempFile, dg);
+        } catch (IpacTableException e) {
+            e.printStackTrace();
+        } catch (IOException e) {
+            e.printStackTrace();
```
loitly: use Logger instead of stdOut.

loitly:
I understand it was needed at first.  But, now that lc app(lc.html) is available, this 'demo' code should be place there instead.

Code looks fine.  Review is complete.

PR #176, opened 2016-09-13
DM-7470: Tri-view coverage now working more like the GWT version

robyww:
fixes issues with DM-7470


ejoliet:
Works as expected. I only saw an issue with the zoom level when combining 2 different images with different scale and opening them in grid expanded mode, it seems to give wrong zoom level to one of the image, making it very tiny.

PR #177, opened 2016-09-13
DM-7462: create dispatch function, rendering style and stored status  for region selection

cwang2016:
Implementation: 
- create region selection dispatch function, dispatchSelectRegion.
- create status to record the text description of selected region and provide function to access the status. 
- add selectMode parameter into dispatchCreateRegionLayer & dispatchAddRegionEntry functions which defines the rendering style, rendering color and line width for the selected region. 5 types of rendering styles are defined, 'UprightBox', 'DottedOverlay', 'SolidOverlay', 'DottedReplace' and 'SolidReplace'. 

Testing: 
- from localhost:8080/firefly/demo/ffapi-highlevel-test.html, 
   1. click 'test open'
   2. find color,  line width text boxes and rendering style dropdown list next to 'Create Region Layer' to set the display features for showing the selected region, then click 'Create Region Layer' to create region layer with the id shown on the box next to it. 
    3. go down to the bottom of the page to select the region (one region description is already put into the textarea box) 
        - select the region layer from the dropdown list, 
        - click 'Select the following region(s)'. (note: current version support single region selection)
    You may delete the region layer and create a new region layer to see the different rendering effect on the selected region. (set the style before clicking 'Create Region Layer') 

- from localhost:8080/firefly
       1. using image select panelt to select two images, m51 & wise, m51 & 2mass
       2. upload region file, test5.reg,  from firefly_test_file/region_test_files
       3. click any region to see the region is highlighted over all plots. 
    (the default rendering style for selected region is applied in here).

### DIFF
--- a/src/firefly/js/drawingLayers/RegionPlot.js
+++ b/src/firefly/js/drawingLayers/RegionPlot.js
@@ -6,19 +6,25 @@ import {makeDrawingDef} from '../visualize/draw/DrawingDef.js';
 import DrawLayer, {DataTypes, ColorChangeType}  from '../visualize/draw/DrawLayer.js';
 import {makeFactoryDef} from '../visualize/draw/DrawLayerFactory.js';
 import {drawRegions} from '../visualize/region/RegionDrawer.js';
-import {addNewRegion, removeRegion} from '../visualize/region/RegionUtil.js';
+import {getRegionIndex, addNewRegion, removeRegion} from '../visualize/region/RegionUtil.js';
 import {RegionFactory} from '../visualize/region/RegionFactory.js';
 import {primePlot, getDrawLayerById} from '../visualize/PlotViewUtil.js';
 import ImagePlotCntlr, {visRoot} from '../visualize/ImagePlotCntlr.js';
 import {MouseState} from '../visualize/VisMouseSync.js';
 import DrawOp from '../visualize/draw/DrawOp.js';
 import DrawLayerCntlr, {DRAWING_LAYER_KEY,
+                        defaultRegionSelectColor,
+                        RegionSelectStyle,
+                        getRegionSelectStyle,
                         dispatchModifyCustomField,
                         dispatchAddRegionEntry,
                         dispatchDeleteRegionLayer,
-                        dispatchRemoveRegionEntry, dlRoot} from '../visualize/DrawLayerCntlr.js';
-import {get, set, isEmpty} from 'lodash';
+                        dispatchRemoveRegionEntry,
+                        dispatchSelectRegion,
+                        dlRoot} from '../visualize/DrawLayerCntlr.js';
+import {get, set, has, isEmpty, isString, isArray} from 'lodash';
 import {dispatchAddSaga} from '../core/MasterSaga.js';
+import {flux} from '../Firefly.js';
 
 const ID= 'REGION_PLOT';
 const TYPE_ID= 'REGION_PLOT_TYPE';
@@ -34,9 +40,10 @@ function* regionsRemoveSaga({id, plotId}, dispatch, getState) {
                                      DrawLayerCntlr.DETACH_LAYER_FROM_PLOT]);
 
             if (action.payload.drawLayerId === id) {
+                var dl;
                 switch (action.type) {
                     case  DrawLayerCntlr.REGION_REMOVE_ENTRY :
-                        var dl = getDrawLayerById(getState()[DRAWING_LAYER_KEY], id);
+                        dl = getDrawLayerById(getState()[DRAWING_LAYER_KEY], id);
                         if (dl && isEmpty(get(dl, 'drawObjAry'))) {
                             dispatchDeleteRegionLayer(id, plotId);
                         }
@@ -78,7 +85,8 @@ function creator(initPayload) {
     };
 
     var actionTypes = [DrawLayerCntlr.REGION_ADD_ENTRY,
-                       DrawLayerCntlr.REGION_REMOVE_ENTRY];
+                       DrawLayerCntlr.REGION_REMOVE_ENTRY,
+                       DrawLayerCntlr.REGION_SELECT];
 
     const id = get(initPayload, 'drawLayerId', `${ID}-${idCnt}`);
     var   dl = DrawLayer.makeDrawLayer( id, TYPE_ID, get(initPayload, 'title', 'Region Plot'),
@@ -87,6 +95,7 @@ function creator(initPayload) {
     dl.regionAry = get(initPayload, 'regionAry', null);
     dl.dataFrom = get(initPayload, 'dataFrom', 'ds9');
     dl.highlightedRegion = get(initPayload, 'highlightedRegion', null);
+    dl.selectMode = get(initPayload, 'selectMode');
 
     dispatchAddSaga(regionsRemoveSaga, {id, plotId: get(initPayload, 'plotId')});
     idCnt++;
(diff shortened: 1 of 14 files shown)


cwang2016:
- Please recompile firefly.

src/firefly/js/visualize/DrawLayerCntlr.js
```diff
@@ -287,27 +301,39 @@ export function dispatchDetachLayerFromPlot(id,plotId, detachPlotGroup=false,
...
- * @param regionAry
- * @param plotId The region layer is created on all plots of the active plot group in plotId is empty
- * @param dispatcher
+ * @summary Create drawing layer based on region file or region description
+ * @param {string} drawLayerId id of the drawing layer to be created, required
+ * @param {string} layerTitle  if it is empty, it will be created internally
+ * @param {string} fileOnServer region file name on server
+ * @param {string[]|string} regionAry  array or string of region description
+ * @param {string[]|string} plotId  array or string of plot id. If plotId is empty, all plots of the active group are applied.
+ * @param {function} dispatcher
```
tgoldina: should be last param

In general, there quite a few eslint warnings/errors (with our .eslintrc) in the changed files, like unused imports, vars, etc. It would be beneficial to fix them.

src/firefly/js/visualize/DrawLayerCntlr.js
```diff
@@ -287,27 +301,39 @@ export function dispatchDetachLayerFromPlot(id,plotId, detachPlotGroup=false,
...
+ * @param {Object} selectMode    redndering features for the selected region
```
tgoldina: Looks like I can not omit selectMode parameter, I always have to specify it, even if I'd like to use the default settings. Unfortunately, eslint does not support the syntax below, which would be beneficial in this case:
{selectStyle = defaultRegionSelectStyle, selectColor = defaultRegionSelectColor, lineWidth = 0 }={}

Perhaps we can do the destructuring assignment in the method body?

We should also add the default values to jsdoc, ex. 
@param {string} [selectMode.selectStyle='UprightBox'] rendering style for the selected region ...

src/firefly/js/visualize/DrawLayerCntlr.js
```diff
@@ -316,33 +342,64 @@ export function dispatchDeleteRegionLayer(drawLayerId, plotId, dispatcher = flux
...
+ * @summary select region from a drawing layer containing regions
+ * @param {string} drawLayerId  id of drawing layer where the region is selected from, required.
+ * @param {string[]|string} selectedRegion array or string of region description, currently only single region is allowed
+ * to be selected if the array contains the description of multiple regions.
+ * @param {function} dispatcher
+ * @public
+ * @func dispatchSelectRegion
+ * @memberof firefly.action
+ */
+export function dispatchSelectRegion(drawLayerId, selectedRegion, dispatcher = flux.process) {
```
tgoldina: How do I know the plotId of the image, which triggered the selection?
Should we add plotId to REGION_SELECT action payload? Or should we allow API user to get/set active plotId via API? (We do have getPrimePlot(), which allows to get active plot, but nothing to set.)

cwang2016: This applies to all plots attached to the region layers as specified.

tgoldina: Also, when clicking on a region, action payload is region object, which user does not know how to parse. Can it include user string? Or normalized string?

tgoldina:
1. Select mode works great.

2. My major concern about REGION_SELECT action is that selectedRegion in its payload is an object. It's not meaningful in the context of user's callback function (like onRegionSelect callback below): 

firefly.util.addActionListener(firefly.action.type.REGION_SELECT, onRegionSelect);

Can we have a string, which represents user region in REGION_SELECT payload?

src/firefly/js/visualize/DrawLayerCntlr.js
cwang2016: - The document is missing parameter type for selectedRegion. selectedRegion could be a string, array of strings, or a region object. I will fix it. 
- when clicking on a region, the action payload will contain the region object. In order to know which region is selected, there is a function reporting the status of the selected region, it is 'getSelectedRegion' from RegionPlot.js which returns the region string, that is the string parsed to build the region object. The string returned from 'getSelectedRegion'  is that either formed (normalized) from a region file on the server  or  from array of strings in the front side. Each returned string has  the coordinate info. (image, physical, j2000, ...) attached  in the front.

cwang2016:
'selectedRegion' in dispatchSelectRegion does take either string, array of strings (for programmatic purpose) or region object (for the usage of UI clicking) . 
Please see how dispatchSelectRegion takes strings in 'ffapi-highlevel-test.html'.

tgoldina:
Yes, dispatchSelectRegion does work as expected - any of the regions added can be  programmatically selected from an external script.

I am talking about different thing: an external script should also be able to react to region selection made by mouse click in the UI. 

An external script can trigger UI changes by calling action dispatchers. But it can also react to any UI action by adding a listener to this action. A listener is a user-defined function, a callback, which is called whenever the action goes through the flux. 

For example, some external script is using Firefly API to plot amplifiers, represented on the image by box regions. Each amplifier is described by unique name and region string, and this 1 to 1 map between name and region string is maintained by the script. When region is clicked, the script displays name.

To do it, the script defines a callback, which takes REGION_SELECT action object and uses the action payload to find which region is selected. The only representation of region the script is aware of is region string. But the payload of REGION_SELECT action (triggered by mouse click) has only selectedRegion object, which is Firefly-internal representation of the region.

cwang2016:
as we discussed in previous meeting, a utility function 'getSelectedRegion' is provided to report the status of the selected region, i.e. the region string. 

Moreover, region string is actually stored in region object. If needed, one more utility besides 'getSelectedRegon' could be implemented to return the region string based on the region object. 

Would that be ok for the callback provided by the external script to use either utility to get  the region string?   dispatchSelectRegion mainly uses region object to form the highlight result. The string(s) passed to the dispatchSelectRegion is used for discovering the region object. Adding both region object and region string to the action payload seems redundant for dispatchSelectRegion itself.

tgoldina:
Sorry, I completely forgot about getSelectedRegion. Yes, the callback can
use it. We need to export it from  ApiUtilImage.jsx though. Thank you for
the explanation.

2016-09-13 23:59 GMT-07:00 Cindy Wang <notifications@github.com>:

tgoldina:
I have tested getSelectedRegion, it works great. All that needs to be done
is to export getSelectedRegion from ApiUtilImage and add it to JSDoc by
adding @public annotation and including RegionPlot.js into the list of
sources in jsdoc_config.json We might also want to mention it in
dispatchRegionSelect JSDoc.

src/firefly/js/drawingLayers/RegionPlot.js
```diff
@@ -319,4 +392,45 @@ function removeRegionsFromData(drawLayer, removedRegions) {
...
+            return prev;
+        }, []);
+    }
+    return selDrawObj.slice(0, stopIndex);
+}
+
+/**
+ * @summary get the region description of the selected region from the specified region layer
+ * @param {string} drawLayerId id of the drawing layer
+ * @return {string}
```
tgoldina: I noticed this method returns null, when no region in drawLayerId is selected.

cwang2016: if the click occurs nowhere (no region under), then the previous selected region will becomes deselected. Passing a 'null' is the way to deselect the region.

tgoldina: I understand, we just need to document it. Looking at the method code, I
though I'd get empty string.

cwang2016: I will document it, but this is not public function. Is that needed in public?

cwang2016: If you pass a 'null' or empty array into 'dispatchSelectRegion', it will work as de-select. I will be documented on the dispatch function. Thanks.

tgoldina:
getSelectedRegion needs to be public. Every method exposed via API should
have public annotation.

2016-09-14 11:20 GMT-07:00 Cindy Wang <notifications@github.com>:

cwang2016:
some addition to the implementation: 
- change the signature of dispatchCreateRegionLayer and dispatchAddRegionEntry - add selectMode to replace the original relevant parameter, the default is empty object if no value is passed. 
- export getSelectedRegion from 'ApiUtilImage'
- update jsdoc for region related dispatch functions.

PR #179, opened 2016-09-14
DM-7547: Remove the extra column on far right-hand side of a table

loitly:
https://jira.lsstcorp.org/browse/DM-7547
Also:
- re-sizes the column selection panel when description is not available.
- fix usage of TablePanel so that only 1 table ui state is created for each table model.


ejoliet:
Looks better. Review done.Thanks!

PR #180, opened 2016-09-14
DM-7594: Display phase column with 8 decimal places precision

loitly:
https://jira.lsstcorp.org/browse/DM-7594
Also, add description meta to that column.


ejoliet:
the displayed phase is now 8 decimal precision and the description appears too.
 
As the column is the last one in the table on the right-hand, the vertical scroll bar hide part of it. Is it possible to make the column wider so it shows the value (8 decimals) fully?
  
Minor (was in the ticket [DM-7595]: I've downloaded the table but i couldn't find any mention on the column 'phase'. Is it possible to add a comment with the column name 'phase' and the description as other columns?

ejoliet:
About hiding columns, yes, probably we don't need all the columns, we can keep the fluxes, and time columns as first approach but i wouldn't change that until we get more details on those aspects.
For now is fine. Thank you!
Review done.

PR #183, opened 2016-09-16
DM-6980:  Fix PNG creation for marker

cwang2016:
Fixing method: 
Marker size is represented by a small pure number (magnitude of thousandth) in term of world coordinate (J2000). World coordinate is a hint to set the size value with 'degree' unit, otherwise the size value was treated with 'pixel' unit, too small to be shown in the PNG.

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-test.html
+++ b/src/firefly/html/demo/ffapi-highlevel-test.html
@@ -317,8 +317,15 @@
                 'j2000;box(47.247072i, 180.445347i, 50p, 20p, 0) # color=blue width=6 text="box 12-3"',
                 'physical;-box point 12 12 # text="pt box 13"',
                 'j2000;polygon(202.564796, 47.281999,202.553758, 47.247072, 202.445347, 47.244296, 202.479687, 47.264027, 202.492153, 47.290841) # color=blue text="polygon 14" offsety=10 width=10 font="helvetica 16 bold"',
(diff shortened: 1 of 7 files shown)


cwang2016:
test:
select any image, click 'marker/footprint' button, select 'Add Marker', click 'save fits, png...' button, and choose 'png' for download, check if the marker is shown in the download PNG.

xiuqin:
1. The color of the marker does not show up in PNG.
2. The Marker label does not  show in PNG 
3. The catalog overlaid on image was square on screen, but cross (X) on PNG.

cwang2016:
More updates are added to fix the issues regarding PNG download from both image search and catalog search, including 
- showing the marker added from marker/footprint dropdown list.
- showing the right color as color changed from drawing layer popup. 
- showing the marker with the right color and right point shape from catalog search.
- showing the marker/footprint label.

ejoliet:
I have tested this fix in http://localhost:8080/firefly/demo/ffapi-highlevel-test.html but i don't see the fix. Am i doing something wrong for testing?

See attached image on left the PNG downloaded and right the browser viewer.
screen shot 2016-09-19 at 3 18 41 pm

ejoliet:
Update: the fix works in firefly viewer. Maybe it doesn't in the API mode?

ejoliet:
Works also great in IRSAVIewer, i've checked.

Really minor: compass rose is included in PNG but not the labels 'N', 'E'.

cwang2016:
more update: 
change the color  of compass labels, from 'black' to 'red'.

PR #184, opened 2016-09-16
DM-7487: add polygon values from active image in catalog search

ejoliet:
Pre-calculated polygon area when this spatial search method is selected in the catalog search.
It will take part of the image and calculate the x,y pairs from active plot.
Please review it.

### DIFF
--- a/src/firefly/js/ui/CatalogSearchMethodType.jsx
+++ b/src/firefly/js/ui/CatalogSearchMethodType.jsx
@@ -16,7 +16,10 @@ import {SizeInputFields} from './SizeInputField.jsx';
 import {InputAreaFieldConnected} from './InputAreaField.jsx';
 import {FileUpload} from '../ui/FileUpload.jsx';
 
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/CatalogSearchMethodType.jsx
```diff
@@ -252,4 +269,78 @@ var initRadiusArcSec = (max) => {
...
         return parseFloat(1 / 3600).toString();
     }
-};
\ No newline at end of file
+};
+
+
+/**
+ * Helper to get polygon from image
+ */
```
loitly: I don't think the helper functions defined below are used.  Also, they contain unused imports as well as undefine function reference.  If you want to keep these, it should be cleaned up and moved to a util type of file.

ejoliet: Removed unused function. Cleanup done.

PR #185, opened 2016-09-16
DM-7630: Display null_string when value is null

loitly:
- delegate IpacTableReader to use DataGroupReader instead.
- fix selectInfo not updating totalRows

Make sure that table with Null Values defined are printed and shown as defined.  ie.  'null' -> 'null'.
If one is not given, empty string is used by default.


ejoliet:
Looks great to me. I'm not sure about 'null' values should be plotted as '0' though. But that is different discussion maybe @tgoldina and @xiuqin can comment on that. For now i would merge it and discuss later.Thanks! 
Reviewed.

PR #186, opened 2016-09-19
DM-7159: add phase folded code, ibe single exp fetch and periodogram …

ejoliet:
Added the usage of the newest LC processor to get the phase folded and periodogram.
Please have a look and let me know.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
+++ b/src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
@@ -19,12 +19,14 @@ import {dispatchMultiValueChange, dispatchRestoreDefaults} from '../../fieldGrou
 import DialogRootContainer from '../../ui/DialogRootContainer.jsx';
 import {PopupPanel} from '../../ui/PopupPanel.jsx';
 import FieldGroupUtils, {revalidateFields} from '../../fieldGroup/FieldGroupUtils';
+import {makeTblRequest,getTblById} from '../../tables/TableUtil.js';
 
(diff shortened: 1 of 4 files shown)


src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
```diff
@@ -361,18 +362,29 @@ function resultsSuccess(request) {
...
+function doPhaseFolding(fields) {
+    let tbl = getTblById(RAW_TABLE);
+
+    console.log(fields);
+
+    var tReq = makeTblRequest('PhaseFoldedProcessor', PHASE_FOLDED, {
+        'period_days': fields.period.value,
+        'table_name': 'folded_table',
+        'time_col_name':fields.timeCol.value,
+        'original_table': tbl.tableMeta.tblFilePath
```
wmiipac: I see a TypeError here
LcPhaseFoldingPanel.jsx:386 Uncaught TypeError: Cannot read property 'tableMeta' of undefined

ejoliet: This is because of the workflow we decided to implement somehow. When the LC viewer is displayed, first thing to do is to bring a 'RWA_TABLE' by uploading one. Then it shouldn't complain.

ejoliet: Maybe @loitly has a better way to try/catch this so the saga doesn't bail out?

wmiipac:
cad do try/catch (!ifUndefined(tableMeta)?

wmiipac:
Yes, I will take over make the ibe part working after E merges.

PR #187, opened 2016-09-19
DM-7116: change the title of drawing layer dialog based on the title of the active image

cwang2016:
Make the tile on the drawing layer dialog to be the same as the image title from either image or catalog search.

### DIFF
--- a/src/firefly/js/visualize/ui/DrawLayerPanel.jsx
+++ b/src/firefly/js/visualize/ui/DrawLayerPanel.jsx
@@ -21,16 +21,12 @@ export const DRAW_LAYER_POPUP= 'DrawLayerPopup';
 
 function getDialogBuilder() {
     var popup= null;
+    const defaultTitle = 'Drawing Layers';
     return () => {
         if (!popup) {
-            var title = 'Drawing Layers';
-            //try {
-            //    if(visRoot()){
-            //        title += ': '+getActivePlotView(visRoot()).plots[0].title;
-            //    }
-            //}catch (E){
-            //    //
-            //}
+            var title = get( getActivePlotView(visRoot()), ['plots', '0', 'title']);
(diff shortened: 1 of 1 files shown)


ejoliet:
Sorry but i don't see the title changing.

cwang2016:
Yes, you have to do 'gradle :firefly:deploy :firefly:war".

cwang2016:
if you switch from DM-6980.

ejoliet:
I did a purge, then a fresh build and deploy on tomcat where i have clean the webapps,etc.
Still don't see the title changing. Is it only me? Can @tgoldina or @robyww confirm?

ejoliet:
Oh, i know. Yes, i can defintively see the title update. What i was confused is that i was looking at the title while i change the image active, and the title doesn't change unless i close and reopen the layer window.
I guess this is a different issue that can be brought later.

For now, it works as described in the ticket.
Review done.

src/firefly/js/visualize/ui/DrawLayerPanel.jsx
```diff
@@ -21,16 +21,12 @@ export const DRAW_LAYER_POPUP= 'DrawLayerPopup';
...
         if (!popup) {
-            var title = 'Drawing Layers';
-            //try {
-            //    if(visRoot()){
-            //        title += ': '+getActivePlotView(visRoot()).plots[0].title;
-            //    }
-            //}catch (E){
-            //    //
-            //}
+            var title = get( getActivePlotView(visRoot()), ['plots', '0', 'title']);
```
robyww: This is not the correct way to get the primary plot.  The plots array should never be access directly.  Beyond that 0 is not always the primary plot.
the `primePlot` function should always be called here.  

The code should be
```
const plot= primePlot(visRoot());
title= plot ? plot.title : null;
```

PR #190, opened 2016-09-22
DM-7336:fixed the algorithm bug in Geom.java

lznakano:
Emmanuel,

Can you please take a look and let me know if you still see problems?

Thanks!


lznakano:
Emmanuel:

Please test it and let me know if it still has problems.  THANKS!

ejoliet:
Looks good to me. I've checked randomly on other image operations and it looks ok.

PR #191, opened 2016-09-23
DM-7665: fixed the region parsing issue regarding region used separators contained in some property values

cwang2016:
Fixed the parsing errors which happen when 
- semicolon, space, or '#' is contained in the values of text property.
- the text (or font) property is not properly quoted with " or ' or {}. 

Test sample, 
upload firefly_test_data/region-test-files/regionText.reg
or 
visit localhost:8080/demo/ffapi-highlevel-test.html by clicking 'test-open' and 'create_region_layer'

### DIFF
--- a/src/firefly/js/visualize/region/RegionFactory.js
+++ b/src/firefly/js/visualize/region/RegionFactory.js
@@ -54,8 +54,11 @@ export class RegionFactory {
      * @returns {null}
      */
     static parseRegionJson(regionData) {
+
+        var globalOptions = Object.assign({}, makeRegionOptions({[regionPropsList.COORD]: defaultCoord}));
+
         return regionData? regionData.reduce ( (prev, region, index) => {
-            const rg = RegionFactory.parsePart(region, index);
+            const rg = RegionFactory.parsePart(region, index, globalOptions);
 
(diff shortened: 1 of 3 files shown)


src/firefly/js/visualize/region/RegionFactory.js
```diff
@@ -74,37 +77,80 @@ export class RegionFactory {
...
     static parseRegionDS9(regionData, bAllowHeader = true, stopAt) {
-        var regionLines = regionData.reduce ((prev, oneLine) => {
-                var resRegions = oneLine.split(';');
+        const sep = ';';
+        const dLeft = ['{', '"', '\''];
+        const dRight = ['}', '"', '\''];
+
+        // collect region lines and each line may contain coordinate, region description or both
+        var regionLines = regionData.reduce( (rLines, oneLine) => {
+
```
robyww: This is a lot of code for a reduce function.  I think you should break it up so that it is a little easier to follow. I see that the old approach was a reduce as well but it was not as big.

cwang2016: Thanks for the comment.

cwang2016:
Breaking up the code as the comment.

PR #193, opened 2016-09-27
DM-7700: Atlas table filtering doesn't work properly

loitly:
https://jira.lsstcorp.org/browse/DM-7700

This is only a problem in Atlas which makes it a little harder to test.
Build locally then deploy it on irsawebdev1 or proxy irsawebdev1 to your localhost if you want to test it.

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -206,19 +206,19 @@
 }
 /* react-split-pane style <END> */
 
-[role="toolbar"] {
+.PanelToolbar {
     display: flex;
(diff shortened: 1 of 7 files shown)


ejoliet:
Checked in atlas that the table is filtered correctly. 
Review done.

loitly:
The help anchor was an introduced side effect.  The xy-plot highlighting icon has always been that way.  But, I'll fix them both.

loitly:
Changes has been committed.  Please verify and approve.

PR #195, opened 2016-09-28
DM-7765 Render the text defined as the 'text' property value of a region on PNG

cwang2016:
The rendering of the region associated text was missing in the PNG. 
This issue is fixed by creating a separate text region  with the location calculated at JS parts, the text value and font defined in the original region, and sending the created text regions as well as the original regions to the server for producing the PNG.

### DIFF
--- a/src/firefly/js/visualize/region/Region.js
+++ b/src/firefly/js/visualize/region/Region.js
@@ -36,17 +36,17 @@ var cloneArg = (arg) => Object.keys(arg).reduce((prev, key) =>
                         }, {});
 
 /**
- *
- * @param wpAry WorldPt array, items contained: line, annulus, ellipse, box: 1, line: 2, polygon: >= 3
- * @param radiusAry RegionValue array, items contained: annulus: at least 2 radii.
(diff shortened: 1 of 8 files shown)


cwang2016:
Test sample: 
firefly_test_data/region-test-files/textText_DM-7765.reg

src/firefly/js/visualize/region/Region.js
```diff
@@ -248,10 +248,10 @@ export var getRegionCoordSys = (coordStr) => (RegionCsys.get(coordStr) || Region
...
 /**
  *
- * @param name string
- * @param point string (font size)
- * @param weight string
- * @param slant string
+ * @param name string - ex: Arieal, Courier, Ties, Helvetica, sans-serif, BRAGGADOCIO
+ * @param point string (font size) ex: 10, 12, 14, 16
+ * @param weight string ex: normal, bold
+ * @param slant string ex: normal, italic
```
robyww: I know the jsdocs were wrong before bug it you add to them you should go ahead and fix them.
Such as:
`@param {string} slant ex: normal, italic`

PR #196, opened 2016-09-28
DM-7488: Text View does not show column's label, only name.

loitly:
- show column's label in 'Text View' if it exists.
 - disable filter when content of the column contains html


ejoliet:
is fixed. reviewed.

PR #198, opened 2016-09-30
DM-7805 reset column options on column change

tgoldina:
This is the fix fo DM-7805:
Changing column to plot doesn't reset the 'flip' option to unchecked state

In this implementation, X or Y column change would result in the reset of the corresponding options (excluding 'grid').


ejoliet:
Works like a charm. Tested also that the reset works ok.
Thanks!
Review done.

PR #199, opened 2016-10-01
DM-7825: Use custom http headers to pass ws connection info

loitly:
https://jira.lsstcorp.org/browse/DM-7825

We were using cookies to pass websocket connection info to the server.  This approach does not work in and embedded mode, ie. Gator.
I've converted to use custom http headers instead.  This also affects python api.  I've modified the code to reflect the changes.

The above url describe how to reproduce the problem.  This change fixed the above problem and should not have any regression issue.  Please confirm.


ejoliet:
I've tested the PR in here:

by loading a large catalog. Is that enough or that won't test the WS fix?

loitly:
This test is not equivalent to the reported bug.  You need to find a case where the html page has a different domain or path then the firefly's script, like /gator and /irsaviewer.

xiuqin:
@ejoliet Shall we add a unit test for this?

ejoliet:
Tested. As i said, it works fine. And can reproduce it in dev but not in this branch.
Review done.

cwang2016:
Review completed. Python API is tested and works fine.

PR #200, opened 2016-10-03
DM-7767 Make Multi image viewer support light curve type layout

robyww:
- mutliple plots coming in with table selection
 - plots comming in as wcs matched
 - wcs functionallity expanded to support new plots, matching zoom and rotation
 - scroll together
 - preference for zoom and stretch
 - rotation will stay wcs matched
 - set number of images (1,3,5,7)


loitly:
- I noticed 'Raw Image' shows up as a layer.  I don't think it should.
- Expanded mode does not work in Safari.  Not sure when it was introduced.
- When image viewer mounted, I get this error in console.
 ` Invalid prop `value` of type `number` supplied to `RadioGroupInputFieldView`, expected `string`.`

Nothing I see here would stop you from merging.  It's a prototype, and there will be issues.
Review is done.  You decide how much of it you want to address.

PR #204, opened 2016-10-07
DM-7783: add HistogramTest class, add the unit tests and end to end test

lznakano:
I added HistogramTest class.  It does the unit test and also end to end test.  The unit test is done through an make-up data.  The end to end test is done using a real image, f3.fits.  First, a main program created the Histogram results using f3.fits and then stored the results in corresponding json file.  When the unit test is run, it reads in the json file and then compare the newly calculated data with the results saved in the json file.  The json file should not be changed and should not be recreated. In the future, when Histogram is changed, if the test fails, that means the new changes introduce bugs.  The coverage of this unit test is 81% which can be seen in IntelliJ.    To run the test:
1. check out DM-7783-HistogramUnitTest branch in  firefly_test_data
2. cd /hydra/cm/firefly
3.  type command:
   gradle :firefly:test


src/firefly/test/edu/caltech/ipac/visualize/plot/HistogramTest.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+    @After
+    /**
+     * Release the memories
+     */
+    public void tearDown() {
+        inData=null;
+        histArray=null;
+        data=null;
+    }
```
loitly: is this necessary?

src/firefly/test/edu/caltech/ipac/visualize/plot/HistogramTest.java
```diff
@@ -0,0 +1,362 @@
...
+
+
+
+    @Test
+    /**
+     * A method was added to calculate the expected value differently here and then compare with the value in Histogram.
```
loitly: Do we have to calculate the expected value 'differently' here?  Shouldn't the expected value be fixed/static?
When the algorithm changes, it has to reflect here as well?

src/firefly/test/edu/caltech/ipac/visualize/plot/HistogramTest.java
```diff
@@ -0,0 +1,362 @@
...
+     * This method is an end to end test.  It takes a FITS file, f3.fits and creates an instance of Histogram.
+     * Compare its results with the ones stored in the f3HistogramResult.json file.  The results stored in
+     * f3HistogramResult.json was generated by the same procedures at the time when HistogramTest class is added.
+     * It ensures that any future changes will not introduce problems.
+     */
+    public void testEndToEnd() throws FitsException, IOException, ParseException {
+        String testTreePath = HistogramTest.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        String  rootPath = testTreePath.split("firefly")[0];
+        String dataPath = rootPath+TEST_DATA_PATH;
+        String inFitsName =dataPath+fileName;
```
loitly: If there is a common method of resolving these paths, will you create a TestUtil class to collect them?  I see them repeated over and over.

loitly:
git pull both firefly and firefly_test_data.
ran gradle test:  gradle clean :firefly:test
got this error:
edu.caltech.ipac.visualize.plot.HistogramTest > testEndToEnd FAILED
    nom.tam.fits.FitsException at HistogramTest.java:265

src/firefly/test/edu/caltech/ipac/visualize/plot/HistogramTest.java
lznakano: I can not think a common method to use.  But it is a good idea to create a utility class somewhere.   I have the similar issues in other classes too.  Some methods need to be used again and again.

src/firefly/test/edu/caltech/ipac/visualize/plot/HistogramTest.java
ejoliet: TearDown is good specially if you have many test after this one (or before, junit test runner doesn't run in order...)

PR #205, opened 2016-10-07
DM-7918 Constraint column render error and the state update error when column constraint changes 

cwang2016:
Fixed the bugs for catalog search constraints, the bugs are: 
- the constraint rendering became empty when 'enter' key is pressed or the focus on the cell is moved away. 
- the state regarding the search constraints is not updated when the constraint changes. 

Note:
If any entered column constraint is checked to be invalid, the error message will be shown in a popup box when the 'search' button is clicked and the search will stop.

### DIFF
--- a/src/firefly/js/tables/FilterInfo.js
+++ b/src/firefly/js/tables/FilterInfo.js
@@ -113,6 +113,16 @@ export class FilterInfo {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


cwang2016:
A feature fixing: 
The column constraints is highlighted in case some invalid constraint string is entered.

ejoliet:
I've tested the fix, seems ok. Thanks!

PR #206, opened 2016-10-12
DM-7920 Refactor chart data and generalize actions

tgoldina:
Refactored chart actions and data to be able to support multiple chart elements connected to different data.

charts.data - is the store object with chart ids as the keys, the values are chart data object, which has chartDataElements field.

chartDataElements - object with chartDataElementId keys
  and chartDataElement object as value

chartDataElement shape:
{id, options, defaultOptions, data, meta, tblId,
  type - to get helper functions like
chartDataElements - object with chartDataElementId keys
  and chartDataElement object as value

chartDataElement shape:
{id, options, defaultOptions, data, meta, tblId, type}
The 'type' is an id of a ChartDataType, two of them are predefined: 'xycols' and 'histogram'.
Each chart data type has fetchData function and fetchParamsChanged functions, and possibly other helper functions.

### DIFF
--- a/src/firefly/js/charts/ChartDataType.js
+++ b/src/firefly/js/charts/ChartDataType.js
@@ -2,8 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {DATATYPE_XYCOLS} from './ChartDataTypeXYCols.js';
-import {DATATYPE_HISTOGRAM} from './ChartDataTypeHistogram.js';
+
 import {logError} from '../util/WebUtil.js';
 
 /**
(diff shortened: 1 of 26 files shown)


src/firefly/js/charts/ChartDataTypeXYCols.js
```diff
@@ -80,119 +89,112 @@ const RESET_ZOOM = `${XYPLOT_DATA_KEY}/RESET_ZOOM`;
...
+export function loadXYPlot({chartId, xyPlotParams, tblId, dispatcher=flux.process}) {
+    //SCATTER
+    dispatchChartAdd({chartId, chartType: 'scatter', groupId: tblId,
+        chartDataElements: [
+            {
+                type: 'xycols', //DATATYPE_XYCOLS.id,
+                options: xyPlotParams,
+                tblId
+            }
+        ], dispatcher});
```
robyww: I am concerned about this as a utility routine. I am not sure we want to make it an example.  It is easier to understand the code if dispatches are being used directly. `loadXYPlot` is effectively a dispatch but you can't tell by looking at the code.  
a couple of other things:
 - `'scatter'` should type a a constant.
 - why not just say `type: DATATYPE_XYCOLS.id`?


Being picky because all this code will be examples for all the other charts.

src/firefly/js/charts/ChartDataTypeHistogram.js
```diff
@@ -0,0 +1,164 @@
...
+    if (!histogramParams) { return []; }
+
+    const serverParams = [];
+    serverParams.push(histogramParams.columnOrExpr);
+    serverParams.push(histogramParams.x && histogramParams.x.includes('log'));
+    serverParams.push(histogramParams.numBins);
+    serverParams.push(histogramParams.falsePositiveRate);
+    //serverParams.push(histogramParams.minCutoff);
+    //serverParams.push(histogramParams.maxCutoff);
+    return serverParams;
```
robyww: why not?
```
if (!histogramParams) { return []; }
return [
     histogramParams.columnOrExpr,
     histogramParams.x && histogramParams.x.includes('log'),
     histogramParams.numBins,
     histogramParams.falsePositiveRate,
];
```

src/firefly/js/charts/ChartDataType.js
```diff
@@ -0,0 +1,50 @@
...
+ * @public
+ * @typedef {Object} ChartDataType - an object which specifies how to get chart data
+ * @prop {string} id - unique chart data type id
+ * @prop {Function} fetchData - function to load chart data element data: fetchData(dispatch, chartId, chartDataElementId)
+ * @prop {Function} fetchParamsChanged - function to determine if fetch is necessary: fetchParamsChanged(oldOptions, newOptions)
+ * @prop {Function} getUpdatedOptions - function to resolve the options, which depend on table or chart data getUpdatedOptions(xyPlotParams, tblId, data)
+ */
+
+const chartDataTypes = [];
+
```
robyww: I am struggle with this code. I am really torn.  It is effectively keeping application state outside of the store. I solved this problem with drawing layer by defining them in ReduxFlux.js (line 110) and them creating the reducing function with the factory (line 129).  However, it does make things more complex.  Might be worth Tatiana, Loi, and Trey talking about tomorrow.

PR #208, opened 2016-10-14
DM-7986: fix circles connecting & add way to reorder image tabs

robyww:
fixing the circles was trivial and did not really need a ticket.  Most of this ticket has the code to reorder the images tabs in the image select panel


cwang2016:
Review completes. Looks very good.

PR #210, opened 2016-10-17
Dm 7800 search processors

lznakano:
Hi Trey,

I was trying to generating some text files to report the issues found in the json data( I don't know it was introduced by converting since Yi was out on Friday).  So I have lot of codes for testing.  I removed half of them and still leave some in case I need them later.  Please ignore the commented codes.   To test it,
  1. pull the DM-7800-searchProcessors in firefly_test_data, it has all the testing data files.
  2. In TestQueriesPanel.jsx,I had hard coded the table names and path.  
  3. http://localhost:8080/firefly/ and then select "Demo Search" and then select "DAX Dummy" tab, 
     there are two buttons in that tab, the first button shows the meta table which does not have  
    data, the second button show the table data and plot. 

Thanks!

Lijun


src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCataLogSearch.java
```diff
@@ -181,18 +190,88 @@ private DataGroup getTableDataFromJson(File jsonFile, TableDef tableDef) throws
...
+            if (fi == null) {
+                throw new DataAccessException("Unable to get file location info");
+            }
+            if (fi.getInternalFilename() == null) {
+                throw new DataAccessException("File not available");
+            }
+            if (!fi.hasAccess()) {
+                throw new SecurityException("Access is not permitted.");
+            }
+            DataGroup sourceDataGroup = DataGroupReader.readAnyFormat(new File(fi.getInternalFilename()));
```
loitly: Instead of sm.getFileInfo, you should have called sm.getDataGroup.  This would have returned a DataGroup instead of reading the file.

src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCataLogSearch.java
```diff
@@ -181,18 +190,88 @@ private DataGroup getTableDataFromJson(File jsonFile, TableDef tableDef) throws
...
+    private  DataType[] getMetaFromMetatable(DataGroup dataGroup){
+
+        DataType[] dataTypes = new DataType[dataGroup.size()];
+        DataObject[] dataObjects=dataGroup.values().toArray(new DataObject[0]);
+        for (int i=0; i<dataObjects.length; i++){
+            dataTypes[i] = new DataType((String) dataObjects[i].getDataElement("Field"),
+                    getDataClass( (String) dataObjects[i].getDataElement("Type") ) );
+            boolean maybeNull = dataObjects[i].getDataElement("Null").toString().equalsIgnoreCase("yes")?true:false;
+            dataTypes[i] = new DataType( (String) dataObjects[i].getDataElement("Field"),
+                    (String) dataObjects[i].getDataElement("description"),
```
loitly: You are using "description" as the column's label.  DataType also has a shortDescription(one sentence) field used to describe the column.  Verify that "description" here is the label and not the description of the column.

src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCataLogSearch.java
lznakano: No, it does not work.  This was what I did yesterday around 6PM:

    private TableDef getMeta(TableServerRequest request){
        SearchManager sm = new SearchManager();

        try {
            DataGroupPart dgp = sm.getDataGroup(request);
            DataGroup metaData = dgp.getData();
            return dgp.getTableDef();
        } catch (Exception e) {
            e.getStackTrace();
        }
        return null;
    }

The dataGroup returned contains empty data.  It tells the number of rows and the TableDef but data.  Actually, I don't understand the reason.  I want to ask you about it later.

loitly:
You need to set TableServerRequest#PAGE_SIZE.  It’s default to zero.

PR #211, opened 2016-10-19
DM-7995: options for lsst

robyww:
- filtering the IRSA master table
  - Zoom to selection
  - then catalog panel do selection or viewport corners
  - Options to allow catalog search to come up with polygon when image
      is selected
  - option to recenter to selected area
  - fixed bug: image is not selected when plotted.
  - Add two new icons


src/firefly/js/ui/CatalogSearchMethodType.jsx
```diff
@@ -272,4 +360,56 @@ var initRadiusArcSec = (max) => {
...
+        return fieldInit();
+    }
+    else {
+        const {fieldKey}= action.payload;
+        const rFields= clone(inFields);
+        if (action.type===VALUE_CHANGE && fieldKey==='polygoncoords') {
+            rFields.imageCornerCalc= clone(inFields.imageCornerCalc, {value:'user'});
+        }
+        else {
+            const {value:cornerCalcV}= inFields.imageCornerCalc||'user';
```
cwang2016: .... || 'user' or ... || {value: 'user'} ?

robyww: yea, that's wrong

src/firefly/js/ui/CatalogSearchMethodType.jsx
```diff
@@ -47,16 +52,33 @@ export class CatalogSearchMethodType extends Component {
         });
     }
 
+    componentWillReceiveProps(nextProps) {
+        // const fields= FieldGroupUtils.getGroupFields(this.nextProps.groupKey);
+        const {coneMax, boxMax, groupKey}= nextProps;
+        if (coneMax && groupKey) {
```
cwang2016: I am not sure if I understand the logic. should the case 'boxMax&&groupKey' be considered?

robyww: should be `coneMax && boxMax && groupKey` - I will fix it.

robyww: found another problem to, I accidentally took out the passing the properties when I was cleaning up.

cwang2016:
Review completed. 
One question regarding the coordinates for the polygon. In 'catalogs' panel, I copy & paste the coordinates from example. and enter 'search', then come back to the 'catalogs' panel, and switch  the radio button for 'Corners' from 'User' to 'ViewPort' and then back to 'User',  the coordinate string is not shown as the original string for 'User' but as the string for 'ViewPort'. Is that as expected?

robyww:
That is the behavior.  We can always modify it later.  Right now if you switch to user, it just means that it will not change it.  If you start editing it will switch to user for you. I am not trying to remember the laster user entry.

PR #212, opened 2016-10-19
Dm 8028 unit test utility

lznakano:
A utility class, FileLoader, is added.  It is in the packageedu.caltech.ipac.firefly.util under the test tree.  Following 8 files are modified accordingly:

HistogramTest.java
ProjectionTest.java
HistogramProcessorTest
StatisticProcessorTest
PhaseFoldedLightCurveTest
IpactableTest
ZscaleTest
GridTest

To test it, please check out of this branch and also the data branch in firefly_test_data. The data branch has exactly the same name.

DM-7783 is depending on this branch.  Once it is merged, the DM-7783 can be finalized.


src/firefly/test/edu/caltech/ipac/firefly/util/FileLoader.java
```diff
@@ -1,40 +1,109 @@
...
+
+            return IpacTableReader.readIpacTable(inFile, null, false, "inputTable");
+        }
+        catch (Exception e){
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+     public static File loadFile(Class cls, String tblFile)  {
```
loitly: This function is poorly named.  Should be resolveFile(Class cls, String fname) or something like it.
You are not loading anything and it has nothing to do with table.

lznakano: It reads a table fie in the test tree  and returns a DataGroup object.   I have named:
  loadFIts: return a FITS
  loadFitsRead: return a FitsRead
  loadFitsReadArray: return a FitsRead array
 loadFile : return a File 
 
Does it make more sense if it is named as loadDataGroup?

PR #214, opened 2016-10-21
DM-4593: background monitor for Firefly

loitly:
https://jira.lsstcorp.org/browse/DM-4593

TO TEST:
http://localhost:8080/firefly/lc.html
Noticed 'Background Monitor' button add to top-right menu bar.
- upload any Phase Folded table, click Search
- upper right corner, click on 'Download' button
- enter the number of images and description then click search
- btw.. this dialog is only for testing purposes.  No need to review.

Background Monitor:
 - support existing features except for listed TODOs.
 - send BackgroundStatus as json action.
 - convert DownloadRequest to json
 - added BackgroundCntlr and related pieces
 - resizable layout.. 1-3 columns

Also:
 - add standardized interfaces for our Flux controllers (see BackgroundCntlr.js)
 - expand CommandServices to allow for HttpCommand
 - separate VisHeader into VisHeader and VisPreview
 - restyle banner.jsx
 - convert code to use SearchServices
 - add LightCurveFileGroupsProcessor stub
 - add VO_TABLE support to DataGroup.readAnyFormat (based on file extension only, should peek into data to detect)

TODOs:
 - downloadScript dialog
 - resend email
 - retain background jobs and statuses after reload
 - email notification is not working for local env


src/firefly/java/edu/caltech/ipac/firefly/server/ServerCommandAccess.java
```diff
@@ -141,13 +117,24 @@ private static void initCommand() {
...
+        return _cmdMap.get(cmd);
+    }
 
     public static abstract class ServCommand {
         public abstract String doCommand(Map<String, String[]> paramMap) throws Exception;
 
         public boolean getCanCreateJson() { return true; }
     }
 
+    public static abstract class HttpCommand extends ServCommand {
```
robyww: I don't understand why we would do another interface that subclasses the original and then disables the main method.  It is confusing.

robyww: I appears that ServCommand would be a simpler version HttpCommand.  Maybe it should be an abstract implementation of HttpCommand. The each user of ServCommand could inherit instead of implement.

robyww:
UI: is the count on the button suppose to be number completed and not yet downloaded or the total number?

src/firefly/js/ui/ProgressBar.jsx
```diff
@@ -0,0 +1,34 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {Component,PropTypes} from 'react';
+
+import sCompare from 'react-addons-shallow-compare';
+import './ProgressBar.css';
+
+export class ProgressBar extends Component {
```
robyww: We might want to let everyone know there is a progress component.

src/firefly/js/ui/InputFieldView.jsx
```diff
@@ -133,7 +134,8 @@ InputFieldView.propTypes= {
     onBlur : PropTypes.func,
     onKeyPress : PropTypes.func,
     showWarning : PropTypes.bool,
-    type: PropTypes.string
+    type: PropTypes.string,
+    placeholder: PropTypes.string
```
robyww: This placeholder is nice, we need to start using it all over.

src/firefly/js/data/ServerParams.js
```diff
@@ -129,6 +131,11 @@ export const ServerParams = {
         TITLE: 'Title',
         JSON_DEEP: 'jsonDeep',
         ACTION: 'action',
+        PACKAGE_REQUEST: 'packageRequest',
+        TABLE_SEARCH: 'tableSearch',
+        SELECTED_VALUES: 'selectedValues',
+        TABLE_SAVE: 'tableSave',
+        JSON_SEARCH: 'jsonSearch',
```
loitly: TABLE_SEARCH is a search that returns a TableModel.
JSON_SEARCH returns a JSON string which can be anything.
It is similar to JSON_DATA but uses the new json-ized TableRequest.
It's not used much, and I do plan to consolidate the two.  Backward compatibility was a concern before.  Not so anymore.

loitly:
Okay, I've pushed 3 more commits to address these issues.  Please check again.
 - add missing stop.gif
 - fix bad temp file location.
 - Correctly make ServCommand extends from HttpCommand
 - Fix cancel button not working consistently.
 - Fix ReduxFlux having to use import * on BackgroundCntrl.

tgoldina:
There is one thing we might want to change: change the status just before we start staging the first file. Currently, the status 'Computing the number of packages'  keeps showing until the second file starts being staged.

PR #215, opened 2016-10-21
DM-7782: add unit test for Crop class

lznakano:
This is the unit test class for Crop.java.  The testing data is located in the firefly_test_data/package.
The end to end test is added.  The header and data are examined.  To test it, please check this branch out and also go to https://github.com/lsst/firefly_test_data to checkout the data branch 
"DM-7782-UnitTestForCrop".

The coverage is about 60%.  The low coverage is due to the fact that Crop.java is dealing with different bitpix. The current test only has bitpix=-32.


ejoliet:
What command should i run and where is the report result to take a look at? how do you get 60%? There must be a place where this is explained. Please, point me to that reference if any.

ejoliet:
I was able to run and test passes. Cases not covered will be addressed, as explained by @lznakano, next class refactor that will be tested again.
Review done.

PR #216, opened 2016-10-31
DM-6286 Charts Container

tgoldina:
The major classes - MultiChartViewer, MultiChartToolbar

Also refactored charts display. ChartPanel.jsx is now generic and gets info via chartType.
See chartTypes/HistogramTblView and chartTypes/ScatterTblView


src/firefly/js/charts/ui/ChartsContainer.jsx
```diff
@@ -4,92 +4,40 @@
...
-                            <ChartsTableViewPanel key={c} expandable={true} chartId={c}/>
-                        );
-                    })}
-                    {defaultCharts.map((c) => {
-                        return (
-                            <ChartsTableViewPanel key={'default-'+c} expandable={true} chartId={c}/>
-                        );
-                    })}
-                </div>
+        const {chartId, expandedMode, closeable} = this.props;
```
loitly: I don't quite get the purpose of this component.  It's a simple wrapper around ChartPanel and MultiChartViewer without any added logic of its own.  When do we use this?

src/firefly/js/charts/chartTypes/HistogramTblView.jsx
```diff
@@ -0,0 +1,156 @@
...
+        }
+    } else if (optionsKey === 'filters') {
+        const {tableModel} = getChartProperties(chartId);
+        return (
+            <FilterEditorWrapper tableModel={tableModel}/>
+        );
+    }
+}
+
+function renderToolbar(chartId, expandable, expandedMode, toggleOptions) {
```
tgoldina: I did make Chart, Toolbar, and Options functional components first, but then I wanted to switch to non-functional component for one chart type, and realized I can not pass it as a function anymore. May be it's not a big deal, and the recipe would be wrapping it into a functional component.

src/firefly/js/charts/ui/ChartsContainer.jsx
tgoldina: This component handles single chart display from API (both standard and expanded mode) and wraps the default MultiChartViewer. Perhaps, at some point it can be eliminated, but I like to have this wrapper so that only chart code changes when chart code is refactored.

PR #219, opened 2016-11-02
DM-7830: Hook image display into raw table as well.

loitly:
https://jira.lsstcorp.org/browse/DM-7830

 - Also show xyPlot on Peak Table.


ejoliet:
I've checked this PR, does what it says. I've noticed new features when using the 'periodogram', xy-plots came up in a grid mode. Is it the expected behavior?

ejoliet:
If the download feature suppose to be connected and working, it works as expected but i have found a minor issue: file image doesn't end with proper suffix (.fits), not sure if it is the case before migration, just notice it.

PR #221, opened 2016-11-04
DM-8189 handling fetch error for the catalog constraint panel.

cwang2016:
The development adds catalog fetch error handling into catalog constraint panel: 
- Create table object with error information  when catalog fetch error occurs. 
- add generic error display into BasicTableView
- move the table with error creator from TableCntlr into TableUtil


cwang2016:
Test: 
- create some case which may cause error to LSST catalog search, like changing the value of 'table_path' in line 490 of file LSSTCatalogSelectViewPanel.jsx
- click 'LSST Catalogs', then the error message will be shown in the catalog constraint panel.

PR #222, opened 2016-11-07
DM-8162: Call IRSAViewer to display images directly via URL

loitly:
https://jira.lsstcorp.org/browse/DM-8162

As described by the title, allow plotImage to be called via url.
In the above link, you'll find more information on how to construct the url with one sample test url.
This PR is to merge into 'rc' branch.  Once it's in rc, we will merge rc back into dev.

### DIFF
--- a/src/firefly/js/visualize/task/PlotImageTask.js
+++ b/src/firefly/js/visualize/task/PlotImageTask.js
@@ -6,7 +6,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {isArray} from 'lodash';
+import {isArray, uniqueId} from 'lodash';
 import {WebPlotRequest, GridOnStatus} from '../WebPlotRequest.js';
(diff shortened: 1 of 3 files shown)


src/firefly/js/visualize/task/PlotImageTask.js
```diff
@@ -65,11 +65,12 @@ function makeSinglePlotPayload(vr, rawPayload ) {
 
 
     if (isArray(wpRequest)) {
-        if (!plotId) plotId= req.getPlotId();
+        if (!plotId) plotId= req.getPlotId() || uniqueId('defaultPlotId-');
```
robyww: nice way to approach this problem.

ejoliet:
Works as intended. 

Is passing the wpRequest object covers all the needs as before? If yes, i'm fine with the PR. Otherwise we need to work on it, specially to match teh need from NED for example.

ejoliet:
I've asked more example from NED team, for now i think is more than fine so you can merge the PR into RC. If there is something more or special, i will raise a ticket. Thanks!

PR #223, opened 2016-11-07
DM-8176: tweeks for lsst entry points and options

robyww:
- added lsst-api-triview.html for api launch with afw.display
 - added lsst-pdac-triview.html for pdac entry point
 - PROCESS_SCROLL action now has a way to turn off bounds checking

The html files might eventually be moved to an lsst repository

### DIFF
--- a/src/firefly/html/lsst-pdac-triview.html
+++ b/src/firefly/html/lsst-pdac-triview.html
@@ -12,25 +12,28 @@
     <script>
         window.firefly = {
             app: {
+                template: 'FireflyViewer',
                 menu: [
                     {label:'LSST Data', action:'LsstCatalogDropDown'},
                     {label:'External Images', action:'ImageSelectDropDownCmd'},
                     {label:'External Catalogs', action:'IrsaCatalogDropDown'},
                     {label:'Add Chart', action:'ChartSelectDropDownCmd'},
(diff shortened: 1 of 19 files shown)


src/firefly/html/lsst-pdac-triview.html
```diff
@@ -0,0 +1,43 @@
...
+                    {label:'External Images', action:'ImageSelectDropDownCmd'},
+                    {label:'External Catalogs', action:'IrsaCatalogDropDown'},
+                    {label:'Add Chart', action:'ChartSelectDropDownCmd'},
+                    //{label:'Help', action:'app_data.helpLoad', type:'COMMAND'},
+                    ]
+            },
+            MenuItemKeys: {maskOverlay:true},
+            imageTabs: [ 'fileUpload', 'url', '2mass', 'wise', 'sdss', 'msx', 'dss', 'iras' ],
+            irsaCatalogFilter: 'lsstFilter',
+            catalogSpacialOp: 'polygonWhenPlotExist'
```
loitly: If it's not already so, these properties under global firelfy should be documented.

robyww: Yes, I agree.  I am thinking about maybe putting them into an object in the store.  Then the rest of the app references them from the store instead of window.firefly.

PR #224, opened 2016-11-09
dm-8216 Histogram for long values

tgoldina:
- pass 14 sig digits, when creating histogram
- collapse bins if overlapping bins are passed to the client because of precision loss
- use column chart if all bins are of 0 width
- cleanup most warnings in HistogramProcessor.java
- use long to store number of points in a bin

To test, use firefly_test_data/DAXTestData/RunDeepForcedSource_limit100.csv
The following columns are long: id, coordHtmId20, objectId, exposureId.
(The display should be fine – compared to dev version – even though the precision can be reduced with respect to the original.) 

The inherent limitation is that Javascript numbers are 64-bit float point, so the 63 bits in Long are converted to 52 bits of Double, hence the precision is inevitably lost if a number is long enough. Even if we find a way to handle arbitrary precision numbers in Javascript, they are not currently supported by Highcharts charting library.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
@@ -9,6 +9,7 @@ import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.server.packagedata.FileInfo;
 import edu.caltech.ipac.firefly.server.util.ipactable.DataGroupReader;
 import edu.caltech.ipac.firefly.server.util.ipactable.DataGroupWriter;
+import edu.caltech.ipac.firefly.util.DataSetParser;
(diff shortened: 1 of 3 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/HistogramProcessor.java
```diff
@@ -362,49 +366,43 @@ private boolean isInSelection( int idx, ArrayList<Integer> list){
 
             binMin[0] = min;
             //assign the left edge to the binMin
-            for (int i = 1; i < nBin; i++) {
-                binMin[i] = bins[i - 1];
-              }
+            System.arraycopy(bins, 0, binMin, 1, nBin-1);
```
lznakano: Nice! Learned a good way to do array copy.

lznakano:
It looks good to me.  It showed long values. Nice work, especially the modification to the java code. 
Thanks!

PR #228, opened 2016-11-15
DM-6617: fix parsing info link

ejoliet:
Fixed the URL parsing when relative URL is found in master catalog table. In order to test, you need to build FF and  redeploy the branch built. 
Then go to 'catalog' search and check that 'info' and inner link in description are ok.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
@@ -262,7 +262,7 @@ public class CatMasterTableQuery extends IpacTablePartProcessor {
                 int endAnchor= s.indexOf(">",start);
                 if (targetStr!=null && endAnchor>0 && url!=null &&
                         !s.substring(beginChar,endAnchor).contains("target=")) {
-                    String rStr= "\\Q" + beginChar+url+beginChar + "\\E";
+                    String rStr = beginChar + url + beginChar;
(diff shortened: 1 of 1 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
```diff
@@ -262,7 +262,7 @@ private static String modifyToFullURL(String inStr, String  targetStr) {
                 int endAnchor= s.indexOf(">",start);
                 if (targetStr!=null && endAnchor>0 && url!=null &&
                         !s.substring(beginChar,endAnchor).contains("target=")) {
-                    String rStr= "\\Q" + beginChar+url+beginChar + "\\E";
+                    String rStr = beginChar + url + beginChar;
```
loitly: Because rStr is used as a regex, simply removing the \Q \E will have a side-effect.  If the url contains '?' for example, the modified code will fail.

ejoliet: I've checked but those seems to not be part of regex characters set. 
My test showed that the code above is also handling this kind of 'inStr' with '?' as mentioned:

res = WISE Preliminary Post-Cryo Single Exposure (L1b)  Source Table (<a href='http://irsa.ipac.caltech.edu/applications/Gator/superseded.html?a=222&b=444'>Superseded</a>

Will merge the branch as it fixes at least the current problem. If something more complex and not resolved happens i will open a new ticket. Thanks for the review!

PR #229, opened 2016-11-15
DM-8006 Add xy chart to phase folding parameter setting panel using Highcharts

cwang2016:
The implementation mainly, 
- add 2D chart on the client side phase folding dialog and the plot changes as any of the phase folding parameters (time column, flux column, zero time point, period) changes.
- update the parameter items in the parameter setting dialog. 

Test:
- localhost:8080/firefly/lc.html
- open 'lc_raw.tbl' from 'Raw Table', click 'search'
- select tab 'Upload/Phase folding' and click button 'Phase Folding'
- experiment the setting of all entries and move slider to set 'period', the plot will be updated as any of the parameter (except Flux error column)  changes 
- click button 'Phase Folded Table' to close the dialog and a table with phase column is created (or updated) based on current setting of period, time column and zero point time. 

Note:
- The time and flux column names are set in default for IRSA cases. Those column names are settable while the phase folding dialog component is used.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPhaseFoldingPopup.jsx
+++ b/src/firefly/js/templates/lightcurve/LcPhaseFoldingPopup.jsx
@@ -14,6 +14,7 @@ import {ValidationField} from '../../ui/ValidationField.jsx';
 import {InputGroup} from '../../ui/InputGroup.jsx';
 import {doUpload} from '../../ui/FileUpload.jsx';
 import {RangeSlider}  from '../../ui/RangeSlider.jsx';
+import {adjustMax} from '../../ui/RangeSliderView.jsx';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
 import {loadXYPlot} from '../../charts/dataTypes/XYColsCDT.js';
 import FieldGroupUtils from '../../fieldGroup/FieldGroupUtils';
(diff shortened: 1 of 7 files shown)


src/firefly/js/templates/lightcurve/LcPhaseFoldingPopup.jsx
```diff
@@ -170,18 +224,193 @@ function PhaseFoldingSetting() {
...
+                    type: 'scatter',
+                    borderColor: '#a5a5a5',
+                    borderWidth: 1,
+                    borderRadius: 5,
+                    zoomType: 'xy',
+                    height,
+                    width
+                },
+                title: {
+                    text: 'Flux vs. Phase'
```
tgoldina: This is just my opinion: I think this title does not add any new info. Also, title and legend together take too much space from the chart. I would put Period=...  in the title, and remove legend.

ejoliet:
Great job! I really like the feature! It works flawless! 

As Tatiana says, removing the 'highchart.com' is good idea. I still think is good to have the legend with the value of the period. The title is fine too. 

I can think about couple of improvement but as i said before, to me is more a (great!) starting point and i don't think we are ready to call it for release unless LC users (i.e. David, Roc, Frank, Vandana, Luisa) can share more thoughts about it. 
Improvements we could consider or user can expect: adding a step input to divide the slider to have more fine control on the period value, be able to change the min/max value, making the popup resizable, be able to add multiple fluxes and keep only one changing while the others stay on (different) fixed period, long etc.

cwang2016:
some update and fixing: 
- fix a bug that wrongly compared the store status and the parameter updates of the phase folding dialog. That bug caused frequent and improper 2D charts rendering and slowed down the performance. 
- re-style the slider in order to make the slider stand out in less complicated look. 
- put the 'phase folding dialog' button under 'phase folding' tab, and the button is used to start the phase folding dialog.

PR #230, opened 2016-11-15
DM-8179: Preserve background jobs and statuses beyond a browser session.

loitly:
https://jira.lsstcorp.org/browse/DM-8179
Also, refactor AppDataCntlr.js

Background jobs are saved based on a userKey cookie which remained the same unless inactive for more than 2 weeks.
To test this, starts one or more downloads.
The downloads should be visible in other firefly instances of the same browser.
It should also remain after you reload the application or quit and come back some time later.
Also, cancel and remove are reflected in all instances as well.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
@@ -40,11 +40,9 @@ public class ServerEvent implements Serializable {
     public ServerEvent(Name name, EventTarget target, Serializable data) {
         this(name, target, DataType.BG_STATUS, data, SERVER_CONN_ID);
     }
-
     public ServerEvent(Name name, EventTarget target, DataType dataType, Serializable data) {
         this(name, target, dataType, data, SERVER_CONN_ID);
     }
-
     public ServerEvent(Name name, EventTarget target, DataType dataType, Serializable data, String from) {
         this.name = name;
         this.target = target;
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
@@ -12,7 +12,6 @@ package edu.caltech.ipac.firefly.data;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.util.event.Name;
 import edu.caltech.ipac.util.StringUtils;
-import org.apache.xpath.operations.*;
 
 import java.io.Serializable;
 import java.lang.String;
@@ -21,9 +20,10 @@ import java.lang.String;
  * @author Trey Roby
  */
 public class ServerEvent implements Serializable {
+
     public static final String SERVER_CONN_ID = "-1";
     private Name name;
-        private EventTarget target;;
+    private EventTarget target;
     private DataType dataType = DataType.STRING;
     private Serializable data;
     private String from;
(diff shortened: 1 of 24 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/events/WebsocketConnector.java
```diff
@@ -114,24 +116,39 @@ public void close() {
      * @param type  action type.  Either "app_data.wsConnAdded" or "app_data.wsConnRemoved"
```
tgoldina: typo in javadoc: 'event though' -> 'even though'

src/firefly/java/edu/caltech/ipac/firefly/server/events/WebsocketConnector.java
```diff
@@ -114,24 +116,39 @@ public void close() {
...
+                        .forEach(q -> {
+                            List<String> l = connInfo.get(q.getChannel());
+                            if (l == null) {
+                                l = new ArrayList<>();
+                                connInfo.put(q.getChannel(), l);
+                            }
+                            if (!l.contains(q.getConnID())) l.add(q.getConnID());
+                        });
+                FluxAction action = new FluxAction(type, connInfo);
+                ServerEventManager.fireAction(action, new ServerEvent.EventTarget(ServerEvent.Scope.SELF, seq.getConnID(), null, null));
```
tgoldina: Just curious: So each client on the same channel will be notified of every new or closed connection? Why do we need it?

src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
```diff
@@ -91,6 +88,10 @@ public String toJsonString() {
         return sb.toString();
     }
 
+    public enum Scope {SELF, CHANNEL, USER, WORLD}
+
+public enum DataType {JSON, BG_STATUS, STRING}
```
loitly: it was not on purpose.  i am guessing intelliJ has something to do with this.

src/firefly/java/edu/caltech/ipac/firefly/data/ServerEvent.java
```diff
@@ -100,11 +101,19 @@ public String toJsonString() {
...
+        private String userKey;
 
         public EventTarget() {}
 
         public EventTarget(Scope scope) {
             this.scope = scope;
+            if (scope == ServerEvent.Scope.CHANNEL) {
+                this.channel = ServerContext.getRequestOwner().getEventChannel();
+            } else if (scope == Scope.USER) {
+                this.userKey =ServerContext.getRequestOwner().getUserKey();
```
loitly: yes.

src/firefly/java/edu/caltech/ipac/firefly/server/events/ServerEventManager.java
```diff
@@ -32,50 +32,57 @@
...
-        else {
-            sev = new ServerEvent(Name.ACTION, scope, ServerEvent.DataType.JSON, actionStr);
-        }
-
+    /**
+     * Send this JSON string action to the client based on the given scope
+     * @param actionStr
+     * @param scope
+     */
+    public static void fireJsonAction(String actionStr, ServerEvent.Scope scope) {
```
loitly: This is intended.  The idea is that you should be able to send an action to self(connId), user, channel, or world without having to create a target.

src/firefly/java/edu/caltech/ipac/firefly/server/events/WebsocketConnector.java
loitly: This is done to support JS external viewer and python API(?).  I am guessing this information may be useful in the future as well.

loitly:
Good catch in regard to email sync.  It's fixed in the latest commit.  As for email not sent, it will be fixed in DM-8182.  Thank you for the review.

PR #232, opened 2016-11-16
DM-8049:

lznakano:
1. Added end to end tests for CropAndCenter and CropFile
   2. Refactored CropFile.java and CropAndCenter.java after the unit test
   3. Deprecated Crop.java and CropTest.java

To run unit test:
 1. check out the testing data at firefly_test_data with the same branch name.
 2. build and run unit test
     gradle :firefly:test


src/firefly/java/edu/caltech/ipac/visualize/plot/Crop.java
```diff
@@ -19,7 +19,7 @@
 import java.io.FileOutputStream;
 import java.io.IOException;
 
-
+@Deprecated
```
robyww: Crop is deprecated but CropFile is calling clone_header? it should be moved.

lznakano: Good catch.  I will fix it. 

Thanks!

ymeiymei:
Tested the branch DM-8049-AllCropClasses: "gradle test" ran successfully.

PR #234, opened 2016-11-17
DM-8182:  send email when package successfully completed.

loitly:
https://jira.lsstcorp.org/browse/DM-8182

 - email notification is working properly
 - it should also resend email when the email field has changed.

To test:
 - goto: firefly/lc.html
 - download something.
 - when email is given, you should receive email notification when job completes.  you should also see 'Email notification sent' in the job status as well.
 - bring up background monitor and enter a different email.  you should get email notifications for all completed jobs.

### DIFF
--- a/config/mail.prop
+++ b/config/mail.prop
@@ -1,0 +1,7 @@
+mail.use.prop.file = true
+mail.transport.protocol = @mail.transport.protocol@
+mail.smtp.host = @mail.smtp.host@
+mail.smtp.auth = @mail.smtp.auth@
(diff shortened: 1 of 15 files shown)


tgoldina:
I got the email when download request was completed, but none of resentEmail worked for me.

loitly:
I've pushed a commit to address the immediate download scenario.  Please confirm.

loitly:
I've pushed a commit to fix the single file scenario not getting email notification. Please confirm.

PR #236, opened 2016-11-18
DM-8303:Use replace scisql functions by qserv functions

lznakano:
For catalog search (RunDeepSource and RunDeepFourcedSource), qserv functions are used in all four type of searches.  For the image meta search, the scisql functions are still used.  They will be changed in the ticket DM-8334.  

To test this branch:
   1. check it out
   2. build and deploy
   3. connect vpn
   4.  use the value below to run the search:
      position: 9.469,-1.152,
     radius = 0.01

    For ellipse:
       semi-major-axis = 58 arcsec
       ratio = 0.62

Thanks!


ymeiymei:
Tested. Both DeepSource and DeepForcedSource worked fine for cone, ellipse, box and polygon. DeepForcedSource is very slow though. 

Passed.

The image tables work fine with scisql function with this "wrong" image center search. (Will improve the query in DM-8334.)

PR #237, opened 2016-11-19
DM-7964: get period from table/plot

ejoliet:
I've added a saga that will listen to table_highlight on peak and periodogram table. Every time a user click on the plot or table, the saga will get the period out of the table and pre-filled the period input field in phase folded panel.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
+++ b/src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
@@ -440,9 +440,9 @@ function handleTableHighlight(action) {
 function getPeriodFromTable(tbl_id) {
     const tableModel = getTblById(tbl_id);
     if (!tableModel || isNil(tableModel.highlightedRow)) return;
-    if ([PERIODOGRAM].includes(tbl_id)) {
+    if (tbl_id === PERIODOGRAM) {
         return getCellValue(tableModel, tableModel.highlightedRow, 'PERIOD');
-    } else if ([PEAK_TABLE].includes(tbl_id)) {
+    } else if (tbl_id === PEAK_TABLE) {
         return getCellValue(tableModel, tableModel.highlightedRow, 'Period');
     }
 }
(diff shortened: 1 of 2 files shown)


src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
```diff
@@ -408,6 +407,42 @@ function makeField1(hide) {
...
+
+/**
+ * gets the period from either peak or periodogram table, these tables doesn't necesary have a period column and same name
+ * @param {string} tbl_id
+ * @returns {string} period value
+ */
+function getPeriodFromTable(tbl_id) {
+    const tableModel = getTblById(tbl_id);
+    if (!tableModel || isNil(tableModel.highlightedRow)) return;
+    if ([PERIODOGRAM].includes(tbl_id)) {
```
tgoldina: This is a bit confusing, why not tbl_id===PERIODOGRAM? Or you think we'll be adding more elements to this array in future?

ejoliet:
@tgoldina I understand the comment about the active_table action to be handled but i have problem to solve it in a easy and quick way now. 
Explanation: 
The workflow suppose to be the following: the user click on the XY-plot and filled the period input field with the selected data point. When switching to a different table, we want to keep the period as selected before, only when the user click again we want to change it. Right? Otherwise, you have to go back to the table,etc. Imagine if you have multiple periodogram or tables...

As the point is already highlighted, there is no action dispatched that let me know "the plot has been clicked (again)" so is actually little bit tricky. Or do i miss something?

That would mean also that the period selection is driven by the active table, which i dislike. 
Maybe we can accept the solution as it is currently and review this feature together with others next Sprint. 
How does it sound?

src/firefly/js/templates/lightcurve/LcPhaseFoldingPanel.jsx
ejoliet: right, that can be simplified. Thanks!

tgoldina:
@ejoliet you are the domain expert, you know better. Whatever you decide is fine with me. :)

I might have misunderstood, I thought we want period field to be in sync with the highlighted row in peak or periodogram table. If so, we can watch TBL_RESULTS_ACTIVE action as we watch TABLE_HIGHLIGHT action, and get highlighted row with
```
const {highlightedIdx} = TableUtil.getTblInfoById(tbl_id);
```

ejoliet:
@tgoldina i see the implementation path but i still don't know if it solve the workflow required. I will leave as it is and go back when we decide how to complete the workflow. Thanks for sharing!!

PR #238, opened 2016-11-23
DM-7784-7963

wmiipac:
DM-7784-7963
* LC image retrieve use url instead of ibe for general purpose LCViewer
* Add user input image cutout size in Phase Fold Panel.


ejoliet:
I've fixed a bug: the cutout size value wasn't passed along the url retrieve mechanism. 

This PR has still couple of issues that should be addressed/discussed:

- Place of the cutout size input field, might go to the image toolbar instead? (next to wcs checkbox?)
- I feel that the change of the cutout size should be applied directly and not after clicking another row in the table (raw or phase folded tables)
- Single exposure image retrieval function (wp request builder) need to be exposed as an entry point to the app if we want to control how the mission get their single exposure fetched (currently, for IRSA, a url column will be included in the LC table to be looked up and fetched directly), _see related ticket DM-7840 about generic retrieval function via URL._

Based on other reviewer(s) feedback, separate issues will need to be created.

ejoliet:
Another issue found is that the change of the cutout size doesn't picked up once the image has been cached with a previous size.

PR #239, opened 2016-11-23
DM-7781. Moved the test class to the package matching the main class …

ymeiymei:
…location.


lznakano:
The unit test should avoid using other classes if it is possible.  I would like to suggest to do the following:
# The best practice in junit is using setUp() method to initialize some needed variables for testing
# Then add a tearDown() method to nullify the variable defined in the setUp()
# Instead of calling 
{code}
 Method m = FitsRead.class.getDeclaredMethod("getZscaleValue", float[].class, ImageHeader.class, RangeValues.class);
        m.setAccessible(true);
{code}

You can call 
{code}
  Zscale.ZscaleRetval zscaleRetval = new Zscale.ZscaleRetval(...).  

{code}
Thus, you don't need FitsRead in the picture. If FitsRead has bugs, it will not affect Zscale's unit test.

PR #241, opened 2016-11-28
DM-8334: change the search sql expression for catalog's image search …

lznakano:
To test the Catalog's image search (for both DeepCoadd and Science CCD):

1. For intersect is CENTER
    target =  9.469,-1.152
2. COVERS:
    target =  9.469,-1.152
    size = 0.01 degree
3. ENCLOSED:
    target =  0, -1
    size = 1 

Thanks!


ymeiymei:
Tested the queries. Compared with my curl commands. Works great! Good job!

cwang2016:
Review and test complete. The client side shows results from LSST image table queries for all testing cases.

PR #244, opened 2016-11-29
DM-8214 - handling server errors in charts

tgoldina:
This pull request is for handling server error in a chart. Prior to it, loading message would not disappear when an error occurred.

The test relies on a bug in 'Clear' options behavior for histogram. When 'Clear' is pressed, all fields histogram options are cleared, including 'Column or expression value'. If options are applied at this point, an empty column is sent to the server, resulting in an error.

### DIFF
--- a/src/firefly/js/charts/ChartsCntlr.js
+++ b/src/firefly/js/charts/ChartsCntlr.js
@@ -311,6 +311,9 @@ function doChartDataFetch(dispatch, payload, getChartDataType) {
  * @param payload.chartId
  * @param payload.chartDataElementId
  * @param payload.isDataReady
+ * @param payload.error
+ * @param payload.error.message
+ * @param payload.error.reason
  * @param payload.data
  * @param [payload.options]
  * @param [payload.meta]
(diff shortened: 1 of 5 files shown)


src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -522,6 +525,25 @@ export function getChartDataElement(chartId, chartDataElementId=FIRST_CDEL_ID) {
...
+    if (chartDataElements) {
+        Object.keys(chartDataElements).forEach((id) => {
+            const error = chartDataElements[id].error;
+            if (error) {
+                errors.push(error);
+            }
+        });
+    }
+    return errors;
+}
```
cwang2016: -just for your reference, using 'reduce' is nicer than 'forEach' in terms of functional nature.

cwang2016:
review completes. good!

src/firefly/js/charts/ChartsCntlr.js
tgoldina: Thank you for the note. Indeed, looks like in my code I use forEach a lot, even where simple for loop would be more performant. (See https://jsperf.com/array-reduce-vs-foreach/9) I just find forEach more readable, especially when the values need to be accumulated in an array.

robyww: FYI - we should not worry about performance for stuff like this unless it is clear there is a need. Most of our objects and arrays are so small the speed difference is negligible anyway.  Those comparisons will do the same thing like 10,000 times and then compare.

tgoldina: I agree. The only place where it matters is when processing arrays of hundred thousands elements.

loitly: Object.entries() is a good alternative.
const errors = Object.entries(chartDataElements)
                                    .filter( ([k,v]) -> v.error)
                                    .map( ([k,v]) -> v.error);

tgoldina:
Looks great, but Object.entries is a feature of EcmaScript2017, which is in
Draft stage now. Is it OK to use?

I don't remember what decision we made and what ourWebPack setting supports.

2016-11-29 16:52 GMT-08:00 loitly <notifications@github.com>:

loitly:
We use babel-polyfill with es2015 and stage-2 presets.  Object entries and values are supported in babel-polyfill.  I think it should be okay.  I've used it in several places.

PR #248, opened 2016-12-06
DM-8543 Activate the image inside the multi-image viewer without clicking the image iteself

cwang2016:
This development intends to solve the issue that no image is highlighted on the coverage map and FITS data panel while clicking back and forth among the image tabs on the triview. The image handling toolbar is  shown properly based on the selected image tab and its associated active image. 

test: 
localhost:8080/firefly/lsst-pdac-triview.html
image->deep source->target: 0, -1, search type: enclosed, search region size: 3600 arcsec

click 'External images' on top of the triview to get two images rendered in FITS Data panel, 
external images -> wise, create new plot, target: m31, search, 
external images ->wise, create new plot, target:  m51, search

click among all three image tabs, and test.

### DIFF
--- a/src/firefly/js/visualize/MultiViewCntlr.js
+++ b/src/firefly/js/visualize/MultiViewCntlr.js
@@ -1,7 +1,7 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import {without,union,difference, get} from 'lodash';
+import {without,union,difference, get, includes, has} from 'lodash';
 import {flux} from '../Firefly.js';
 import {clone} from '../util/WebUtil.js';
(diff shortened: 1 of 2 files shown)


xiuqin:
IS it possible to remember the active image in previous tab (with multiple images) and highlight it?

cwang2016:
Yes and there is some exception. More description is added in the ticket. 


Cindy

tgoldina:
Just a note: MultiViewCntlr is used by both MultiImageViewer and
MultiChartViewer.

Currently, for charts I keep the activeItemId in the
customData.activeItemId. I also allow it to be undefined, which simplifies
things a bit.

It would be nice to have the same way of dealing with active item in both
images and charts.


2016-12-06 13:47 GMT-08:00 Cindy Wang <notifications@github.com>:

PR #249, opened 2016-12-08
DM-8181: Add script download dialog

loitly:
https://jira.lsstcorp.org/browse/DM-8181

To test:
- launch /firefly/lc.html
- upload a raw table
- click download on top right corner
- enter at least 20 images, so that more than one download parts appear in the background monitor
- when package completes, click on "Get Download Script"
- play around with the different parameters to make sure it's working as intended


cwang2016:
All look good except finding two issues,  
- In the 'Download Retrieval Script' panel, the option 'Don't unzip the file' for 'Uncompression' is not selectable. 
- for the case only one download part shown in 'A sample download', click 'Download Now', the spinning doesn't go away when the download ends.

PR #251, opened 2016-12-14
DM-7966: Default table sorting in LC viewer

loitly:
https://jira.lsstcorp.org/browse/DM-7966 

- set default sorting on raw, phase, and peak tables
 also:
 - allow for raw table upload only, but added option to specify time and flux column
 - code cleanup
 - fix table index out-of-bound when navigate via arrow keys
 - change popup phase folded values from 3 decimal places to 8 to match the server-side version

Use /firefly/lc.html to test.

### DIFF
--- a/src/firefly/js/tables/SortInfo.js
+++ b/src/firefly/js/tables/SortInfo.js
@@ -5,6 +5,17 @@ export const SORT_ASC = 'ASC';
 export const SORT_DESC = 'DESC';
 export const UNSORTED = '';
 
+
+/**
+ * convenience function to create the serialized string representation of SortInfo
+ * @param {string} colName column's name
+ * @param {boolean} [isAscending=true] false for descending order, otherwise it's default to ascending.
+ */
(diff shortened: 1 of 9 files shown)


ejoliet:
Great! Sorted tables works as expected. 

Minor thing that worth be noticing but I don't think it is part of the ticket, once i choose a flux column, i would expect to see the same in the other panels (period finding, phase folded,etc.). What do you think? Maybe this should be part of the layout change and how we use the same column name elsewhere.

2 minor issues though:

1. When i close one table, the other one is unresponsive. 
2. if i upload another raw lc, i would expect the rest of the tables to be cleared out because they are not related but i don't know if this is a right workflow. Could be considered.

Maybe would be good to check at least the first issue before merging merge.

loitly:
The issues you raised are existing issues that should be resolved in other pull requests especially since we have open tickets to revamp the flow and layout of this app.  I recommend merging this pull request as is, and address these issues in the other tickets.  What do you think?

PR #253, opened 2016-12-15
DM-7779: add unit test for FitsRead class

lznakano:
To test:
          1. check out this branch
          2. check out the data in firefly_test_data with the same branch name
          3. build and then run the unit test by :
                 gradle :firefly:test

Thanks!


ejoliet:
Great test!
Overall is a good start. I have couple of comments that could make the test more complete and improve the coverage:

- Coverage of the methods of FitsRead class is low (41%) that means that changes in the class won't be tested/detected in CI tests: either check that the uncovered methods are not often used elsewhere in the code, in that case it should be removed and refactor at some point or in the contrary, the test should be covering those method. Ex: createFitsReadWithGeom() or getDataType or cases in updateRefHeader
- Most of the exceptions and extreme cases are not tested (nulls, not found, size==0,etc): in general is good to test those cases too and that it can be applied to any particular unit test

In general, i would advocate to add tests and improve them later. I'm happy to see that important classes are going under test and will help us detect changes in the future.

Hence, for now, i would advice to merge the branch.

PR #254, opened 2016-12-16
DM-3612 scatter plot errorbars

tgoldina:
In this ticket, the option is provided to enter a symmetric error  for X and Y column. Asymmetric errors will be handled by a separate ticket.

A column or an expression can be entered next to X and Y field in plot options.
Good test is WISE color-color plots: use w[1,2,3,4]sigmpro as an uncertainty for w[1,2,3,4]mpro. 

If less than 20 points are displayed, the error bar will have whiskers, otherwise no whiskers.

To test, bring up a catalog search in Firefly, choose WISE. After the default chart comes up, open the options, choose w1mpro for X, w1sigmpro for X Err, w2mpro for Y, w2sigmpro for Y Err. Do the search, zoom to see error bars, check tooltip to see the uncertainty, when available. (If not available, the corresponding cell in the table should show null.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "react-addons-perf" : "15.3.1",
     "react-component-resizable": "1.0.1",
     "react-grid-layout": "0.13.3",
-    "react-highcharts": "10.0.0",
+    "react-highcharts": "11.0.0",
     "react-split-pane": "0.1.44",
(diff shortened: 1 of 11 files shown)


cwang2016:
review completes. looks good. test is good.

PR #255, opened 2016-12-16
DM-8542 Add support for changing draw layer symbol and the size

cwang2016:
The development mainly adds support for the user to change the symbol and its size when overlaying a catalog on an image including the features: 
- catalog default symbol is square and size 3 to 20, size means the size shown on UI. 
- add shape options for the symbol, 
   x, square, cross, diamond, circle, arrow, boxcircle, and dot. 
- create a dialog for changing the symbol and its size, each drawing layer has its own dialog. 
- the change in color will update symbol content in symbol dialog and drawing layer panel in real time. 
- the change in symbol, size, and color will update the catalog symbol overlaying the image in real time.

### DIFF
--- a/src/firefly/js/ui/PointShapeSizePicker.jsx
+++ b/src/firefly/js/ui/PointShapeSizePicker.jsx
@@ -29,14 +29,44 @@ const PointOptions = [ DrawSymbol.CIRCLE, DrawSymbol.SQUARE, DrawSymbol.DIAMOND,
 
 const MINSIZE = 3;
 const MAXSIZE = 20;
-const popupId = 'ShapePickerDialog';
+const popupIdBase = 'ShapePickerDialog';
 const ARROW_UP = 38;
 const ARROW_DOWN = 40;
 
+var popupId;
+
+
+const colours = {"aliceblue":"#f0f8ff","antiquewhite":"#faebd7","aqua":"#00ffff","aquamarine":"#7fffd4","azure":"#f0ffff",
+    "beige":"#f5f5dc","bisque":"#ffe4c4","black":"#000000","blanchedalmond":"#ffebcd","blue":"#0000ff","blueviolet":"#8a2be2","brown":"#a52a2a","burlywood":"#deb887",
+    "cadetblue":"#5f9ea0","chartreuse":"#7fff00","chocolate":"#d2691e","coral":"#ff7f50","cornflowerblue":"#6495ed","cornsilk":"#fff8dc","crimson":"#dc143c","cyan":"#00ffff",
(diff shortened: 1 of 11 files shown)


src/firefly/js/ui/PointShapeSizePicker.jsx
```diff
@@ -0,0 +1,511 @@
...
+    "navajowhite":"#ffdead","navy":"#000080",
+    "oldlace":"#fdf5e6","olive":"#808000","olivedrab":"#6b8e23","orange":"#ffa500","orangered":"#ff4500","orchid":"#da70d6",
+    "palegoldenrod":"#eee8aa","palegreen":"#98fb98","paleturquoise":"#afeeee","palevioletred":"#d87093","papayawhip":"#ffefd5","peachpuff":"#ffdab9","peru":"#cd853f","pink":"#ffc0cb","plum":"#dda0dd","powderblue":"#b0e0e6","purple":"#800080",
+    "rebeccapurple":"#663399","red":"#ff0000","rosybrown":"#bc8f8f","royalblue":"#4169e1",
+    "saddlebrown":"#8b4513","salmon":"#fa8072","sandybrown":"#f4a460","seagreen":"#2e8b57","seashell":"#fff5ee","sienna":"#a0522d","silver":"#c0c0c0","skyblue":"#87ceeb","slateblue":"#6a5acd","slategray":"#708090","snow":"#fffafa","springgreen":"#00ff7f","steelblue":"#4682b4",
+    "tan":"#d2b48c","teal":"#008080","thistle":"#d8bfd8","tomato":"#ff6347","turquoise":"#40e0d0",
+    "violet":"#ee82ee",
+    "wheat":"#f5deb3","white":"#ffffff","whitesmoke":"#f5f5f5",
+    "yellow":"#ffff00","yellowgreen":"#9acd32"};
+
```
robyww: are this colors for you background algorithm? You might want to clean it up so it uses single quotes and the keys are not quoted.

robyww: I understand more this is a named color map to RGB. I think you want to this objects and the background utility routines to util/Color.js.

src/firefly/js/ui/PointShapeSizePicker.jsx
```diff
@@ -0,0 +1,511 @@
...
+            hsv.saturation /= 100.0;
+            hsv.value /= 100.0;
+
+
+            let i = Math.floor(hsv.hue);
+            let f = hsv.hue - i;
+            let a = hsv.value * (1.0 - hsv.saturation);
+            let b = hsv.value * (1.0 - hsv.saturation * f);
+            let c = hsv.value * (1.0 - hsv.saturation * (1-f));
+            let d = hsv.value;
```
robyww: you could use `const` for these instead of `let`

ejoliet:
Yay!! Great extra feature for catalogs use on images. I think the options are clear enough and the fact that the change is applied immediately is also a extra bonus for usability. I like it a lot. Good job!

cwang2016:
Some update:
- only one symbol dialog is rendered on the selected symbol on the draw layer panel.
- move  color related routines to Color.js

PR #256, opened 2016-12-16
DM-7970: Image settings for Light curve

robyww:
- WCS target match on by default.
 - The image stretch and color is saved as preference
 - object position should be centered in the display by default.
 - object is overlaid

To test: just read in the raw table.  The changes are visible.


ejoliet:
It works perfectly fine and adding the overlay on active target is also a good idea as improvement. 
Minor thing, if i deselect the 'target match', it shouldn't enable by itself when i switch to a different row.

The only twist i would suggest to add is:
- if i zoom, the zoom level should be kept - i think Vandana/Gregory when they talked about 'stretch' was about the zoom level, you might want to ask them.

What do you think?
With both minor fixes, i think you could merge.

loitly:
It looks great and works as described.

PR #258, opened 2016-12-17
DM-7965 make consistent highlight and selection colors

tgoldina:
This ticket is for making highlight and selection colors consistent in tri-view. (Highlight matching orange image border color, selection is yellow.)

### DIFF
--- a/src/firefly/js/charts/ui/Histogram.jsx
+++ b/src/firefly/js/charts/ui/Histogram.jsx
@@ -5,7 +5,7 @@
 import React, {PropTypes} from 'react';
 import ReactHighcharts from 'react-highcharts';
 import numeral from 'numeral';
-import {set} from 'lodash';
+import {get, set} from 'lodash';
 import {getFormatString} from '../../util/MathUtil.js';
 import {logError} from '../../util/WebUtil.js';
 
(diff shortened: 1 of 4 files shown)


loitly:
Changes are fine.  It works as expected.
However, the highlighted(orange) and selected(yellow) dots in plot look almost identical.  Also, when there is only 2 images, I can't tell which one is highlighted.  The blue border is just as strong as the orange one.
Of course, this is only my opinion.  It should not stop you from merging.

ejoliet:
Color seems fine. But I was randomly checking and the tooltip in histograms is broken, an error log message says:

> Histogram.jsx:376` Uncaught TypeError: Cannot read property 'name' of undefined
>     at Object.formatter (Histogram.jsx:376)
>     at a.Tooltip.refresh (highcharts.js:188)
>     at a.Pointer.runPointActions (highcharts.js:198)
>     at a.Pointer.onContainerMouseMove (highcharts.js:206)
>     at HTMLDivElement.c.onmousemove (highcharts.js:208)

ejoliet:
Looks that the problem is not part of the branch, but was introduced before (building 'dev' shows the same problem (and others)
A hint from @tgoldina : a change from `"react-highcharts": "10.0.0"`, to `"react-highcharts": "11.0.0"` in `package.json`.

tgoldina:
Emmanuel, thank you for finding the histogram tooltip issue. I think, it was a side effect of migrating to Highcharts 5. Anyway, I have fixed this issue in this pull request. 

I decided to merge it, so that people have something to look at: the highlighted color is "orangy" in all 3 views, and selected color is "yellowish" in both image and plot.

The colors are easy to adjust as long as we decide what they should be. However we should be aware that the same color looks different on different background. I made table highlight lighter shade of orange, and adjusted the saturation of xy plot highlight to make it different enough from the yellow of the selected points. I realize that it might not be perfect, given that color perception is different from person to person. Still, it's more consistent than it was. :)

PR #262, opened 2017-01-04
DM-8548 fixed the wrong deepCoaddId in the image title

lznakano:
For ScienceCCD, its scienceCCDExposureId is the base id which does not contain the filterId, but for DeepCoaddId, it is the baseId + filterId.    For both, I used 
title = id +'_'+filterName.

To test it, please
  1. check it out and build
  2. connect vpn 
  3. start tomcat and then open this the http://localhost:8080/firefly/lsst-pdac-triview.htm
  4. select image, then table name
  5 enter 9.469,-1.152 at the Name or Position text box 

Thanks,


ymeiymei:
The bug is fixed: The IDs in the image title for the science ccd exposure images are correct now.

PR #265, opened 2017-01-05
DM-8302 options for opposite axis location: X on top, Y on right

tgoldina:
Separated reversed and opposite ("top" for X, "right" for Y) axis options.
Refactored code to use ColumnOrExpression field in histogram or other charts.

https://jira.lsstcorp.org/browse/DM-8302

To test, open chart options and play with axis options in collapsible for X or Y axes.

### DIFF
--- a/src/firefly/js/charts/ui/Histogram.jsx
+++ b/src/firefly/js/charts/ui/Histogram.jsx
@@ -77,7 +77,6 @@ export class Histogram extends React.Component {
                 let doUpdate = false;
                 if (height !== nextProps.height || width !== nextProps.width ) {
                     chart.setSize(nextProps.width, nextProps.height, false);
-                    return false;
                 }
 
                 if (desc !== nextProps.desc) {
@@ -99,8 +98,8 @@ export class Histogram extends React.Component {
                 this.error = error;
                 chart.showLoading(error);
             }
-            return false;
         }
+        return false;
     }
 
     /*
--- a/src/firefly/js/charts/ui/Histogram.jsx
+++ b/src/firefly/js/charts/ui/Histogram.jsx
@@ -6,6 +6,7 @@ import React, {PropTypes} from 'react';
 import ReactHighcharts from 'react-highcharts';
 import numeral from 'numeral';
 import {get, set} from 'lodash';
+import shallowequal from 'shallowequal';
 import {getFormatString} from '../../util/MathUtil.js';
 import {logError} from '../../util/WebUtil.js';
 
@@ -51,6 +52,9 @@ export class Histogram extends React.Component {
      * @param {number} props.width - width of the chart in pixels
      * @param {number} props.height - height of the chart in pixels
      * @param {string} [props.logs] - can have values 'x', 'y', or 'xy'
+     * @param {Object} [props.xAxis] - Highcharts xAxis properties
+     * @param {Object} [props.yAxis] = Highcharts yAxis properties
+     * @param {string} [props.opposite] - can have values 'x', 'y', or 'xy'
      * @param {string} [props.binColor='#d1d1d1'] - darker bin color
      * @param {string} props.desc - description
      * @public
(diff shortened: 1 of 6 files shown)


src/firefly/js/charts/ui/Histogram.jsx
```diff
@@ -63,39 +67,38 @@ export class Histogram extends React.Component {
...
-            let doUpdate = false;
-            if (height !== nextProps.height || width !== nextProps.width ) {
-                chart.setSize(nextProps.width, nextProps.height, false);
-                return false;
-            }
+            try {
+                let doUpdate = false;
+                if (height !== nextProps.height || width !== nextProps.width ) {
+                    chart.setSize(nextProps.width, nextProps.height, false);
+                    return false;
```
loitly: this function is a bit confusing to me.  beside returning should update, it also does a chart.redraw.
but here, a setSize() does not need to be redrawn?

src/firefly/js/charts/ui/Histogram.jsx
```diff
@@ -63,39 +67,38 @@ export class Histogram extends React.Component {
...
+                    doUpdate = true;
+                }
+
+                if (doUpdate) { chart.redraw(false); }
+            } catch (error) {
+                this.error = error;
+                chart.showLoading(error);
             }
-            if (doUpdate) { chart.redraw(false); }
             return false;
```
loitly: this return false statement should be moved outside of the if(chart) block.

src/firefly/js/charts/ui/Histogram.jsx
```diff
@@ -459,17 +458,20 @@ export class Histogram extends React.Component {
...
+    xAxis: {},
+    yAxis: {}
 };
 
 
 // when more than one histogram is defined use series
 Histogram.propTypes = {
     series: PropTypes.arrayOf(PropTypes.object), // array of objects with data, binColor, and name properties
+    xAxis: PropTypes.object,
+    yAxis: PropTypes.object,
```
loitly: It may be useful if you made this a shape type with defined props.  just a thought.  I am not sure how useful it is.

src/firefly/js/charts/ui/XYPlot.jsx
```diff
@@ -22,7 +22,7 @@ export const axisParamsShape = PropTypes.shape({
     label : PropTypes.string,
     unit : PropTypes.string,
     error: PropTypes.string,
-    options : PropTypes.string // ex. 'grid,log,flip'
+    options : PropTypes.string // ex. 'grid,log,flip,opposite'
```
loitly: can you use oneOf here?

tgoldina:
Regarding the first comment. Highcharts separate redraw and reflow. Reflow
happens automatically on container resize, it's more efficient than redraw.
But you are right that return should not be there, because other props can
change too.

------------------------------
>
> In src/firefly/js/charts/ui/Histogram.jsx
> <https://github.com/Caltech-IPAC/firefly/pull/265#pullrequestreview-15413470>
> :
>
> >          const chart = this.refs.chart && this.refs.chart.getChart();
>          if (chart) {
> -            let doUpdate = false;
> -            if (height !== nextProps.height || width !== nextProps.width ) {
> -                chart.setSize(nextProps.width, nextProps.height, false);
> -                return false;
> -            }
> +            try {
> +                let doUpdate = false;
> +                if (height !== nextProps.height || width !== nextProps.width ) {
> +                    chart.setSize(nextProps.width, nextProps.height, false);
> +                    return false;
>
> this function is a bit confusing to me. beside returning should update, it
> also does a chart.redraw.
> but here, a setSize() does not need to be redrawn?
>

tgoldina:
>
> In src/firefly/js/charts/ui/Histogram.jsx <https://github.com/Caltech-IPAC/firefly/pull/265#pullrequestreview-15413470>:
>
> >  };
>
>
>  // when more than one histogram is defined use series
>  Histogram.propTypes = {
>      series: PropTypes.arrayOf(PropTypes.object), // array of objects with data, binColor, and name properties
> +    xAxis: PropTypes.object,
> +    yAxis: PropTypes.object,
>
> It may be useful if you made this a shape type with defined props. just a
> thought. I am not sure how useful it is.
>
I was hoping that xAxis and yAxis can hold any relevant Highcharts option.
For now, we pass "reversed" and "opposite", but if you use a component
directly as in
http://localhost:8080/firefly/demo/ffapi-highlevel-charttest.html, you can
supply any option for xAxis or yAxis via these properties.

I am still not sure what would be the best way to supply random Highcharts
properties. If we allow to pass the whole config, how do we map series to
the data? We create series dynamically, based on  parameters. In addition
to data, there could be error series, selected points series, a highlight.
How do we give user an ability to tweak options for each of those?

PR #267, opened 2017-01-09
DM-8684: allow for others types of image expanded mode

robyww:
- Light curve viewer now has it own expanded mode


loitly:
Changes look fine.  However, after you expand image, select image count: 1, then close, you'll see an additional 'expand' button on that image.  You should fix this before merging.

PR #269, opened 2017-01-12
Dm 8957 plot abort

robyww:
- handle plot aborting

Also fixed the following:

- now plots with same id dow not get overridden by previous request
-  fixed problem with image url that has parameters
- ibe image retrieve does not do locking or caching, required changing several files
- cleaned up LockingVisNetwork some. First steps to remove the VisNetwork layer
- better optimize image loads with ImageMetaDataWatcher
- removed redundent for AnyFitsGetter
- image coverage overlay now have a thicker selected box (3 px)
- fixed very small bug with image coverage overlay show symbol change options
- fix PlotTitle.js from calling function that does not exist


ymeiymei:
I have built the branch dm-8957-plot-abort and verified the test case in the description field. The bug is fixed: When I clicked any row while downloading images, the whole tri-view changed to the row I just clicked. When I clicked back the row which was downloading the image, the downloading resumed.

cwang2016:
Review completes. Works well!

PR #272, opened 2017-01-19
DM-8668 lc period finder

cwang2016:
The development includes:
- period finding layout before and after periodogram which contains the panels 
   (LcPeriods.jsx)
    . to enter values of period finding parameters
    . to show phase folding preview of xy plot
    . to show periodogram and peak tables and XT plots (after periodogram)
- Find periodogram dialog  which inherits some parameters from period finding panel
   (LcPeriodogram.jsx)
- Revamp LcManager to work with period finding and periodogram finding flow 
   (LcManager.js)
- add 'keepMounted' property to FieldGroup component which allows the parameters under the field group to be kept mounted while the component is unmounted. 
- update 'RangeSlider' component by removing the text input field attached to the slider. 
Test:
- localhost:8080/firefly/lc.html. 
- choose file - lc_raw.tbl ->search, 'Periof Finding' button is shown on the top menu after the search. 
- click button 'Period Finding' to get period finding layout
- enter values or move the slider in period finding panel to see the change of phase folding XY chart and the change of the buttons at the bottom ( 'Back to Period', 'Accept Period') 
- click button 'Find Periodogram' to get Periodogram popup dialog
   . 'reset' button restores the values to those as the popup is started. 
   . 'Periodogram Calculation' button exits the dialogs and calculates periodogram/peak tables and charts. 
  . 'cancel' button restores the field values and exits the dialog. 
- select the period from the periodogram/peak tables or charts to see the change of phase folding  plot.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcManager.js
+++ b/src/firefly/js/templates/lightcurve/LcManager.js
@@ -2,10 +2,11 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {isUndefined, get,isNil} from 'lodash';
+import {isUndefined, get,isNil, has, set} from 'lodash';
 import {take} from 'redux-saga/effects';
-
-import {LO_VIEW, LO_MODE, SHOW_DROPDOWN, SET_LAYOUT_MODE, getLayouInfo, dispatchUpdateLayoutInfo, dropDownHandler} from '../../core/LayoutCntlr.js';
+import {flux} from '../../Firefly.js';
+import {LO_VIEW, LO_MODE, SHOW_DROPDOWN, SET_LAYOUT_MODE, SET_LAYOUT_DISPLAY,  getLayouInfo,
+        dispatchUpdateLayoutInfo, dropDownHandler, dispatchLayoutDisplayMode} from '../../core/LayoutCntlr.js';
 import {TBL_RESULTS_ADDED, TABLE_LOADED, TBL_RESULTS_ACTIVE, TABLE_HIGHLIGHT} from '../../tables/TablesCntlr.js';
 import {getCellValue, getTblById, makeTblRequest} from '../../tables/TableUtil.js';
 import {updateSet} from '../../util/WebUtil.js';
@@ -21,8 +22,10 @@ import {ServerRequest} from '../../data/ServerRequest.js';
 import {CHANGE_VIEWER_LAYOUT} from '../../visualize/MultiViewCntlr.js';
 import {LcPFOptionsPanel, grpkey} from './LcPhaseFoldingPanel.jsx';
 import FieldGroupUtils, {revalidateFields} from '../../fieldGroup/FieldGroupUtils';
+import {VALUE_CHANGE} from '../../fieldGroup/FieldGroupCntlr.js';
 import {makeWorldPt} from '../../visualize/Point.js';
 import {CoordinateSys} from '../../visualize/CoordSys.js';
+import {getMenu, dispatchSetMenu} from '../../core/AppDataCntlr.js';
 
 export const LC = {
     RAW_TABLE: 'raw_table',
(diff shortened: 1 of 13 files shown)


ejoliet:
BTW, i think the `back period` feature is nice but not required ;-) 
I'm also not sure about how making the user aware of recalculate the periodogram. I saw the button `change periodogram` but i wonder if there is another way to do it. We can discuss later about the improvements.

src/firefly/js/templates/lightcurve/LcPeriod.jsx
```diff
@@ -0,0 +1,1035 @@
...
+        );
+
+    }
+}
+
+/**
+ * @summary light curve phase folding FieldGroup rendering
+ * @param period
+ * @param minPeriod
+ * @param maxPeriod
```
loitly: this jsdoc is not valid.  there's only one param in this function.  there are also a few more places where validation fail (intelliJ 'red' indicators).

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -5,14 +5,15 @@
 
 import React, {Component, PropTypes} from 'react';
 import sCompare from 'react-addons-shallow-compare';
-import {pickBy} from 'lodash';
+import {pickBy, startsWith} from 'lodash';
```
loitly: same comment.  is lodash's startsWith different from string.startsWith?

loitly:
I have these comments from testing the UI:
- There should not be a tab panel with input fields in the results view.
- We should have a title with some simple description the 'Period Finding' panel.
- The 'period = x.xx(day)' chart should be resizeable and it is not.
- Selecting period from table row seems to be more sluggish then selecting period from xyplot.  I am guessing some extra steps was necessarily done.

Bug:
- load raw table -> select 'Period Finding' -> select 'Cancel' -> results is blank.
  - I expect to still see the raw table with images and xyplot for the raw table.
- load raw table -> select 'Period Finding' -> select 'Find Periodogram' -> then cancel Period Finder while Periodogram popup is visible.  Periodogram does not go away.  It stays on top of the results view.

loitly:
There are things that need to be fixed.  But, it can be done at a later stage.  Feel free to merge if you plan to fix them in another ticket.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -2,10 +2,11 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {isUndefined, get,isNil} from 'lodash';
+import {isUndefined, get,isNil, has, set, lowerCase} from 'lodash';
```
cwang2016: change to use string.toLowerCase()

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -100,19 +144,44 @@ export function* lcManager(params={}) {
...
+                }
+                break;
         }
 
         if (newLayoutInfo !== layoutInfo) {
             dispatchUpdateLayoutInfo(newLayoutInfo);
-
+        }
+        if (bInitPeriod) {
+            periodPageMode(LC.PERIOD_PAGE);
```
cwang2016: dispatchUpdateLayoutInfo is preferred to be called first in order to get better visual effect for  the page transition.

src/firefly/js/templates/lightcurve/LcViewer.jsx
cwang2016: yes, string.startsWith is used instead.

ymeiymei:
Reviewed. Agree with Loi's comments on the UI.

Two more comments.

In the current version, it seems like we have interaction between the period finding panel and the result panel: click a row in the table or a dot in the xyplot, the period finding changes; Select a period in the period finder, the table and xyplot change. But when period = 48 is selected, the xyplot stays at 3.27, also the table. I guess that is because no values between period = 3.27 and 193 for this calculation. 

Also it would be better to put the Help tab at the most right side. The content of Help can link to IRSA's LC Help?

Cindy told me she was doing very major modification. I'll stop reviewing this version. If my comments above conflict with the new design, just ignore them.

cwang2016:
fixed the following bugs, 
- load raw table -> select 'Period Finding' -> select 'Cancel' -> results is blank.
- load raw table -> select 'Period Finding' -> select 'Find Periodogram' -> then cancel Period Finder while Periodogram popup is visible. Periodogram does not go away. It stays on top of the results view.

will fix the issues,
- make table/chart consistent with the period value
- add help link

PR #273, opened 2017-01-20
DM-8367: Line chart

tgoldina:
Added support for line plot
- added plotStyle ('points' and 'line', defaults to 'points') and sortColOrExpr (defaults to x.colOrExpr) to XY plot parameters. For smaller (<5000 rows) tables, plotStyle shows in options; 
- added support for sortColOrExpr to XYWithErrorsProcessor;
- improved the logic to see if server fetch is needed, the function now has access to the server call parameters, with which the data were obtained 

To test, load a catalog less then 5000. Open options and select 'Line' as a plot style.

https://jira.lsstcorp.org/browse/DM-8367

### DIFF
--- a/src/firefly/js/charts/dataTypes/XYColsCDT.js
+++ b/src/firefly/js/charts/dataTypes/XYColsCDT.js
@@ -16,6 +16,7 @@ import {colWithName, getNumericCols, SCATTER} from './../ChartUtil.js';
 import {serializeDecimateInfo} from '../../tables/Decimate.js';
 
 export const DT_XYCOLS = 'xycols';
+const DECI_ENABLE_SIZE = 5000; // matching QueryUtil.DECI_ENABLE_SIZE
 
 /**
  * Chart data type for XY columns
@@ -231,7 +232,7 @@ function serverParamsChanged(oldParams, newParams, chartDataElement) {
 
 function isLargeTable(tblId) {
     const {totalRows}= getTblById(tblId);
-    return (totalRows >= 5000);
+    return (totalRows >= DECI_ENABLE_SIZE);
 }
 
 //function errorsOrSortRequested(xyPlotParams) {
--- a/src/firefly/js/charts/dataTypes/XYColsCDT.js
+++ b/src/firefly/js/charts/dataTypes/XYColsCDT.js
@@ -197,31 +197,53 @@ export function setZoom(chartId, chartDataElementId, selection=undefined) {
     }
 }
 
-function serverParamsChanged(oldParams, newParams) {
+function serverParamsChanged(oldParams, newParams, chartDataElement) {
     if (oldParams === newParams) { return false; }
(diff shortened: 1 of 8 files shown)


ejoliet:
Great stuff! Works smoothly! No problem found. 

I have 2 minor comments/open questions:

- probably we want to keep the `dots` points that connects the lines (maybe need some feedback from LSST/IRSA about that)
- Do we want to make the option sticky? I mean, if i decide to see a line chart with one catalog search, after bringing a new search, do i expect to see a line chart too? Right now, the option is by default coming with 'scatter' plot.

ejoliet:
BTW, about the first comment: it seems that the dots (points) are visible or not depending on the zoom. I think it probably need to be consistent and be always visible. What do you think?

src/firefly/js/charts/ui/XYPlot.jsx
```diff
@@ -743,11 +746,25 @@ export class XYPlot extends React.Component {
...
+                        }
+                    }
+                }
+            },
             title: {
                 text: desc
             },
             tooltip: {
-
+                snap: 10,
```
ejoliet: This is the same option ('`snap`') that is used for the cross-hair feature BTW.

src/firefly/js/charts/dataTypes/XYColsCDT.js
```diff
@@ -197,32 +197,54 @@ export function setZoom(chartId, chartDataElementId, selection=undefined) {
...
 }
 
-function errorsRequested(xyPlotParams) {
-    return xyPlotParams &&
-        (xyPlotParams.x.error || xyPlotParams.y.error ||
-        xyPlotParams.x.errorLow || xyPlotParams.y.errorLow ||
-        xyPlotParams.x.errorHigh || xyPlotParams.y.errorHigh);
+function isLargeTable(tblId) {
+    const {totalRows}= getTblById(tblId);
+    return (totalRows >= 5000);
```
ejoliet: I think it would make sense to convert the `5000` value into a constant, what do you think?

src/firefly/js/charts/ui/XYPlotOptions.jsx
```diff
@@ -449,7 +455,7 @@ export class XYPlotOptions extends React.Component {
     render() {
         const { colValStats, groupKey, xyPlotParams, defaultParams, onOptionsSelected}= this.props;
 
-        const noLogOption = possibleDecimatedTable(colValStats);
+        const largeTable = possibleDecimatedTable(colValStats);
```
ejoliet: What is the difference with `isLargeTable(...)`?

tgoldina:
The recent commit 
- separates 'Line' and 'Connected Points' options ('line' and 'linepoints' plot style 
- fixes bug in radio input group, which would cause plot style show no selection, when plot style was modified in Charts dropdown while some charts options are open.
 - updates API, so that new options can be used

PR #274, opened 2017-01-24
DM-7840 Change LC WISE image to ibe call, plus get image band from th…

wmiipac:
DM-7840
This change is for WISE LC, we are using ibe call for image retrieve, and the image band and cutout size are passed from the phase folding panel.


wmiipac:
DM-9078 should be related ticket to make the image display work

ejoliet:
Problem was fixed in DM-8985

> fixed FileInfo null retured bug in last pull request

The branch can be merged.

PR #275, opened 2017-01-27
DM-8660: add unit test for Geom class

lznakano:
Please check out the data branch at firefly_test_data with the same branch name to run the tests.
The coverage of this unit test is 85% of the method and 57% of lines. There are still more room to improve.  But for now, I think it is sufficient.


ymeiymei:
Reviewed the unit test. Great job! Good coverage (the exception part need more coverage though).

The Geom.java needs refactor for sure!

PR #276, opened 2017-01-27
DM-8985: server now finding image with mask extension, mask UI implemented

robyww:
DM-8985: server now finding image with mask extension

 - server discovers mask layers in fits file
 - mask and other overlay information returned as addtional data for image load
 - give user options to load mask in drawing layers panel
 - add show all and hide all buttons
 - improvement in consistent color choosing of mask layers
 - mask can be lazy loaded
 - enable all mask in group, on/off in group,
 - fixed FileInfo null retured bug in last pull request
 - rotate ui: matching group, matching last plot
 - fixed zoom locking now consistently working, added LEVEL as plot zoom options to disable locking
 - made the default ZoomType TO_WIDTH_HEIGHT
 - fixed small toolbar layout background size issue

**To Test**:
 - use validation_data_hsc repository:
 - look in directory validation_data_hsc/DATA/00533/HSC-R/corr
 - load any CORR*.fits
    interesting mask:
           - bit number 4, 5,  and 11 are interesting to turn on.
           - CORR-0903342-100.fits has bad pixels so try bit number 1 as well
 - bring up layer control
 - enable the mask layers, the layers should appear
 - turn any on and off 
- try the show all or hide all buttons
- try zoom
- try rotate and flip though they are a little flaky

**known bugs**:
 - flip does not work if you flip before you turn on a mask bit
 - flip and rotate are a fragile in general, might need to look into it more deeply in another ticket

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
@@ -406,6 +406,15 @@ public class WebPlotReader {
         }
     }
 
+    /**
+     * This method attempts to find how data might be related in a multi-extension fits file. I expect it will grow
+     * and become more advanced over time.
+     * Currently it looks of the extension type marked 'IMAGE' and makes that the based. Any extensions marked 'MASK' or
+     * 'VARIANCE' are related data.
+     * @param f the fits file name
+     * @param frAry the array of FitsRead objects that came from the file.
+     * @return the data relations
+     */
     private static List<List<RelatedData>> investigateRelations(File f, FitsRead frAry[]) {
         if (frAry.length>1) {
             int imageIdx= -1;
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
@@ -3,8 +3,9 @@
  */
 package edu.caltech.ipac.firefly.server.visualize;
 
-import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.data.FileInfo;
+import edu.caltech.ipac.firefly.data.RelatedData;
+import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.visualize.Band;
 import edu.caltech.ipac.firefly.visualize.VisUtil;
 import edu.caltech.ipac.firefly.visualize.WebPlotRequest;
@@ -27,8 +28,10 @@ import nom.tam.fits.FitsException;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 
@@ -91,7 +94,7 @@ public class WebPlotReader {
         modFileWriter= checkUnzip(imageIdx,Band.NO_BAND,originalFile, modFileWriter);
 
         retval= new FileReadInfo(originalFile, fitsRead, Band.NO_BAND, imageIdx,
-                                 fd.getDesc(), uploadedName, modFileWriter);
+                                 fd.getDesc(), uploadedName, null, modFileWriter);
 
         Map<Band, FileReadInfo[]> retMap= new HashMap<>(1);
         retMap.put(Band.NO_BAND, new FileReadInfo[] {retval});
@@ -124,7 +127,7 @@ public class WebPlotReader {
 
         if (fd.isBlank())  {
             retval= new FileReadInfo[] { new FileReadInfo(null, PlotServUtils.createBlankFITS(req)[0],
-                                                          band, 0, fd.getDesc(), null, null) };
+                                                          band, 0, fd.getDesc(), null, null, null) };
         }
         else {
             File originalFile = fd.getFile();
@@ -136,7 +139,7 @@ public class WebPlotReader {
             FitsRead frAry[]= FitsCacher.readFits(originalFile);
             retval = new FileReadInfo[frAry.length];
 
-            if (needsPipeline(req)) { // if we need to use the pipline make sure we create a new array
+            if (needsPipeline(req)) { // if we need to use the pipeline make sure we create a new array
                 FitsRead newFrAry[]= new FitsRead[frAry.length];
                 int imageIdx;
                 for (int i = 0; (i < frAry.length); i++) {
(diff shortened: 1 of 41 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
```diff
@@ -401,5 +406,48 @@ private static double getImageSize(WebPlotRequest r) {
...
+                    imageIdx = i;
+                }
+            }
+            if (imageIdx>-1) {
+                List <RelatedData> relatedList= new ArrayList<>();
+                for(int i=0; (i<frAry.length); i++) {
+                    String extType= frAry[i].getExtType();
+                    if (extType!=null) {
+                        if (!extType.equalsIgnoreCase("IMAGE")) {
+                            if (frAry[i].getExtType().equalsIgnoreCase("MASK")) {
```
tgoldina: can you combine these two if statements into one: 
if (extType.equalsIgnoreCase("MASK")?

src/firefly/js/visualize/RelatedDataUtil.js
```diff
@@ -0,0 +1,190 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import {difference,get, flatten, values, uniq, isEmpty} from 'lodash';
```
tgoldina: isEmpty not used

src/firefly/js/visualize/RelatedDataUtil.js
```diff
@@ -0,0 +1,190 @@
...
+function dataTypeMatches(r,idx, pv) {
+    const dataType= get(r,'dataType');
+    if (!RDConst.SUPPORTED_DATATYPES.includes(dataType)) return false;
+
+    if (dataType===RDConst.IMAGE_MASK) {
+        const opvAry= pv.overlayPlotViews.filter((opv) => opv.relatedDataId===r.relatedDataId);
+        if (!opvAry.length) return true;
+        const maskNumberAry= values(r.availableMask);
+        if (maskNumberAry.length!==opvAry.length) return true;
+    }
```
tgoldina: Comments would be useful here. When true is returned?

src/firefly/js/visualize/task/ImageOverlayTask.js
```diff
@@ -2,22 +2,33 @@
...
 
 function *getColor()  {
-    const autoColor= ['#FF0000','#00FF00','#0000FF','#91D33D', '#AE14DB','#FF0000', '#EBAA38', '#F6E942'];
     var nextColor=0;
     while (true) {
-        yield autoColor[nextColor % autoColor.length];
+        yield colorList[nextColor % colorList.length];
         nextColor++;
     }
 }
```
tgoldina: Do you use this generator function? Looks like you are getting color based on the mask number.

src/firefly/js/visualize/task/PlotImageTask.js
```diff
@@ -398,3 +422,37 @@ function initBuildInDrawLayers() {
     DrawLayerCntlr.dispatchCreateDrawLayer(ActiveTarget.TYPE_ID);
 }
 
+/**
+ *
+ * @param {String[]} plotIdAry
+ * @param {Object.<string, OverlayPlotView>} oldOverlayPlotViews
```
tgoldina: Is it one OverlayPlotView or an array? enableMatchingRelatedData seems to expect an array.

src/firefly/js/visualize/task/PlotImageTask.js
```diff
@@ -398,3 +422,37 @@ function initBuildInDrawLayers() {
...
+    const vr= visRoot();
+    plotIdAry
+        .map( (plotId) => getPlotViewById(visRoot(), plotId))
+        .filter( (pv) => pv)
+        .forEach( (pv) => {
+            const opvMatchArray= uniqBy(flatten(getPlotViewIdListInGroup(vr, pv.plotId)
+                                                       .filter( (id) => id!== pv.plotId)
+                                                       .map( (id) => getPlotViewById(vr,id))
+                                                       .map( (gpv) => gpv.overlayPlotViews)),
+                                   'maskNumber' );
```
tgoldina: Looked up docs for uniqBy. Shouldn't the last arg be a function?

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/WebPlotReader.java
```diff
@@ -401,5 +406,48 @@ private static double getImageSize(WebPlotRequest r) {
...
+        if (frAry.length>1) {
+            int imageIdx= -1;
+            List<List<RelatedData>> retList= new ArrayList<>(frAry.length);
+            for(int i=0; (i<frAry.length); i++) {
+                retList.add(null);
+                String extType= frAry[i].getExtType();
+                if (extType!=null && extType.equalsIgnoreCase("IMAGE")) {
+                    imageIdx = i;
+                }
+            }
```
robyww: I will add some. This function could be there start of something far more advanced.  Right now the way I have it is that the last image with the header 'EXTTYPE' == 'IMAGE' is the based.  I reality that works out to the the first extension.

src/firefly/js/visualize/RelatedDataUtil.js
robyww: Ok, I also need to add a typedef for RelatedData

src/firefly/js/visualize/task/ImageOverlayTask.js
robyww: Not using it any more but I don't want to delete since I can see using it in the future

src/firefly/js/visualize/task/PlotImageTask.js
robyww: uniqBy can take a function or a string to use as the object key. In that case it will make sure the array of objects are unique by the key in each object. it is in the docs.
[http://devdocs.io/lodash~4/index#uniqBy](url)

tgoldina:
I have downloaded validation_data_hsc/DATA/rerun/20160805/00533/HSC-R/corr/CORR-0903332-100.fits

After image is shown, mask layers are shown as advertised. Zoom works on all 3 extensions simultaneously. Flip, Rotate, Rotate North do not work as expected. 

1. If the mask layer was shown during flip, it appears correctly when the image is flipped. Otherwise it would show unflipped.

After rotate the mask does not overlap with the image. They seem to be shown at different zoom levels. Zooming in then out brings them together. After that all mask layers show in the right place.

2. Another test: mask is enabled but not shown. Rotated Image, then mask, then variance first before showing the mask

Rotate works lazily on each image, which means it does not start until I go to the image. And I can not return to already rotated image, because it is covered by the mask.

I have rotated all 3 images (image, mask, extension) and tried bit 5. The mask does not overlap with the image. It appears to have different rotation angle.

After that I tried rotate North. Image and mask layer got rotated, but they appear to be at different zoom levels. (It seems to be tile generation issue, because zooming in and out does not resolve this problem.) 

3. The image is 85MB, but zooming in and out the rotated image with all mask layers, I quickly got out of memory on the server, even though I have
MIN_JVM_SIZE=1g
MAX_JVM_SIZE=4g

Also I noticed an exception in server log:

NFO: Server startup in 8787 ms
Warning: multiple occurrences of key:AR_NAME
Warning: multiple occurrences of key:AR_NAME
Warning: multiple occurrences of key:AR_NAME
Warning: multiple occurrences of key:AR_NAME
java.lang.ArrayIndexOutOfBoundsException: 1
        at edu.caltech.ipac.firefly.server.rpc.VisServerCommands$FileFluxCmdJson.doCommand(VisServerCommands.java:125)
        at edu.caltech.ipac.firefly.server.ServCommand.processRequest(ServCommand.java:27)
        at edu.caltech.ipac.firefly.server.servlets.CommandService.processRequest(CommandService.java:69)
 
Not sure if it's

ymeiymei:
I think I found a problem. After loading the image, either rotate or flip or north up, I can't add mask anymore. The error message is "Overlay failed: Could not find extension number 1. The file contains only one extension. Plot failed." And clicking "ok" will not resume that image.

robyww:
After looking at the issues with rotation and flipping with mask, I think that the system is to fragile to those actions.  I am going to disable rotation and flipping when mask are active.  We can pursuit fixing it in another ticket.

PR #278, opened 2017-01-30
DM 8916: Junit Test on IPACTableReader.

ymeiymei:
Please review the Junit test for IPACTableRead.

Need to use the branch, DM-8916-JUnitTestForIPACTableRead, on firefly_test_data.

 This test is mainly to test if IPACTableReader.java reads the IPAC table correctly. The IRSA IPAC table reader requirements are in
       http://irsa.ipac.caltech.edu/applications/DDGEN/Doc/ipac_tbl.html
 
 An IPAC table is composed by three parts: Attributes which includes keywords and comments; header which includes column names, data types, data units, and null values; and data rows.
 
The test cases below test if the reader can parse all parts in a table correctly and also if the reader can throw exceptions correctly when a table violate the requirements.

In this version, there are several known issues:
1) So far all the data comparison is done to compare strings (.toString() )  Will improve this.
2) Need to finalize the rules to validate an IPAC table. Or how strict we should accept/reject a table. Then we can update the test cases.

### DIFF
--- a/src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
+++ b/src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
@@ -1,0 +1,801 @@
+package edu.caltech.ipac.astro;
+
+
+import edu.caltech.ipac.firefly.util.FileLoader;
+import edu.caltech.ipac.util.DataGroup;
+import edu.caltech.ipac.util.DataObject;
+import edu.caltech.ipac.util.DataType;
+import edu.caltech.ipac.util.StringUtils;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.Reader;
+import java.util.*;
+
+/**
+ * Created by ymei on 1/26/17.
+ *
+ * DM-8916-JUnitTestIPACTableRead
+ *
+ * This test is mainly test if IPACTableReader.java reads the IPAC table correctly. The IRSA IPAC table reader requirements are in
+ *      http://irsa.ipac.caltech.edu/applications/DDGEN/Doc/ipac_tbl.html
+ *
+ * An IPAC table is composed by three parts: Attributes which includes keywords and comments; header which includes column names,
+ *      data types, data units, and null values; and data rows.
+ *
+ * The test cases below test if the reader can parse all parts in a table correctly and also if the reader can throw exceptions correctly
+ *      when a table violate the requirements.
+ *
+ */
+public class IpacTableReaderTest {
+
+    //Input parameters:
+    private static final String catName = "catName";
+    String[] onlyColumns = null;
+    private boolean useFloatsForDoubles = false;
+    private boolean isHeadersOnlyAllow = false;
+    private boolean noAttributes = false;
+    private int estFileSize = 0;
+
+    //Expected parameters in the data group:
+    private String[] attributeKeys = null;
+    private String[] attributeValues = null;
+    private String[] attributeComments = null;
+    private String[] colTitles = null;
+    private String[] colKeyNames = null;
+    private String[] colDataTypeDesc = null;
+    private String[] colDataTypes = null;
+    private String[] colDataUnits = null;
+    private String[] colNullStrings = null;
+    private String[][] dataValues = null;
+
+    //The IPAC table reader store the table content in the dataGroup:
+    private DataGroup dataGroup;
+
+    @Before
+    /**
+     *
+     */
+    public void setUp() throws IpacTableException, ClassNotFoundException, java.io.FileNotFoundException {
+        //Read different tables in each test case below.
+    }
+
+    @After
+    /**
+     * Release the memories
+     */
+    public void tearDown() {
+        dataGroup = null;
+    }
+
+
+    @Test
+    /**
+     * This test calls the methods IpacTableReader.readIpacTable to read all the columns from a valid input IPAC table, IPACTableSample1.tbl,
+     * and verify if the produced dataGroup contains the correct information from the table.
+     *
+     * The input parameters to call IpacTableReader.readIpacTable:
+     * File
+     * onlyColumns
+     * useFloatsForDoubles
+     * catName = "catName"
+     * isHeadersonlyAllow
+     */
+    public void testValidIpacTable1() throws IpacTableException, ClassNotFoundException, java.lang.NullPointerException {
+
+        //Read all the columns, other parameters are "false".
+
+        /* IPACTableSample1.tbl:
+
+        \catalog1 = 'Sample Catalog1'
+        \catalog2 = 'Sample Catalog2'
+        \ Comment1
+        \ Comment2
+        |   ra      |    dec    |   n_obs  |    V     |   SpType   |
+        |   double  |    double |   int    |   real   |   char     |
+        |   deg     |    deg    |          |   mag    |            |
+        |   null    |    null   |   null   |   null   |   null     |
+          165.466279  -34.704730      5       11.27       K6Ve
+          123.4       5.67            9       8.9         K6Ve-1
+
+        */
+
+        //Inputs:
+        String tableName = "IPACTableSample1.tbl";
+        File inFile = new File(FileLoader.getDataPath(IpacTableReaderTest.class) + tableName);
+        //onlyColumns = null;
+
+        //Set the expected values:
+        setExpectedValues_Input1_full();
+
+        //Read the table and generate the dataGroup:
+        dataGroup = IpacTableReader.readIpacTable(inFile, catName);
+
+        //Check the result:
+        checkResult(dataGroup);
+
+        //Test result: This IPAC table is read correctly.
+        // Passed.
+        // Known issue: The data type "real" is currently not supported. It is treated as a String. See the ticket https://jira.lsstcorp.org/browse/DM-9026
+
+    }
+
+    @Test
+    /**
+     * This test calls the methods IpacTableReader.readIpacTable to read all the columns from a valid input IPAC table, IPACTableSample1.tbl,
+     * and verify if the produced dataGroup contains the correct information from the table.
+     *         useFloatsForDoubles = true; isHeadersOnlyAllow = true;
+     *
+     */
+    public void testValidIpacTable2() throws IpacTableException, ClassNotFoundException, java.lang.NullPointerException
+
+    {
+
+        //Read all the columns but useFloatsForDoubles = true; isHeadersOnlyAllow = true;
+
+        /* IPACTableSample1.tbl:
+
+        \catalog1 = 'Sample Catalog1'
+        \catalog2 = 'Sample Catalog2'
+        \ Comment1
+        \ Comment2
+        |   ra      |    dec    |   n_obs  |    V     |   SpType   |
+        |   double  |    double |   int    |   real   |   char     |
+        |   deg     |    deg    |          |   mag    |            |
+        |   null    |    null   |   null   |   null   |   null     |
+          165.466279  -34.704730      5       11.27       K6Ve
+          123.4       5.67            9       8.9         K6Ve-1
+
+        */
+
+        //Inputs:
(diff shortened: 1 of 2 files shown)


src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
+package edu.caltech.ipac.astro;
+
+
+import edu.caltech.ipac.firefly.util.FileLoader;
+import edu.caltech.ipac.util.DataGroup;
+import edu.caltech.ipac.util.DataObject;
+import edu.caltech.ipac.util.DataType;
+import edu.caltech.ipac.util.StringUtils;
```
lznakano: remove import edu.caltech.ipac.util.StringUtils;

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+    private String[] attributeValues = null;
+    private String[] attributeComments = null;
+    private String[] colTitles = null;
+    private String[] colKeyNames = null;
+    private String[] colDataTypeDesc = null;
+    private String[] colDataTypes = null;
+    private String[] colDataUnits = null;
+    private String[] colNullStrings = null;
+    private String[][] dataValues = null;
+
```
lznakano: The best practice is to build an expected DataGroup in your setUp method, and then write it to a temporary file. Use this expected DataGroup to compare with the DataGroup read in from the file by IpacTableReader.

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+        |   null    |    null   |   null   |   null   |   null     |
+         */
+
+        //Read a table without data
+        String tableName = "IPACTable6_nodata.tbl";
+        File inFile = new File(FileLoader.getDataPath(IpacTableReaderTest.class) + tableName);
+
+        try{
+            dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName, isHeadersOnlyAllow);
+        } catch (IpacTableException e){
```
lznakano: This is the only Exception the IpacTableReader has and the one to be tested.  The correct way to test it is using
     try{
            dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName,  isHeadersOnlyAllow);
            Assert.fail("If no data error is not caught");
        } catch (IpacTableException e){
            System.out.println("IpacTableException from the table " + tableName + ": "  + e.getMessage());
        }

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+    //public static DataGroup readIpacTable(File f, String onlyColumns[], boolean useFloatsForDoubles, String catName)
+    //public static DataGroup readIpacTable(File f, String onlyColumns[], boolean useFloatsForDoubles, String catName, boolean isHeadersOnlyAllow)
+
+    //public static DataGroup readIpacTable(Reader fr, String catName)
+    //public static DataGroup readIpacTable(Reader fr, String catName, long estFileSize)
+    //public static DataGroup readIpacTable(Reader fr, String catName, String onlyColumns[],boolean useFloatsForDoubles,long estFileSize)
+    //public static DataGroup readIpacTable(Reader fr, String catName, String onlyColumns[],boolean useFloatsForDoubles,long estFileSize, boolean isHeadersOnlyAllow)
+
+
+}
```
lznakano: This unit test serves more like a verification of the algorithm and less like a unit test.  However since building a small DataGroup (table) is very easy,  the unit test can really be written as a unit test.  

For unit test:
1. There is a one to one correspondence for each public method in the main class.  
For example, there should be:
    testIpacTableReaderTwoArgs()
    testIpacTableReaderThreeArgs()
    testIpacTableReaderFourArgs()
    testIpacTableNoDataException()

2. Unit test does not check if the algorithm is correct, it checks if the algorithm does what it should do. 
    For example, no matter what table passed to the reader, if the reader reads it correctly, the  test is pass.

3. If there is duplicated columns, it is the issue in TableDef. If the TableDef allows to have duplicated columns, it is OK, if it does not allow duplication, TableDef should check it.

4.  The no header is also the issue with the TableDef.  It should catch such issue. 

Overall, I think this test can serve as a wonderful use cases verifications.  It correctly identify a few holes in the current algorithm such duplication in the column names, and empty headers etc.  GREAT Job!

To make it a self-contained unit test,
   1. build needed dataGroup dynamically.  The built dataGoup will be the expected DataGroup.  
   2. Let IpacTableReader reads the dataGroup from the output file by the the same dataGroup and do the comparison.
  3. In the tearDown method:
   remove the temporary output files used for the test. 

Then the unit test is self-contain, neat and clean.

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+        */
+
+        //Inputs:
+        String tableName = "IPACTableSample1.tbl";
+        File inFile = new File(FileLoader.getDataPath(IpacTableReaderTest.class) + tableName);
+        //onlyColumns = null;
+
+        //Set the expected values:
+        setExpectedValues_Input1_full();
+
```
loitly: This @Test gives the impression that it's simple and small based on the amount of code in it.  However, it is not the case.  setExpectedValues_Input1_full() has key information that require investigation to understand.  It also rely on the input, therefore the comment is repeated again in this function.  Now, you need to keep in sync the 2 comments and the input file.

checkResult() is a share function which is clever but is relying on setExpectedValues_Input1_full() to fully set up its expected values.  The reuse is good, but the dependencies may lead to bugs.

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+        setExpectedValues_Input1_radecOnly();
+
+        //Read the table and generate the dataGroup:
+        dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName, isHeadersOnlyAllow);
+
+        //Check the result:
+        checkResult(dataGroup);
+
+        //Test result: "onlyColumns" works fine. Only those two columns are read.
+        // Passed.
```
loitly: I will use this test case to post an example of my prefer method:
For me, the example below is simple and easy to read.  The input and the expected results for this test case are all in one place. 
```
        String input =
        "\\catalog1 = 'Sample Catalog1'\n" +
        "\\catalog2 = 'Sample Catalog2'\n" +
        "\\ Comment1\n" +
        "\\ Comment2\n" +
        "|   ra      |    dec    |   n_obs  |    V     |   SpType   |\n" +
        "|   double  |    double |   int    |   real   |   char     |\n" +
        "|   deg     |    deg    |          |   mag    |            |\n" +
        "|   null    |    null   |   null   |   null   |   null     |\n" +
        " 165.466279  -34.704730      5       11.27       K6Ve       \n" +
        " 123.4       5.67            9       8.9         K6Ve-1     \n";

        DataGroup dg = IpacTableReader.readIpacTable(new StringReader(input), "onlyColumns",  new String[]{"ra", "dec"}, false, 0);

        Assert.assertEquals("There should be 2 columns", 2, dg.getDataDefinitions().length);
        Assert.assertEquals("There should be 2 rows", 2, dg.size());
        Assert.assertEquals("ra for row 1", 165.466279, (Double)dg.get(0).getDataElement("ra"), 0.001);
        Assert.assertEquals("dec for row 2", 5.67, (Double)dg.get(1).getDataElement("dec"), 0.001);
```

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -0,0 +1,801 @@
...
+ *      http://irsa.ipac.caltech.edu/applications/DDGEN/Doc/ipac_tbl.html
+ *
+ * An IPAC table is composed by three parts: Attributes which includes keywords and comments; header which includes column names,
+ *      data types, data units, and null values; and data rows.
+ *
+ * The test cases below test if the reader can parse all parts in a table correctly and also if the reader can throw exceptions correctly
+ *      when a table violate the requirements.
+ *
+ */
+public class IpacTableReaderTest {
```
loitly: Test should extends from ConfigTest based on @ejoliet's implementation to take advantage of logging and future additional features.  System.out should then should be replaced with logger.

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
loitly: I agree Assert.fail() should be added.  However, we said that it should be commented and replaced with a message to indicating it needs to be fixed with referenced ticket number.  Also, I don't think printout in catch clause is needed.

ymeiymei:
I have improved the IPAC table reader test unit based on reviewers' suggestions:
 
    1)Cleaned import
    2)Put the input table and expected values together in each test case
    3)Improved the exception handling
    4)Added JIRA tickets and the code after the tickets (commented out)
    5)Let the class extend ConfigTest
    6)Use logging system in ConfigTest. No System.out.println

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -54,243 +50,218 @@
...
-        onlyColumns = null;
-        setExpectedValues_Input2();
-        dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName, isHeadersOnlyAllow);
-        checkResult(dataGroup);
+            Assert.assertEquals("The last two lines will be ignored as they have no leading space nor '=' ", 1, attributes.size());
 
-        //Test result: It is okay for an IPAC table not having any attributes. The table is read correctly.
-        // Passed.
-        // Known issues: The trailing zeros to the right of the decimal point is truncated, eg. 1309.0000 to 1309.0.
+        }catch(IpacTableException e){
```
loitly: Same, fail test if encounter exceptions.

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -302,423 +273,234 @@ public void testWrongAttributes1() throws IpacTableException, ClassNotFoundExcep
...
+            "  165.466279  -34.704730      5       11.27       K6Ve      \n" +
+            "  123.4       5.67            9       8.9         K6Ve-1    ";
+
+
 
         try{
-            dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName, isHeadersOnlyAllow);
+            DataGroup dataGroup = IpacTableReader.readIpacTable(new StringReader(input), catName);
+            // After DM-9332 is implemented uncomment this:
+            //Assert.fail("No exception thrown out");
```
loitly: I thought we are supposed to print a warning message for the time being

src/firefly/test/edu/caltech/ipac/astro/IpacTableReaderTest.java
```diff
@@ -302,423 +273,234 @@ public void testWrongAttributes1() throws IpacTableException, ClassNotFoundExcep
...
-        //setExpectedValues_Input7_full();
 
         try{
-            dataGroup = IpacTableReader.readIpacTable(inFile, onlyColumns, useFloatsForDoubles, catName, isHeadersOnlyAllow);
-        } catch (IpacTableException e){
-            System.out.println("IpacTableException from the table " + tableName + ": "  + e.getMessage());
-        }
+            DataGroup dataGroup = IpacTableReader.readIpacTable(new StringReader(input), catName);
+            //After https://jira.lsstcorp.org/browse/DM-9333 is implemented, use this:
+            //Assert.fail("No exception is thrown out");
```
loitly: add warning message for the time being

PR #279, opened 2017-02-01
DM-9206 - XY Plot: set precision for expression columns

tgoldina:
This is a fix for a bug described in https://jira.lsstcorp.org/browse/DM-9206

Test procedure and test table are in the ticket.

Set precision for expression columns to 14 seg digits. 
Tooltip will include 4 sig digits in the difference between min and max of the plot.


ejoliet:
Tatiana, i can see the problem of the line to be solved but the tooltip isn't showing the same value as in the table.
Example: -count/time=0.465, time=781.765648, tooltip shows: 782, 0.465.

The rounding is on purpose in the tooltip?

tgoldina:
You are right. It's not intended, needs to be fixed.

2017-02-01 18:33 GMT-08:00 Emmanuel Joliet <notifications@github.com>:

tgoldina:
The latest commit fixes the bug: non-expression columns should match the table (except trailing zeros). Also fixed a bug in calculation of min values for negative numbers.

PR #281, opened 2017-02-02
DM-8670 change LC result area as new UI design for time series viewer and update period/periodogram area based on the review feedback.

cwang2016:
The development includes: 
- revamp result area including finder panel, chart, table and image areas
     . finder panel includes entries which are mission related and non-related. 
     . synchronize among table highlight, image selection, chart highlight and entries changes. 
     . add 'Period finding' button in the finder panel to move to period area to find period and the phase folding table. 
     . Reorder the phase folding table by moving phase related columns to the left side of the table. 
     . add indication regarding table sort columns and sort order. 
- update period area according to the feedback to the flow and the layout of DM-8668 
    . show the time and flux field as read only data (as set in result area). 
    . put period min and max entries to the end of the period slider. 
    . highlight the period field. 
    . change the layout of phase folding chart, periodogram tables and charts. 
    . put 'revert to' button by 'period' field in the finder panel. 
    . put 'change periodogram' button into the priodogram table/chart area. 
    . update the validation functions for some fields in the finder panel, 
        - default:  period min = 0.001, and  period max = 365. 
        - period min >= 0.001, period max > period min, period >= period min and period <= period max
        - extend period max or period min in case period is set beyond the min and max range. 
        - adjust period in case period min is set greater than period or  period max is set less than period. 
        - zero point time >= 0.0. 
    . update the slider to work for any period range scale. 
    . make the look of phase folding chart consistent with  those of periodogram table and charts. 
    . no slider and period related items changes until the periodogram popup is closed.    
- update Periodogram popup  
     . update field validation which allows period min, period max and step size fields to be blank

### DIFF
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -58,7 +58,7 @@ export function makeTblRequest(id, title, params={}, options={}) {
  */
 export function makeFileRequest(title, source, alt_source, options={}) {
     const id = 'IpacTableFromSource';
-    var req = {startIdx: 0, pageSize: 100000};
+    var req = {startIdx: 0, pageSize: 100};
     title = title || source;
     const tbl_id = options.tbl_id || uniqueTblId();
     options = omit(options, 'tbl_id');
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -58,7 +58,7 @@ export function makeTblRequest(id, title, params={}, options={}) {
  */
 export function makeFileRequest(title, source, alt_source, options={}) {
     const id = 'IpacTableFromSource';
-    var req = {startIdx: 0, pageSize: 100};
+    var req = {startIdx: 0, pageSize: 100000};
     title = title || source;
     const tbl_id = options.tbl_id || uniqueTblId();
     options = omit(options, 'tbl_id');
--- a/src/firefly/js/templates/lightcurve/LcPhaseTable.js
+++ b/src/firefly/js/templates/lightcurve/LcPhaseTable.js
@@ -2,13 +2,12 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {cloneDeep, get, set, omit, slice, replace, isArray, isString} from 'lodash';
+import {get, set, omit, slice, replace, isArray, isString} from 'lodash';
 import {doUpload} from '../../ui/FileUpload.jsx';
 import {loadXYPlot} from '../../charts/dataTypes/XYColsCDT.js';
 import {sortInfoString} from '../../tables/SortInfo.js';
 import {dispatchTableSearch} from '../../tables/TablesCntlr.js';
-import {makeTblRequest,getTblById, getTblIdsByGroup, tableToText, makeFileRequest} from '../../tables/TableUtil.js';
-import {dispatchTableReplace} from '../../tables/TablesCntlr.js';
+import {makeTblRequest,getTblById, tableToText, makeFileRequest, getColumnIdx} from '../../tables/TableUtil.js';
 import {LC, getValidValueFrom} from './LcManager.js';
 import {getLayouInfo} from '../../core/LayoutCntlr.js';
 import FieldGroupUtils from '../../fieldGroup/FieldGroupUtils';
@@ -28,7 +27,11 @@ export function uploadPhaseTable(tbl, flux) {
     const file = new File([new Blob([ipacTable])], `${tbl_id}.tbl`);
 
     doUpload(file).then(({status, cacheKey}) => {
-        const tReq = makeFileRequest(title, cacheKey, null, {tbl_id, sortInfo:sortInfoString(LC.PHASE_CNAME)});
+        const tReq = makeFileRequest(title, cacheKey, null,
+                                     {tbl_id, sortInfo:sortInfoString(LC.PHASE_CNAME),
+                                      pageSize: LC.FULL_TABLE_SIZE
+                                     });
+
         dispatchTableSearch(tReq, {removable: true});
 
         const xyPlotParams = {
@@ -84,12 +87,11 @@ export function getPhase(time, timeZero, period,  dec=DEC_PHASE) {
  * @returns {TableModel}
  */
 function addPhaseToTable(tbl, timeName, fluxName, tzero, period) {
-    var {columns} = tbl.tableData;
-    var tIdx = timeName ? columns.findIndex((c) => (c.name === timeName)) : -1;
+    var tIdx = timeName ? getColumnIdx(tbl, timeName) : -1;
 
     if (tIdx < 0) return null;
 
-    var tPF = Object.assign(cloneDeep(tbl), {tbl_id: LC.PHASE_FOLDED, title: 'Phase Folded'},
+    var tPF = Object.assign({}, tbl, {tbl_id: LC.PHASE_FOLDED, title: 'Phase Folded'},
                                             {request: getPhaseFoldingRequest(period, timeName, fluxName, tbl)},
                                             {highlightedRow: 0});
     tPF = omit(tPF, ['hlRowIdx', 'isFetching']);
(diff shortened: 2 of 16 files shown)


cwang2016:
test:
-http://localhost:8080/firefly/lc.html
- upload the test sample lc_raw.tbl => serach
- in result page: table, chart and image are shown on the raw table, you may
     - update flux, cutout size to see the change of the images and the chart. 
     - change the highlight of  table row, chart, and image to see if all parts are synchronized. 
- click 'Period finding' to move to period finding page. 
- click 'find periodogrm' to get periodogram popup dialog and 'periodogram calculation' button will exit the dialog and produce periodogram table peak table based on the setting on the dialog. 
- update the zero point time, period min, period max, period slider, or period fields to the change of the phase folding chart. 
- 'accept period' button will exit period finding page and render the phase folding table/chart/image in the result page.

ymeiymei:
Great work!

A problem: When selecting a period in the slider panel, the periodogram (the left lower table) and the period-power chart are not responding properly.  For example, period = 180.

ejoliet:
Wow!! So far i must say the change is awesome and i think is taking into account all the feedback we received. Great job.
The periodogram form is fine for now. In a different branch, i have some couple of changes to include.
One question: what happen with the {{error column}} next to the button {{Period Finding}}, is it suppose to do something or is it just a dummy field currently for later more work to add it from some other ticket/branch pending?

Apart from that, all works as expected and is ready to merge in my opinion.

tgoldina:
The flow looks very nice to me! 

Just a few things I noticed when "Accept Period" button is clicked.

- After first Accept Period, a warning appears:
Failed to fetch XY plot data: TypeError: Cannot read property 'tblId' of undefined
(Not sure if it's related: I see that the XY plot data for "phase_folded" are fetched twice, one time produces this error.)

- If I do Find Period again, I get a bunch of Warnings "setState(...): You passed an undefined or null state object; instead, use forceUpdate()" LcImageToolbar.jsx:46

- Also for each image, I get warning:
Warning: setState(...): Can only update a mounted or mounting component... Please check the code for the ImageViewerView component.

ejoliet:
@loitly the browser differences is a good catch! Thanks for that!

ejoliet:
Some weirdness found in the flux column selector: when i'm changing the flux column and start typing `w`, the drop-down list is hidden by the panel below, see picture below. If it's easy to fix fine, otherwise don't make it a high priority right now i would say (there is a workaround).

screen shot 2017-02-02 at 6 10 57 pm

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -0,0 +1,232 @@
...
+ */
+function addPhaseToTable(tbl, timeName, fluxName, tzero, period) {
+    var {columns} = tbl.tableData;
+    var tIdx = timeName ? columns.findIndex((c) => (c.name === timeName)) : -1;
+
+    if (tIdx < 0) return null;
+
+    var tPF = Object.assign(cloneDeep(tbl), {tbl_id: LC.PHASE_FOLDED, title: 'Phase Folded'},
+                                            {request: getPhaseFoldingRequest(period, timeName, fluxName, tbl)},
+                                            {highlightedRow: 0});
```
robyww: is cloneDeep really necessary here?  Seems very heavy if the table gets big.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -0,0 +1,232 @@
...
+    return tPF;
+}
+
+
+/**
+ * @summary change the content of each row of the table by relocating some columns
+ *          to be after or replace certain column
+ * @param {TableModel} tbl
+ * @param {array|string} srcCols
+ * @param {number|string} targetCol
```
robyww: what does a number mean here? Maybe another line of documentation on this parameter? Same with strCols.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * @summary duplicate the phase cycle to the phase folded table
+ * @param {string} timeCol
+ * @param {number} period
+ * @param {TableModel} phaseTable
+ */
+function repeatDataCycle(timeCol, period, phaseTable) {
+    var tIdx = phaseTable.tableData.columns.findIndex((c) => (c.name === timeCol));
+    var fIdx = phaseTable.tableData.columns.findIndex((c) => (c.name === LC.PHASE_CNAME));
+    var {totalRows, tbl_id, title} = phaseTable;
```
robyww: Code might be a little cleaner if you also pull `tableData` our of `phaseTable`  and moved this  line to the beginning of the function so that did not have to keep repeating `phaseTable.tableData`

```
var {totalRows, tbl_id, title, tableData} = phaseTable;
```

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -2,56 +2,97 @@
...
+const cTimeSeriesKeyDef = {
+    time: {fkey: LC.META_TIME_CNAME, label: 'Time Column'},
+    flux: {fkey: LC.META_FLUX_CNAME, label: 'Flux Column'},
+    timecols: {fkey: LC.META_TIME_NAMES, label: ''},
+    fluxcols: {fkey: LC.META_FLUX_NAMES, label: ''},
+    cutoutsize: {fkey: 'cutoutSize', label: 'Cutout Size (deg)'},
+    errorcolumn: {fkey: 'errorColumn', label: 'Error Column'}
+};
+
+const defValues = {
```
robyww: There is a lot of indirection here. Maybe a couple of lines of documentation would help.

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -2,56 +2,97 @@
...
+    [cTimeSeriesKeyDef.time.fkey]: Object.assign(getTypeData(cTimeSeriesKeyDef.time.fkey, '',
+                                                'time column name',
+                                                `${cTimeSeriesKeyDef.time.label}:`, labelWidth),
+                                                {validator: null}),
```
robyww: The value of this object look similar.  Could that be a in function to make the code a little more readable?

src/firefly/js/tables/TableUtil.js
```diff
@@ -58,7 +58,7 @@ export function makeTblRequest(id, title, params={}, options={}) {
  */
 export function makeFileRequest(title, source, alt_source, options={}) {
     const id = 'IpacTableFromSource';
-    var req = {startIdx: 0, pageSize: 100};
+    var req = {startIdx: 0, pageSize: 100000};
```
robyww: This is an important issue.  It could affect performance in several places.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
loitly: Should use TableUtil.getColumnIdx(tableModel, colName) instead.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
loitly: I think you do.  The resulting table will have more columns and twice as many rows as the original.

robyww:
## UI Notes
 
- `Period Finding` button does not stand out enough.  I think you need make i justified to the right. and vertically centered (using flexbox layout, `align-self: center`).
- _Note_: Error column input will be mission specific.
- Slider is much better
- I agree with loi, no horizontal resizer is necessary between table and form on the main display.
-  _Optional_: You should use a disclosure panel that defaults to open for the form. It might be necessary in the future if we have a lot of mission specific parameters.
- Overall I think it looks really, really good.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
cwang2016: just thought this parameter could be a column name or a column index meaning a table column that srcCols are relocated to.

cwang2016:
fixed the bugs and the features based on the reviews. 
one issue is left unchanged, 
- the periodogram table highlight is not synchronized with the change of period field -- will be updated in the future if needed. 
one issue found, 
- phase table upload by using Blob on ipac table and File doesn't work for Safari. Some alternative method will be further investigated for uploading file from Safari in the future.

PR #282, opened 2017-02-02
DM-7780:JUnit Test for image flip (FlipXY)

lznakano:
1.  Add unit test for FlipXY class
 2.  Add a utility class to generate a FITs file for unit test
 3.  Modified the FitsValidation.class to test each values when the HeaderCard contains numberical values

### DIFF
--- a/src/firefly/test/edu/caltech/ipac/firefly/util/FitsGenerator.java
+++ b/src/firefly/test/edu/caltech/ipac/firefly/util/FitsGenerator.java
@@ -1,0 +1,218 @@
+package edu.caltech.ipac.firefly.util;
+
+import nom.tam.fits.*;
+
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


ymeiymei:
Good work!  Good coverage.  Can be merged.

ymeiymei:
Thanks for adding testDoFlipXWithSmallData and testDoFlipYWithSmallData. Those tests really test the flipped image data.

Still don't think the tests, validateFlipX and validateFlipY, are good. They validate the double flipped image against the original image. What if FlipXY.java never does any flip? The double flipped image will match the input image for sure but it doesn't mean the image is flipped correctly.

ymeiymei:
Thanks for adding endToEndTest which is a very good test case (enough data for histogram). I think you can merge.

PR #283, opened 2017-02-03
DM-9196: fix cutout size parameter is not passed to server

loitly:
https://jira.lsstcorp.org/browse/DM-9196

Use lc.html for testing.  No visible changes.  
You will be able to verify the passed parameter if you set a breakpoint in the debugger in LightCurveFileGroupsProcessor.java


ejoliet:
I debugged it and see the attribute in the server side request. Thanks.
Looks good!

PR #284, opened 2017-02-04
DM-8309: Enable histogram in IRSAViewer

loitly:
https://jira.lsstcorp.org/browse/DM-8309

Added fixedAlgorithm prop to HistogramOptions.  Set it to true for IRSA.

to test:
- checkout branch in firefly AND ife:
- build irsaviewer in ife
- any catalog search, then clicks on 'Charts' top menu.
- under Histogram, there should only be one option for Algorithm.


ejoliet:
Looks good. Can be merged.
Thanks!

PR #285, opened 2017-02-07
Dm 7195 lc download images

ejoliet:
I've completed and adapted the server and client side for downloading WISE images (cutout or original) single exposure (product level 1b) from Time Series viewer. 

Selected images are downloaded in a zip file previous packaging in the server.

Downloading selected row from raw table should work the same way as from the phase folded table.

To test, upload lc raw table and phase fold, select rows and check that the images have a name that match the row frameid and cutout size, size that match the cutout IBE call as well as original images.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcResult.jsx
+++ b/src/firefly/js/templates/lightcurve/LcResult.jsx
@@ -159,6 +159,8 @@ const buttonW = 650;
 // eslint-disable-next-line
 const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables, settingBox}) => {
 
+    let {cutoutSize} = settingBox.props.generalEntries || '0.3';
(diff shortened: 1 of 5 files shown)


cwang2016:
Review completes. Downloading is tested on all selection cases and works good!

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -159,6 +159,8 @@ const buttonW = 650;
 // eslint-disable-next-line
 const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables, settingBox}) => {
 
+    let {cutoutSize} = settingBox.props.generalEntries || '0.3';
+    //let csize = get(generalEntries, 'cutoutsize, '0.3');
```
loitly: the commented-out code makes more sense to me.  generalEntries should be passed into StandardView so it can make use of cutoutSize.

PR #291, opened 2017-02-08
DM-9326 clean up time series viewer 

cwang2016:
update: 
- change filename from lc.html to ts.html
- mission name is capitalized in previous ticket. (DM-7195). 

test: 
rebuild server
localhost:8080/firefly/ts.html


ejoliet:
Looks good. Thanks. This can be merged and the change of the HTML name from `lc.html` to `ts.html` should be notified to @Caltech-IPAC/firefly developers.

PR #292, opened 2017-02-08
DM-9248 Implement 'All Sky' mode in LSST catalog and image search catalogs

cwang2016:
Development includes
[CW DM-9248]
- the 'all sky' feature was built in the LSST spatial search component. 
- fixed some code syntax errors issued by IntelliJ. 
- fixed the setting of some spatial search parameters from float to string format prior to building the table search request. 

[LZ DM-9247]
- Add the AllSky search in Catalog search
- Add the allSky search in Image Catalog search
- Add the guard to prevent from allSky search without constraints.

test for 'AllSky': 
- localhost:8080/firefly/lsst-pdac-triview.html
- In catalog search, select "allSky"
Enter the constrains at the field:
For deep source at the id field enter
=3219370314566605
For forced deep source at the id field enter:
=225486482582800165

- in image search
for Deep coadd: 
deepCoaddId =1610612744
for ccd 
scienceCcdExposureId = 2649410121

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCataLogSearch.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCataLogSearch.java
@@ -12,22 +12,33 @@ import edu.caltech.ipac.util.DataType;
 import edu.caltech.ipac.util.StringUtils;
 import edu.caltech.ipac.visualize.plot.CoordinateSys;
 import edu.caltech.ipac.visualize.plot.WorldPt;
-
 import java.util.List;
 import java.util.Locale;
 
 
 /**
+ * Feb-9-2017
+ * DM-9247
+ *   Implement search processor(s) to support All Sky mode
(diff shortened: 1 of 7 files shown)


ymeiymei:
The catalog search works fine. 

Please add "AllSky" for image search.

cwang2016:
'ALLSKY' option is added for image search. 'ALLSKY' option is also needed to be added at the server side.

ymeiymei:
Reviewed. Looks great!

One comment. When I give only one coaddId or scienceCcdExposureId, I get one row of data in the table, but I get 5 images from all 5 filters. This the feature for cone search (when no specific filter is given). I need to double check with Trey and Gregory see if we should keep this when filters are specified.

lznakano:
Yi,

Cindy and I just went to discuss with Trey.  Trey said it is fine to show all related filters. 
To prevent this from happening, UI has to hard codes several columns such as filterId, filterName and coaddId etc.  

FYI

PR #294, opened 2017-02-09
DM-9141: cleanup periodogram form

ejoliet:
Please, review the periodogram form and let me know if you find any problem.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
+++ b/src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
@@ -69,11 +69,11 @@ const labelWidth = 150;
 
 var defValues = {
     [pKeyDef.min.fkey]: Object.assign(getTypeData(pKeyDef.min.fkey,
-        '', 'minimum period',
+        '', 'minimum period (> 0)',
         `${pKeyDef.min.label}:`, labelWidth),
         {validator: null}),
     [pKeyDef.max.fkey]: Object.assign(getTypeData(pKeyDef.max.fkey,
-        '', 'maximum period',
+        '', 'maximum period (> 0)',
         `${pKeyDef.max.label}:`, labelWidth),
         {validator: null}),
     [pKeyDef.algor.fkey]: Object.assign(getTypeData(pKeyDef.algor.fkey,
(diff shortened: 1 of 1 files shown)


src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
```diff
@@ -424,6 +430,26 @@ function stepsizeValidator(description) {
...
+        var retval = Validate.isInt(description, valStr);
+
+        if (!retval.valid) return retval;
+
+        var val = parseInt(valStr);
+        const min = 0;
+        var bVal = val > min;
+
+        return bVal ? {valid: true} :
+        {valid: false, message: description + `: must be greater than ${min}`};
```
tgoldina: So the number of peaks can be as whatever big? I tried 1000000000, it worked, I tried 10000000000 - it failed.

src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
```diff
@@ -256,7 +257,7 @@ class PeriodogramOptionsBox extends Component {
                         </button>
                         <br/>
                         <div style={{marginTop: 10}}>
-                            {'Fields of "Period Min", "Period Max" and "Step Size" are allowed to be blank.'}
+                            {'Leave the fields blank to use default values.'}  <HelpIcon helpId={'periodogram.options'}/>
```
tgoldina: Currently, the help button appears right under the help text. I think, it would be better if help button is with other buttons. In fact, FormPanel.jsx has a support for help button when you pass help_id property.

Also, there is too much white space under help text. It might be better if the help text was just above the lower form border.

ejoliet: This is not a FormPanel component. Plus, I will get more review on those things next week and will do a ticket for all anyway. Prefer to differ the change next week.

src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
ejoliet: API is not ready, default values are not communicated. For now, is open to the user to know what they enter. Once we get better understanding of the API, we will add validation to input fields.

src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
ejoliet: Will fix the validator message.

ejoliet:
@tgoldina i've updated couple of minor misspelled message and fix the validation (min period should be greater than 0). let me know.
Thank you!

src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
tgoldina: I meant, it should not hard to convert it to FormPanel when we have another chance. :)

PR #296, opened 2017-02-10
DM-9340: raise size limit to 100 and use radio buttons instead of checkboxes

ejoliet:
The ticket request was to raise the shape size limit to 100 pixels and change checkboxes to radio buttons for the choice selection.

By raising the size, i had to take out the current canvas. If you find a better way to adjust the window with the canvas, let me know, i couldn't.

Fixing the window width or any other style/css approach is not worth the effort because it could change again if the max size is not yet enough for future cases.

Maybe @robyww or @Caltech-IPAC/firefly developers might have a better solution. Come to me to discuss it if you want or feel free to commit directly to my branch. 

Thanks.


robyww:
Instead of removing the symbol resize, you so it at some max and the give a text indication that is is bigger. So it will resize when it is under 30 and just the text indication will change when it is over 30.
30plus

PR #297, opened 2017-02-10
DM-9349: filtering by ROWID is off by 1

loitly:
https://jira.lsstcorp.org/browse/DM-9349

In Firfefly (+IRSAVIewer) development version, when the user select table rows and filter them out, the result table is not matching the selected rows before applying the filter.
OPS is fine though.

Steps to reproduce:
do a catalog search i.e., 2mass, 100". From the result table, select a row and apply filter. Then check that the resulting row is not the one you selected.

TG 02/10/2017 The row returned is actually the previous row. 
The related issue is filtering a point from an image. A previous row is returned as a result.


ejoliet:
Looks good. Fixed.
You can merge.

PR #300, opened 2017-02-13
DM-7971: Sync highlighted row for raw and phase-folded tables.

loitly:
https://jira.lsstcorp.org/browse/DM-7971

also:
- fix missing rows.
- fix phase folded table sort issue

To test:
- Using ts.html, upload a raw table.
- add a phase-folded table.
- the highlighted row of the two tables should be in synced.
Note:  when going from raw to phase-folded, it will highlight the first occurrence.


cwang2016:
Review completes. Great to have those issues fixed.

PR #301, opened 2017-02-13
DM-7946: Unit test for CentralPoint

lznakano:
Add unit test for CentralPoint class
 Change posdis to private since it is only used inside the class

While I was working on this test, I found that the CentralPoint class did not use the Elzinga-Hearn Algorithm as described at the beginning of the class. There are no further document about the
algorithm.


ymeiymei:
Reviewed. Good job. Good coverage.

PR #303, opened 2017-02-15
DM-9425: several python interface issues

robyww:
- cleaned up ExternalAccessCntlr
 - fixed point selection improperly turning on and off
 - fixed extension added.  It will now update if the extension id already exist
 - extension response data sent to server now contains selected table row information
    if a table is associated with an image.
 - new plots are active by default, it was not defaulting correctly with remote actions
 - improve image meta data table uploads a little, made DataSource work correctly

Notes:
- this is really going to be hard to test until we get it on  the pdac.  I would sugest that the code review just look over the code the David S. and I will test it with his python code.
- ignore the python notebook file in the review

### DIFF
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -654,22 +654,23 @@ export function getTblInfo(tableModel, aPageSize) {
 
 
 /**
- *
+ * Return the row data as an object keyed by the column name
  * @param {TableModel} tableModel
  * @param {Number} [rowIdx] = the index of the row to return, default to highlighted row
+ * @return {Object<String,String>} the values of the row keyed by the column name
  */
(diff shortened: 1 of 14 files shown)


src/firefly/js/tables/TableUtil.js
```diff
@@ -652,6 +652,28 @@ export function getTblInfo(tableModel, aPageSize) {
...
+
+/**
+ *
+ * @param {TableModel} tableModel
+ * @param {Number} [rowIdx] = the index of the row to return, default to highlighted row
+ */
+export function getTblRowAsObj(tableModel, rowIdx= undefined) {
+    if (!tableModel) return {};
+    if (isUndefined(rowIdx)) rowIdx= tableModel.highlightedRow;
+    if (rowIdx<0) return {};
```
loitly: Since you're checking, you should check for >= data.lenght as well.

PR #308, opened 2017-02-17
DM-9430 Fix Compass Layout 

cwang2016:
the development includes
- Locate the compass at the left corner of the image when the image is scrolled, zoomed, re-centered, rotated, flipped, etc. 

test: 
- find a image by using 'Images' panel.
- click 'compass' button from the image tool bar
- apply image processing related buttons and see how the compass is rendered.


ymeiymei:
Tested with both LSST and WISE images. After the compasses are added to the images,  zooming, moving, rotating, flipping....the images,  the compasses keep staying in the images. Good work!

PR #310, opened 2017-02-18
DM-9272: Clean up IRSA Time series viewer to prepare RC

loitly:
https://jira.lsstcorp.org/browse/DM-9272

- fix SplitPane resizes beyond browser's width/height
- fix Periodogram expanded mode sizing issue.  Content is larger the window's size causing scrollbars to appear
- removing multiple redundant SplitPanels
- remove 'SplitPanel' based styling when components are not resizable.

To test:
http://localhost:8080/firefly/ts.html

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -86,6 +86,7 @@
 
 .button.std {
     height: 23px;
+    white-space: nowrap;
 }
 
 .button.hl {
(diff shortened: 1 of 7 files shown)


ejoliet:
Looks nicer, definitively an improvement. There is another ticket DM-9469 mentioning a need to have a title, just in case you are able to add it as part of the improvements.

Otherwise that's good to be merged. Thanks!

cwang2016:
Review completes. Good to have the fixing and cleaning.

loitly:
@ejoliet, in regard to DM-9469, where should the title be?  It says 'first page'.  Is that the results page where the mission and time column fields are?

PR #312, opened 2017-02-22
DM-9543: calling showXYPlot API without xcol/ycol name fails

loitly:
Bug fixed in rc.  Please review.


ejoliet:
Works as expected. Thanks for quick check and fix!

PR #313, opened 2017-02-22
DM-9348 LSST - time series viewer results (part 2)

tgoldina:
The related ticket: [https://jira.lsstcorp.org/browse/DM-9348](https://jira.lsstcorp.org/browse/DM-9348)
Sample raw table for LSST_SDSS is attached to the ticket.

Refactored the code to support multiple missions (currently WISE and LSST SDSS)
- LcConverterFactory.js defines properties storing mission specific information
- It is assumed that each raw table contains time and at least one periodic column (we call it flux, but it can be anything)
- Each mission can have different UI to define those columns (upper part of result view)
- For LSST SDSS, the uploaded table contains all the bands, raw table is always filtered: band selection triggers table filter. It would be nice to have hidden filter for this case, which user can not remove. Currently, user can remove the filter, and band selection will not match the content of the table.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcConverterFactory.js
+++ b/src/firefly/js/templates/lightcurve/LcConverterFactory.js
@@ -2,11 +2,14 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
+import {get} from 'lodash';
 import {TitleOptions} from '../../visualize/WebPlotRequest.js';
 import {logError} from '../../util/WebUtil.js';
 
+import {DefaultSettingBox, defaultOnNewRawTable, defaultOnFieldUpdate, defaultRawTableRequest} from './DefaultMissionOptions.js';
 import {getWebPlotRequestViaWISEIbe} from './wise/WisePlotRequests.js';  //WiseRequestList.js
 import {makeLsstSdssPlotRequest} from './lsst_sdss/LsstSdssPlotRequests.js'; //LsstSdssRequestList.js';
+import {LsstSdssSettingBox, lsstSdssOnNewRawTable, lsstSdssOnFieldUpdate, lsstSdssRawTableRequest} from './lsst_sdss/LsstSdssMissionOptions.js';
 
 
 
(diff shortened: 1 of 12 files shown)


src/firefly/js/templates/lightcurve/LcConverterFactory.js
```diff
@@ -19,36 +22,77 @@ import {makeLsstSdssPlotRequest} from './lsst_sdss/LsstSdssPlotRequests.js'; //L
  * @param {Object} params - mission specific parameters
  */
 
+/**
```
loitly: The documentation below is very useful.  I am so glad you put the effort into adding it.

loitly:
I agree with the filter remove problem.  As a workaround, I suggest adding an 'all' option into band so that when one removes the filter, the application will still function properly.  At the moment, when that happens, the results are bad.

Another really minor thing.  In Safari, the line LSST Band and Periodic Column is not aligned.  It's off by s couple of pixels.

This is a great pull request.  Feel free to merge if you don't agree to my suggestion.

src/firefly/js/templates/lightcurve/LcConverterFactory.js
```diff
@@ -60,15 +104,19 @@ export function getAllConverterIds() {
...
-        logError(`Unable to find dataset converter ${datasetConverterId}`);
+        logError(`Unable to find dataset converter ${converterId}`);
         return;
     }
     return converter;
 }
 
+export function getMissionName(converterId) {
+    const converterData = converterId && converters[converterId];
+    return get(converterData, 'missionName', converterId);
```
cwang2016: should convertId to be capitalized? like get(converterData, 'missionName', convertId.toUpperCase())?

tgoldina:
I think it does not matter, because if you care you just specify
missionName for this converterId.

I made missionName a separate property from the key converterId. For
example, missionName for wise is 'WISE', missionName for lsst_sdss is 'LSST
SDSS'. If we add 'unknown', we might want to see 'Unknown' as mission name.

2017-02-22 11:06 GMT-08:00 Cindy Wang <notifications@github.com>:

src/firefly/js/templates/lightcurve/LcConverterFactory.js
ejoliet: I think missionName as it is is good, you might want to pass a mission name that has camel humps, like NExSci ;-)

BTW, a generic mission `unknown` mission name could be also called our `user defined` case - @tgoldina @robyww @cwang2016 DM-9491

cwang2016:
Review completes.  Very nice! Good to have time series viewer work dynamically on multiple missions.  Just a minor thought, On LcViewer, Instead of selecting the mission,  could the raw table loading determine the mission directly? or the selection of the mission will show a list of raw light curve table for selection?

tgoldina:
I'll treat the filter issue in a separate request. May be I should create s search processor, which takes filters as one of parameters (not as filters).

PR #314, opened 2017-02-22
DM-9550:

lznakano:
Fixed the color band radio button bug
  Fixed saving color band FITS file issue
  Correct the alignment issue
  Hide the color band radio buttons when the file type is png or reg

Please test to see if it works as expected.

Thanks!

### DIFF
--- a/src/firefly/js/ui/FitsDownloadDialog.jsx
+++ b/src/firefly/js/ui/FitsDownloadDialog.jsx
@@ -5,6 +5,11 @@
  *   propType: define all the property variable for the component
  *   this.plot, this.plotSate are the class global variables
  *
+ * Work History
+ *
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/FitsDownloadDialog.jsx
```diff
@@ -334,7 +350,7 @@ function resultsSuccess(request, plot) {
 
     var band = Band.NO_BAND;
     if (bandSelect) {
-        band = Band[bandSelect];
+        band = Band[bandSelect.toUpperCase()];
```
robyww: Since Band is an Enum object, the right way to do this line is:

`band= Band.get(bandSelect)`

That converts the string (case insensitive) into the enum

PR #315, opened 2017-02-22
DM-9552: Fixed: Reported mouse position on mouse click is a couple of pixels off

robyww:
_to test-_ choose lock by click option and click on the image symbol should be (within a pixel) of centered on the mouse


cwang2016:
Review completes. It works well.

PR #321, opened 2017-02-27
DM-9343:scatter and line chart zoom changed when changing between "points" and "connected points"

lznakano:
Fixed bug after zoom

Pease check:

1.  If the bug is fixed.
2. If the way of fixing is appropriate.

Thanks!

### DIFF
--- a/src/firefly/js/charts/ui/XYPlotOptions.jsx
+++ b/src/firefly/js/charts/ui/XYPlotOptions.jsx
@@ -52,7 +52,7 @@ function getUnit(colValStats, colname) {
 }
 */
 
-export function resultsSuccess(callback, flds, tblId) {
+export function resultsSuccess(callback, flds, tblId, zoom) {
     const xName = get(flds, ['x.columnOrExpr']);
     const yName = get(flds, ['y.columnOrExpr']);
     const xErr = get(flds, ['x.error']);
@@ -107,7 +107,8 @@ export function resultsSuccess(callback, flds, tblId) {
         shading: flds.shading || undefined,
         x : { columnOrExpr : xName, error: xErr, label : xLabel, unit : xUnit, options : xOptions},
         y : { columnOrExpr : yName, error: yErr, label : yLabel, unit : yUnit, options : yOptions},
-        tblId
+        tblId,
+        zoom,
     }, isUndefined);
 
     if (xErr || yErr) {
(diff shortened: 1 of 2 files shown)


ejoliet:
It works great for me. 
I've noticed that the reset button does not re-select (radio button UI) the default initial type of line. But when clicking 'apply' it surely did. Is just that the option is not visible. Weird.

This might be introduced with your change, not sure, i just wanted to mention it to you. 

Otherwise, please, raise a separate ticket for the reset problem to be looked in the future.

src/firefly/js/charts/dataTypes/XYColsCDT.js
```diff
@@ -194,7 +194,9 @@ export function setZoom(chartId, chartDataElementId, selection=undefined) {
             if (decimatedUnzoomed || isUndefined(decimatedUnzoomed)) {
                 noFetch = false;
             }
-            dispatchChartOptionsUpdate({chartId, chartDataElementId, updates: {zoom: xyPlotParams.selection, selection: undefined}, noFetch});
+            dispatchChartOptionsUpdate({chartId, chartDataElementId, updates: {zoom: xyPlotParams.boundaries, selection: undefined}, noFetch});
```
lznakano: Can you explain to me why it messed up setting min and max?  I looked at the algorithm used and did not see it messing up the setting.  Maybe I did not understand the algorithm.

tgoldina: Your approach "if the zoom is in the old options, copy it to the new options" is correct. Just do it in XYPlotOptions and only use "zoom" field. Currently, every time we press "Apply" button in options dialog, zoom is gone. We want to preserve it, unless X or Y columns changed.

We use "zoom" field to track zoom and "boundaries" field to track data or user boundaries. "boundaries" is a calculated field and should never be set directly. 

Because you clear boundaries, not zoom, the unzoom icon does not disappear when the chart is unzoomed.

When I said that using boundaries messes up setting min and max values, I was wrong. I noticed that zooming up and then setting min/max value does not work. But the real cause of this behavior is a bug in lines 376-377 of XYPlot. If both zoom and data boundaries are present, min/max of the two should be taken as an axis min/max. I would appreciate if you fix it too. (Or I can fix it later in your branch.)

src/firefly/js/charts/ui/XYPlotOptions.jsx
```diff
@@ -107,7 +107,8 @@ export function resultsSuccess(callback, flds, tblId) {
         shading: flds.shading || undefined,
         x : { columnOrExpr : xName, error: xErr, label : xLabel, unit : xUnit, options : xOptions},
         y : { columnOrExpr : yName, error: yErr, label : yLabel, unit : yUnit, options : yOptions},
-        tblId
+        tblId,
+        zoom,
```
tgoldina: Please, do not preserve zoom if X or Y column changed.

lznakano: You are right.  When the columns change, the zoom should not be preserved. Thanks for pointing it out.  I did not realize it.

PR #323, opened 2017-03-01
DM-9371 Display cutouts in LSST timeseries viewer

tgoldina:
https://jira.lsstcorp.org/browse/DM-9371

- Display image cutouts in LSST TS viewer 
To test, http://localhost:8080/firefly/ts.html, use LSST_SDSS and raw file attached to DM-9348

- Enable option to return cutouts on image search (at the moment, DAX supports cutouts image cutouts only Science CCD Exposure table, not yet for Deep Coadd) 
To test, http://localhost:8080/firefly/lsst-pdac-triview.html, select Images, set Return Image Size or leave it blank for full image.

- Updated SizeInputField to allow empty values.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTImageSearch.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTImageSearch.java
@@ -1,23 +1,27 @@
 package edu.caltech.ipac.firefly.server.query.lsst;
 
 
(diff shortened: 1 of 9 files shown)


cwang2016:
Review completes. All is good except the  concern for SizeInfputField.jsx.

tgoldina:
The latest commit addressed Cindy's concern: back to no formatting when size is converted to degrees in SizeInputField.jsx.

PR #324, opened 2017-03-01
DM-9581:MSX image rotation not computed correctly.

lznakano:
The reason that the rotation is wrong is that the incoming coordinate in m16 is galactic. But the out image created by rotation is in EQ_2000.  I think the rotation should be done in the same coordinate as the original image.  However, I maybe wrong since I don't know the reason why the EQ_2000 is hard coded there. 

But the bug is fixed after I changed the coordinate.  I also tried other rotation related function and seemed to me all are well.

Please review this and let me know if I am doing the wrong way.  Thanks!

To test the bug,
 1.  In image search, enter target "m16"
 2. Select MSX tab
 3. Search
 4. Click rotation icon and enter 180


robyww:
Don't merge yet I stilll want to look at the pull request.

PR #325, opened 2017-03-01
DM-7803: add time series launch capability to suit app.

loitly:
https://jira.lsstcorp.org/browse/DM-7803

On LSST PDAC, when searching for 'Deep Forced Source' data, there should be a button to view the time series data for the highlighted objectId.
If you upload a table with time series's metadata, there should be a button to view the table as time series.

To Test:
- Check out 'DM-7803_triview_timeseries' branch in both firefly and suit repo.
- Build suit and then goto http://localhost:8080/suit/
- Search for "Deep Forced Source", then click on 'View Time Series' button at upper right corner to view the data in a separate window.
- When uploading a table with time series metadata, i.e. 'timeCName', there should also be a 'View as Time Series' button at the same location

### DIFF
--- a/src/firefly/js/tables/ui/ActiveTableButton.jsx
+++ b/src/firefly/js/tables/ui/ActiveTableButton.jsx
@@ -1,0 +1,78 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {Component,PropTypes} from 'react';
+import sCompare from 'react-addons-shallow-compare';
+
+import {flux} from '../../Firefly.js';
+import {getTblInfoById, getActiveTableId} from '../TableUtil.js';
+
+/**
+ * This button will track the highlighted row of the given tbl_id or of the active table of the given tbl_grp.
+ * There are 2 callbacks available as props.
+ * onClick: This function is called when a user click on this button.
+ * onChange({tbl_id, highlightedRow}): This function is called when highlightedRow changes.
+ *                                     The returned object of this function will be set into state.
+ *                                     Props like show or title can be changed via this function.
+ */
+export class ActiveTableButton extends Component {
+
+    constructor(props) {
+        super(props);
+        this.state = {...props};
+        this.onClick = this.onClick.bind(this);
+    }
+
+    componentDidMount() {
+        this.removeListener= flux.addListener(() => this.storeUpdate());
+    }
+
+    componentWillUnmount() {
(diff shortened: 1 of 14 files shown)


src/firefly/js/tables/ui/ActiveTableButton.jsx
```diff
@@ -0,0 +1,73 @@
...
+import {getTblInfoById, getActiveTableId} from '../TableUtil.js';
+
+/**
+ * This download button does 2 things:
+ * 1. track a table for selected rows, then changes style to reflect whether or not it can be clicked
+ * 2. when clicked, display a popup dialog of the given download options panel
+ *    or a message prompting the user to select rows first
+ */
+export class ActiveTableButton extends Component {
+
```
robyww: I had a little trouble figuring out how this all worked.  The key is the onChanges function that takes a table_id and a highlighted row and return a boolean and a title.  I think this should be documented.

loitly:
Changes since last pushed:
- fix highlighted row out of sync after sorting or filtering.
- for lsst, show only first of multiimage. - trey
- add bolded text to highlight the time series button.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -321,13 +328,35 @@ function handleAddChart(layoutInfo, chartId) {
...
+ * handle logic when a new search is initiated.
+ * @param {LayoutInfo} layoutInfo layoutInfo
+ * @param {string} action
+ * @returns {LayoutInfo} the new layoutInfo
+ */
+function handleNewSearch(layoutInfo, action) {
+    const tbl_id = get(action, 'payload.request.META_INFO.tbl_id');
+    if (tbl_id === LC.RAW_TABLE) {
+        removeTablesFromGroup();
+        removeTablesFromGroup(LC.PERIODOGRAM_GROUP);
```
tgoldina: Looks like now we clear the derived tables only on the new search. But if we don't remove the derived tables, when raw table changes, how do we keep raw table in sync with the related tables? If user filters Raw and Phase Folded tables separately, we can not synchronize the highlight index between them, can we?

loitly: > tgoldina: ... how do we keep raw table in sync with the related tables? If user filters Raw and Phase Folded tables separately, we can not synchronize the highlight index between them, can we?

You are right.  I think a filter of a raw table should trigger a reset condition.

hello

PR #326, opened 2017-03-02
DM-9471: remove error flux column input field in setting box area

ejoliet:
error column input field removed from setting box. Please check that this is the case.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LCPanels.css
+++ b/src/firefly/js/templates/lightcurve/LCPanels.css
@@ -52,5 +52,5 @@
     background-color: #dcdcdc;
     border: 1px solid #cccccc;
     margin-bottom: 3px;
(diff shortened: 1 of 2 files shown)


ejoliet:
Yes, will be in a different ticket but lower priority. Thanks Wei.

src/firefly/js/templates/lightcurve/LCPanels.css
```diff
@@ -52,5 +52,5 @@
     background-color: #dcdcdc;
     border: 1px solid #cccccc;
     margin-bottom: 3px;
-    padding: 5px 10px;
+    padding: 5px 10px 10px 5px;
```
loitly: minor.. adding 10px 5px causes left/right padding mismatch.  I would think you want 10px 10px, instead.

PR #330, opened 2017-03-08
DM-9638 Remove unexpected image tab after a column for histogram is selected

cwang2016:
This development is to
- Remove unexpected image tab and not unmount the column select ion popup after a column for Histogram is selected.
- skip to check if the table used for selecting the column is a catalog table or an image meta table after the table is loaded.

test: 
- start a catalog search (or lsst catalog search from lsst-pdac-triview.html)
- click 'Charts' (or 'Add Charts' for lsst)
- click 'Histogram' or 'Scattered plot' radio button.  click 'Cols' and select a column from the popup table. 
- the select column name is shown in the field for columns in chart select panel.

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-test.html
+++ b/src/firefly/html/demo/ffapi-highlevel-test.html
@@ -182,8 +182,7 @@
                     { pageSize: 15,
                         META_INFO: {
                             datasetInfoConverterId : 'SimpleMoving',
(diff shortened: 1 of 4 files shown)


ejoliet:
works nice now. Thanks for fixing that.

cwang2016:
Remove the tag 'onlyData' and use the updated isMetaDataTable to check if the loaded table is meta data table or not.

cwang2016:
The issues for both popup tables on 'fits header' and 'column selection' are tested ok.

PR #331, opened 2017-03-09
DM-9594:The default radio button (point) in scatter plot is not selected after line style changes.

lznakano:
Minor changes to have the radio button selected.


lznakano:
I did not fix the default button behavior.  What I did is to keep the selected radio button as it is. 
I like Emmanuel to review to see if this behavior is OK.

PR #332, opened 2017-03-09
DM-9271: fix unselected columns appearing in downloaded tables.

loitly:
https://jira.lsstcorp.org/browse/DM-9271

Please read ticket for reported bug and how to duplicate it.


ejoliet:
Fixed. File name fixed too, was also wondering about it why we had 'undefined.tbl'. 
I guess the cleaning of metadata (META_INFO,...) and the documentation will be part of the improvement we said we want in the near future.

All good. 
Please, merge.Thanks!

PR #334, opened 2017-03-13
DM-9589: Fixed WISE case and add basic case with no images

ejoliet:
This is a fix for the basic case and neowise tables.

I've introduced a basic case which suppose to handle all numerical column of a general table uploaded in order to allow user to find the period and phase fold, with no images to be displayed.

I have added to the converter cases, 2 new functions, one about the images layout and another one to validate the table uploaded.

The case for WISE is trickier because we need to support tables that contains either frame_id or source_id or scanId and frame_num to be able to display WISE single exposure images. I've changed the WISE plot request and add specific function to validate the table. If the table doesn't have those column, the viewer will bailout in a message (the current message will be reviewed and is not definitive) to let the user know that although the table upload is not valid for WISE/NEOWISE, it can still upload by selecting the basic option.

On table load has changed in order to get the layoutInfo object updated which is acting as changing state to for layout result display, where showImages control the final layout.

To test, please, upload table for different cases and see if the above make sense, otherwise, come to me and let me know your questions.

WISE and Basic case are those that should go for release next week or so.
Generic case has more advanced options and should be reviewed to be included in future releases.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcManager.js
+++ b/src/firefly/js/templates/lightcurve/LcManager.js
@@ -379,7 +379,8 @@ function clearResults(layoutInfo) {
         periodState: LC.PERIOD_PAGE,
         missionEntries: null,
         generalEntries: null,
-        fullRawTable: null
+        fullRawTable: null,
+        error:''
     });
 }
 
--- a/src/firefly/js/templates/lightcurve/LcManager.js
+++ b/src/firefly/js/templates/lightcurve/LcManager.js
@@ -25,6 +25,7 @@ import {getConverter} from './LcConverterFactory.js';
 import {sortInfoString} from '../../tables/SortInfo.js';
 import {makeMissionEntries, keepHighlightedRowSynced} from './LcUtil.jsx';
 import {dispatchMountFieldGroup} from '../../fieldGroup/FieldGroupCntlr.js';
+import {ERROR_MSG_KEY} from '../lightcurve/generic/errorMsg.js';
 
 export const LC = {
     RAW_TABLE: 'raw_table',          // raw table id
@@ -41,6 +42,7 @@ export const LC = {
 
     META_TIME_CNAME: 'ts_timeCName',
     META_FLUX_CNAME: 'ts_fluxCName',
+    META_FLUX_BAND: 'ts_bandName',
     META_ERR_CNAME: 'ts_errorCName',
     META_COORD_XNAME: 'ts_coordXName',
     META_COORD_YNAME: 'ts_coordYName',
@@ -60,10 +62,10 @@ export const LC = {
 
     META_TIME_NAMES: 'ts_timeNames',
     META_FLUX_NAMES: 'ts_fluxNames',
+    META_URL_NAMES: 'ts_datasources',
     META_ERR_NAMES: 'ts_errorNames',
 
     META_MISSION: MetaConst.TS_DATASET,
-
     MISSION_DATA: 'missionEntries',
     GENERAL_DATA:'generalEntries',
 
@@ -75,6 +77,10 @@ const plotIdRoot= 'LC_FRAME-';
 
 var defaultCutout = '0.2';
 //var defaultFlux = '';
+
+function getFluxBandName(layoutInfo) {
+    return get(layoutInfo, ['missionEntries', LC.META_FLUX_BAND]);
+}
 
 function getFluxColumn(layoutInfo) {
     return get(layoutInfo, ['missionEntries', LC.META_FLUX_CNAME]);
@@ -107,10 +113,21 @@ function getDataSource(layoutInfo) {
 function getImagePlotParams(layoutInfo) {
     return {timeCol: getTimeColumn(layoutInfo),
             fluxCol: getFluxColumn(layoutInfo),
+            bandName: getFluxBandName(layoutInfo),
             dataSource: getDataSource(layoutInfo),
             ra: getRA(layoutInfo),
             dec: getDEC(layoutInfo),
             coordSys: getCoordSys(layoutInfo)};
+}
+
+function getUserInputParams(layoutInfo) {
+    return {timeCol: getTimeColumn(layoutInfo),
+        fluxCol: getFluxColumn(layoutInfo),
+        bandName: getFluxBandName(layoutInfo),
+        dataSource: getDataSource(layoutInfo),
+        ra: getRA(layoutInfo),
+        dec: getDEC(layoutInfo),
+        coordSys: getCoordSys(layoutInfo)};
 }
 
 export function getConverterData(layoutInfo=getLayouInfo()) {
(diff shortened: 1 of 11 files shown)


src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -124,52 +142,78 @@ const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables,
...
+                    <SplitContent>
+                        <div style={{display: 'flex', flexDirection: 'column', height: '100%'}}>
+                            <div className='settingBox'>{settingBox}</div>
+                            <div style={{flexGrow: 1, position: 'relative'}}>
+                                <div className='abs_full'>{tables}</div>
+                            </div>
+                        </div>
+                    </SplitContent>
+                    <SplitContent>{xyPlot}</SplitContent>
+                </SplitPane>
```
loitly: The table/chart <SplitPane...../> is the same in both cases.. you should reuse it, instead of duplicating it,

src/firefly/js/templates/lightcurve/LcUtil.jsx
```diff
@@ -42,7 +42,8 @@ export function makeMissionEntries(tableMeta, layoutInfo={}) {
         [LC.META_ERR_CNAME]: get(tableMeta, LC.META_ERR_CNAME, converterData.defaultYErrCname),
         [LC.META_TIME_NAMES]: get(tableMeta, LC.META_TIME_NAMES, converterData.timeNames),
         [LC.META_FLUX_NAMES]: get(tableMeta, LC.META_FLUX_NAMES, converterData.yNames),
-        [LC.META_ERR_NAMES]: get(tableMeta, LC.META_ERR_NAMES, converterData.yErrNames)
+        [LC.META_ERR_NAMES]: get(tableMeta, LC.META_ERR_NAMES, converterData.yErrNames),
+        [LC.META_FLUX_BAND]: get(tableMeta, LC.META_ERR_NAMES, converterData.bandName)
```
loitly: This doesn't look right... FLUX -> ERR.  Will you take a look?

src/firefly/js/templates/lightcurve/wise/WiseMissionOptions.js
```diff
@@ -29,46 +32,85 @@ export class WiseSettingBox extends Component {
...
             <ValidationField key={key} fieldKey={key} wrapperStyle={wrapperStyle}
                              style={{width: 80}}/>
         );
 
         const missionKeys = [LC.META_TIME_CNAME, LC.META_FLUX_CNAME];
         const missionOtherKeys = [LC.META_ERR_CNAME];
+        const imageDataSource = [LC.META_URL_CNAME]; // use meta_url from generic case, see LcManager.js:320 for why
         const missionListKeys = [LC.META_TIME_NAMES, LC.META_FLUX_NAMES];
-
+        const topZ = 9999;
```
loitly: This is problematic.  9999 is too high.  The validation error message for the same field is showing underneath the suggestion box.

src/firefly/js/templates/lightcurve/LcUtil.jsx
ejoliet: good one. Will fix.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -107,12 +113,23 @@ function getDataSource(layoutInfo) {
...
+function getUserInputParams(layoutInfo) {
+    return {timeCol: getTimeColumn(layoutInfo),
+        fluxCol: getFluxColumn(layoutInfo),
+        bandName: getFluxBandName(layoutInfo),
+        dataSource: getDataSource(layoutInfo),
+        ra: getRA(layoutInfo),
+        dec: getDEC(layoutInfo),
+        coordSys: getCoordSys(layoutInfo)};
+}
+
```
cwang2016: the content of getUserInputParams & getUserInputParams look the same?

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -585,24 +626,32 @@ export function setupImages(layoutInfo) {
...
-            .forEach( (pv) => dispatchDeletePlotView({plotId:pv.plotId, holdWcsMatch:true}));
-    } catch (E){
-        console.log(E.toString());
+            .filter((pv) => pv.plotId.startsWith(plotIdRoot))
+            .filter((pv) => pv.plotId !== vr.mpwWcsPrimId)
+            .filter((pv) => !maxPlotIdAry.includes(pv.plotId))
+            .forEach((pv) => dispatchDeletePlotView({plotId: pv.plotId, holdWcsMatch: true}));
+
+        // Decide whether or not to show images, mission specific:
+        const imagesShown = converterData.shouldImagesBeDisplayed(layoutInfo);
```
cwang2016: can this line occur move to the earlier part of this function?

src/firefly/js/templates/lightcurve/LcPeriod.jsx
```diff
@@ -649,9 +649,9 @@ function LcPFOptions({fields, lastPeriod, periodList=[]}) {
                                  border: highlightBorder}}>
                         <ValidationField fieldKey={fKeyDef.period.fkey} />
                         <div>
-                            <button type='button' className='button std hl' onClick={revertPeriod}>
+                            {/*<button type='button' className='button std hl' onClick={revertPeriod}>
                                 {revertPeriodTxt}
-                            </button>
+                            </button>*/}
```
ejoliet: This is to address ticket DM-9768

ejoliet:
I've fixed the popup problem (Loi's catch) and also move the validation of the table to the viewer before the result layout kick in. So it takes an error from `LayoutInfo` object and check if empty.
Let me know if you find more problems. Thanks!

src/firefly/js/templates/lightcurve/LcManager.js
ejoliet: no, it needs to be after the plot request object being called.

src/firefly/js/templates/lightcurve/wise/WiseMissionOptions.js
ejoliet: Fixed by changing to topZ= 3.

src/firefly/js/templates/lightcurve/LcManager.js
wmiipac: I think you mean getUserInputParams and getImagePlotParams are the same, right?

loitly:
It only handles error for WISE.
If you upload a bad ipac table, you will get uncaught javascript errors in console and a generic table load error message in where the table would be.  Same goes for LSST.  
Once this happens, a page reload is needed to recover.

If not fixed here, a ticket(s) should be created to handle all errors.

ejoliet:
Error handling was raised here: DM-9770 and is covered for WISE.
For other cases, i think we need to raise a ticket. There is a general ticket here DM-4592 but i guess we need something specific for LC and uploaded table, right? 
Will merge this and raise ticket for the other cases.

PR #335, opened 2017-03-13
DM-9342:histogram option input dialog

lznakano:
DM-9342:
   Add the bin width,min and max options

Please check the UI and let me if they look OK with you or need more changes.

### DIFF
--- a/src/firefly/js/charts/ui/HistogramOptions.jsx
+++ b/src/firefly/js/charts/ui/HistogramOptions.jsx
@@ -56,7 +56,7 @@ export function setOptions(groupKey, histogramParams) {
 }
 
 const algorithmOptions = [  {label: 'Bayesian blocks', value: 'bayesianBlocks'},
-                            {label: 'Fixed size', value: 'fixedSizeBins'} ];
+                            {label: 'Uniform binning', value: 'fixedSizeBins'} ];
 const binSizeOptions = [  {label: 'Number of bins:', value: 'numBins'},
     {label: 'Bin width:', value: 'binWidth'} ];
 
--- a/src/firefly/js/charts/ui/HistogramOptions.jsx
+++ b/src/firefly/js/charts/ui/HistogramOptions.jsx
@@ -19,10 +19,12 @@ import {getAppOptions} from '../../core/AppDataCntlr.js';
 
 export const histogramParamsShape = PropTypes.shape({
          algorithm : PropTypes.oneOf(['fixedSizeBins','bayesianBlocks']),
-         numBins : PropTypes.oneOfType([React.PropTypes.string,React.PropTypes.number]),
+         numBins : PropTypes.string,
+         binWidth : PropTypes.string,
+         fixedBinSizeSelection : PropTypes.oneOf(['numBins','binWidth']),
          falsePositiveRate : PropTypes.string,
-         minCutoff : PropTypes.number,
-         maxCutoff : PropTypes.number
+         minCutoff : PropTypes.string,
+         maxCutoff : PropTypes.string
       });
 
 
@@ -35,20 +37,25 @@ export function resultsFail() {
     // TODO: do I need to do anything here?
 }
 
-export function setOptions(groupKey, histogramParams) {
+export function setOptions(groupKey, histogramParams, key=null) {
     const flds = [
         {fieldKey: 'columnOrExpr', value: get(histogramParams, 'columnOrExpr')},
         {fieldKey: 'x', value: get(histogramParams, 'x', '_none_')},
         {fieldKey: 'y', value: get(histogramParams, 'y', '_none_')},
         {fieldKey: 'algorithm', value: get(histogramParams, 'algorithm', 'fixedSizeBins')},
         {fieldKey: 'falsePositiveRate', value: get(histogramParams, 'falsePositiveRate','0.05')},
+        {fieldKey: 'fixedBinSizeSelection', value: get(histogramParams, 'fixedBinSizeSelection','numBins')},
         {fieldKey: 'numBins', value: get(histogramParams, 'numBins','50')}
+
+
     ];
     dispatchMultiValueChange(groupKey, flds);
 }
 
 const algorithmOptions = [  {label: 'Bayesian blocks', value: 'bayesianBlocks'},
                             {label: 'Fixed size', value: 'fixedSizeBins'} ];
+const binSizeOptions = [  {label: 'Number of bins:', value: 'numBins'},
+    {label: 'Bin width:', value: 'binWidth'} ];
 
 export class HistogramOptions extends React.Component {
 
@@ -100,6 +107,7 @@ export class HistogramOptions extends React.Component {
         this.iAmMounted= true;
     }
 
+
     renderAlgorithmParameters() {
         const {groupKey, histogramParams} = this.props;
         const {fields} = this.state;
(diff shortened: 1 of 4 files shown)


ejoliet:
The option added works as expected. Thanks for adding it quikly!
Now, to answer Tatiana comments:
There was a mockup document (PPT) that was sent around about UI histograms based on Gregory/David and Vandana ideas. 
And i think this was the idea behind this implementation. My comments:
- Didn't understand the origin problem? Bin width is size of the bin in unit of the column (fixed range bin of magnitude for instance).
- One of the requirement was to have a 'Uniformly binned' algorithm to plot histogram based on 2 options: 'bin size' or 'bin width'. We could be more clever and add more behavior to the radio button selection but i'm not sure if it's worth doing it.  
- Probably we should be consistent with other settings layout and make the X min, max value similar to other places.
- Agree on the reset/clear fix

I will ask Vandana to take a look before accepting this merge anyway about wording and labels.

lznakano:
The UI layout was based on the PPT I got from Xiuqin.    I will wait for your final decision about 
the UI.

Good catch on the reset/clear.  I forgot about it and did not test it there.  I will make change now.

lznakano:
I fixed the clear/reset issue.   I will make change in the UI after talking to Emmanuel.

tgoldina:
I meant by the origin the left edge of the first bin. Do we start at user
defined minimum or – if it's not defined – data minimum? Or we start at the
center of user/data range? In the last case, the first bin left edge will
depend on the bin size, and user can not change it easily.

Would it be possible to attach the PPT to the ticket? I did not see it.

(Just thought when the selection changes from "Number of bins" to "Bin
size", and both input boxes are visible and non-compliant, it's a bit
confusing.)

2017-03-13 16:55 GMT-07:00 Lijun Zhang <notifications@github.com>:

ejoliet:
Only label change feedback: 
Algorithm 'Fixed size' to be renamed 'Uniform binning'.

There is a concern about the usefulness of the 'bin width' if the app doesn't provide default value after column is selected, the same for min,max. 

Whether is more work or easy, we should provide a default value once the user choose the column for min, max and bin width/size depending on the option selected. Same if the user change the number of bins or bin width, both option should be updated. 

I know that's require more work and we need to talk about if we want to include the branch as it is now for RC or we just postpone the merge after release and plan more work to do. 

@xiuqin Let's talk tomorrow during CCB and release March goals. 

Please, put on hold this branch for now. Will decide tomorrow. 

Thanks!

ejoliet:
@tgoldina i see your point and i think you are right. As you know, LSST/IRSA would like to see **in the future** the number of point that have been discarded/included on the left and right - this is for the overflow/outlier discussion. And your remarks will make more sense when we look at that story. 

I think we need to go slowly anyway. For now, i think it would be good to have the options as implemented here and make progress in the near future about more advanced feature that are expected from the histograms.

lznakano:
@tgoldina I don't have the PPT.   Xiuqin printed a copy and drew the components to be added. 
 I agree with you and Emmanuel for all your concerns.  When I tested it myself, I just realized the issues.  I discussed with Cindy.  We decided to put it on to let people see it.   I did see the "colValStats" has min/max information.  I believe the using state or listeners, we should be able to refill the min/max/binWidth as requested.  @tgoldina may have better view on this.

src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -42,13 +44,21 @@ export function setOptions(groupKey, histogramParams) {
         {fieldKey: 'y', value: get(histogramParams, 'y', '_none_')},
         {fieldKey: 'algorithm', value: get(histogramParams, 'algorithm', 'fixedSizeBins')},
         {fieldKey: 'falsePositiveRate', value: get(histogramParams, 'falsePositiveRate','0.05')},
-        {fieldKey: 'numBins', value: get(histogramParams, 'numBins','50')}
+        {fieldKey: 'fixedBinSizeSelection', value:'numBins'},
```
tgoldina: The value should be `get(histogramParams, 'fixedBinSizeSelection', 'numBins')`

src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -224,6 +255,52 @@ export class HistogramOptions extends React.Component {
     }
 }
 
+//Make the fixed bin layout
+function renderFixedBinSizeOptions(groupKey, histogramParams){
+    return (
+      <div style={{display: 'flex', flexDirection: 'row', padding: '5px 0 15px'}} >
+         <RadioGroupInputField
+            initialState= {{
+                                value: 'bin size selection',
```
tgoldina: Here the value should also be `get(histogramParams, 'fixedBinSizeSelection', 'numBins')` - otherwise first time rendering would not be correct. Example: create histogram from drop down 'Charts' menu with bin width option, then open the side options.

tgoldina:
Populating min, max, and binWidth might require server call, because we support column expressions. But I agree, for single column, we can get min/max from column values statistics (colValStats).

src/firefly/js/charts/ui/HistogramOptions.jsx
lznakano: Fixed

lznakano:
@tgoldina You are right.  If the expression is not a linear relation, using min/max can not get min/max for the expression.  We still need the server calls.

lznakano:
@ejoliet @tgoldina 

 This commit includes the following:
  1. Disable/enable the input bin width/number of bins based on the radio button state
 2. Fill the min/max and binWidth based on data min/max and (max-min)/50. 

Please review it and let me know if I need to make change changes.  

Thanks!

src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -60,6 +60,41 @@ const algorithmOptions = [  {label: 'Bayesian blocks', value: 'bayesianBlocks'},
...
+                        const dataMin = colValStats[i].min;
+                        const dataMax = colValStats[i].max;
+                        const binWidth = (dataMax - dataMin)/50.0;
+                        inFields = updateSet(inFields, ['minCutoff', 'value'],`${dataMin}`);
+                        inFields = updateSet(inFields, ['maxCutoff', 'value'], `${dataMax}`);
+                        inFields = updateSet(inFields, ['binWidth', 'value'], `${binWidth}`);
+                        break;
+
+                    }
+                }
```
tgoldina: When colName is an expression (it can not be found in the colValStats), the values should be set to undefined. (Otherwise we have wrong suggestions, which user should remove one by one.)

src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -60,6 +60,41 @@ const algorithmOptions = [  {label: 'Bayesian blocks', value: 'bayesianBlocks'},
...
+            // when column name changes, update the min/max input
+            fieldKey = get(action.payload, 'fieldKey');
+            const colName = action.payload.value;
+            if (fieldKey === 'columnOrExpr') {
+
+                for (var i=0; i<colValStats.length; i++){
+                    if (colName=== colValStats[i].name){
+                        const dataMin = colValStats[i].min;
+                        const dataMax = colValStats[i].max;
+                        const binWidth = (dataMax - dataMin)/50.0;
```
tgoldina: What if the value of numBins is not 50? Perhaps, it should be divided by the current value of 'numBins' field.

Another improvement would be adjust the value of 'binWidth' when all of 'numBins', 'minCutoff', 'maxCutoff' are populated and one of them changes.

src/firefly/js/charts/ui/HistogramOptions.jsx
lznakano: Good suggestion.  I just fixed it.

tgoldina:
Another bug I noticed (which was there, but less noticeable). When user sets Min and Max in Charts dropdown, creates chart and then opens options, Min and Max are the suggestions, not the values user to create histogram. The way to cure it is to add
```
        if (histogramParams) {
            defer(setOptions, groupKey, histogramParams);
        }
```
to the end of componentDidMount() - similar to how we do it for XYPlot.

src/firefly/js/charts/ui/HistogramOptions.jsx
lznakano: I filled width by the (max-min)/50 since the 50 is the default.   My thought is to give user a reasonable bin width to start with.  The users can change it later.   The bin width can be any valid float value.  It does not have to be bounded with (max-min)/numBins.

lznakano:
@tgoldina 
I made the changes as you request (change bin width when min/max or number of bins changes) since the implementation was  just adding a switch statement.  I also fixed the bug you noticed.  I did rebase and messed up my environment.  It took me some time to clean it and pushed the update here.  Please review it one more time to see if you can spot anything.  Thanks!

PR #336, opened 2017-03-15
IRSA-174: TSV mission list should display only WISE and Basic for now

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-174

To test:
git checkout this branch on both firefly and ife.
build irsaviewer in ife.
goto http://localhost:8080/irsaviewer/timeseries to verify that only WISE and Basic option in dropdown.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcViewer.jsx
+++ b/src/firefly/js/templates/lightcurve/LcViewer.jsx
@@ -82,7 +82,7 @@ export class LcViewer extends Component {
     }
 
     render() {
-        var {isReady, menu={}, appTitle, appIcon, altAppIcon, dropDown,
+        var {isReady, menu={}, appTitle, appIcon, altAppIcon, dropDown, missionOptions,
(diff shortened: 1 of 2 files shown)


ejoliet:
Works as expected.
Thanks!

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -198,8 +198,9 @@ BannerSection.propTypes = {
  */
 export function UploadPanel(props) {
     const wrapperStyle = {margin: '5px 0'};
+    const {missionOptions=getAllConverterIds()} = props || {};
 
-    const options = getAllConverterIds().map((id) => {
+    const options = missionOptions.map((id) => {
```
ejoliet: I guess this will be the object that has the column time, value, etc, right?

ejoliet:
Before merging, please, changed the options list to ['wise', 'other']. 
{{basic}} has been renamed to {{other}} (see DM-9827) but i thought this PR was already merged in IFE so the changed is already in Firefly {{dev}}.

PR #337, opened 2017-03-15
IRSA-194: Handle bad ipac table upload

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-194

To ensure our application works properly, we will convert uploaded tables to fixed-length ipac format before displaying.

Also part of this pull request:
- mapped AnyFileUpload into the CommandService scheme.
- clean and combine multiple web.xml.

Some files from ife and suit were also changed.  The changes are in the same branch.

Below is a sample catalog file that showed up with missing rows.
```
| ra          | dec        |
| double      | double     |
  32.073988     55.074964
 238.594530     22.404794
 250.679708     56.165833

```


ejoliet:
Seems to work in tri-view but my test from local gator.html using API didn't work, is still shows 2 rows. Can you confirm that the fix is also in API used by Gator/Atlas?

loitly:
> Seems to work in tri-view but my test from local gator.html using API didn't work, is still shows 2 rows. Can you confirm that the fix is also in API used by Gator/Atlas?

You are right.  I've made changes to also clean ipac table coming from external sources.  This should fix the problem.  Please verify.

ejoliet:
Last commit fix also the problem in API. Thanks!

PR #339, opened 2017-03-17
xyplot: support asymmetric errors

tgoldina:
https://jira.lsstcorp.org/browse/DM-8685

To test: 
1. WISE catalog search (ex. m31, 100 arcsec)
2. Chart dropdown -> Scatter plot -> set "Error" options (can be const or col expressions)
3. Check on chart

### DIFF
--- a/src/firefly/js/charts/ui/XYPlotOptions.jsx
+++ b/src/firefly/js/charts/ui/XYPlotOptions.jsx
@@ -3,7 +3,7 @@
  */
 import React, {PropTypes} from 'react';
 
-import {get, isEmpty, isUndefined, omitBy, defer} from 'lodash';
+import {defer, get, isEmpty, isUndefined, omitBy, set} from 'lodash';
 
 import ColValuesStatistics from '../ColValuesStatistics.js';
 import CompleteButton from '../../ui/CompleteButton.jsx';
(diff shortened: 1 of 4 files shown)


src/firefly/js/charts/ui/XYPlotOptions.jsx
```diff
@@ -109,16 +109,36 @@ export function resultsSuccess(callback, flds, optionParameters) {
...
         zoom
     }, isUndefined);
 
-    if (xErr || yErr) {
-        if (xErr) { xyPlotParams.x.error = xErr; }
-        if (yErr) { xyPlotParams.y.error = yErr; }
-    }
+    // set error related parameters
+    ['x', 'y'].forEach((axis) => {
+        const errorsType = get(flds, [`${axis}.errorsType`]);
```
loitly: get() take an array of paths or a string of path.
nothing is wrong with your statement.
but alternatively, you can do 
```
get(flds, `${axis}.errorsType`)  or
get(flds, [axis, 'errorsType'])
```

tgoldina: I can not use the alternative ways you have suggested, because the fieldKey contains dot in it, ex. 'x.errorsType'. Since every fieldKey is converted into an object property, this is an example of the property, which can only be accessed using the square bracket notation. 

I did it for convenience of mapping between fields and plot parameters, which are not flat as fields. In plot parameters,  x  is an object with all x-related parameters.

PR #341, opened 2017-03-28
 DM-9803:Multi-object search in PDAC

lznakano:
Add mutli-object search processor that uses Java Concurrent to do a parallel search
 DM-9920:
     Fixed bug in CatalogSelectViewPanel.  The IRSA's catalog search does not work due to this bug
 Fixed bug in LSSTCatalogSelectViewPanel
 Eable multi-object search in LSSTCatalogSelectViewPanel.jsx
 DM-9964:
   Changed the empty result behavior in LSSTCatalogSearch so that the multi-object search is not blocked.

DM-9803:
  Modified UI part of the codes
  Rename the search processor

 DM-9803:
   Fixed bugs in CatalogSelectViewPanel.jsx and LSSTCatalogSelectViewPanel.jsx
   Implement multi-object search

 DM-9803:
   Modified LSSTQuery so that the empty table will not stop the mutl-pjbect search.

DM-9803:
  Add code to prevent from forever thread

Several tickets were implemented (shown as above )in this branch.  The bugs were found while implementing this ticket.  Without fixing the bugs, the implementation can not be worked.  So the tickets were combined. 

Please note:
I could add this implementation into the LSSTCatalogSearch.  I prefer to have a class for Concurrent 
implementation.  Please evaluate it.  If there an is objection, I can merge it to the existing LSSTCatalogSearch.  

It took me a lot of time to test it due to the PDAC 's issues. After several days's frustrations, finally, I made fiver test input files work.  Please download the attached file and test it.
[testData.tar.gz](https://github.com/Caltech-IPAC/firefly/files/877197/testData.tar.gz)

Thanks!


src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTMultiObjectSearch.java
```diff
@@ -0,0 +1,354 @@
...
+     * @throws ExecutionException
+     * @throws InterruptedException
+     * @throws IOException
+     */
+   private  DataGroup[]  getSearchResult(TableServerRequest request, DataGroup inDg) throws ExecutionException, InterruptedException, IOException, EndUserException, TimeoutException {
+
+
+       DataObject[] rows = inDg.values().toArray(new DataObject[0]);
+
+        int nThread = request.containsParam("nThread")? Integer.parseInt(request.getParam("nThread")) :2;
```
robyww: i don't think the number of threads should come from the request.  I think is should be property the the search processes keeps

src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTMultiObjectSearch.java
```diff
@@ -0,0 +1,354 @@
...
+               if (list.get(i).get(timeout, TimeUnit.SECONDS) != null) {
+                //   if (list.get(i).get() != null) {
+                   dgArrayList.add(list.get(i).get());
+               }
+           }
+           catch (InterruptedException e) {
+               e.printStackTrace();
+           } catch (ExecutionException e) {
+               e.printStackTrace();
+           }
```
robyww: Using these Future objects is pretty cool!

robyww:
Testing:
I tried upload3.tbl
```
\ EQUINOX = 'J2000.0'
|   ra     |   dec    |  major |  ratio | angle |
|   double |   double | double |  double| double|
 9.469000    -1.152000   100.0     0.25    0.5
 9.500000    -1.230000    50.0     0.5     0.15
```
The search failed

lznakano:
I split the LSSTMultiObjectSearch.java two two classes and of which is taking a request and the number of thread and then does the parallel search and returns a DataObject.  Please take a look and let me know if it looks OK.  

Thanks!

robyww:
When I said break it up into reusable parts I did not mean just do more inheritance. I don't think it is really any different than before.  Don't add more inheritance. I wanted you to make utility routines out of the parts that to the concurrent searching and reassembly. 

You did some nice work.  We want to be able to reuse those parts.

PR #342, opened 2017-03-31
 DM-8854:Create a Unit test for Rotate class

lznakano:
DM-8854:
  Add unit test for Rotate class
  Add the algorithm descriptions

The coverage is 80%.  There is a main program inside which made the coverage lowered.  To 
test it, please run:
   gradle :firefly:test

Thanks!

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/Rotate.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/Rotate.java
@@ -5,8 +5,46 @@ package edu.caltech.ipac.visualize.plot;
 
 public class Rotate
 {
-
-public static boolean debug = false;
+    /**
+     * DM-8845
+     * While working on this ticket, the algorithm is added for later references
+     *
+     * Basic rotation
+     * A basic rotation (also called elemental rotation) is a rotation about one of the axes of a
+     * Coordinate system. The following three basic rotation matrices rotate vectors by an angle θ
+     * about the x, y, or z axis, in three dimensions, using the right-hand rule—which codifies their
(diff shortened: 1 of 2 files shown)


src/firefly/java/edu/caltech/ipac/visualize/plot/Rotate.java
```diff
@@ -5,8 +5,46 @@
...
+     *
+     *     1            0            0
+     *     0       cos(angle)     - sin(angle)
+     *     0       sin(angle)       cos(angle)
+     *
+     *     }
+     *
+     * Ry(angle) = {
+     *
+     *      cos(angle)    0         - sin(angle)
```
cwang2016: should this be -sin(angle) or +sin(angle)?

lznakano: Yes, I have the correct document in RotateTest.java but Roate.java.  There is nothing documented there when I started working on this ticket.  I added the algorithm in RotateTest.java.   I also added some in Rotate.java.  I created another ticket for Rotate.java.  I will update the document and the code then.

cwang2016:
review complete.  Here is some concern,  
- the unit test contains the computation to calculate the expected output used for baseline, how would this piece of computation to be verified?  The class to be tested, rotate.class,  intends to build rotation matrix for the purpose to rotate any given world point to be aligned with x axis at 180 deg direction (as I traced its code), so the unit test should just check if input point used for the baseline is rotated to (-1, 0, 0) while testing 'do_rotate' function in Rotate.class. Since the functional description is not seen from   the document of the tested class, so it is understandable why the algorithm is coded and put inside the unit test.  
You may merge the branch. Thanks.

PR #343, opened 2017-04-05
DM-9943 Create Plotly version of histogram

cwang2016:
This development made a histogram conversion from hi-chart to Plotly based, including
- construct Plotly accepted data  and layout JSON structure based on the parameters passed to  the 'Histogram' component. 
- Try to make the Plotly based histogram with similar styling as current hi-chart based histogram.

note:  scatter chart on phase folding in light curve (LcPeriodPlotly.jsx) is trimmed and cleaned a little bit. 

test:
start a catalog search
click 'charts' from the menu in the tri-view page to select 'histogram' and the column to make histogram chart


src/firefly/js/charts/chartTypes/HistogramTblView.jsx
```diff
@@ -38,7 +40,9 @@ function Chart(props) {
     if (!TblUtil.isFullyLoaded(tblId) || !chartData || !heightPx || !widthPx) {
         return (<div/>);
     }
-    const { isDataReady, data:histogramData, options:histogramParams} = ChartsCntlr.getChartDataElement(chartId);
+    var { isDataReady, data:histogramData, options:histogramParams} = ChartsCntlr.getChartDataElement(chartId);
+    const HistogramInstance = get(getAppOptions(), 'charts.chartEngine') !== 'plotly' ? Histogram : HistogramPlotly;
```
tgoldina: I suggest renaming Histogram to HistogramHighcharts and have Histogram.jsx handle instance creation, like you do here. This way all references to Histogram will be switched to plotly references without changes in the code that uses it. (We have a few references besides HistogramTblView.)

tgoldina:
Another thing I noticed – we can address it in a separate bug ticket if it takes long time to fix – X log option produces _Invalid data passed to Histogram: TypeError: Cannot read property '1' of undefined_ message and histogram is not displayed. 

Test case: 
catalog search m31, radius 200 arcsec, 
create fixed bin histogram for w1mpro, then apply X log option.

src/firefly/js/charts/ui/HistogramPlotly.jsx
```diff
@@ -0,0 +1,304 @@
...
+
+    render() {
+        this.error = undefined;
+        const {dataUpdate} = this.state;
+        const {plotlyData, plotlyDivStyle, plotlyLayout}= this.chartingInfo;
+        const {width, height} = this.props;
+
+        return (
+            <div style={{width: width? Number(width): undefined,
+                         height: height? Number(height): undefined}}>
```
loitly: Is there a reason why you are limiting width/height to numeric values only?  They cannot use '100px' or 'calc(100% - 20px)'?

cwang2016:
Regarding the x log issue, it is found that no data is returned from the request (cmd: tableSearch and request for HistogramProcessor) sent to the server side. It needs more investigation to find out the reason that causes the problem. I will put this issue into another ticket.

tgoldina:
I think, Histogram X log scale got broken after https://jira.lsstcorp.org/browse/DM-9342, when we added bin width option. When X log is requested, bin width and the related options should be disabled or all the values should be logs. @ejoliet needs to check it out.

ejoliet:
@tgoldina good point, i'll check with IRSA what is the default behavior to follow if log is enabled. It seems anyway that Log in XY plot is not working properly either ( David Ciardi raised the issue in IRSA jira ticket IRSA-316).

cwang2016:
Updates per review comment, 
- Histogram.jsx will automatically switch to HistogramPlotly or HistogramHichart based on the chart engine setting. 
- Add plotly click event handler to resolve the histogram chart selection issue. However, the 'click' event is not able to be captured when the mouse click occurs on the gap between two bars if there is. 
- update the left/right and top/bottom margins based on the setting of x and y axis. 

Additional update: 
- Update the left margin (or right margin if the yaxis title on the right side of the chart) dynamically based on the width of the text string for y ticks.

PR #345, opened 2017-04-07
IRSA-327: fix rounding problem

ejoliet:
I've set a default min precision for VOTable converter to data group, plus added the reading of a precision attribute from VOTable if it exists.
The guessing part of the format is only made in the first row. We need to improve that also for the IPAC table reader which is related to ticket raised during GOODS data-set release ([IRSA-128](https://caltech-ipac.atlassian.net/browse/IRSA-128)).

Please check that this make sense and doesn't affect other things.

To test: do a periodogram calculation with period min=1, period max=2, rest as default for example, the first value of the VOTAble from the API is `1`, the second is 1.000... 
This value should appear  in the table/xy plot view with at least `8` digit which is the default minimum precision.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/IpacTableUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/util/IpacTableUtil.java
@@ -174,11 +174,15 @@ public class IpacTableUtil {
         }
     }
     public static void guessFormatInfo(DataType dataType, String value) {
+       guessFormatInfo(dataType, value, 0);
+    }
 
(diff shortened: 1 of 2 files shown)


ejoliet:
Precision is not yet generated from back-end service, Walter needs to add it. But the code is prepare to handle it ;-)

xiuqin:
@ejoliet  if the precision is not set, the default is 8. Correct?

ejoliet:
Correct, but we can always change the default precision later. This is only applied to VOTable reader. We still have the problem for IPAC tables...

PR #347, opened 2017-04-08
IRSA-336: reorder periodogram tables

ejoliet:
Change the order of the dispatch so the periodgoram is the first table from left to right.

Test: run periodogram from TSViewer and check that periodogram table is the first from the left, peaks the second.


wmiipac:
screen shot 2017-04-10 at 1 08 49 pm

ejoliet:
Yes, i know. This is a bug related to table guessing format, see [IRSA-327](https://github.com/Caltech-IPAC/firefly/pull/345) PR.
Will merge this one. Thank you!

PR #348, opened 2017-04-10
DM-10065: Client side Rotation and Flipping

robyww:
Branch covers the following:

 - Uses html5 canvas to render a rotated or flipped  image
 - package.json: added transformation-matrix-js
 - package.json: added point-in-polygon
 - new TileDrawerCanvas replaces TileDrawer
 - thumbnail working
 - new DevicePt coordinate system
 - removed ViewportPt coordinate system
 - drawing layers mostly working
 - select and selection related operations working with rotation
 - select now has a second box that shows when image is rotated, this is used for cropping and image stats
 - rotate dialog changed to use new rotate, flip button uses new flip
 - WCS match working
 - target match working
 - Time series working
 - Compass working, required adding more logic to work with rotate
 - code cleanup of almost all  the affected files
 - removed some old unused code.

_Testing_:
Here are some functions that need special attention.

 - wcs match
 - target match
 - time series viewer
 - image the are both flipped and rotated
 - compass
 - large zoomed images - zoom it the flip and rotate
 - looking at the thumbnail view
 - magnifier (I don’t think it is right)
 - catalog layers - then rotate
 - testing with images that not dss, wise, or m31

_Known Issues that will be handle in separate ticket_:

 - footprint rotate handles off when rotated, see:  https://jira.lsstcorp.org/browse/DM-10188
 - marker resize handles off when rotated, see:  https://jira.lsstcorp.org/browse/DM-10188
 - png download will not show a rotated or fliped file. - _fixed in second commit on this branch_


robyww:
Files to review by reviewer:

@loitly :
 - package.json
 - visualize/iv/PlotTitle.jsx
 - visualize/iv/TileDrawer.jsx
 - visualize/iv/TileDrawerCanvas.jsx
 - visualize/iv/TileDrawHelper.jsx
 - visualize/iv/EventLayer.jsx
 - visualize/iv/ExpandedTools.jsx
 - visualize/iv/ImageViewerDecorate.jsx
 - visualize/ui/ThumbnailView.jsx
 - visualize/reducer/HandlePlotChange.js
 - visualize/reducer/HandlePlotCreation.js

@cwang2016 :
 - visualize/draw/DirectionArrowDrawObj.js
 - visualize/draw/Drawer.js
 - visualize/draw/FootprintObj.js
 - visualize/draw/MarkerFootprintObj.js
 - visualize/draw/PointDataObj.js
 - visualize/draw/SelectBox.js
 - visualize/draw/ShapeDataObj.js
 - drawingLayers/Catalog.js
 - drawingLayers/FootprintTool.js
 - drawingLayers/MarkerTool.js
 - drawingLayers/NorthUpCompass.js
 - drawingLayers/SelectArea.js
 - visualize/CsysConverter.js

@tgoldina :
 - visualize/ChangePrime.js
 - visualize/ImagePlotCntlr.js
 - visualize/ImageViewerLayout.jsx
 - visualize/PlotPostionUtil.js
 - visualize/PlotState.js
 - visualize/Point.js
 - visualize/WebPlot.js
 - visualize/reducer/PlotView.js
 - visualize/task/WcsMatchTask.js
 - visualize/saga/MouseReadoutWatch.js
 - visualize/VisUtil.js

_Less Important/ Small changes_:
 - visualize/region/RegionDescription.js
 - visualize/RelatedDataUtil.js
 - visualize/ui/VisToolbarView.jsx
 - visualize/task/PlotImageTask.js
 - visualize/task/PlotChangeTask.js
 - ui/FitsRotationDialog.jsx
 - api/ApiUtilImage.jsx
 - core/ExternalAccessUtils.js
 - ui/ZoomOptionsPopup.jsx
 - ui/TargetFeedback.jsx
 - ui/TargetPanel.jsx
 - visualize/saga/ImagePlotter.js
 - visualize/ui/VisCtxToolbarView.jsx
 - visualize/task/ImageOverlayTask.js

_Removed unused files:_
 - ui/ModalDialog.jsx
 - ui/AllDialogsContainer.jsx

ejoliet:
I've built and test randomly the tri-view. So far i've found the following:
- North-up doesn't work if first image is highlighted: i brought 2 images side by side and rotate them to 45 degrees, then apply north-up doesn't change the images

ejoliet:
Another issue is that i've noticed that the footprint rotation is broken.

robyww:
>> Another issue is that i've noticed that the footprint rotation is broken.

The was noted in _known issues_ at the top of the pull request.  There is a list 3 issues we will do in separate tickets.

src/firefly/js/visualize/draw/FootprintObj.js
```diff
@@ -210,48 +209,51 @@ function ptSegDistSq(x1, y1, x2, y2, px, py) {
...
+	let pt0;
+	let pt;
+	const {color,lineWidth} = drawParams;
+	if (!onlyAddToPath) DrawUtil.beginPath(ctx,color,lineWidth, renderOptions);
 	for (wpt of footprint) {
-		pt0 = plot.getViewPortCoords(wpt0);
-		pt = plot.getViewPortCoords(wpt);
+        pt0 = plot ? plot.getDeviceCoords(wpt0) : wpt0;
+        pt = plot ? plot.getDeviceCoords(wpt) : wpt;
 		if (!pt0 || !pt) return;
```
cwang2016: should here return with 'if (!onlyAddToPath) DrawUtil.stroke(ctx);'?

ymeiymei:
Found a bug: After rotation, the restore key can't rotate the image back to the default. How to repeat:
External image, wise, target: 9.6 -1.1, cutout 0.15deg, band 1. Rotate the image by 45deg, then restore it to the defaults. Fail.

robyww:
>> Found a bug: After rotation, the restore key can't rotate the image back to the default.
Good catch. I did not even think of restore

ymeiymei:
Found another bug: Now the rotation is done by the given angle relative to the original position. For example, rotate 45deg, then rotate 90deg, the result is only rotating 90 deg. Or rotate 45 deg, then another 45 deg, the image has no change for the second rotation. Before the rotation is relative to the current position.

cwang2016:
It seems that the rotate angle entered into the  'Rotate Image' popup window means the angle relative to the original image. Will that be good to have some number shown on the popup window indicating the current rotate angle, then the user could enter the rotation angle in addition to current rotate angle?

robyww:
>>  Will that be good to have some number shown on the popup window indicating 

It is already showing on the title. We could do it in the dialog as well but I think we might want the to be another ticket.

robyww:
>> Found another bug: Now the rotation is done by the given angle relative to the original position. 

The new way is more consistent with DS9.  I think it is the right way to go.

src/firefly/js/visualize/ui/ThumbnailView.jsx
```diff
@@ -39,6 +41,14 @@ export class ThumbnailView extends Component {
         return this.drawData;
     }
 
+    onImageLoad(e) {
+        if (e) {
+            e.onload= () => {
+                this.setState( () => ({imWidth: e.width, imHeight:e.height }));
```
robyww: Loi and I just talked about this.  This is a new way to setState in react.

tgoldina:
Indeed. From
https://facebook.github.io/react/docs/react-component.html#setstate

*The first argument can be an object (containing zero or more keys to
update) or a function (of state and props) that returns an object
containing keys to update.*
Though in your case, is there an advantage of using a function rather than
an object?

2017-04-11 16:11 GMT-07:00 Trey Roby <notifications@github.com>:

robyww:
No advantage for what I was doing it.  However, for the next major release of react they are encouraging this to be the main way of setting state.  There is some performance optimization this helps enable.

cwang2016:
review completes. 
- Find some issue for region rendering. If a region is defined on image domain, the rotation angle (and flip) should be applied to the shape rendering. I will look into this issue and fix it in another ticket. 
- Could 'rotate angle' be a negative number?

robyww:
>> Could 'rotate angle' be a negative number?

Yes, I will look into it.

ymeiymei:
More issues:

* Not all the footprints can be rotated by the image rotation (WFIRST can).
*The footprints can not be rotated by hand more than once.
*The rotation direction now is clockwise but IRSA's convention is counter-clockwise. See detail in IRSA's help:
"Rotating the image to any angle
This feature allows you to rotate the image to any angle of your choice, in degrees. It will rotate the image counter-clockwise (to the left) from the current view, not necessarily the original image. For example, entering "45" in the rotation pop-up and hitting "rotate" will rotate the image 45 degrees counter-clockwise relative to its original orientation. Then selecting the icon again, and entering "180" in the pop-up (followed by hitting "rotate") will rotate the image an additional 180 degrees counter-clockwise. To exit the pop-up without making further changes, hit the blue 'x' in the upper right of the pop-up."

*Need further discussion about the rotation be relative to the original or additional to the current position. In other words, we should follow DS9 way or IRSA way (see IRSA help above).

Continue to test ...

ymeiymei:
Also the rotation angle should be between 0 and 360 degrees. Negative angle is not allowed. Input a negative angle, one should see the error message.

robyww:
Yi wrote:
>> Not all the footprints can be rotated by the image rotation 

Footprints are being handle in another ticket. see: https://jira.lsstcorp.org/browse/DM-10188

>> rotation direction now is clockwise but IRSA's convention is counter-clockwise

Will fix this one.

ymeiymei:
The post-rotation cropping frames are confusing: one dashed and one solid. It seems the cropping is done by the dashed one. Do we need the solid frame? See attached image.
screen shot 2017-04-12 at 10 18 42 am

robyww:
>> The post-rotation cropping frames are confusing: one dashed and one solid. It seems the cropping is done by the dashed one. Do we need the solid frame? See attached image.

>> I guess the solid frame is for showing the user's mouse dragging?

I am open to suggestions. We need to crop on the original image.  Therefore when you make a selection now there are some actions that can happen on the selected area like selecting points but crop cropping need to be in original image bounds.

ymeiymei:
Post rotation flipping issue:

After rotation, the flip is done on the north-south direction (if the image is north up originally ) or the original up-down direction. A bug? 

In IRSA, the flip is alway on the Y axis (up and down direction) before or after rotation.

robyww:
>> After rotation, the flip is done on the north-south direction (if the image is north up originally ) or the original up-down direction. A bug?

This way is more consistent with DS9. I think firefly was doing it wrong before. I am not sure what others do but I think users will expect it to be more like how DS9 does it.

ymeiymei:
Before when multiple images are locked as a group, rotation is still done for the selected image only. 

This version the rotation is done for the whole group which locked. New feature of bug?

ymeiymei:
After loading some images, right click and select "Reload", this is for reloading the images (IRSA wise way) or going back to search (the current version)? Even it's for going back to search, the search panel's "Cancel" key is not working.

robyww:
>> This version the rotation is done for the whole group which locked. New feature of bug?

The way we did it before has been criticized.  We only did it that way because it was so CPU on the server. Now it seems to make sense to keep the group lock like everything else.

src/firefly/js/visualize/PlotState.js
```diff
@@ -141,26 +141,22 @@ export class PlotState {
...
      * @return {boolean}
      */
     isFlippedY() { return this.flippedY; }
 
-    /**
-     *
-     * @return {Number}
-     */
-    getRotationAngle() { return this.rotationAngle; }
+    // getRotationAngle() { return this.rotationAngle; }
```
tgoldina: How should can we get the rotation angle now? (I know we rely on PlotState a lot in API.)

src/firefly/js/visualize/Point.js
```diff
@@ -22,7 +22,17 @@ var Point = {  SPT, IM_PT, IM_WS_PT, VP_PT, PROJ_PT, W_PT, OFFSET_PT};
  *
  * @prop {Number} x
  * @prop {Number} y
- * @prop {String} type one of 'ScreenPt', 'ImagePt', 'ImageWorkSpacePt', 'WorldPt', 'ViewPortPt', 'ProjectionPt', 'OffsetPt'
+ * @prop {String} type one of 'RenderedPt', 'ScreenPt', 'ImagePt', 'ImageWorkSpacePt', 'WorldPt', 'ViewPortPt', 'ProjectionPt', 'OffsetPt'
```
tgoldina: Is ViewPortPt completely gone? It is still referenced here.

src/firefly/js/visualize/VisUtil.js
```diff
@@ -734,7 +750,131 @@ export function formatFluxValue(value) {
...
+        cc.getDeviceCoords(makeImagePt(1, plot.dataHeight))
+    ];
+
+
+    const polyPtsAsArray= polyPts.map( (p) => [p.x,p.y]);
+
+    return testPts.every( (p) => pointInPolygon([p.x,p.y], polyPtsAsArray));
+}
+
+export function findIntersectionPt(x1, y1, x2, y2, x3, y3, x4, y4) {
```
tgoldina: We need jsdoc or comments here to tell which pairs of (x,y) make first and second segments

src/firefly/js/visualize/saga/MouseReadoutWatch.js
```diff
@@ -235,15 +241,22 @@ function showSingleBandFluxLabel(plot, band) {
...
-        .then((result) => {
-            return result;
-        })
-        .catch((e) => {
-            console.log(`flux error: ${plotView.plotId}`, e);
-            return ['', '', ''];
-        });
+    const plot= primePlot(plotView);
+    const {x,y}= iPt;
+    if (x >=0 && y >= 0 && x < plot.dataWidth && y < plot.dataHeight) {
```
tgoldina: What is this check for? To make sure it's a valid image point?

src/firefly/js/visualize/ImagePlotCntlr.js
```diff
@@ -99,12 +95,14 @@ const ROTATE_START= `${PLOTS_PREFIX}.RotateChangeStart`;
 const ROTATE= `${PLOTS_PREFIX}.RotateChange`;
 const ROTATE_FAIL= `${PLOTS_PREFIX}.RotateChangeFail`;
```
robyww: I will remove the old constants

src/firefly/js/visualize/ImageViewerLayout.jsx
```diff
@@ -270,59 +259,58 @@ ImageViewerLayout.propTypes= {
...
+    ];
+
+    let found=  devAsScreenAry.some( (pt) =>
+                   contains(0,0,screenSize.width, screenSize.height,pt.x,pt.y)  ||
+                   intersects(0,0,screenSize.width, screenSize.height,pt.x,pt.y,1,1));
+
+    if (!found) {
+        found= screenAsDevAry.some( (pt) => contains(0,0,viewDim.width, viewDim.height,pt.x,pt.y) ||
+                                            intersects(0,0,viewDim.width, viewDim.height,pt.x,pt.y,1,1));
+    }
```
robyww: I will write a description.  I had not because it took me awhile to convince myself the concept worked.

src/firefly/js/visualize/PlotState.js
robyww: plotView.rotation and plotView.flipY

src/firefly/js/visualize/Point.js
robyww: yes, I will clean up the docs

src/firefly/js/visualize/VisUtil.js
robyww: I will document it.

src/firefly/js/visualize/saga/MouseReadoutWatch.js
robyww: is is support to do the same as line 140.  I meant to fix both. It check to see it the image point is on the plot. It should call `pointInPlot`

src/firefly/js/visualize/PlotPostionUtil.js
```diff
@@ -0,0 +1,131 @@
...
+
+
+
+
+
+
+
+
+
+
```
robyww: good idea.  This will was a work-in-progress for a long time. I will give it a better name.

ymeiymei:
I found sometimes the rotation button has no response when I click it. This happened after rotation, zoom, crop, was/target match, etc. 

I can't repeat the problem. See the error messages I got in the attached file.

[rotationButtonNoResponseErr.txt](https://github.com/Caltech-IPAC/firefly/files/920278/rotationButtonNoResponseErr.txt)

robyww:
I see the place in the code but I am not sure how it could get there.  If you could figure our how to repeat is would help. I wonder if you have a series of exceptions and it is really the first exception that is creating this problem.

src/firefly/js/visualize/ImagePlotCntlr.js
robyww: I think I will make the just to be ROTATE and FLIP

ymeiymei:
Sorry I tried but I really can't repeat the problem. 

I think I am done with this version's testing. Found some minor issues but I rather to double check them in next version. They are hard to repeat also.

ymeiymei:
Completed the test on this new version. This version fixed a lot.  My test report:
1)Rotation and flip can be restored. 
2)All rotations are relative to the original position (New feature).
3)Flip is on the original Y axis, even after rotation, no matter where North is. (New feature)
4)Rotation direction is fixed to counter-clockwise now.
5)Post rotation cropping: TBD.
6)A bug(?) :   When right click an image and select "Reload", the current page should be reloaded (that is the Chrome reload), but instead, the image search panel pops up. Also this search panel can't be cancelled.

PR #349, opened 2017-04-11
Irsa 142-WCS match behavior is odd

lznakano:
To test it,

1. build firefly
2. build irsaviewer in ife
3. Please follow the procedures in the video Louisa created (https://youtu.be/B-DfHOtWFGo)

For some reason, the image  from irsaviewer is not the same as the one from the irsatest.  I attached the old image in the ticket. 

Thanks for testing!

### DIFF
--- a/src/firefly/js/visualize/ui/ImageSelectPanelResult.js
+++ b/src/firefly/js/visualize/ui/ImageSelectPanelResult.js
@@ -6,7 +6,7 @@ import {keyMap, panelKey, computeCurrentCatalogId, rgbFieldGroup, getTabsIndexes
         PLOT_NO, RED, GREEN, BLUE, PLOT_CREATE, PLOT_CREATE3COLOR, rgb} from './ImageSelectPanel.jsx';
 import WebPlotRequest from '../WebPlotRequest.js';
 import {dispatchPlotImage, visRoot } from '../ImagePlotCntlr.js';
-import {findViewerWithItemId, getMultiViewRoot, IMAGE} from '../MultiViewCntlr.js';
+import {findViewerWithItemId, getViewer, getMultiViewRoot, IMAGE} from '../MultiViewCntlr.js';
 import {parseWorldPt} from '../Point.js';
 import {getPanelCatalogs} from './ImageSelectPanelProp.js';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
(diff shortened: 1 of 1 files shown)


wmiipac:
the correct steps should be:
*checkout IRSA-142 branch in firefly
*checkout rc branch in ife
*build irsaviewer in ife, and follow Luisa's youtube 

I built Atlas with irsaviewer(with Lijun's fix) point to my localhost
http://irsadev.ipac.caltech.edu:9072/data/SPITZER/C2D/

the wcs match problem is  fixed. 
There seem to be a problem when loading image from the irsaviewer, the image size is being ignored, the loading image panel had default size for wise set at 0.15 degree, see attached screenshot,  but the image loaded is full size. irsatest has the correct wise image size.
screen shot 2017-04-11 at 10 50 29 am

src/firefly/js/visualize/ui/ImageSelectPanelResult.js
```diff
@@ -308,8 +308,17 @@ export function resultSuccess(plotInfo, hideDropdown = false) {
...
 
-        if ((plotInfo.addPlot&create) && plotInfo.viewerId) { // create plot case: only for fits viewer
-            groupId = plotInfo.viewerId;
+        if ((plotInfo.addPlot&create) && plotInfo.viewerId) { // create plot case: only for fits
+            //IRSA-142 LZ 4/07/17
+            //When the image from different groups, the wcsMatch does not work since the match only matches the image within the same group.
+            //If the group id exists, add the image into the same group
+            var vr = visRoot();
+            if (vr.activePlotId && vr.plotGroupAry[0]){
+               groupId =  vr.plotGroupAry[0].plotGroupId;
```
robyww: This is not correct, you want to do the following:
```
 if ((plotInfo.addPlot&create) && plotInfo.viewerId) {
       groupId = plotInfo.viewerId;
       const viewer = getViewer(getMultiViewRoot(), plotInfo.viewerId);

       if (viewer && viewer.itemIdAry[0]) {
           const pv = getPlotViewById(visRoot(), viewer.itemIdAry[0]);
           if (pv) {
               groupId = pv.plotGroupId;
           }
       }
} 
else {
    ...

Right now you are getting just the first plotGroupId that is a risk. This way it will get the first PlotView in the active viewer and then get the plotGroup from that.


`getActivePlotView` is imported from `PlotViewUtil.js`
`getViewer`, `getMultiViewRoot` is imported from `MultiViewCntrl.js`

lznakano: Thank you for pointing out the potential risk.  I will make the changes as you recommended.

PR #350, opened 2017-04-12
DM-9944 XYPlot with plotly library

tgoldina:
- Converted XYPlot into XYPlotHighcharts and XYPlotly implementations.
- Added lost click handling, to be able to select plotly chart from multi-chart viewer by a mouse click.


src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -111,8 +118,12 @@ export class PlotlyWrapper extends React.Component {
 
     render() {
         const {style}= this.props;
+        // note: wrapper div is the target for the simulated click event
+        // when the original click event is lost and plotly_click is emitted instead
```
tgoldina: > For the density, when you hover over the active point, it loses the depth (3rd) line.

It's not exactly a bug. We point the point from the highlighted row, which might not be present in the decimated data. Though I can still find a bin to which the highlighted point belongs and add "represents".

PR #351, opened 2017-04-13
DM-9395 Showing histogram on external viewer 

cwang2016:
This development fixes the following: 
- If the external viewer shows charts only, 'Charts' drop-down button is not working to make another histogram since there is no active table from the viewer. 
- the 'close' button is removed from the external viewer if only charts are on the external viewer. 
  the 'close' button exists when the chart is expanded from an external viewer with table, image and charts. Clicking the 'close' button, the page will go back to the original external viewer. 

test:
go to the demo page: 

click on 'show histogram'. an external viewer opens a new tab with a histogram chart on the page.

### DIFF
--- a/src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
+++ b/src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
@@ -129,12 +129,11 @@ function onAnyAction(layoutInfo, action, views) {
 
 function handleNewTable(layoutInfo, action) {
     const {tbl_id} = action.payload;
-    var {images={}, showImages, showTables, hasTables} = layoutInfo;
+    var {images={}, showImages, showTables} = layoutInfo;
     var {coverageLockedOn, showFits, showMeta, showCoverage, selectedTab, metaDataTableId} = images;
     const isMeta = isMetaDataTable(tbl_id);
-    const isDisplayTable = (hasTables && showTables);
 
-    if ((isMeta || isCatalogTable(tbl_id)) && isDisplayTable  ) {
+    if ((isMeta || isCatalogTable(tbl_id)) && showTables ) {
         if (!showFits) {
             // only show coverage if there are not images or coverage is showing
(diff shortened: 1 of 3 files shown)


src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
```diff
@@ -129,11 +129,12 @@ function onAnyAction(layoutInfo, action, views) {
 
 function handleNewTable(layoutInfo, action) {
     const {tbl_id} = action.payload;
-    var {images={}, showImages} = layoutInfo;
+    var {images={}, showImages, showTables, hasTables} = layoutInfo;
     var {coverageLockedOn, showFits, showMeta, showCoverage, selectedTab, metaDataTableId} = images;
-
     const isMeta = isMetaDataTable(tbl_id);
-    if (isMeta || isCatalogTable(tbl_id)) {
+    const isDisplayTable = (hasTables && showTables);
```
loitly: I don't think isDisplayTable is needed because that's what showTables is supposed to represent.  showTables should never be true unless hasTables is true.

loitly:
Changes look fine.  Good to merge.

ejoliet:
I've built the branch locally and when i click on the link {{show histogram}} the external viewer launched but shows the image search dialog and no histogram is shown. Is it only me?
(I'm using Safari {{Version 10.1 (12603.1.30.0.34)}} on OSX Sierra)

ejoliet:
Ok, got it. Now that I can see external viewer , I've played around. One problem is when I click the 'charts' button. It sent the user to an error message page but there is no way I can go back to the histogram. If the drop down doesn't work or not usable, should it be removed? Or adds way to go back.

cwang2016:
Emmanuel wrote: 
 It sent the user to an error message page but there is no way I can go back to the histogram.

Resolve this issue by adding a 'OK' button on the dropdown which shows the error message. You may go back to the viewer by clicking the 'OK' button.

PR #352, opened 2017-04-13
DM-10208 Resolve histogram rendering when x is set to be 'log' on Chart dropdown window

cwang2016:
This development: 
- blank values of fields MIn, Max, Bin width on 'Charts' select dropdown window when histogram, a single column (not column expression) and x log are selected. 

test:
-localhost:8080/firefly
-do a catalog search and get tri-view page
-click 'Charts' to get a dropdown window
-select histogram, a single column and select 'log' for Options/X. 
  => fields Min, Max and Bin Width become blank. 
- click OK to get a histogram chart with x on log scale.

### DIFF
--- a/src/firefly/js/charts/ui/HistogramOptions.jsx
+++ b/src/firefly/js/charts/ui/HistogramOptions.jsx
@@ -87,7 +87,7 @@ var columnNameReducer= (colValStats) => {
                 case 'columnOrExpr':
                     const colName = action.payload.value;
                     if (colName ) {
-                        if (isSingleColumn(colName, colValStats)) {
+                        if (isSingleColumn(colName, colValStats) && get(inFields, ['x', 'value'], '') !== 'log') {
                             for (var i=0; i<colValStats.length; i++){
                                 if (colName=== colValStats[i].name) {
                                     const dataMin = colValStats[i].min;
@@ -144,6 +144,16 @@ var columnNameReducer= (colValStats) => {
                         }
                     }
 
+                    break;
+                case 'x':
+                    const xVal = get(inFields, ['x', 'value'], '');
+
+                    if (xVal && xVal.includes('log')) {
+                        inFields = updateSet(inFields, ['binWidth', 'value'], '');
+                        inFields = updateSet(inFields, ['minCutoff', 'value'], '');
+                        inFields = updateSet(inFields, ['maxCutoff', 'value'], '');
+
+                    }
                     break;
                 default:
                     break;
@@ -374,7 +384,7 @@ function renderFixedBinSizeOptions(groupKey, histogramParams, disabled){
          <RadioGroupInputField
             initialState= {{
                                 value: get(histogramParams, 'fixedBinSizeSelection', 'numBins'),
-                                tooltip: 'Please select number of bins or bin width',
+                                tooltip: 'Please select number of bins or bin width'
                                 //label: 'BinSize:'
                             }}
             options={binSizeOptions}
@@ -387,8 +397,7 @@ function renderFixedBinSizeOptions(groupKey, histogramParams, disabled){
                  initialState= {{
                                   value: get(histogramParams, 'numBins', '50'),
                                   validator:Validate.intRange.bind(null, 1, 500, 'numBins'),
-                                  tooltip: 'Number of bins',
-
+                                  tooltip: 'Number of bins'
                              }}
                  disabled = {disabled}
                  fieldKey='numBins'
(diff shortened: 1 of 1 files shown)


src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -145,6 +145,16 @@ var columnNameReducer= (colValStats) => {
                     }
 
                     break;
+                case 'x':
+                    const xVal = get(inFields, ['x', 'value'], '');
+
+                    if (xVal && xVal.includes('log')) {
```
loitly: xVal && is not necessary because it will never be undefined.

loitly:
The comments left for code is nitpicking and does not need to be corrected.
However, the fix to this ticket, in my option, is not ideal.  It's true the value are blank-out.  But, a user can still enter values which will still produce the same errors that this ticket raised.  I think if the fields are not applicable, they should be hidden as well.

ejoliet:
Just a note: I don't if this is the right ticket to deal with but in case of log option selected, min,max and bin width values should come back updated with the right preset values, the same way the linear option is working. Not sure if it's the case though.

cwang2016:
Loi wrote: It's true the value are blank-out. But, a user can still enter values which will still produce the same errors that this ticket raised.

Per Lijun's opinion: If x is selected to be 'log', it is treated equivalently as a column expression ('log(column_name)'), therefore fields of min, max and bin width are set to be blank accordingly. In that case, the user can still input a value (i.e. log of minimum of 'dec' for field 'min' if 'dec' is the target column) to any of those fields. 
Previously the minimum and maximum values of the selected column were shown in min, max fields and that resulted in the wrong return from the server.  This ticket fixes this problem by reentering 'blank' values to min, max and bin width in case x is selected to be 'log'.

lznakano:
Before the request is sent to the server, the client side has an array of table columns where each array item contains the column's name, min and max value of the column.  Thus, if the column is an expression, the min/max are not available and can not be pre-filled.

ejoliet:
I don't think the user expect to see the values cleared out. The expectation is that if i click on 'log' option, i want to see the min/max and bin width updated. Maybe the resolution could be in a different ticket, but clearing the fields is a loss of functionality. Should be improved or fixed.

cwang2016:
I am open to make updates for the specs. you suggest:  
Option 1: If a single column is selected and x is log, then min & max & bin width are updated => this is compliant to the case of single column and linear x. 
Option 2: if a single column is selected and x is log, then set min, max, bin width blank => this is compliant to column expression cases. 

Please, 1, 2, or  others?

ejoliet:
Option 1 looks more consistent. 
If it's an expression, values could be cleared regardless the option selected.

lznakano:
In general, if the histogram column is an expression, the server request is needed to get the min and max values.  However, for log, the client side can calculate the min and max.  If the min <0, change the min=0, or a smallest positive number.  I think that log(singleColumn) can be treated as the single column with little changes in the UI.

tgoldina:
When switching between log and linear scale, it does not matter if we are plotting expression or not. If min, max, and width are populated, we can calculate these values. 

The difficult part is when something is missing. Also, because Javascript is using float, and Java uses double, we might loose precision and cut off the edge points. Also, it's not quite clear what "user expects". It's very possible that when user is switching to log scale, he does not expect the min and max to change.  Not sure about "bin width" though... 

In my opinion, the root of the problem is that server side is not aware of which scale is requested. When user switches to log scale, the client converts the content of the column field into an expression log(x) and when it gets results back, it converts bins back to non-log values.  

If we move this logic to the server, the same min and max should work, no matter which scale is selected. (As of bin width, we can hide it for log scale to avoid confusion.)

Check this link out: http://blogs.sas.com/content/graphicallyspeaking/2014/08/18/histograms-on-log-axis/

loitly:
I agree with Tatiana.  The problem needs to be fixed on the server-side.  Also, in my opinion, min/max should not be affected when one chooses log scale.  That's a presentation preference.

ejoliet:
User expectation is that if i change scale, my value are either the same or log applied. maybe keeping the same values would make more sense.

If you change to log scale, min max value doesn't have the same meaning. 
Min, max value are (min bin range, max bin range) a way to filter the histogram., it's an input to the algorithm chosen. in the future there could be non-unform binning too.
Note that ultimately what will be needed is to give the statistic of the algorithm computed for the histogram: how many data fall outside the min,max range, etc. The story is wider and need to be discussed outside github with scientists input.

PR #353, opened 2017-04-14
IRSA-355: add extra style option to the LC viewer

ejoliet:
This PR is adding additional style to the title so it can be layout on the right position of the icon.
Plus, the title will change depending on the page the user is. @ layouts have 2 different title.
See IRSA-355 for details.

Test: download both branches IFE and Firelfly related to IRSA-355. Build Firefly and IRSAViewer and check both LC viewer:

http://localhost:8080/firefly/ts.html   (doesn't have icon, only title)
http://localhost:8080/irsaviewer/timeseries (have banner icon and title shifted)


lznakano:
It looks good to me.

PR #354, opened 2017-04-15
IRSA-310 improve upload lc panel

ejoliet:
Based on the feedback input in IRSA-310, i've changed the upload panel in LC viewer.

I took also the chance to fix the overlapping `No file chosen` on the `Browse...` button in Firefox.

Please, check that the upload looks ok in different browser and devices sizes


ejoliet:
While trying to fix the label overlap in firefox, i've realized that we are using style `color:transparent` to hide the `No file chosen` by default from `<input type='file'>` and using style `margin-left=-150px` to get the filename label after `Browse` button. 
Is ok for short names but if i choose long names the space between the filename and the button is too much. I would suggest to leave it like this and raise another ticket to improve the `FileUpload` component.

PR #355, opened 2017-04-17
Irsa 311-Update "Result" layout of Time Series Viewer

lznakano:
The changes are only seen in WISE mission.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
@@ -145,14 +145,11 @@ public class IpacTableFromSource extends IpacTablePartProcessor {
             UserCatalogQuery.addCatalogMeta(defaults,columns,request);
         }
         if (request.containsParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
-            if (request.getBooleanParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
 
-                String fileName = ((UploadFileInfo) UserCache.getInstance().get(new StringKey(request.getParam(ServerParams.SOURCE)))).getFileName();
+             String fileName = ((UploadFileInfo) UserCache.getInstance().get(new StringKey(request.getParam(ServerParams.SOURCE)))).getFileName();
 
-                defaults.setAttribute("title",  fileName);
+             ( (TableServerRequest) request).getMeta().put("title", fileName);
 
-
-            }
         }
     }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
@@ -9,7 +9,9 @@ import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.data.table.TableMeta;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.data.FileInfo;
+import edu.caltech.ipac.firefly.server.cache.UserCache;
 import edu.caltech.ipac.firefly.server.query.*;
+import edu.caltech.ipac.firefly.server.util.multipart.UploadFileInfo;
 import edu.caltech.ipac.util.DataType;
 import edu.caltech.ipac.util.FileUtil;
 import edu.caltech.ipac.util.StringUtils;
@@ -24,6 +26,12 @@ import java.net.URL;
 import java.util.List;
 
 
+/**
+ * History
+ * 4/13/17 LZ
+ * IRSA-311
+ *   Add a title meta data based on the setting of  USE_UPLOADED_FILENAME_AS_TABLE_TITLE
+ */
 @SearchProcessorImpl(id = "IpacTableFromSource")
 public class IpacTableFromSource extends IpacTablePartProcessor {
     public static final String TBL_TYPE = "tblType";
@@ -135,6 +143,16 @@ public class IpacTableFromSource extends IpacTablePartProcessor {
         String type = request.getParam(TBL_TYPE);
         if (type == null || type.equals(TYPE_CATALOG)) {
             UserCatalogQuery.addCatalogMeta(defaults,columns,request);
+        }
+        if (request.containsParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
+            if (request.getBooleanParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
+
+                String fileName = ((UploadFileInfo) UserCache.getInstance().get(new StringKey(request.getParam(ServerParams.SOURCE)))).getFileName();
+
+                defaults.setAttribute("title",  fileName);
+
+
+            }
         }
     }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
@@ -146,9 +146,9 @@ public class IpacTableFromSource extends IpacTablePartProcessor {
         }
         if (request.containsParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
 
-             String fileName = ((UploadFileInfo) UserCache.getInstance().get(new StringKey(request.getParam(ServerParams.SOURCE)))).getFileName();
+                String fileName = ((UploadFileInfo) UserCache.getInstance().get(new StringKey(request.getParam(ServerParams.SOURCE)))).getFileName();
 
-             ( (TableServerRequest) request).getMeta().put("title", fileName);
+                defaults.setAttribute("title",  fileName);
 
         }
     }
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
@@ -9,9 +9,7 @@ import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.data.table.TableMeta;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.data.FileInfo;
-import edu.caltech.ipac.firefly.server.cache.UserCache;
 import edu.caltech.ipac.firefly.server.query.*;
-import edu.caltech.ipac.firefly.server.util.multipart.UploadFileInfo;
 import edu.caltech.ipac.util.DataType;
 import edu.caltech.ipac.util.FileUtil;
 import edu.caltech.ipac.util.StringUtils;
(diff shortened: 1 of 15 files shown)


ejoliet:
Before i look more deep into the changes, just to remind you that the ticket is about changing all the cases (at least for IRSA release) which are 'WISE' and 'Other', not only the 'WISE' case. Are you going to create another ticket for the 'Other' case?

ejoliet:
Plus, the PR has to be requested against `RC` not `dev`. Please change by editing the base branch for this PR. thanks.

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -230,24 +230,27 @@ class SettingBox extends Component {
...
+                  </div>
+
+              </div>
+              <div >
+                <CompleteButton
+                     style={{ width:'80px'}}
+                     groupKey={groupKey}
+                     onSuccess={setViewerSuccess(periodState)}
+                     onFail={setViewerFail()}
+                     text={'Period Finder'}
```
ejoliet: Can you add the ellipsis at the end of `Finder`: `...` 

The text should looks like `Period Finder...` so it gives the user a hint that is going to move to a different window.

src/firefly/js/charts/ui/XYPlot.jsx
```diff
@@ -47,7 +47,8 @@ export const plotParamsShape = PropTypes.shape({
     nbins : PropTypes.shape({x : PropTypes.number, y : PropTypes.number}),
     shading : PropTypes.oneOf(['lin', 'log']),
     x : axisParamsShape,
-    y : axisParamsShape
+    y : axisParamsShape,
+    plotTitle:PropTypes.string
```
loitly: plotTitle added here.. is it used anywhere?

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -267,28 +272,37 @@ export function* lcManager(params={}) {
...
+function updateRawTableChart(timeCName, fluxCName, mission='') {
     var chartX = get(getChartDataElement(LC.RAW_TABLE), ['options', 'x', 'columnOrExpr']);
     var chartY = get(getChartDataElement(LC.RAW_TABLE), ['options', 'y', 'columnOrExpr']);
 
     if (chartX === timeCName && chartY === fluxCName) return;
 
     if (timeCName && fluxCName) {
-        const xyPlotParams = {x: {columnOrExpr: timeCName}, y: {columnOrExpr: fluxCName, options: 'grid,flip'}};
+
+        const  title = (mission && (mission.toLowerCase()==='wise' || mission.toLowerCase()==='other') )? 'Input Data':'';
```
loitly: why is 'wise' hard-coded here?  it should not.  this logic needs to be moved into the converter.  that's why they are there.  this applies to all of the 'wise' and 'other' references in this file.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -133,8 +135,11 @@ export function getConverterData(layoutInfo=getLayouInfo()) {
     const converterId = get(layoutInfo, [LC.MISSION_DATA, LC.META_MISSION]);
     return converterId && getConverter(converterId);
 }
+export function getConvertId(layoutInfo=getLayouInfo()) {
+    return get(layoutInfo, [LC.MISSION_DATA, LC.META_MISSION]);
```
loitly: this logic is used in many places and is referred to as converterId.  if this is a convenience function, you should name is as getConverterId, not getConvertId.  changing all other places to use this function may be a good idea if it's truly needed.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -33,11 +32,13 @@ export function uploadPhaseTable(tbl, flux) {
                                      });
 
         dispatchTableSearch(tReq, {removable: true});
-
+        const mission =get(getLayouInfo(), [LC.MISSION_DATA, LC.META_MISSION]);
+        const  plotTitle = (mission && (mission.toLowerCase()==='wise' ||mission.toLowerCase()==='other') )? 'Phase Folded Data':'';
```
loitly: again, use converter instead.

src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
```diff
@@ -136,6 +144,16 @@ public void prepareTableMeta(TableMeta defaults, List<DataType> columns, ServerR
         if (type == null || type.equals(TYPE_CATALOG)) {
             UserCatalogQuery.addCatalogMeta(defaults,columns,request);
         }
+        if (request.containsParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
+            if (request.getBooleanParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
```
lznakano: Only the request for Wise and Other mission has the ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE set.  For generic and LSST, the request does not contain this parameter.

loitly: I know.. in which case, getBooleanParam() will return false and you only care when it's true.

src/firefly/js/templates/lightcurve/LcManager.js
ejoliet: Agree. This need to change. We discussed yesterday the topic. Lijun will work on that refactoring.

Let's be clear(er) here, the agreement from the CCB was to change what the ticket explains to all the cases in LC. Yesterday I talked to Lijun so she can _make the common code not mission specific_, and only keep what is specific to mission **separately** in the already existing files which are used to control images and layout specific mission control panel. There shouldn't be any specific code in `LcManager.jsx`, `LcResult.jsx` and `LcViewer.jsx`, that was the original design/ideas. 
But, Lijun needs to add specific code that will deal with the change request so she can handle the specific mission as needed.
That said, i'm not sure she will have time to change also `LSST` and `Generic` case but as is not part of the release, that could be done anytime, not necessarily now unless she managed to do it or somebody have time to do it by end of this month. 
_Bottom line_: The code she is adding, to support the changes, shouldn't affect the other mission specific cases as they are currently but it has to support the changes for WISE and Other at very least and for sure the near **future cases such as PTF**. 

If you need to discuss further, please come to me.

src/firefly/js/charts/ui/XYPlot.jsx
lznakano: Yes, that how the plot title is passed into the XYPlot.

src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
```diff
@@ -56,15 +56,17 @@ public static JSONObject toJsonTableModel(DataGroupPart page, TableServerRequest
...
+      /*  if (request.containsParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE) &&
                 request.getBooleanParam(ServerParams.USE_UPLOADED_FILENAME_AS_TABLE_TITLE)){
 
                 tableModel.put("title", page.getTableDef().getAttribute("title").getValue());
 
         }
         else {
             tableModel.put("title", page.getData().getTitle());
-        }
+        }*/
```
loitly: The code was added by this PR.  Then it was deemed not necessary.  Why would you not removed it, but instead leaving it commented out?  How would this block of code ever get cleaned up?

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -273,30 +274,34 @@ export function* lcManager(params={}) {
...
 
     if (chartX === timeCName && chartY === fluxCName) return;
 
     if (timeCName && fluxCName) {
 
-        const  title = (mission && (mission.toLowerCase()==='wise' || mission.toLowerCase()==='other') )? 'Input Data':'';
+        const rawTable = getTblById(LC.RAW_TABLE);
+
+        //const  title = (mission && (mission.toLowerCase()==='wise' || mission.toLowerCase()==='other') )? 'Input Data':'';
+        const title = getConverter(converterId).showPlotTitle? rawTable.request.META_INFO.title :'';
```
loitly: The intent of the converter is to provide dynamic data so that the core(common) logic does not need to know where/how the data came from.  So, if the logic is to display a plot title when available, shouldn't there be a converter.getPlotTitle().  Then, you display it when not empty.  But instead, it requires that the title comes from meta_info of the raw_table's request which is used as the table's title.  What if you want them to be different or you want to display something else?  Anyways, it's working, so you can leave it as is if you like.

src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -273,30 +274,34 @@ export function* lcManager(params={}) {
...
+function updatePhaseTableChart(flux, converterId) {
     var chartY = get(getChartDataElement(LC.PHASE_FOLDED), ['options', 'y', 'columnOrExpr']);
 
     if (chartY === flux) return;
 
     if (flux) {
 
-        const title= (mission && (mission.toLowerCase()==='wise' || mission.toLowerCase()==='other'))?'Phase Folded Data':'';
+       // const title= (mission && (mission.toLowerCase()==='wise' || mission.toLowerCase()==='other'))?'Phase Folded Data':'';
+        const title = getConverter(converterId).showPlotTitle?'Phase Folded Data':'';
```
loitly: You see what I mean?  What if other missions want to show something other than 'Phase Folded Data'?

src/firefly/js/templates/lightcurve/LcManager.js
ejoliet: For the records, we discussed that earlier today. 
I agree with Loi. 
We agree that Lijun will change and add getPlotTitle callback in the converter so any mission can pass in any plot title.

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -250,13 +251,23 @@ UploadPanel.defaultProps = {
     name: 'LCUpload'
 };
 
+export function getUploadedFileName(){
+    var fields = FieldGroupUtils.getGroupFields(vFileKey);
+    const displayName = fields.rawTblSource.displayValue.split('\\');
+    return ( displayName && displayName[displayName.length-1])?displayName[displayName.length-1]:'';
```
loitly: Using the value from input field as if it's static is not accurate.  The value can be changed.
Try select file 'a', then upload.  Then select file 'b', but click cancel.  Then, click on 'Period Finder' -> cancel, then you will see the uploaded file name is now 'b' when it should have been 'a'.

src/firefly/js/templates/lightcurve/wise/WiseMissionOptions.js
```diff
@@ -110,16 +109,19 @@ export class WiseSettingBox extends Component {
...
 
         return (
             <FieldGroup groupKey={groupKey}
                         reducerFunc={wiseOptionsReducer(missionEntries, generalEntries)} keepState={true}>
 
                 <div >
-                    <div style={{ fontWeight:'bold', display:'inline-block'}} > Column Selection</div>
-                    <div style = {{paddingLeft:'10px', display:'inline-block'}}>{tblModel.tableMeta.title} </div>
-                    <div style = {{fontWeight:'bold',paddingLeft:'82px', display:'inline-block'}}>Images</div>
+                    <div style={{ with:{labelWidth}, fontWeight:'bold', display:'inline-block', margin: '3px 0 6px 0'}} > Column Selection</div>
```
loitly: same here

src/firefly/js/templates/lightcurve/LcViewer.jsx
lznakano: I tried to load two different files and did not notice the problem.  So I chose to do this way. 
 Another way that  I can think of  is to add a parameter in the request.  Then I need to modify the TableUtil.  I will discuss with you tomorrow.  Thanks!

PR #356, opened 2017-04-19
IRSA-337: updating highchart to latest version

ejoliet:
Using latest version of Highcharts and the option to find nearest point fix the problem found by users during the round of LC viewer test with the selection of the period on the periodogram plot.

Test: Bring LC, do a periodogram and check that you can select any peak period (point dots). Previously it was impossible or highly difficult.


tgoldina:
Just a note for for merging in: Highcharts specific code was moved to XYPlotHighcharts.jsx

PR #357, opened 2017-04-19
DM-10149 fix the phase value range of phase folded chart

cwang2016:
the development includes:
- fixes the phase value range on the phase folded curve to be [0, 2)
- plotly based phase folded chart doesn't react properly as the periodogram table is highlighted on a row with invalid period value. 

test:
- start time series upload page, i.e. localhost:8080/firefly/ts.html
- load a time series file
- select 'find period' in the result page, go to period page
- make the zero point time value bigger by adding a couple of digit '0' before the decimal point
- slide the slider to see if the phase falls between 0 and 2
- click 'calculate periodogram'
- select some period value which is either valid or invalid
- check if the period entry and the phase folded chart react correctly. 
(valid period shows a phase folded chart, invalid period shows no chart).

### DIFF
--- a/src/firefly/html/firefly.html
+++ b/src/firefly/html/firefly.html
@@ -21,7 +21,7 @@
                 options : {
                     MenuItemKeys: {maskOverlay:true},
                     catalogSpacialOp: 'polygonWhenPlotExist',
-                    charts: {chartEngine: ''}
+                    charts: {chartEngine: 'plotly'}
                 }
             }
         };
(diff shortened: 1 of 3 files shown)


ymeiymei:
Tested:

- started time series upload page, i.e. localhost:8080/firefly/ts.html
- loaded a time series file (lc_raw.tbl for WISE)
- selected 'find period' in the result page, went to period page
- made the zero point time value bigger by adding a couple of digit '0' before the decimal point
- slid the slider and saw the phase falling between 0 and 2. 
           Suggestion:  For next version, when sliding the period bar, the phase plot has fixed range[0,2].
- clicked 'calculate periodogram'
- selected some period value in the Period chart which is either valid or invalid (within or beyond the period range defined in the period bar)
- checked and found the period entry and the phase folded chart react correctly:
          When the period is valid, the phase plot shows; when the period is invalid (say < 0.001), the phase plot shows nothing and the period in the period selection turns to red.

Pass with one suggestion for future fix.

PR #359, opened 2017-04-21
Irsa 320 disable period change on table active

ejoliet:
This PR is removing the constraint of updating the period value every time the user switch from one active table to another. Plus, it also set no default value for the period.

Check that when you go to calculate periodogram in LC viewer, switching from one table to another back and forth doesn't change the value of the period, only if you click on the plot or the table.
Also noticed that no value is pre-filled in the period input field.

Let me know if you find any problem.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
+++ b/src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
@@ -527,7 +527,7 @@ function periodogramSuccess(popupId, hideDropDown = false) {
             peaks: get(request, [pKeyDef.peaks.fkey]),
             table_name: LC.PEAK_TABLE,
             sortInfo: sortInfoString('SDE', false)                 // sort peak table by column SDE, descending
-        }, {tbl_id: LC.PEAK_TABLE, pageSize: parseInt(peak), noPeriodUpdate: true});
+        }, {tbl_id: LC.PEAK_TABLE, pageSize: parseInt(peak)}); //, noPeriodUpdate: true //see LcManager#getPeriodFromTable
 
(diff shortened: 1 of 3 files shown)


src/firefly/js/templates/lightcurve/LcPeriodogram.jsx
```diff
@@ -563,7 +568,8 @@ function periodogramSuccess(popupId, hideDropDown = false) {
             const xyPlotParams = {
                 x: {columnOrExpr: LC.PEAK_CNAME, options: 'grid'},
                 y: {columnOrExpr: LC.POWER_CNAME, options: 'grid'},
-                plotStyle: 'linepoints'
+                plotStyle: 'linepoints',
+                plotTitle: `First ${get(request, [pKeyDef.peaks.fkey])} Peaks`
```
loitly: I don't see the added plotTitle appearing.

ejoliet: Thanks to catch that. It will appear once IRSA-311 is merged ;-)

ejoliet:
Thanks for reviewing!!

Related to your comment about the highlight and choosing the period, yes, agree, you are right. Will need to figure out in the future how to deal with selection without highlight or a combination of other triggers that we don't expose/have now. 

The other issue wrt the accept/cancel button is true and I've noticed it yesterday while playing in the laptop screen. This is not going to be very friendly in a (small) laptop screen. A ticket has been raised for that issue: [IRSA-396](https://caltech-ipac.atlassian.net/browse/IRSA-396).

I think is good enough to merge now, and address the issues later on.

PR #360, opened 2017-04-24
DM-10188 Fix drawing related issues on markers and footprints with new rotation scheme

cwang2016:
The development includes 
- fixing the bug regarding footprint rotation handles off while rotating
- fixing the bug regarding marker resize handles off while resizing
- rendering the footprints, markers and regions based on the rotation angle of the image. 
- updating footprint and marker rendering on multiple plots. 

test:
do image search 
add footprint or marker onto the image
operate the footprint or marker (rotate/move the footprint or resize/move the marker)
rotate the image before or while operating on the footprint/marker


cwang2016:
Additional fixing:
- When image is flipped the rotation handle on the footprint is fixed
- the center information on drawing layer ui for the footprint is moved to a new line.

PR #363, opened 2017-04-25
IRSA-395:fix linkage between histogram params: Number of bins, bin width, min, max

lznakano:
This implementation fixed the following issues:
  1. User changes Bin width. Number of bins changes. 
  2. User changes Min or Max. Number of bins changes. Bin width does not. 
  3. Fixed bug in Validate to validate the integer at the given range.

### DIFF
--- a/src/firefly/js/charts/ui/HistogramOptions.jsx
+++ b/src/firefly/js/charts/ui/HistogramOptions.jsx
@@ -81,70 +81,64 @@ var columnNameReducer= (colValStats) => {
         }
         let fieldKey = undefined;
         if (action.type === VALUE_CHANGE) {
-            // when column name changes, update the min/max input
             fieldKey = get(action.payload, 'fieldKey');
-            switch (fieldKey){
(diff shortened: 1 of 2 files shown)


src/firefly/js/charts/ui/HistogramOptions.jsx
```diff
@@ -81,70 +81,64 @@ var columnNameReducer= (colValStats) => {
...
-                                inFields = updateSet(inFields, ['binWidth', 'value'], '');
-                            }
-                            break;
-
-                        }
+               //When numBins changes, update binWidth
+                case 'numBins':  {
+                         const numBins = get(inFields, ['numBins', 'value'], 50);
+                         const min = get(inFields, ['minCutoff', 'value'], '');
+                         const max = get(inFields, ['maxCutoff', 'value'], '');
```
lznakano: I don't expect a value unless the column name is a single column.  For expression, those two fields are not empty.

PR #364, opened 2017-04-27
DM-10369: client side mask color change

robyww:
**Changes**:

  - Processing image on the client side to set mask color
  - Add image processing framework that can be used in future of color bar changing
  - Processed images can be added to store as data png
  - include new functions in `WebUtil.js`: `loadImage` and `requestIdleFrame`
  - small clean up of PlotViewUtil

**Code to pay attention to:**

 - `ImageViewerLayout.jsx` - small changes but it definea function that modifies image
 - `ImageRender.jsx` - Image drawing react component - mostly wrapper around CanvasTileDrawer
 - `ImageProcessor.js` - image retrieval and  processing framework
 - `CanvasTileDrawer.js` - Does the actually canvas rendering and managing cached images

**To Test**

 - Use a LSST image that has mask. 
 - Overlay the mask
 - change the mask color - affects will be immediate
 - talk to @robyww is you need the fits files.


src/firefly/js/util/Color.js
```diff
@@ -32,13 +32,13 @@ const colours = {
...
+    const num = parseInt(color[0]==='#' ? color.substr(1) : color, 16);
     return [num >> 16, num >> 8 & 255, num & 255, alpha];
 }
 
 function toRGB(/* String */ color) {
-        var num = parseInt(color, 16);
-        return [num >> 16, num >> 8 & 255, num & 255];
+    const  num = parseInt(color[0]==='#' ? color.substr(1) : color, 16);
+    return [num >> 16, num >> 8 & 255, num & 255];
 }
```
cwang2016: you may also call function getRGBA() in Color.js which handles the color string like '#...', 'rgb(...)', 'rgba(...)' or a pure color name.

src/firefly/js/visualize/ImageViewerLayout.jsx
```diff
@@ -316,19 +317,43 @@ function makeTileDrawers(pv) {
...
 }
 
+
+function makeMaskColorProcessor(colorStr) {
+    return (imageData) => {
+        const cAry= Color.toRGB(colorStr);
+        const {data}= imageData;
+        const len= data.length;
+        for(let i= 0; i<len; i+=4) {
+            if (data[i] || data[i+1] || data[i+2]) {
```
cwang2016: Does this mean if data[i], data[i+1] and data[i+2] are all zero, then the condition for 'if' is 'false'?

robyww: Thanks for making me really think about this. You found a bug, if the user every turns it black then the mask gets stuck at that color.

It should be 
`if (data[i] || data[i+1] || data[i+2] || data[i+3]) { `

However, now that I think about it, in the mask case the only element in the array that matters for detecting if I need to set the color is `data[i+3]` so I can optimize it to

` if (data[i+3]) {`

In the array `data[i], data[i+1], data[i+2]` is a RGB and `data[i+3]` is opacity. The images from the server set all 4 to zero because we want full transparency (no mask bit) and `data[i+3]` to 255 if there is a mask bit. So I can just look at the opacity index in the `if`, which should be more efficient. The image arrays are big so efficiency matters.


## _Thank you!_

cwang2016:
Review completes. The mask color change works well.

src/firefly/js/util/Color.js
robyww: ok, I will use that.

PR #365, opened 2017-04-27
DM-10115 Reset the column select dropdown and column stats on new catalog search

cwang2016:
This development fixes the following issues after starting a new catalog search: 
- The column selection dropdown shows column stats of previous active table
- the added chart shows the content relevant to the previous active table, not current active table. 

test: 
- start a catalog search
- add a chart by clicking 'Charts' button, selecting chart type (Histogram or scatter) and selecting the column name from column select form.
- check if the added chart reflects the selected column and its stats.
- start another catalog search
- repeat the column selection steps to add the chart
- check if the added chart reflects the selected column and its stats from the second catalog search

### DIFF
--- a/src/firefly/js/charts/ui/ColumnOrExpression.jsx
+++ b/src/firefly/js/charts/ui/ColumnOrExpression.jsx
@@ -70,9 +70,11 @@ export function ColumnOrExpression({colValStats,params,groupKey,fldPath,label,la
     };
 
     var val = get(params, fldPath);
-    const onColSelected = (colName) => {
-        val = colName;
-        dispatchValueChange({fieldKey: fldPath, groupKey, value: colName, valid: true});
+    const onColSelected = (colValStats) => {
+        return (colName) => {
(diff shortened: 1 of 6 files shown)


cwang2016:
additional update:
- reset the chart select dropdown content when the active table in the triview panel is changed. 

Note:
- calling the chart select dropdown reset function in saga for firefly viewer manager, this will reset the chart select dropdown when a new table is created or the active table is changed. 
- table column value statistics is changed when there is  active table change. We may attach the table specific statistics with the action payload whenever the relevant field group value change request is dispatched. (it is no need to update the reducer function for that field group, and there seems no bug issue in FiedlGroupCntlr.js')

tgoldina:
I have fixed the bug in FieldGroupCntlr.js which resulted in reducer function not being updated with new colValStats property in HIstogramOptions.jsx and removed the workaround Cindy had: passing colValStats in VALUE_CHANGED payload.

I think, now it works as expected.

PR #368, opened 2017-05-02
IRSA-399 remove x-log options from histogram

ejoliet:
Option in histogram setting panel has been removed for now. See detail in the ticket.
I have kept the code for the (near) future.

Please, confirm that the option is no longer available.


ejoliet:
You are right, this is wrong. I have to remove only 'log'. I will change it! 
Thank you for catching that!

ejoliet:
@tgoldina should be good now.

PR #369, opened 2017-05-02
 DM-6501:Regrid needed  for WebGrid (coordinate grid overlay on image)

lznakano:
1. Refactor codes to impove the performance.
  2. Add LinearInterpolator.js
  3. Add Regrid.js
  4. Use Regrid to make more gride lines and more points in a line
  5. Rotate the label direction according to the angle of the line
     Label location is calcualted at the time when the line angle is calculated.
  6. Re-orgainze and clean up the codes
  7. add the jsdoc info

### DIFF
--- a/src/firefly/js/drawingLayers/ComputeWebGridData.js
+++ b/src/firefly/js/drawingLayers/ComputeWebGridData.js
@@ -703,7 +703,7 @@ function drawLabeledPolyLine (drawData, bounds,  label,  x, y, aitoff,screenWidt
                     var wpt1 = cc.getScreenCoords(ipt0);
                     var wpt2 = cc.getScreenCoords(ipt1);
                     const slope = (wpt2.y - wpt1.y) / (wpt2.x - wpt1.x);
-                    slopAngle = Math.atan(slope) * 180 / Math.PI;
+                    slopAngle =  Math.atan(slope) * 180 / Math.PI;
 
                     //difficult to find a padding to work for all coordinates. Adding deltaX(Y) does not make a big difference
                     labelPoint = makeImageWorkSpacePt(x[i]+deltaX , y[i]+deltaY);
--- a/src/firefly/js/drawingLayers/ComputeWebGridData.js
+++ b/src/firefly/js/drawingLayers/ComputeWebGridData.js
@@ -1,7 +1,14 @@
-/*
+/**
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  * Lijun
  * 4/14/16
+ *
+ * 05/02/17
+ * DM-6501
+ *   Add the Regrid and LinearInterpolator in the utility
+ *   Use the Regrid to adding more gride lines and the number of the points in the line
+ *   Rotate the labels based on the angles
+ *   Cleaned up the codes 
  */
 
 import { makeWorldPt, makeImagePt,makeImageWorkSpacePt} from '../visualize/Point.js';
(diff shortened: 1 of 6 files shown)


ymeiymei:
Tested. 

Yes, the labels are along the grid lines now. Also more grid lines than before (irsawebdev1).

Problem: When rotate the image or flip it, the labels are not along the grid lines any more.

lznakano:
When the rotation is done, only the grid lines are rotated but labels.  The rotation is done outside the WebGrid.  This is the side effect to have the labels with angles other than a fixed direction.

src/firefly/js/visualize/draw/ShapeDataObj.js
```diff
@@ -214,13 +214,14 @@ function makePolygon(ptAry, drawObjAry=null) {
...
- * @param text
+ * @desc make a text
+ *  @param   pt
+ *  @param  text
+ *  @param  rotationAngle - the rotation angle + deg
  * @return {*}
  */
-function makeText(pt, text) {
-    return Object.assign(make(ShapeType.Text), {pts:[pt],text});
+function makeText(pt, text, rotationAngle='0deg') {
```
robyww: pass as undefined for the default

src/firefly/js/visualize/draw/ShapeDataObj.js
```diff
@@ -229,14 +230,15 @@ function makeTextWithOffset(textOffset, pt, text) {
...
-    const fontStyle= drawObj.fontStyle || def.fontStyle || 'normal';
+    var style= drawObj.style || def.style || Style.STANDARD;
+    var lineWidth= drawObj.lineWidth || def.lineWidth || DEF_WIDTH;
+    var textLoc= drawObj.textLoc || def.textLoc || TextLocation.DEFAULT;
+    var unitType= drawObj.unitType || def.unitType || UnitType.PIXEL;
+    var fontName= drawObj.fontName || def.fontName || 'helvetica';
+    var fontSize= drawObj.fontSize || def.fontSize || DEFAULT_FONT_SIZE;
+    var fontWeight= drawObj.fontWeight || def.fontWeight || 'normal';
+    var fontStyle= drawObj.fontStyle || def.fontStyle || 'normal';
+    var rotationAngle = drawObj.rotationAngle||'0deg';
```
robyww: default to  undefined

src/firefly/js/visualize/draw/DrawUtil.js
```diff
@@ -99,12 +99,13 @@ function drawInnerRecWithHandles(ctx, color, lineWidth, inX1, inY1, inX2, inY2)
...
  * @param backGroundColor
  * @param padDing
+ * @param rotationAngle
  */
 function drawText(drawTextAry,text, x,y,color,
                   renderOptions,
                   fontFamily='helvetica', size='9px',
                   fontWeight='normal', fontStyle='normal',
-                  backGroundColor, padDing) {
+                  backGroundColor, padDing,rotationAngle='0deg' ) {
```
robyww: pass as undefined as the default

src/firefly/js/visualize/draw/DrawUtil.js
```diff
@@ -126,7 +127,9 @@ function drawText(drawTextAry,text, x,y,color,
         fontStyle,
         'MozBorderRadius': '1px',
         'borderRadius': '1px',
-        'WebkitBorderRadius': '1px'
+        'WebkitBorderRadius': '1px',
+        transform: 'rotate('+rotationAngle+')'
```
robyww: pass as undefined as the default and then if defined add to to the style like `backGroundColor` on line 135.   This way you don't add unnecessary stuff to the style.

src/firefly/js/util/Interp/Regrid.js
```diff
@@ -0,0 +1,95 @@
...
+    var outData = new Array(outSize);
+
+    const outDataCoord = generateNewCoordinates(inData, outSize, allowExtrapolation);
+
+    const inCoords = Array.from(new Array(inData.length), (x,i) => i);
+    const linearInterpolator = new LinearInterpolator(inCoords, inData, allowExtrapolation);
+
+    for (var i=0; i<outSize; i++){
+        outData[i]=linearInterpolator.getInterpolatedValue(outDataCoord[i]);
+    }
```
robyww: Are you doing a `for` loop to get better performance with a large array. Otherwise you might want to use map.
```
const outData= outDataCoord.map( (c) => linearInterpolator.getInterpolatedValue(c));
```

src/firefly/js/util/Interp/Regrid.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+export function Regrid(inData, outSize, allowExtrapolation=false){
+
+
+    var outData = new Array(outSize);
+
+    const outDataCoord = generateNewCoordinates(inData, outSize, allowExtrapolation);
+
+    const inCoords = Array.from(new Array(inData.length), (x,i) => i);
```
robyww: Never seen this before, cool!

src/firefly/js/util/Interp/Regrid.js
```diff
@@ -0,0 +1,95 @@
...
+
+    if (allowExtrapolation ) {
+        RegridFact = inData.length / outArraySize;
+    }
+    else { //MINUS_ONE
+        RegridFact = (inData.length - 1) / (outArraySize - 1);
+    }
+
+    for (let i=0; i<outArraySize; i++) {
+        coordinate[i] =i * RegridFact;
```
robyww: This could be a map

src/firefly/js/util/Interp/LinearInterpolator.js
```diff
@@ -0,0 +1,157 @@
...
+
+        // determine "x" is ascending or descending
+        var max = ((x0 > xn) ? x0 : xn);
+        for (let i = 1; i < this.n; i++) {
+            if (max > x0) {
+                // check if the data is strictly monotonically increasing
+                if (this.x[i] <= this.x[i - 1]) {
+                    const  x1 = i - 1;
+                    throw new UserException(
+                        'Data must be strictly monotonically increasing or decreasing' + '\n    x['
```
robyww: This should simple be `throw new Error('....');`. Don't define a UserException, by throwing Error you are carrying a stack trace. Same in other places as well.

src/firefly/js/util/Interp/LinearInterpolator.js
```diff
@@ -0,0 +1,157 @@
...
+function UserException(message) {
+    this.message = message;
+    this.name = 'UserException';
+}
+
+/**
+ * LinearInterpolator class
+ */
+export class LinearInterpolator {
+    constructor(x, y, allowExtrapolation) {
```
robyww: I don't think this needs to be a class.  I should be a function closure that returns a function that will do that work (`getInterpolatedValue`). if there is an invariant it could return a null.  The other function could be in the closure since they are private anyway.

src/firefly/js/drawingLayers/ComputeWebGridData.js
```diff
@@ -748,31 +695,40 @@ function drawLabeledPolyLine (drawData, bounds, nLevel0,  label, labelLocations,
...
-                drawData.push(ShapeDataObj.makeLine(ipt0,ipt1));
+                drawData.push(ShapeDataObj.makeLine(ipt0, ipt1));
+
+                //find the middle point of the line, index from 0, so minus 1
+                if (i===Math.round(x.length/2)-1 ) {
+
+                    var wpt1 = cc.getScreenCoords(ipt0);
+                    var wpt2 = cc.getScreenCoords(ipt1);
+                    const slope = (wpt2.y - wpt1.y) / (wpt2.x - wpt1.x);
+                    slopAngle = Math.atan(slope) * 180 / Math.PI;
```
robyww: You could use `cc.getDeviceCoords` to get the real slope  after image rotation. I think that would fix you rotation problem.

lznakano:
The following changes are done:
  1. Fixed the label rotation issues
  2. Changed the codes styles as suggested

ymeiymei:
Tested: WISE, ra=45, dec=89; add grid, flip; 
Result: The labels on the grids are along the lines after flipping but the RA label are upside down. Need to fix.

lznakano:
Label locations are fixed but the distances to the lines are not fixed.  I need Trey's input about this issue.

ymeiymei:
Tested again. After flipping or rotating, the labels are along the grid lines and no upside down problem. Great work!

robyww:
This is looking really nice now. With labels aligning with rotation and angle of the line it look very, very good. Much better than I anticipated.  I am glad you did it.

There are still cases when some of the label are upside down.  To see it:

 - wise, m51, 1.5 deg, rotate 270
 - A couple will be upside down
 - When you flip even more  are upside down.

robyww:
_One more note:_ The speed is very good.

PR #371, opened 2017-05-03
DM-10370: Optimize server memory usage for fits images

robyww:
DM-10370: Optimize server memory usage for fits images

 - Server memory size cut in half
 - Optimized drawing so that user sees working message
 - Optimized scrolling so the it does not block when loading images
 - beta value (used for asinh strech) now only computed when required
 - added a getBetaValue service call
 - Uploading fits files now show masking while uploading
 - removed addColorBand call from server (not used)
 - Added way to push out memory in cache before large data is read in
 - Fixed bug in computing in CsysConvert.pointInPlot with rectangular images

This is a very difficult test it requires jconsole to monitor the memory with large fits images. It could take awhile just to get set up.  Mostly the review should just look over the code.

_Target code to look at:_
 
 - FitsCacher.java - _@lznakano this is the key file to check closely_
 - VisServerOps.java
 - FitsRead.java
 - ImageDataGroup.java

@loitly _would you look at:_
  - EhcacheProvider.java
  - ObjectSizeEngineWrapper.java

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
@@ -40,6 +40,7 @@ public class EhcacheProvider implements Cache.Provider {
     private static long curConfModTime = 0;
 
     static {
+
         URL url = null;
         File f = getConfFile("ehcache.xml");
         if (f != null && f.canRead()) {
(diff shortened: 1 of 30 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/FitsCacher.java
```diff
@@ -59,9 +64,13 @@
                     else {
                         Fits fits= null;
                         try {
+                            if (memCache != null) {
+                                memCache.put(key, new ObjectSizeEngineWrapper.BluffSize(fitsFile.length()));
+                                memCache.put(key, null);
```
lznakano: I am not quite understand what it is doing here? Why the same key was put twice with different value?

robyww: This is how I am clear space in the cache before I actually read the file.  If there is a 5G file in the cache by doing this I can potentially push out the 5G from the cache before the 5G file is read. Since the act of reading would takes 10G we would go as high as 15G. With this change it will only go as high as 10G the down to 5G.

To summarize- replace a f5G file in cache:
before:
5G => 15G => 5G
after
5G => 10G => 5G

The `ObjectSizeEngineWrapper.BluffSize` object appears to the cache to be big but it is not.

src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
```diff
@@ -65,7 +66,14 @@
             File ignoreSizeOf = getConfFile("ignore_sizeof.txt");
             System.setProperty("net.sf.ehcache.sizeof.filter", ignoreSizeOf.getAbsolutePath());
 
-            sharedManager = CacheManager.create(sharedConfig.getAbsolutePath());
+            try {
+                System.setProperty("net.sf.ehcache.sizeofengine.shared.VIS_SHARED_MEM",
+                                   "edu.caltech.ipac.firefly.server.cache.ObjectSizeEngineWrapper");
```
loitly: I think using ObjectSizeEngineWrapper.class.getName() here would be even cleaner.  It will fail on setProperty so clearing it is not necessary.  It'll also help find/refactor class when needed.

PR #372, opened 2017-05-03
DM-10379: fixed: After a catalog search with a globe coverage map, the catalog search doesn't work anymore.

robyww:
Fixed.

also fixes: DM-9824


ymeiymei:
Tested DM-10379:

The bug has been fixed. 

After getting catalogs from both m31 and m81, I can still go back to do the catalog search. The image search panel can be cancelled also. Good.

Also tested DM-9824. After loading the catalog table file in the ticket, I can keep doing catalog search and receive data.

PR #373, opened 2017-05-03
download plotly image as PNG

tgoldina:
Added save icon to download plotly chart as a png image. 
The size of the images will match the size on display.

Known issue: in Safari, an image is not downloaded, but opened in the browser, it can be saved from there. Plotly team says it behaves this way because download attribute is not supported in Safari (http://caniuse.com/#feat=download) and they open the link instead.

Link to the ticket:
https://jira.lsstcorp.org/browse/DM-10275
(As discussed at SUIT meeting, we want to save PNG without an extra dialog and we prefer png to jpeg because it is lossless.)

### DIFF
--- a/src/firefly/js/charts/ui/PlotlyWrapper.jsx
+++ b/src/firefly/js/charts/ui/PlotlyWrapper.jsx
@@ -3,19 +3,16 @@
  */
 
 import React, {Component, PropTypes} from 'react';
+import {get} from 'lodash';
 import {getPlotLy} from '../PlotlyConfig.js';
 import {logError} from '../../util/WebUtil.js';
 import Enum from 'enum';
 
 const PLOTLY_BASE_ID= 'plotly-plot';
 var counter= 0;
-var downloadCounter = 0;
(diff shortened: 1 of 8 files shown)


src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -25,6 +27,21 @@ const defaultConfig= {
...
 };
 
+export function downloadChart(chartId) {
+    getPlotLy().then( (Plotly) => {
+        const chartDiv = document.getElementById(chartId);
+        if (chartId && chartDiv) {
+            downloadCounter++;
+            Plotly.downloadImage(chartDiv, {
+                format: 'png',
+                filename: `plotly-${chartId.replace(/\-.*$/, downloadCounter)}`
```
loitly: I think 'replace' and counter tracking is unnecessary.  Even 'plotly', I don't think it's needed.
Filename should be chart.title or chart.chartId if title not available.

PR #375, opened 2017-05-05
IRSA-331:  update default cutout size units in TST

wmiipac:
What changed:
1. the default cutout size is set to 5 arcmin for all missons
2. added  cutout size conversion from arcmin to deg before sending WebPlotRequest

To Test
* upload a WISE/LSST time series table, the upper left form panel should have the default 5arcmin cutout size, the lower image panel will show size in arcmin unit in title

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcResult.jsx
+++ b/src/firefly/js/templates/lightcurve/LcResult.jsx
@@ -26,6 +26,7 @@ import {dispatchMultiValueChange, dispatchRestoreDefaults}  from '../../fieldGro
 import {logError} from '../../util/WebUtil.js';
 import {getConverter, getMissionName} from './LcConverterFactory.js';
 import {ERROR_MSG_KEY} from '../lightcurve/generic/errorMsg.js';
+import {convertAngle} from '../../visualize/VisUtil.js';
 
 const resultItems = ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'searchDesc', 'images',
     LC.MISSION_DATA, LC.GENERAL_DATA, 'periodState'];
@@ -130,6 +131,9 @@ const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables,
     const mission = getMissionName(converterId) || 'Mission';
     const showImages = isEmpty(imagePlot);
 
+    // convert the default Cutout size in arcmin to deg for WebPlotRequest, expected to be string in download panel
+    var cutoutSizeInDeg = (convertAngle('arcmin','deg', cutoutSize)).toString();
+
     var tsView = (err) => {
 
         if (!err) {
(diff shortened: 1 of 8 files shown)


src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -126,7 +126,7 @@ const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables,
 
     const converterId = get(settingBox, ['props', 'missionEntries', LC.META_MISSION]);
     const convertData = getConverter(converterId);
-    const cutoutSize = get(convertData, 'noImageCutout') ? undefined : get(settingBox, 'props.generalEntries.cutoutSize', '0.3');
+    const cutoutSize = get(convertData, 'noImageCutout') ? undefined : get(settingBox, 'props.generalEntries.cutoutSize', '5');
```
ejoliet: `cutoutSize` is used elsewhere (line 177) expecting the unit of the value in degree. The value is passed to the download component and the returned images are bigger than 5 arcminutes cutout because '5' (or any value input by user) is interpreted in the server in degree. 
You need to convert `cutoutSize` in degree before it get used in the download part.

ejoliet: This will also fix the LSST case automatically.

wmiipac: You are right, I neglected download size conversion, will fix it. thanks for catching it!

src/firefly/js/templates/lightcurve/wise/WisePlotRequests.js
```diff
@@ -26,15 +27,15 @@ export function makeWisePlotRequest(table, rowIdx, cutoutSize) {
...
+     set default cutout size 5 arcmin
      const url = `http://irsa.ipac.caltech.edu/ibe/data/wise/merge/merge_p1bm_frm/${scangrp}/${scan_id}/${frame_num}/${scan_id}${frame_num}-w1-int-1b.fits`;
      */
     const serverinfo = 'http://irsa.ipac.caltech.edu/ibe/data/wise/merge/merge_p1bm_frm/';
     const centerandsize = cutoutSize ? `?center=${ra},${dec}&size=${cutoutSize}&gzip=false` : '';
     const url = `${serverinfo}${scangrp}/${scan_id}/${frame_num}/${scan_id}${frame_num}-w${band}-int-1b.fits${centerandsize}`;
     const plot_desc = `WISE-${frameId}`;
     const reqParams = WebPlotRequest.makeURLPlotRequest(url, plot_desc);
-    const title = 'WISE-' + frameId + (cutoutSize ? ` size: ${cutoutSize}(deg)` : '');
+    const title = 'WISE-' + frameId + (cutoutSize ? ` size: ${cutoutSize}(arcmin)` : '');
```
wmiipac: This was the older version of image retrieve via url, it is not used for WISE anymore. Since we kept code here, I will convert the size to degree just to be consistent with the IBE image retrieve.

ejoliet: If it's not used, might be good to remove then so we are not tempted to maintain something useless.

ejoliet:
Works fine. Merge approved. Thanks!

PR #377, opened 2017-05-09
Creates a generic plot.ly chart.

loitly:
https://jira.lsstcorp.org/browse/DM-9945

Key points:
- plot any plot.ly chart
- ability to bind the data to our table model
- ability to update chart after it's been created (change options, add new trace, etc.)

Trey, I've made some changes to FieldGroupConnector.  Can you take a look to make sure it will not break existing code?

To test:
The last 2 charts are created using this API.  You can use the UI to update the chart.
Optionally, you can bring up DevTool and update the chart via our API.
The 2 functions worth mentioning are:
firefly.showPlot(targetDiv, {chartId, chartType, data, layout});
This function create a chart using the given parameters and place it in the targetDiv.
  - chartType:  use 'plot.ly'
  - chartId: a unique string identifier
  - data, layout are js objects documented here: https://plot.ly/javascript/

firefly.action.dispatchChartUpdate({chartId, changes:{'path': updates}});
This function update the chart based on the given chartId.
  - changes are keyed by a path string (location of the object to update).
  - updates is the object to replace it with.


tgoldina:
Some observations so far:
- Selection box is sometimes gone when switching between unexpanded and expanded, while selection icons stay.
- When switching charts or going from one to multiple views in chart multiviewer. the chart is rendered twice: first with smaller hight, then with the right height. Same when switching to expanded mode.
- In multiviewer, the options are shown in the chart space, not under settings icon. Expand is also per chart, not per multiviewer. In general, multiviewer in API probably needs more work.

src/firefly/js/charts/ui/ChartPanel.css
```diff
@@ -76,3 +76,82 @@
...
+    background-image: url("/images/icons-2014/24x24_GearsNEW.png");
+}
+
+.ChartToolbar__expand {
+    background-image: url("/images/icons-2014/24x24_ExpandArrows.png");
+}
+
+.ChartToolbar__auto-scale {
+    background-image: url("/images/icons-2014/Zoom1x.png");
+}
```
loitly: Icons updated.
Will create a new ticket for save and filter.

src/firefly/js/charts/ui/PlotlyChartPanel.jsx
```diff
@@ -0,0 +1,151 @@
...
+        const {activeTrace=0}  = ChartsCntlr.getChartData(chartId);
+        const [xMin, xMax] = get(evData, 'range.x', []);
+        const [yMin, yMax] = get(evData, 'range.y', []);
+        const points = get(evData, 'points', []).filter((o) => {
+            return o.curveNumber === activeTrace;
+        }).map((o) => {
+            return o.pointNumber;
+        });
+
+        dispatchChartUpdate({chartId, changes:{'selection': {points, range: {x:[xMin, xMax], y: [yMin, yMax]}}}});
```
loitly: Will create a new ticket for this.

src/firefly/js/charts/ui/options/BasicOptions.jsx
```diff
@@ -0,0 +1,150 @@
...
+            ...fieldProps,
+        },
+        ['layout.yaxis.title']: {
+            fieldKey: 'layout.yaxis.title',
+            value: get(layout, 'yaxis.title'),
+            tooltip: 'Y axis title',
+            label : 'Yaxis Title:',
+            ...fieldProps,
+        },
+    };
```
loitly: Can't duplicate.

src/firefly/js/fieldGroup/FieldGroupUtils.js
```diff
@@ -133,6 +133,10 @@ export const getFieldGroupState= function(groupKey) {
     return fieldGroupMap[groupKey] ? fieldGroupMap[groupKey] : null;
 };
 
+export function getFieldVal(groupKey, fldName, defval=undefined) {
+    return get(getGroupFields(groupKey), [fldName, 'value'], defval);
+};
+
```
loitly: getFldValue expects fields and fieldName.  In order to get fields, you need to call getGroupFields with groupKey.  
The above function is a convenience way to get the value using just one import, groupKey, and fieldName.  getFldValue is used everywhere and is more optimized when used multiple times.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -215,3 +226,155 @@ export function makeHistogramParams(params) {
...
+        if (isEmpty(traceTS)) {
+            tablesources[idx] = oldTraceTS;     // if no updates.. move the previous one into the new tablesources
+        } else {
+            const {tbl_id, ...rest} = traceTS;
+            if (!getTblById(tbl_id)) return;
+
+            if (oldTraceTS && oldTraceTS._cancel) oldTraceTS._cancel();   // cancel the previous watcher if exists
+            //creates a new one.. and save the cancel handle
+            updateChartData(tbl_id, chartId, idx, rest);
+            traceTS._cancel = watchTableChanges(tbl_id, [TABLE_FILTER, TABLE_SORT, TABLE_HIGHLIGHT, TABLE_SELECT], (action) => updateChartData(tbl_id, chartId, idx, rest, action));
```
loitly: Cleared data on TABLE_REMOVE and clear saga on CHART_REMOVE.

loitly:
I've addressed these issues:
- replace expanded, resetZoom icon
 - change autoscale to reset-zoom
 - add docs to submitChanges()
 - set new trace as active
 - clear data when table is removed and remove saga when chart is removed

I've created a new ticket for the rest.  https://jira.lsstcorp.org/browse/DM-10509

PR #379, opened 2017-05-11
DM-10491:WebGrid has missing lines

lznakano:
It seemed no one recalls why the algorithm was used in WegGrid.java.  It could be some special case.  That algorithm caused line missing in several cases.  I changed to the simple, and correct algorithm.  There is no more missing lines.

The other way to fix the missing lines is to do the forced regrid.  But since the cause is the algorithm, I chose to change the algorithm.

To test:
   ra=45, dec=0, it has a missing line at the right side of the image before.  Now it is there.
   ra=45, dec=90, it has a missing line in the bottom of the image before, it is there now.

### DIFF
--- a/src/firefly/js/drawingLayers/ComputeWebGridData.js
+++ b/src/firefly/js/drawingLayers/ComputeWebGridData.js
@@ -58,12 +58,25 @@ export function makeGridDrawData (plot,  cc, useLabels, numOfGridLines=11){
 }
 
 function adjustLevels(levels,numOfGridLines){
+
     const nL0 = levels[0].length;
+
     const nL1 = levels[1].length;
     var newLevels = levels;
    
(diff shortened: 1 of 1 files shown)


ymeiymei:
Tried WISE with target (0, 0) and (0, 89). Still some lines are missing.

lznakano:
Good catch!!!  Yi is a great tester!!! 

This is caused by the discontinuity of the ra data.  By regriding the data levels first and then adjust the ra values to the range [0-360], the discontinuity was taking care of.  I had to change the algorithm a bit to achieve it.

Please test again.

ymeiymei:
Tested again. The missing line problem has been fixed. Good job!

There are still some minor issues with grid line labels. I am going to issue an improvement ticket for future.

PR #381, opened 2017-05-11
DM-10477: wcs match reset images when it turns off

robyww:
DM-10477: wcs match reset images when it turns off

Including bug fixes:

  - Fixed a rotation issue: clicking rotate north on north up images fliped them south
  - Fixed: when north is up, clicking north up makes the image rotate south.

_To test:_

 - load m81 - 2mass, wise 1a, dss, sdss
 - click on any image turn wcs-match on then turn off
 - click on another image and do the same things
 - image should rotation to match then unrotated
 - Also, while WCS matched, rotate an image, they should rotate together and stay matched

_To test North up bug:_

  - load 2mass click on rotate north - nothing should happen

### DIFF
--- a/src/firefly/js/visualize/VisUtil.js
+++ b/src/firefly/js/visualize/VisUtil.js
@@ -170,7 +170,7 @@ export function computeCentralPointAndRadius(inPoints) {
  * @param {number} dec  the equatorial DEC in degrees of the second object
  * @return {number} position angle in degrees between the two objects
  */
-function getPositionAngle(ra0, dec0, ra, dec) {
+export function getPositionAngle(ra0, dec0, ra, dec) {
     let sind, sinpa, cospa;
 
     const alf = ra * DtoR;
@@ -200,7 +200,7 @@ function getPositionAngle(ra0, dec0, ra, dec) {
     if (dec0===-90) pa = 0.0;
 
     return pa;
-};
+}
 
 /**
  * Rotates the given input position and returns the result. The rotation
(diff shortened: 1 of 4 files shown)


ymeiymei:
Tested:
Image search, got three images:
WISE@m81, SDSS@m81, WISE@m31

Select SDSS@m81 and click "WCS Match", saw:
The WISE@m81 rotates and shrinks and drops lower so that it matches SDSS@m81 on orientation, target location and the pixel size.  The WISE@m31 image disappeared and a "Recenter" sign is shown. 

Then deselect "WCS Match", saw: 
WISE@m81 rotates back to the original orientation, still keeps the image size. WISE@m31 comes back and with the original orientation and the same pixel size as the other two images.

This bug of "WCS Match" off has been fixed.

A new problem:
When "WCS Match" is on and then off, and then back on, the images which have far away target are still shown, even with "Recenter" sign on them. 

Another problem:
When "WCS Match" is off, all the three images are locked together, click "North Up" button, those three images are not North Up together. Also when keep clicking NorthUP, images take turn to be NorthUp.

robyww:
Fixed the follow bugs:

- [x] When "WCS Match" is on and then off, and then back on, the images which have far away target are still shown, even with "Recenter" sign on them.
- [x] When "WCS Match" is off, all the three images are locked together, click "North Up" button, those three images are not North Up together. Also when keep clicking NorthUP, images take turn to be NorthUp.

ymeiymei:
All the problem found are fixed. Thanks!

PR #382, opened 2017-05-12
DM-10523: period finder layout up to date with plotly version

ejoliet:
The pull request take care of a version left over from RC and include it in the plotly recently created in dev.
The period finder layout has changed according to the feedback during the test campaign and the version needed to be merged in the plotly version.

Test: bring the LC tool, and go to period finder. The button of the form are now under the control panel, no period value is set by default and title have been added to the plots.

### DIFF
--- a/src/firefly/js/charts/ui/XYPlotPlotly.jsx
+++ b/src/firefly/js/charts/ui/XYPlotPlotly.jsx
@@ -34,6 +34,7 @@ const selectionRectColorGray = 'rgba(165, 165, 165, 0.5)';
 
 const Y_TICKLBL_PX = 90;
 const X_TICKLBL_PX = 60;
+const TITLE_PX = 30;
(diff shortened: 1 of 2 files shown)


tgoldina:
I have pushed a fix for title and selection rectangle not being shown with log scale.

I still see at least one issue: when you select period by clicking on the periodogram, the period in 'Set period' section is populated correctly, but the phase finding chart reflects the previous period value, not the current one. You have to hover over a point for the chart to update to reflecting the current period. (Same behavior with peak table.)

Also, when you first click 'Period Finder', the chart shows with period=NaN day.

ejoliet:
Ok, i will try the fix @tgoldina suggest. Thank you.

About the title `NaN (days)` there are more issue about it and it should be in another ticket. For example, what happen if my data is not in days, etc. See     IRSA-354.

PR #383, opened 2017-05-16
DM-10543: improved plotly performance during resize and add masking

robyww:
DM-10543: improved plotly performance during resize and add masking 

   - react component not has properties to control how it renders during resize
   - react component will now mask during slower work
   - moved all the resize debouncing to the PlotlyWrapper, Firefox is debounced more
   - LcPeriodPlotly will debounce the regen data for Firefox

### DIFF
--- a/src/firefly/js/charts/ui/ChartPanel.jsx
+++ b/src/firefly/js/charts/ui/ChartPanel.jsx
@@ -34,7 +34,7 @@ class ChartPanelView extends PureComponent {
 
         this.onResize = (size) => {
             if (this.state.widthPx === 0 || isPlotly()) {
-                defer(normal, size);
+            defer(normal, size);
             } else {
                 debounced(size);
             }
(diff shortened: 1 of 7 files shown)


xiuqin:
@tgoldina @robyww  The periodagram plotting does not use the scatter plot. We need to change that. I talked to @cwang2016  about it before.

robyww:
Just push change for fixes and some better approaches

- [x] The periodagram plotting does not use the scatter plot. We need to change that. 
- [x] When clicking on the periodogram, I noticed that it does not always disappear when the chart is updated.
- [x] Also in certain uses cases with PlotlyWrapper it can detect when an update involves a resize.

tgoldina:
I noticed the scrollbars when sizing down ChartPanel. It creates the impression that resize is slower then before.

Compare resizing 4 charts on a catalog before (irsawebdev1) and after. Now the charts are rendered with the mask and scrollbars first,  then resized, then mask disappears.  No scrollbars when sizing down before.

Otherwise works great.

robyww:
We can get better performance if the charts lets the wrapper set the size. this is the way the the phase folding is doing it. In this case the size is only changed in the style. I am not sure why we need to use the chart width  and height in the layout.

tgoldina:
Fixed the following issues: 

1. Scrollbars showing when scrolling down. (Set overflow=‘auto’ only if the chart and container dimensions are different. (See the note below).

2. When using X/Y ratio=1 and stretch=fill, we are able to scroll the chart vertically. When doing vertical resize, chart dimensions do not change, so there should be no resizing. But is still see the mask coming up. We should have set doingResize based on the chart size, not container’s.

To handle various X/Y ratios and stretch (by height or by width), we place chart into a container with the size matching Resizable size and overflow=auto. This way, whenever the chart size is larger than the size of the container, the scroll bars would show up. Chart dimensions are calculated in calculateChartSize of XYPlot.jsx  (It is the only chart with stretch parameters as of now.) We might want to autosize all plotly charts and move the logic calculating chart dimensions based on X/Y ratio and stretch to PlotlyWrapper.

loitly:
In regard to 'X/Y ratio', I think it will be phased out.  Physical appearance can be achieved by dragging the corners of the chart. 
It is still useful for decimation.  In our next iteration, we will use the decimation logic to create heatmap.  In that case, it may be presented as an input for creating heatmap.  The way I see it, we would not decimate scatter plot at all.  We will support it to a very large number, then offer heatmap option beyond that.

robyww:
> If you have time, it would be great not to show mask if it comes up for a very short time. (It makes resize feel slower.)

Ok, I added a delay before that mask appears.  The chart still does not update during resizing the only change is that the mask just does not appear for the first 4/10 of a second.

PR #384, opened 2017-05-18
IRSA-257: make use of ops internal from properties

ejoliet:
Added reading properties for cases iwise and future iptf, internal ops images and fixing properties related to Gator in ceres with basic authentication.

Note that in order to build, you will need to checkout both branches in Firefly AND IFE.

Test: build IV with `ops_int` environment (`-Penv=ops_int`), then try to get wise image and catalog search working. Searches will go to ceres and should be working. Attached in the ticket IRSA-257 is a LC table with yr4 data that will show images in TimeSeries tool in this case. Check also the download part of images is correct.

If you have question, let me know.


loitly:
I tested it with iwise and it seems to be working fine.  Good to go.

PR #385, opened 2017-05-20
DM-10560: clean of image retrieval error messages

robyww:
DM-10560: clean of image retrieval error messages

  - Error message are more accurate and short
  - The right message is passed though exception, instead of have a generic message
  - http codes are translated into messages for failure cases
  - removed 2 old Exception classes

_To Test:_

 - Try msx image with target m81
 - Try sdss image with target m31
 - Take your computer offline an try a wise image or a dss image
 - Load http://localhost:8080/firefly/demo/ffapi-highlevel-test.html notice the error message on one of the plots
 - There are many more error messages but they are hard to test when the servers are up


ymeiymei:
Reviewed and tried all the test cases suggested. They all have better error messages than before. 

Only one small issue. For WISE, the error message mentioned "IbeImageGetter Call failed with IOException". Comparing with other missions, the class names have some information about the missions. Possible to add "WISE" in the class name IbeImageGetter.java? No big deal though. Just would like to give users more info and be consistent with other missions.

robyww:
> Only one small issue. For WISE, the error message mentioned "IbeImageGetter Call failed with IOException". 

Your right, that is not the right way to do that message

xiuqin:
IbeImageGetter is not only for WISE images, it could be used for PTF, 2MASS images too. I would suggest we put in better error message, not just use the class name.

cwang2016:
Review completes. The error messages for image retrieval looks much more readable. 
some thoughts, 
- the error message for offline case (ex. Failed - IbeImageGetter Call failed with IOException)  is not quite readable. 
-  The error message style for image retrieval could also be applied to other search engine, such as LSST catalog and image search. 
- we may some documentation which lists the protocol for composing the error messages and the error messages for the image and catalog search.

robyww:
> we may some documentation which lists the protocol for composing the error messages and the error messages for the image and catalog search.

 I am not sure I am ready to write documentation yet, but we should discuss in the W
ednesday meeting.

PR #387, opened 2017-05-22
 IRSA-371-histogram - numbins sometimes ignored

lznakano:
Change the default behavior of Histogram to show all the bins including the empty bins

This is a minor change to avoid confusion to the users.  As I commented in the ticket the reserveEmptyBins can be added to the UI and the parameter.


lznakano:
Emmanuel, 

Do you want me to implement it as what the comments are?  If so,  I can completely remove the REAERVE_EMPTY_BINS parameter.  The histogram will always show all the bins. 

Thanks!

PR #388, opened 2017-05-22
DM-9254 Make Firefly to work on Windows browser IE11

cwang2016:
This development: 
-  import 'babel-polyfill' in Firefly.js to fix the fallback of some string functions on IE11. 

Test, 
you may install Windows on your Mac using VirtualBox to run IE11. You need, 
- Oracle VirtualBox(with extensions), 
   https://www.virtualbox.org/
- Windows 10 ISO 
   https://www.microsoft.com/en-us/software-download/windows10ISO

To do:
- download VirtualBox (and extensions), and create virtual machine with Windows 10. 
   (if you need more detail about installation, please let me know). 
- Starts IE11 from the created virtual machine and launch Firefly by using "http://10.0.2.2/firefly" (this is relevant to Firefly from your local machine, localhost:8080/firefly)

### DIFF
--- a/src/firefly/js/FFEntryPoint.js
+++ b/src/firefly/js/FFEntryPoint.js
@@ -4,7 +4,6 @@
 
 
 import {get} from 'lodash';
-
 import {firefly, Templates} from './Firefly.js';
 import {HELP_LOAD} from './core/AppDataCntlr.js';
 
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -3,8 +3,7 @@
  */
 /*global __$version_tag*/
 
-
-// import 'babel-polyfill';
+import 'babel-polyfill';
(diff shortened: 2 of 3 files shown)


ejoliet:
Thanks for taking car of this compatibility issue.

I tested in IE11 and seems to be working almost everything.
But somehow, when tested in LC, the period finder 'Accept' button doesn't return the phase folding table/result layout. 

I can see also some error such as:
'{{offsetWidth}} property undefined or null reference' when i swithc layout 'tri-view' to 'img-xy' for example.

xiuqin:
@ejoliet Could you please show Cindy on your laptop if needed?  She mad a virtual machine to develop and test.

cwang2016:
The issue regarding the phase folding table in LC could not be reproduced, and @ejoliet didn't reproduce it either.

PR #391, opened 2017-05-30
DM-10462 Adapt scatter plot for multi-trace chart architecture

tgoldina:
Test page:
(w[1,2,3,4]mpro and w[1,2,3,4]sigmpro are good fields to plot)

Major changes:

- Made Plotly chart a part of MultiChartContainer (shared options, taking all length of the container)

- Used XYPlotWithErrorProcessor for a proof of concept how the derived data can be handled in the chart. tablesource now stores mappings, optional function fetchData and last fetch options. The data fields that can not be mapped to known plotly trace data are stored in data[traceNum].firefly object.

- The derived table can be smaller (less rows) than the original table. data[traceNum].firefly.rowIdx provides the connection between the original table rows and chart points. Chart highlight/selection is using point indexes, table highlight/selection - table row indexes.

- Handling highlight of a selected point. plotly_onclick only gives the topmost point clicked. Since we do not store selected and highlighted traces, I had to use plotly trace data when active trace is not matching the clicked point trace.

- Updated plotly to the latest version (1.27.1) Unfortunately, onclick still returns only the topmost point. (According to https://github.com/plotly/plotly.js/issues/1648 it should be fixed. Am I missing something?)

- Adjusted data flow to avoid data refetch, when the updates to data do not require server call (ex. marker color or shape change) Since the updates are passed as the changes to the data, they overwrite existing data. I save old data and pass them to the table connector to be able to restore them. (Maybe, we can come with a better solution later.)

- Made adjustments to avoid double updates on highlight and select. Before, chart was triggering table events directly and responded to table events. Now we have chart specific dispatchers which update the chart AND trigger table action. (Added chartTrigger to action payload.)

- Added Save [Image] button to the new plotly chart toolbar. It turned out that in API we can have 2 components with the same chart id - one expanded, one non-expanded - adjusted the method to handle both.

- Updated ScatterOptions to use table statistics and column lookup and expressions. Added UI to define errors. (ScatterOptions still need a lot of work.)


src/firefly/js/charts/ui/options/NewTracePanel.jsx
```diff
@@ -125,47 +128,61 @@ function fieldReducer({data, layout, activeTrace, tablesources={}}) {
...
+             value: get(tablesourceMappings, ['error_x.array'], ''),
+             //tooltip: 'X error',
+             label: 'X error\u2191:',
+             ...fieldProps
+         },
+         [`_tables.data.${activeTrace}.error_x.arrayminus`]: {
+             fieldKey: `_tables.data.${activeTrace}.error_x.arrayminus`,
+             value: get(tablesourceMappings, ['error_x.arrayminus'], ''),
+             //tooltip: 'X error',
+             label: 'X Err\u2193:',
```
loitly: Err instead of error, is that on purpose or typo?

tgoldina:
Thank you for the comments. Here are  some clarifications.

> I am trying to make use to SORT and FILTER directly instead of relying on invokedBy in LOADED. This cannot be done using SORT or FILTER?

For larger tables that don't load right away, I need to watch TABLE_LOADED. If I also watch TABLE_SORT and TABLE_FILTER – they are preceded by TABLE_LOADED –  I will trigger the data fetch twice when we filter or sort the table.

> rowIdx – I thought the data points can have a one-to-one relationship with the same indices. I am not sure this is needed.

It's not always needed, but we might have less chart points that table rows and still want to preserve connection between them. At the moment we our search processor is not returning the data with missing x or y. Another example: decimated chart point only connects to one (representative) table row. In future, we might be skipping some rows or doing the aggregation by rows rather than columns, so this feature is useful.

> activeTrace is different from activeDataTrace if a selected point highlighted, for example - I don't understand this

Suppose we are highlighting a selected point and out layout has an attribute hovermode: 'closest'. Then plotly_onclick event data would only contain the topmost point. CurveNum will be the selected trace number, which does not have the tablesource associated with it.

src/firefly/html/demo/plotly-concepts.html
```diff
@@ -9,7 +9,7 @@
     <link rel=”apple-touch-startup-image” href=”images/fftools-ipad_splash_768x1004.png”>
     <title>Plotly Concept</title>
 
-    <script  type="text/javascript" src="../plotly-1.25.0.min.js"></script>
+    <script  type="text/javascript" src="../plotly-1.27.1.min.js"></script>
```
ejoliet: Can't we not having an alias and get it from somewhere else that we maintain separately? github or directly from its CDN?

tgoldina:
Two items moved to TODO list: 
- separating firefly specific things from plotly trace data
- not storing fetchData function in tablesource (though I realized that _cancel is also a function which we store in the tablesource)

ejoliet:
Please @tgoldina , raise ticket for those items if not yet done. Thanks!

PR #394, opened 2017-06-10
DM-10831 Histogram multi-trace chart

tgoldina:
This PR is for supporting Firefly Histogram chart (with the column data aggregated into bins on the server) in multi-trace plotly chart. (Currently, multi-trace chart is only available in API.)

I've changed histograms to solid color to make multiple histograms look good together, and so that the trace color can be changed easily.
For comparison, I have added options for Plotly histogram (client side histogram calculation). Firefly histogram (backed by Plotly bar chart) and Plotly histogram do not work well together, but it's good to see what options are available and see if we can support both charts with one set of options.

Ticket:
https://jira.lsstcorp.org/browse/DM-10831

Test:
Check the first chart, try to change options, add new traces, reset. You can also add histogram trace to scatter plot just above the table.

### DIFF
--- a/src/firefly/html/demo/ffapi-highlevel-charttest.html
+++ b/src/firefly/html/demo/ffapi-highlevel-charttest.html
@@ -91,32 +91,34 @@
 //             **/
 //            firefly.util.renderDOM('histogramHere', firefly.ui.Histogram, props);
 
-            var dataH = [{marker: {color: '#993399'},opacity: 0.8},
-                          {marker: {color: '#669900'},opacity: 0.7}];
+            var dataH = [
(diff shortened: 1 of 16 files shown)


cwang2016:
Review completes. look very good. 
some minor thought: 
for histogram options, are plot title. xaxis title, yaxis title common to every trace? from the option panel, it is not known that the setting of those items is applied to entire histogram.

cwang2016:
one more thought, 
color is an essential visual effect for multi-trace histogram, especially when the bar width are various among traces. how about in addition to the bar color, the border color of the bar is settable? just my personal thought.

PR #395, opened 2017-06-10
DM-7405 Add  VOTable support while uploading table 

cwang2016:
The development includes:
- Update votable lib (starlink)   to work for parsing votable with BINARY2 format.
  (from directory jars/starlink,  rayreg.jar, stil.jar are updated, regclient.jar has no change). 
- replace linefeed (\n) character in the data content of the table cell which is 'char' type to be a space while converting votable (or any other format table) to IPAC file. 

test:
start catalogs search panel, go to 'load catalog' tab, select the votable sample from firefly_test_data/votable-samples. The samples of 'binary2_votable.xml',  'stilbinary.xml' are with BINARY2 format.


src/firefly/java/edu/caltech/ipac/util/IpacTableUtil.java
```diff
@@ -100,6 +100,7 @@ public static void writeRow(PrintWriter writer, List<DataType> headers, DataObje
             if (v == null && dt.getKeyName().equals(DataGroup.ROWID_NAME)) {
                 v = dt.getFormatInfo().formatData(row.getRowIdx());
             }
+            v = v.replaceAll("\n", " ");
```
robyww: I am very nervous about make any change to IpacTableUtil unless loi and/or tatiana approve it. Have you talked to them?

tgoldina:
I am OK with this solution. The root problem is that IPAC table format does not allow line break in the cell values, while other table formats (FITS binary, CVS, StarTable) allow it. So we can either replace line break characters on parse or on writing out the IPAC table. The tradeoff is between the efficiency (avoiding string substitution in every line) and preserving data as they were in the source until we write them out.

tgoldina:
Loi might have a different opinion, so if the merge can wait until Wednesday, let's wait for his input.

ejoliet:
@tgoldina the problem is not IPAC table per se, the problem is our back-end datamodel to cache/parse the uploaded file into IPAC table format. 
If we want to give more flexibility to change table (add column, etc) and we also care about handling large tables, I think what we should do is to plan how we move away from our data-model based on ipac table (ASCII) to more binary or (distributed) database internally. This is upcoming stories anyway that Loi have been already working on. We can discuss it next meeting.

ejoliet:
BTW, a good and easy to use service to download sample data in VOTable in binary, binary2, fits and tabledata can be found in Gaia from here:
http://gaia.ari.uni-heidelberg.de/tap.html

The ADWL can be as simple as : ```SELECT TOP 1000 source_id, ra, dec, pmra, pmdec, parallax
FROM gaiadr1.tgas_source```

cwang2016:
linefeed in the middle of the row affects the ipac row structure. linefeed (\n) is only  allowed in the end of the row. Strictly speaking, should other non-printable or  escape character like tab be also checked before writing the data content into IPAC file?

cwang2016:
@ejoliet the sample, binary2_votable.xml, is seen to have 1019 rows, though some text string contains non-printable character.

loitly:
Cindy and I have talked about this.  We already have a function that deals with these cases.  Let's use it.  edu.caltech.ipac.util.StringUtils#convertExtendedAscii

cwang2016:
- Update function, edu.caltech.ipac.util.StringUtils#convertExtendedAscii, by handling cases in which the character is with ascii value less than 32 (space). 
- check and convert the text cell by using the above function before writing into IPAC file.

cwang2016:
@loitly, @ejoliet  would you please review the update again? Thanks.

PR #397, opened 2017-06-23
DM-10830: Optimize the performance of multi-trace chart updates

loitly:
https://jira.lsstcorp.org/browse/DM-10830
Explores ways to improve scatter chart performance, especially in Firefox.
To test:
- build firefly in this branch
- goto http://localhost:8080/firefly/demo/ffapi-highlevel-charttest.html
  - edit line 155 of this file from 'radius: 300' -> 'radius: 1000' to get 4k rows

Issues to consider:
- plot update instead of redraw did not increase performance enough, especially on larger plots.
- scattergl support is not complete.
  - latest update v1.28.2 released just a few days ago
  - known bugs:  https://github.com/plotly/plotly.js/issues/130
     - affects us:
        - reversed axis not supported
        - yaxis right+log does not work all the time.
        - zoom select on color map with size very sluggish.
        - 'open' symbol types not working.  ie.  'circle-open'
- what should we do?

Below is a summary of measured time for a single row highlight:
```
     500 rows        GL                SVG
     browser    update    redraw    update    redraw
     chrome     <15       150       40-45     50
     firefox    17-22     150       170-200   210

     4k rows        GL                SVG
     browser    update    redraw    update    redraw
     chrome     15-20     180       215       230
     firefox    20-25     180       1200      1400
```

The code is committed with no scattergl and partial update enabled.
To test the various modes yourself, set sessionStorage variables in debugger console:
        _chartRedraw: set to always redraw full plot_
        _scatterGL: set to convert scatter to scattergl_
```
   sessionStorage.setItem('chartRedraw', 'true' | 'false')
   sessionStorage.setItem('scatterGL', 'true' | 'false')
   sessionStorage.clear()

```


tgoldina:
when trying to add another firefly histogram to histogram chart, I get
```
TypeError: Cannot convert undefined or null to object
    at entries (<anonymous>)
    at doAdd (NewTracePanel.jsx:78)
```

ejoliet:
I didn't quite understand how to test with or without gl, let's talk about it so you can show me. 

What i did is i open the Time series tool and play with it. Looks much more faster and responsive but still quite slow in Firefox with 3000 row of table. I think the partial update fix is a right approach but there must be plotly code that is not optimized. I think we need to have a meeting and discuss all about this issue of performance to plan what to do and have clear what we can fix and what we have to live with.
 
Also during my test, i discover a bug when the plot is a density plot, the time series plot get frozen and even uploading a new table doesn't make the plot goes away. I needed to refresh the browser. I'll test other tables.

src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -166,24 +167,58 @@ export class PlotlyWrapper extends Component {
         return true;
     }
 
+    optimize(graphDiv, renderType, {chartId, data, layout, dataUpdate, layoutUpdate, dataUpdateTraces, ...rest}) {
+        const {lastInputTime} = layout;
+        if (lastInputTime && renderType ===  RenderType.NEW_PLOT && graphDiv.data) {
+            // omitting 'firefly' from data[*] for now
+            data = data.map((d) => omit(d, 'firefly'));
```
tgoldina: I think the mismatch in clicked and highlighted point when highlighting a selected point is caused not passing firefly rowIdx array to plotly div. (For selected points I get row idx from plotly div, because we do not save them in data.) I think, we should omit firefly object for the purposes of calculating the difference, but it should not be dropped completely. We should still pass it with the partial or full update.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -548,6 +548,19 @@ export function applyDefaults(chartData={}) {
     chartData.layout = merge(defaultLayout, chartData.layout);
 }
 
+const TRACE_COLORS = ['lightblue', 'green', 'purple', 'brown', 'yellow', 'red'];
+export function getNewTraceDefaults(type='', traceNum=0) {
+    if (type.includes(SCATTER)) {
```
tgoldina: I think, when the type is not SCATTER this function returns undefined, which is causing the type error, which I mentioned before (when we are trying to use the result as an array).

loitly:
I've committed changes to this pull request.  Please review again.
- defaults to partial update and no scattergl.
- fix getNewTraceDefaults() returns undefined when type is not scatter
from tgoldina:
- fixes highlighting selected, bug due to shared data

ejoliet:
When i click around in timeseries tool result layout (from raw to pahse folded table back and forth), i see error in console:
```uncaught at masterSaga 
 at lcManager 
 TypeError: Cannot read property '0' of undefined
    at http://localhost:8080/firefly/firefly-dev.js:65411:53
    at http://localhost:8080/firefly/firefly-dev.js:62641:29
    at Array.reduce (native)
    at http://localhost:8080/firefly/firefly-dev.js:62640:29
    at Array.findIndex (native)
    at findIndex (http://localhost:8080/firefly/firefly-dev.js:62639:67)
    at keepHighlightedRowSynced (http://localhost:8080/firefly/firefly-dev.js:195735:35)
    at handlePlotActive (http://localhost:8080/firefly/firefly-dev.js:194393:43)
    at lcManager$ (http://localhost:8080/firefly/firefly-dev.js:194217:38)
    at tryCatch (http://localhost:8080/firefly/firefly-dev.js:26266:41)
```

loitly:
> When i click around in timeseries tool result layout (from raw to pahse folded table back and forth), i see error in console:
timeseries is still using highchart.  I don't think this pull request has an affect on it.  Have you tried the same on 'dev' or 'rc' build to see if there's a regression bug there already?

ejoliet:
Timeseries in firefly/dev is using plotly.

loitly:
> Timeseries in firefly/dev is using plotly.

Okay, I found the errors now.  Although this is an existing bug which can be duplicated in ops and is not related to this PR, I went ahead and fixed it.

PR #398, opened 2017-06-23
DM-8707: scatter options in multi-trace charts

tgoldina:
- Added options for plot boundaries, color map, and xyratio/stretch (for all chart types) 
- Added color picker to Color field. (Later we should have color input field.)
- Added field validation for boundary fields and logic to clear the fields on x and y changes.
- Color map and Size map can be column expressions. 
  (If Size map evaluates to a constant, it won't be connected to the table.) 

The issues that need to be resolved in the future tickets:
- Color input field
- Color bar display when color map is specified. (There is a number of issues with display: two tick sets, no update when color scale changes. We also need to make sure it's on the opposite side of yaxis.)
- Size map scaling: to make sure all points are visible, we might need to scale point size in pixels so that it falls into the valid range (4px and up), otherwise the points will be invisible.

Known issues to be resolved in future:

Currently it is possible to set boundaries from the options, but not when the chart is added. The reason is that as of now Plotly does not support setting upper or lower boundary for an axis, but it is likely that it [will be added in the future](https://github.com/plotly/plotly.js/issues/400).

You can still set the axis range in layout, but it's not that convenient, because it has to follow plotly limitations: both values in range need to be there, if the axis is reversed, the values need to be reversed, if the axis type is "log", the values need to be logs.  

https://jira.lsstcorp.org/browse/DM-8707

### DIFF
--- a/src/firefly/js/charts/dataTypes/FireflyScatter.js
+++ b/src/firefly/js/charts/dataTypes/FireflyScatter.js
@@ -38,6 +38,14 @@ export function getTraceTSEntries({traceTS}) {
     } else {
         options['yErrColOrExpr'] = get(mappings, ['error_y.array']);
     }
+    const colorColOrExpr = get(mappings, ['marker.color']);
+    if (colorColOrExpr) {
+        options['colorColOrExpr'] = colorColOrExpr;
+    }
+    const sizeColOrExpr = get(mappings, ['marker.size']);
+    if (sizeColOrExpr) {
+        options['sizeColOrExpr'] = sizeColOrExpr;
+    }
 
     return {options, fetchData};
 }
@@ -61,7 +69,7 @@ function fetchData(chartId, traceNum, tablesource) {
             const yTipLabel = yLabel.length > 20 ? 'y' : yLabel;
 
             const {tableMeta} = tableModel;
-            const validCols = ['rowIdx', 'x', 'y', 'sortBy', 'xErr', 'xErrLow', 'xErrHigh', 'yErr', 'yErrLow', 'yErrHigh'];
+            const validCols = ['rowIdx', 'x', 'y', 'sortBy', 'xErr', 'xErrLow', 'xErrHigh', 'yErr', 'yErrLow', 'yErrHigh', 'color', 'size'];
             tableModel.tableData.columns.forEach((col) => {
                 const name = col.name;
                 if (validCols.includes(col.name) && tableMeta[name]) {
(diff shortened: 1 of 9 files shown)


src/firefly/js/charts/dataTypes/FireflyScatter.js
```diff
@@ -90,20 +118,19 @@ function fetchData(chartId, traceNum, tablesource) {
 
 function addOtherChanges({changes, xLabel, xTipLabel, xUnit, yLabel, yTipLabel, yUnit, chartId, traceNum}) {
 
-    // TODO: if we set the default label when there is only one trace, it needs to be cleared when we add more traces
     // set default title if it's the first trace
     // and no title is set by the user
```
loitly: Chart sent from API is not affected by this, right?

src/firefly/js/charts/ui/options/ScatterOptions.jsx
```diff
@@ -38,23 +39,24 @@ export class ScatterOptions extends SimpleComponent {
...
-                <FieldGroup className='FieldGroup__vertical' keepState={false} groupKey={groupKey} reducerFunc={fieldReducer({data, layout, activeTrace, tablesources})}>
+                <FieldGroup className='FieldGroup__vertical' keepState={false} groupKey={groupKey} reducerFunc={fieldReducer({chartId, activeTrace})}>
                     <ListBoxInputField fieldKey={`data.${activeTrace}.mode`} options={[{value:'markers'}, {value:'lines'}, {value:'lines+markers'}]}/>
                     <ListBoxInputField fieldKey={`data.${activeTrace}.marker.symbol`}
                                        options={[{value:'circle'}, {value:'circle-open'}, {value:'square'}, {value:'square-open'}, {value:'diamond'}, {value:'diamond-open'},
                                                  {value:'cross'}, {value:'x'}, {value:'triangle-up'}, {value:'hexagon'}, {value:'star'}]}/>
                     {tablesource && <TableSourcesOptions {...{tablesource, activeTrace, groupKey}}/>}
                     <br/>
-                    <BasicOptionFields {...{layout, data, activeTrace}}/>
+                    <BasicOptionFields {...{activeTrace, groupKey}}/>
```
loitly: I am slightly concern with naming this attribute 'groupKey'.  From chart's perspective, what is it?  Is it traceId?  It's okay to use the value from groupKey, but maybe it should be named according to its usage?

tgoldina:
Thank you for all the suggestions. We do want to handle axes scale ratio in
the future. I'll create a ticket for it.

1. Regarding boundaries, I was tempted to map it directly to plotly range,
but I think it is not convenient for the user, because range is always a
pair of values which has values dependent on the axis type (log or linear)
or display (ex. reverse). Boundaries allow to set a single limit, which
does not depend on the axis display parameters. I also thought that we do
want to preserve these limits through user's interactions with the chart
(panning or zooming).

2. Regarding setting default axis labels, the idea is that the plot is
meaningless unless you specify what each axis means, so by default, if only
one trace is present, we want to use table columns (and its units) as axis
labels, if they are not specified.

3. Regarding groupKey - this is FieldGroup key, which we pass around to be
able to set/get specific fields values for more complex interactions with
the fields (choosing color, showing/hiding stretch when xyratio changes,
etc.)

2017-06-26 9:54 GMT-07:00 loitly <notifications@github.com>:

PR #402, opened 2017-06-30
IRSA-522:Migrate Finderchart: pass the information in UI to the search processor

lznakano:
I changed the TableUtils.js to make the SDSS request for Finderchart.  This branch should work with IRSA-522-finderchart in ife.


src/firefly/js/tables/TableUtil.js
```diff
@@ -120,7 +120,7 @@ export function makeIrsaCatalogRequest(title, project, catalog, params={}, optio
     title = title || catalog;
     options.use = options.use || 'catalog_overlay';
     const tbl_id = options.tbl_id || uniqueTblId();
-    const id = 'GatorQuery';
+    const id = params.processorName||'GatorQuery';
```
loitly: Can you please undo this change?  This function is called makeIrsaCatalogRequest which is a request to the gator service at IRSA.  Passing a different 'id' will void all of the defined parameters.  There are other function you can use when you want to supply your own 'id', ie. makeTblRequest

lznakano: I did know that, so that I introduced a new parameter in my code "processorName", and the 
" const id = params.processorName||'GatorQuery';" will always return 'GatorQuery' except in Finderchart because the processorName  is only set in Finderchart.   All the existing calls will have "params.processorName" as undefined.   Of course, I can add one method to just do that.  But since this line does not hurt anything else, I just modified the existing method.

lznakano: Thanks Loi's explanation that makeIrsaCatalogRequest is specified for GatorQuery.  The TableUntil is revert to the dev version.  The corresponding changes are made in Finderchart.jsx.  The branch " IRSA-522-Finderchart" is under ife  located in the server. 
Please verify it again.  Thanks!

lznakano: This branch can be deleted.  There is no change in firefly.  The changes are all in finderchart.

lznakano:
I will remove TABLE-SEARCH and push and then merge.  Thanks!

lznakano:
I forgot that I changed hydraManager.  So I will push my changes here and then merge.

ejoliet:
Great! thanks. Remember to do the merge in dev and clean up the branch in IFE too ;-)

PR #403, opened 2017-06-30
DM-10349 Chart expression logarithm to be the same as other languages

tgoldina:
https://jira.lsstcorp.org/browse/DM-10349

Use the following functions for logarithms in Firefly column expressions: 
   log, ln – for natural logarithm
   log10, lg – for common logarithm
 
(The meaning of 'log' has changed from lg to ln to match definitions in JS or Python.)

Added documentation for expressions to both code and tooltips.

Reduced the number of suggest box updates to avoid flickering.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/expr/expressions.html
+++ b/src/firefly/java/edu/caltech/ipac/util/expr/expressions.html
@@ -1,0 +1,228 @@
+<!--
+Copyright 2002 by Darius Bacon <darius@wry.me>
+Altered
+-->
+<!DOCTYPE html>
+
+<html>
+<head><meta name="generator" content="Latte 2.1"><title>Expressing Formulas</title></head>
+   <body bgcolor="ivory"><p><h1>Expressing formulas</h1>
+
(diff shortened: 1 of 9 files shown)


src/firefly/java/edu/caltech/ipac/util/expr/expressions.html
```diff
@@ -0,0 +1,228 @@
+<!--
+Copyright 2002 by Darius Bacon <darius@wry.me>
+Altered
+-->
+<!DOCTYPE html>
+
+<html>
```
loitly: I am not sure this is the right place to put this file.  But, I can't think of a better place for it.

tgoldina:
It's actually from the original expression library, I have modified it just
slightly to be in sync with our code, so I thought I should be stored with
the other code from the same library.

I don't think it should go into the online docs as is, but it gives an
overview of what the expression library does support.

2017-06-30 15:47 GMT-07:00 loitly <notifications@github.com>:

xiuqin:
A modified version should be in online help eventually.

PR #404, opened 2017-06-30
IRSA-504 workspace server api

ejoliet:
I've checked in the manager for workspaces that should work for different implementation. Part of the code was already implemented few years ago so i needed to adjusted and add/fix webdav specific implementation.

**The idea** is to have a common interface that will be used from UI client to interact with workspaces back in the server. Right now the implementation is WebDAV and local FS (not fully implemented) for testing purposes. I've kept original method that was used in other places already (such as to get the ws directory or staging files directly from FF app) - _`I'm not sure if we want to stage files directly to workspace without the consent from the user first, such that the user has to deliberately set the default mechanism to save the result into workspaces`_

**The API** exposes `get, put, create parent, rename, delete and list files` from user workspace independently of the implementation in the workspace server. There is also a way to set and get metadata, which will help in particular to get/set custom properties in the future. 

I've added a bunch of tests that covers the implementation of the manager and aslo helps to understand how the api works. Typically a request to the server will give back a metadata object which contains the properties of the node(s). This object should be also generic to cover any workspace implementation in the future. 

**The tests** covers the authenticated user and public (anonymous) space as well. I've omitted the test to authenticate with cookies (with IRSA SSO server).

Finally, i've added some code to start the **skeleton of the server commands** but i feel that this will be part of the other ticket IRSA-474 (File picker to use server worskspace api). 

Couple of things need to be addressed and **can't be done at this stage**: 

- how the response json object would look like when used by the file picker to list the files for example.
- Should we create by default the workspace user home if not present? Right now the implementation is creating a parent (see constructor of the manager).
- Would the file picker only show files if logged in? Should we always use anonymous space?

**Need to discuss further also:** I'm not sure if we want to stage files directly to workspace without the consent from the user first, such that the user has to deliberalty set the default mechanism to save the result into workspaces

Part of the _**answer of many questions could be found in IRSA-457**_: workspace reqs and concepts, when ready.

> **To TEST**:
> 
> Checkout branches from firefly and firefly_test_data: `IRSA-504-workspace-server-api`.
> In order to see logs, you need to change level from WARN to INFO in `config/log4j-test.properties`.
> Then run tests `gradle test` or from your favorite IDE. Tests will take files from `test_data` here: `edu/caltech/ipac/firefly/ws/`. One fits file is used for SSO only because pubspace has limit to 1Mb. The tests use a common user id: `test-ws` with no password (anonymous).
> 
> Then go to some tests and see how you can call the manager via a handler that will keep the manager mapping the user id. Each handler belongs to a protocol: WEBDAV, LOCAL or VOSPACE. Manager can be of different implementation: webdav or local.
> 
> Another way to see in action is to comment out the tear down (@After in test) of the test and look for url:  https://irsa.ipac.caltech.edu/pubspace/test-ws/
> This area should be populated and clean up as long you run those tests.

Let me know if you have trouble or find problems.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
@@ -6,6 +6,7 @@ import edu.caltech.ipac.firefly.server.ws.WsCredentials;
 import edu.caltech.ipac.firefly.server.ws.WsException;
 import edu.caltech.ipac.firefly.server.ws.WsResponse;
 import edu.caltech.ipac.firefly.server.ws.WsUtil;
+import edu.caltech.ipac.util.AppProperties;
 import org.apache.commons.io.FileUtils;
 
 import java.io.File;
@@ -21,6 +22,8 @@ import java.nio.file.Paths;
  * Created by ejoliet on 6/21/17.
  */
 public class LocalFSWorkspace implements WorkspaceManager {
+    //TODO We might need to change those properties if we need to deal with 2 or more workspaces at the same time
+    String WS_HOST_URL = AppProperties.getProperty("workspace.host.url", "https://irsa.ipac.caltech.edu");
 
 
     private final WsCredentials wscred;
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
@@ -15,6 +15,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.function.Consumer;
 
 /**
  * LOCAL filesystem implementation of workspace- NOT FULLY IMPLEMENTED
@@ -75,6 +76,11 @@ public class LocalFSWorkspace implements WorkspaceManager {
 
     @Override
     public WsResponse getList(String parentUri, int depth) throws IOException {
+        throw new IllegalArgumentException("not implemented");
+    }
+
+    @Override
+    public WsResponse getFile(String fileUri, Consumer consumer) throws WsException {
         throw new IllegalArgumentException("not implemented");
     }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
@@ -1,0 +1,163 @@
+package edu.caltech.ipac.firefly.server;
+
+import edu.caltech.ipac.firefly.data.WspaceMeta;
+import edu.caltech.ipac.firefly.server.util.Logger;
+import edu.caltech.ipac.firefly.server.ws.WsCredentials;
+import edu.caltech.ipac.firefly.server.ws.WsException;
+import edu.caltech.ipac.firefly.server.ws.WsResponse;
+import edu.caltech.ipac.firefly.server.ws.WsUtil;
+import org.apache.commons.io.FileUtils;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * LOCAL filesystem implementation of workspace- NOT FULLY IMPLEMENTED
+ * For testing purposes so far
+ * Created by ejoliet on 6/21/17.
+ */
+public class LocalFSWorkspace implements WorkspaceManager {
+
+
+    private final WsCredentials wscred;
+
+    public LocalFSWorkspace() {
+        String userHome = System.getProperty("user.home");
+        String userKey = userHome;
+        if (userHome.lastIndexOf(File.separator) > 0) {
+            userKey = userHome.substring(userHome.lastIndexOf(File.separator) + 1, userHome.length());
+        }
+        wscred = new WsCredentials(userKey);
+    }
+
+    public LocalFSWorkspace(WsCredentials cred) {
+        wscred = cred;
+    }
+
+    public static final String WS_ROOT_DIR = ".";
+
+    private static final Logger.LoggerImpl LOG = Logger.getLogger();
+
+    private void createIfNotExist(String parent) {
+        new File(parent).mkdirs();
+    }
+
+
+    @Override
+    public String getProp(PROPS prop) {
+        PROPS propFound = null;
+        for (PROPS p : PROPS.values()) {
+            if (p.name().equalsIgnoreCase(prop.name())) {
+                propFound = p;
+            }
+        }
+        String propVal = prop + " doesn't exist";
+        if (propFound.equals(PROPS.ROOT_URL)) {
+            propVal = WS_HOST_URL;
+        } else if (propFound.equals(PROPS.ROOT_DIR)) {
+            propVal = WS_ROOT_DIR;
+        } else if (propFound.equals(PROPS.AUTH)) {
+            propVal = "NONE";
+        } else if (propFound.equals(PROPS.PROTOCOL)) {
+            propVal = PROTOCOL.WEBDAV.name();
+        }
+
+        return propVal;
+    }
+
+
+    @Override
+    public WsResponse getList(String parentUri, int depth) throws IOException {
+        throw new IllegalArgumentException("not implemented");
+    }
+
+    @Override
+    public WsResponse getFile(String fileUri, File outputFile) throws WsException {
+        try {
+            byte[] bytes = Files.readAllBytes(Paths.get(getResourceUrl(fileUri)));
+            FileUtils.writeByteArrayToFile(outputFile, bytes);
+            return WsUtil.success(200, "Downloaded", outputFile.getAbsolutePath());
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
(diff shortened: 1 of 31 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Get file relative to ws home
+     *
+     * @param wspaceRelFile remote URI relative file path , i.e. "x/y/z/aaa.tbl" will be located under WS_HOME at WS_URL
+     * @param outputFile    local file downloaded
+     * @return {@link WsResponse} response string should contain the abs path of the file TODO could contain more metadata
+     * @throws WsException handle ws error
+     */
+    WsResponse getFile(String wspaceRelFile, File outputFile) throws WsException;
```
loitly: This is fine if you want to save the content into a file.  What if you wanted to do something else with it, like putting it into a database, or converting it into a json object, etc?
I suggest overloading this method with one that takes a Consumer<InputStream>.  That way, you allow other usages without giving up connection/state management.

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
```diff
@@ -0,0 +1,81 @@
...
+    WsResponse delete(String wspaceRelPath) throws WsException;
+
+    /**
+     * Create parent folder relative to ws home under
+     *
+     * @param newRelativeFolderPath relative to ws home folder path
+     * @return {@link WsResponse}
+     * @throws WsException handle ws error
+     */
+    WsResponse createParent(String newRelativeFolderPath) throws WsException;
```
loitly: Is there a reason why you call it 'Parent' instead of folder?  Also, javadoc can be improve.
- should include behavior when intermediate folders are missing.  auto create, or excaption?

src/firefly/test/edu/caltech/ipac/firefly/ws/WsPutGetTest.java
```diff
@@ -0,0 +1,118 @@
...
+        try {
+            byte[] bytes = jackrabbitGet(testUrl);
+            assertTrue("Size uploaded " + size + " not the same found " + bytes.length, bytes.length == size);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+    }
+
+    byte[] jackrabbitGet(String uri) throws Exception {
```
loitly: I would expect to see our 'getFile' test here instead of another method which is not used in our code.

src/firefly/test/edu/caltech/ipac/firefly/ws/client/WebDAVClient.java
```diff
@@ -0,0 +1,375 @@
...
+import org.apache.jackrabbit.webdav.client.methods.VersionControlMethod;
+import org.apache.jackrabbit.webdav.version.report.ReportInfo;
+import org.apache.jackrabbit.webdav.version.report.ReportType;
+
+
+/**
+ * Created by ejoliet on 6/16/17.
+ * Example of Webdav Client.
+ */
+public class WebDAVClient {
```
loitly: How is this used?  I don't get it.  It looks like another implementation of some of the functions in the WsManager.  What is it trying to test?

src/firefly/java/edu/caltech/ipac/firefly/server/ws/WebDAVWorkspaceHandler.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    Map<String, WorkspaceManager> wsPool;
+
+    public WebDAVWorkspaceHandler() {
+        wsPool = new HashMap<>();
+    }
+
+    public WorkspaceManager withCredentials(WsCredentials cred) {
+
+        synchronized (cred) {
```
ejoliet: i've put it around synchronized in case we use it as a thread in a background stage. But as long as the object is not null, is fine, it will block.
The pool keys are the id (which is the unique key for identifying the manager. As a string you don't need to overwrite equals. 
The webdav implementation using http client rely on the id key, so unless the password/cookies are null, the workspace manager will access as anonymous with an id string. id can't be null. I was thinking that from the UI we could pass the cookie session as id or a random string. That needs to be discussed and added together with the file picker (probably need a saga for that i guess... what do you think?

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
ejoliet: That's a good idea. How the consumer would look like? Do you wnat me to add it to the interface such as: `getFile(String relFile, WsConsumer<InputStream> pc)` and the runnable class leaving empty run to implement? Do you want to add a database one already if you have that in mind?

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
ejoliet: By default it will create the parent recursively, but i guess i can add a boolean `isRecursive` or something and fail if `false`. What do you think?

src/firefly/test/edu/caltech/ipac/firefly/ws/WsPutGetTest.java
ejoliet: i can change that, but i also want to keep jackrabbit ;-)

src/firefly/test/edu/caltech/ipac/firefly/ws/client/WebDAVClient.java
ejoliet: Just for test purposes, that's why is located under `test`folder. It's a client webdav using jackrabbit implementation.

src/firefly/java/edu/caltech/ipac/firefly/server/query/IpacTablePartProcessor.java
```diff
@@ -616,11 +594,22 @@ protected File createFile(TableServerRequest request) throws IOException {
...
+        File file = null;
         if (doCache()) {
             if (useWorkspace) {
-                file = ServerContext.getRequestOwner().getWsManager().createFile(
-                        getWspaceSaveDirectory(), getFilePrefix(request), fileExt);
+                //TODO can we discuss if this is the right behaviour to put the temp file in ws?
+                //TODO The idea was to use maybe pubspace for temp but the user ws should be populated only if the user request it
+                //TODO Meaning that the UI needs to expose a control for that
+                // Put it in WS via local FS
+                ServerContext.getRequestOwner().getWsManager().createWsLocalFile(getWspaceSaveDirectory(), getFilePrefix(request), fileExt);
```
ejoliet: So i guess we can remove the code?

loitly: Yes, I think it should be removed.

src/firefly/java/edu/caltech/ipac/firefly/server/ws/WebDAVWorkspaceHandler.java
loitly: Sure, we can talk about it when the time comes.

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
loitly: `Consumer<T>` is an actual jdk interface made for this type of usecase.  You can use it as is.. ie.  `getFile(String relFile, WsConsumer<InputStream> consumer)`

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
loitly: Or, just document its behavior.

src/firefly/java/edu/caltech/ipac/firefly/server/ws/Workspaces.java
ejoliet: yes, i think is used in producer/consumer pattern but we can add wsexceptions and more workspace related. I will add the method for now and no implementation, how about that?

ejoliet:
I've made changes, added tests, clean up code and also fix the file loader for tests to find the path relative to the running directory. I'll wait for more comments from others that might want to checkout the code.

src/firefly/java/edu/caltech/ipac/firefly/server/LocalFSWorkspace.java
```diff
@@ -79,6 +80,11 @@ public WsResponse getList(String parentUri, int depth) throws IOException {
     }
 
     @Override
+    public WsResponse getFile(String fileUri, Consumer consumer) throws WsException {
+        throw new IllegalArgumentException("not implemented");
+    }
```
loitly: Consumer should be `Consumer<InputStream>`.
And if you want to implement it, it's pretty straight forward.  Move logic to new method, then delegate old one like below.
```
    public void getFile(String fileUri, File outputFile) throws WsException {
        getFile(fileUri, inStream -> {
            try {
                FileUtil.writeToFile(inStream, outputFile);
            } catch (IOException e) {
                throw new WsException(e);
            }
        });
    }
```

ejoliet: Thanks for the review. Yes, I've left out the parameter type to consume because i thought it could be anything we want, not only inputstream. 
I will merge and we can add the implementation and specifying the generic type at the same time.

PR #409, opened 2017-07-11
IRSA-540, IRSA-541:Add PTF case to Time Series Tool

lznakano:
Two tickets were implemented together.  A new PTF form is created and added. Corresponding actions in the form were implemented.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcConverterFactory.js
+++ b/src/firefly/js/templates/lightcurve/LcConverterFactory.js
@@ -14,7 +14,7 @@ import {LsstSdssSettingBox, lsstSdssOnNewRawTable, lsstSdssOnFieldUpdate, lsstSd
 import {DefaultSettingBox, defaultOnNewRawTable, defaultOnFieldUpdate, defaultRawTableRequest} from './generic/DefaultMissionOptions.js';
 import {BasicSettingBox, basicOnNewRawTable, basicOnFieldUpdate, basicRawTableRequest, imagesShouldBeDisplayed} from './basic/BasicMissionOptions.js';
 import {WiseSettingBox, wiseOnNewRawTable, wiseOnFieldUpdate, wiseRawTableRequest,isValidWiseTable} from './wise/WiseMissionOptions.js';
-import {PTFSettingBox, ptfOnNewRawTable, ptfOnFieldUpdate, isValidPTFTable} from './ptf/PTFMissionOptions.js';
+import {PTFSettingBox, ptfOnNewRawTable, ptfOnFieldUpdate, ptfRawTableRequest,isValidPTFTable} from './ptf/PTFMissionOptions.js';
 
 import {LC} from './LcManager.js';
 
@@ -125,8 +125,8 @@ const converters = {
         missionName: 'PTF',
         MissionOptions: PTFSettingBox,
         onNewRawTable: ptfOnNewRawTable,
-        onFieldUpdate: wiseOnFieldUpdate,  //same as in wise
-        rawTableRequest: wiseRawTableRequest, //the request is same now.
+        onFieldUpdate: ptfOnFieldUpdate,
+        rawTableRequest: wiseRawTableRequest, //the r
         yErrNames: '',
         dataSource: 'pid',
         webplotRequestCreator: getWebPlotRequestViaPTFIbe,
--- a/src/firefly/js/templates/lightcurve/LcConverterFactory.js
+++ b/src/firefly/js/templates/lightcurve/LcConverterFactory.js
@@ -13,8 +13,8 @@ import {basicURLPlotRequest} from './basic/BasicPlotRequests';
 import {LsstSdssSettingBox, lsstSdssOnNewRawTable, lsstSdssOnFieldUpdate, lsstSdssRawTableRequest} from './lsst_sdss/LsstSdssMissionOptions.js';
 import {DefaultSettingBox, defaultOnNewRawTable, defaultOnFieldUpdate, defaultRawTableRequest} from './generic/DefaultMissionOptions.js';
 import {BasicSettingBox, basicOnNewRawTable, basicOnFieldUpdate, basicRawTableRequest, imagesShouldBeDisplayed} from './basic/BasicMissionOptions.js';
-import {WiseSettingBox, wiseOnNewRawTable, wiseOnFieldUpdate, wiseRawTableRequest, isBasicTableUploadValid} from './wise/WiseMissionOptions.js';
-import {PTFSettingBox, ptfOnNewRawTable, ptfOnFieldUpdate, ptfRawTableRequest} from './PTF/PTFMissionOptions.js';
+import {WiseSettingBox, wiseOnNewRawTable, wiseOnFieldUpdate, wiseRawTableRequest,isValidWiseTable} from './wise/WiseMissionOptions.js';
+import {PTFSettingBox, ptfOnNewRawTable, ptfOnFieldUpdate, isValidPTFTable} from './ptf/PTFMissionOptions.js';
 
 import {LC} from './LcManager.js';
 
@@ -24,7 +24,6 @@ export const coordSysOptions = 'coordSysOptions';
 /**
  * A function to create a WebPlotRequest from the given parameters
  * @callback WebplotRequestCreator
- * @param {TableModel} tableModel
  * @param {number} hlrow
  * @param {number} cutoutSize
  * @param {Object} params - mission specific parameters
@@ -94,7 +93,7 @@ const converters = {
         dataSource: 'frame_id',
         webplotRequestCreator: getWebPlotRequestViaWISEIbe,
         shouldImagesBeDisplayed: () => {return true;},
-        isTableUploadValid: isBasicTableUploadValid,
+        isTableUploadValid: isValidWiseTable,
         yNamesChangeImage: [],
         showPlotTitle:getPlotTitle
     },
@@ -114,7 +113,7 @@ const converters = {
         yErrNames: ['magErr', 'tsv_fluxErr'],
         webplotRequestCreator: makeLsstSdssPlotRequest,
         shouldImagesBeDisplayed: () => {return true;},
-        isTableUploadValid: () => {return true;},
+        isTableUploadValid: () => {return {errorMsg:undefined, isValid:true};},
         yNamesChangeImage: []
     },
     'ptf': {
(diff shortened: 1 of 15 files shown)


src/firefly/js/templates/lightcurve/PTF/PTFMissionOptions.js
```diff
@@ -0,0 +1,292 @@
...
+        const tblModel = getTblById(LC.RAW_TABLE);
+        var getList = (val, type) => {
+            var colType = (!type || type === 'numeric') ?
+                ['double', 'd', 'long', 'l', 'int', 'i', 'float', 'f'] : ['char', 'c', 's', 'str', 'double', 'd', 'long', 'l', 'int', 'i', 'float', 'f'];
+
+            return get(tblModel, ['tableData', 'columns']).reduce((prev, col) => {
+                if ((colType.includes(col.type)) &&
+                    (!has(col, 'visibility') || get(col, 'visibility') !== 'hidden') &&
+                    (col.name.startsWith(val))) {
+                    prev.push(col.name);
```
loitly: Same comment here.  This is copy/paste code that should be in util.

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -105,6 +105,12 @@ export class LcViewer extends PureComponent {
                     errorMsg = `The uploaded table is not valid. The ${getMissionName(converterId)} option requires frame_id, source_id, or both scan_id and frame_num.
                     Please select the "Other" upload option for tables that do not meet these requirements.`;
                 }
+
+                if (converterId === 'ptf') {
```
loitly: LcViewer is generic code and should not have 'ptf' references.

src/firefly/js/templates/lightcurve/LcUtil.jsx
```diff
@@ -16,6 +16,23 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
...
+ */
+export function getOnlyNumericalCol(rawTbl) {
+
+    var colType = ['double', 'd', 'long', 'l', 'int', 'i', 'float', 'f'];
+    return get(rawTbl, ['tableData', 'columns']).reduce((prev, col) => {
+        if ((colType.includes(col.type)) &&
+            (!has(col, 'visibility') || get(col, 'visibility') !== 'hidden')) {
+            prev.push(col.name);
+        }
+        return prev;
```
loitly: These sort of functions should be in TableUtil.js.  It would also be easier to read if written as:
```
        return get(rawTbl, ['tableData', 'columns'])
                .filter((col) => colType.includes(col.type) && get(col, 'visibility') !== 'hidden')
                .map((col) => col.name);


```

src/firefly/js/templates/lightcurve/PTF/PTFMissionOptions.js
```diff
@@ -0,0 +1,292 @@
...
+    }, {});
+}
+
+
+/**
+ * Pregex pattern for wise, at least to find mjd and w1mpro if present
+ * @type {string[]}
+ */
+const xyColPattern = ['\\w*jd\\w*', 'w[1-4]mpro\\w*'];
+export function ptfOnNewRawTable(rawTable, missionEntries, generalEntries, converterData, layoutInfo) {
```
loitly: > xyColPattern = ['\\w*jd\\w*', 'w[1-4]mpro\\w*'];

This is wise's specific.  I doubt it'll work for ptf.

src/firefly/js/templates/lightcurve/LcUtil.jsx
```diff
@@ -16,6 +16,23 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
     };
 }
 
+/**
+ * Returns only numerical column names form raw lc table
+ * @param {Object} rawTbl
+ * @returns {TableColumn[]} - array of table columns objects
```
loitly: This is wrong.  It's returning string[] - array of column names, not TableColumn[].

src/firefly/js/templates/lightcurve/PTF/PTFMissionOptions.js
wmiipac: Lijun, As I have commented before both PTFMissionOptions and PTFplotRequests will be rewritten for PTF data when implementing IRSA-514,  don't need to make any changes as far as to dealing with PTF data specifics.

lznakano:
Refactor and rearrange the codes will be done after real PTF specific parameters are in place.

ejoliet:
Also, the band label should display a lower-case 'g' not 'G'. 'R' is fine. 
Thanks!

src/firefly/js/templates/lightcurve/LcViewer.jsx
ejoliet: Loi, this code has been introduced to deal with specific error and we said last we will cleanup and refactor the code but never get to it. I think we have time to do it and i propose to Lijun to go ahead and spend time to at least do this part and cleanup as much as possible, specially because we will use again for ZTF.

@lznakano , my proposal is the following: make use of {{converters.wise.isTableUploadValid}} function from the `LcConverterFactory.js`.
I've introduced this function in the past but is not used in the common code, only in `WiseMissionOption.js`. 

Among other cleanup that you might want to do and simply the code, I'd suggest at least to replace this piece of code:

```javascript
if (converterId === 'ptf') {
                    errorMsg = `The uploaded table is not valid. The ${getMissionName(converterId)} option requires pid.
                    Please select the "Other" upload option for tables that do not meet these requirements.`;
                }
```
with:

```js
errorMsg =  converterData.isTableUploadValid(converterId)
```
or something similar.

Right now, the function `isTableUploadValid` returns a boolean value but you can return a function with ```{errorMsg:'yada yada', isValid:boolean}``` and use it in `LcViewer.jsx` to show the error message if validation fails. 
Of course, you will need to add a specific function for PTF in `PTFMissionOptions.js`.

What do you think? @loitly @lznakano

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -96,7 +96,7 @@ export class LcViewer extends PureComponent {
         const LcPeriodInstance=  get(getAppOptions(), 'charts.chartEngine')==='plotly' ? LcPeriodPlotly : LcPeriod;
 
         var mainView = (err,converterId) => {
-            if (!isEmpty(error) && converterId) {
+            if (error!='undefined' && !isEmpty(error) && converterId) {
```
ejoliet: i thought `isEmpty` functionn from lodash covered the case ```error!='undefined'```, is not? WOW, I'm surprised! i've used elsewhere wrongly then. @loitly @robyww did you know that?

src/firefly/js/templates/lightcurve/PTF/PTFMissionOptions.js
ejoliet: @wmiipac for ticket IRSA-514, you only need to care about `PTFPlotRequests.js`. Why do you care about `PTFMissionOptions.js` ? You shouldn't, right?
I agree that the request will be different of course but is fair to add code as part of the PR so you get a code to change only ;-)

src/firefly/js/templates/lightcurve/LcConverterFactory.js
```diff
@@ -5,14 +5,17 @@
...
 import {getWebPlotRequestViaWISEIbe} from './wise/WisePlotRequests.js';
 import {makeLsstSdssPlotRequest} from './lsst_sdss/LsstSdssPlotRequests.js';
 import {makeURLPlotRequest} from './generic/DefaultPlotRequests.js';
 import {basicURLPlotRequest} from './basic/BasicPlotRequests';
 import {LsstSdssSettingBox, lsstSdssOnNewRawTable, lsstSdssOnFieldUpdate, lsstSdssRawTableRequest} from './lsst_sdss/LsstSdssMissionOptions.js';
-import {WiseSettingBox, wiseOnNewRawTable, wiseOnFieldUpdate, wiseRawTableRequest, isBasicTableUploadValid} from './wise/WiseMissionOptions.js';
 import {DefaultSettingBox, defaultOnNewRawTable, defaultOnFieldUpdate, defaultRawTableRequest} from './generic/DefaultMissionOptions.js';
 import {BasicSettingBox, basicOnNewRawTable, basicOnFieldUpdate, basicRawTableRequest, imagesShouldBeDisplayed} from './basic/BasicMissionOptions.js';
+import {WiseSettingBox, wiseOnNewRawTable, wiseOnFieldUpdate, wiseRawTableRequest, isBasicTableUploadValid} from './wise/WiseMissionOptions.js';
+import {PTFSettingBox, ptfOnNewRawTable, ptfOnFieldUpdate, ptfRawTableRequest} from './PTF/PTFMissionOptions.js';
```
ejoliet: Please, change to lower-case 'ptf' directory. it make more sense and is cleaner. Thanks.

src/firefly/js/templates/lightcurve/LcViewer.jsx
loitly: isEmpty(undefined) will return true.  but here, it's checking for a string 'undefined'.

src/firefly/js/templates/lightcurve/LcViewer.jsx
loitly: `error` is a string and meant to be used as an error message.  So, the code here can simply display the error message if exists.  It should not even care about converterId.  Mission specific message can be set in `onNewRawTable`.  You can see `wiseOnNewRawTable` as an example.

src/firefly/js/templates/lightcurve/LcUtil.jsx
lznakano: This method was in WiseMissionOption.  I moved it here without making any changes.  I will make the necessary changes and then put it in TableUtils.js.

lznakano:
I have done lot of refactoring.  There are so many duplications.   I combined some methods to one which can be used by all missions and also moved some common methods to the LcUtil.jsx.  After PTF images are added,  it is possible to do further refactoring.

I have pushed ife branch with the same name "IRSA-540-IRSA-541-PTF".  Please check out firefly and ife branches and have a look.  Thanks!

ejoliet:
I've built it and didn't see any problem. PTF case is in IRSAViewer.
Code refactored looks cleaner!
You might want to ask quick review by @loitly  before merging though.
Once merge, let @wmiipac know to add/fix the IBE function to get the images.
Thank you!

lznakano:
There are some duplications in wise and ptf options.  I removed them last Friday but added back today as place holders. RawTableRequest and FiledUpdate are such functions.  I would wait until the PFT image loaded to see if these two functions in two mission need to be different or not.

src/firefly/js/templates/lightcurve/LcUtil.jsx
```diff
@@ -146,10 +146,10 @@ export function getInitialDefaultValues(labelWidth, missionName) {
...
-                    'Select WISE band for images to be displayed',
        'Image display:', 70)),
-                [LC.META_ERR_CNAME]: Object.assign(getTypeData(LC.META_ERR_CNAME, '',

-                    'value error column name',
       'Error Column:', labelWidth))
+                [LC.META_FLUX_BAND]: Object.assign(getTypeData(LC.META_FLUX_BAND, '',        '' +
+                    'Select WISE band for images to be displayed',        'Image display:', 70)),
+                [LC.META_ERR_CNAME]: Object.assign(getTypeData(LC.META_ERR_CNAME, '',
+                    'value error column name',       'Error Column:', labelWidth))
```
loitly: FYI: I removed a few non-ascii characters from here.  However, the code could use some cleaning up.  Object.assign() is used with only 1 parameter, why?  Also, why is there a `'' + 'Select...'`

PR #410, opened 2017-07-12
DM-10948: Slate View: beginning work on

robyww:
DM-10948: Slate View

   - support a window/grid like layout
   - can be controlled remotly
   - removed unused WebPlotReuest: set/getMultiImageSupport
   - started using reactGridLayout package (already in package.json)
   - added react-sizeme
   - add demo/ffapi-slate-test.html

_To Test:_

- http://localhost:8080/firefly/demo/ffapi-slate-test.html
- start clicking on links

_Known issues and todo_

- Charts, coverage, and Image Metadata viewer are not removable
- Not yet a way to add a general chart that will connect to all tables in main group as in triview.
- Net yet a python binding.


src/firefly/js/templates/fireflyslate/FireflySlateManager.js
```diff
@@ -0,0 +1,216 @@
...
+
+
+        if (newLayoutInfo !== layoutInfo) {
+            dispatchUpdateLayoutInfo(newLayoutInfo);
+        }
+    }
+}
+
+
+function startSepcialViewerSaga(action, alreadyStarted) {
```
loitly: Is there a stop?  I see `dispatchRemoveCell`, but not where saga were stopped.

src/firefly/js/templates/fireflyslate/GridLayoutPanel.jsx
```diff
@@ -0,0 +1,231 @@
...
+
+        const normalLayout= makeNormalLayout(this.props.gridView);
+        const nextNormalLayout= makeNormalLayout(nP.gridView);
+        // todo compare them - make sure everything are integers
+    }
+
+    renderedLayoutUpdated(layout) {
+        console.log('onLayoutChange');
+        console.log(layout);
+        console.log(this.normalLayoutTMP);
```
loitly: Don't forget to remove debugging logs.

src/firefly/js/visualize/ui/ImageSelectPanelResult.js
```diff
@@ -348,13 +349,18 @@ export function resultSuccess(plotInfo, hideDropdown = false) {
             });
         }
 
+        if (request[panelKey].createNewCell==='newCell' && request[panelKey][keyMap['plotmode']].startsWith('create'))  {
```
loitly: Props are pretty deep in the request.  There's no chance of TypeError?

loitly:
Too many missing files from checkin.  Cannot continue testing.  Please confirm and push the missing files:
```
- './react-grid-layout_styles.css';
- './react-resizable_styles.css';
- slate.html

```

loitly:
From UI testing:
- jsdoc is missing from width/height parameters.  It was not clear to me what unit they were in while reviewing the code.  Only after looking at example, did I see that they are row/col span.

- row/col parameters does not work as expected. 
  - fixed row, toggle col between 0 and 1. ` loadHistogram(0,0,1,1)`, then `loadXY(0,1,1,1)` works, but when I reverse the order, `loadXY(0,1,1,1)`, then  loadHistogram(0,0,1,1) they appears wrong.
  - same when I toggle the row parameter.  `loadHistogram(1,0,1,1)`, then `loadXY(0,0,1,1)` they appeared wrong

- width parameter need more explanation.
  - `loadHistogram(0,0,2,1)`, then `loadXY(0,1,1,1)`, I would expect them to be side-by-side with histogram having 2x the width.  But, they are  of equals width.

- height parameter works inconsistently.
  - `loadHistogram(0,0,1,1)`, then `loadXY(0,1,1,2)` works as expect.  side-by-side with XY having 2x the height
  - but, `loadHistogram(0,0,1,2)`, then `loadXY(0,1,1,1)` does not work.  they both have the same height.

The added feature is amazingly powerful.  However, we'll need to include good documentation to explain clearly how it suppose to work.

src/firefly/js/templates/fireflyslate/FireflySlateManager.js
robyww: I don't think it needs a stop since the saga's don't really take that much resources. I also don't think the basic use case will be to be bringing all different types and removing them.  The user will get a layout he likes for his work put up the data

src/firefly/js/visualize/ui/ImageSelectPanelResult.js
robyww: I just modified the panel.  This seems to be how it was written.

tgoldina:
It's not quite clear how resize works: it seems that first cell added is nor resizable or movable.
Here is what I did: loaded entry point html, clicked 'load Moving Table' link, tried to reduce the size of the cell in the viewer – no effect, clicked on 'Image metadata viewer' - table in viewer is now minimal size (but resize started to work in most cases). Added XYPlot - the upper two cells got much smaller.

tgoldina:
Some components (responding to mouse drag) are hard to move. For example, Coverage can be moved only if you drag by the border.

src/firefly/html/demo/ffapi-slate-test.html
```diff
@@ -0,0 +1,364 @@
...
+Add Different types of image displays
+<ul>
+    <li>
+        <a href="javascript:loadImageDataDisplay(2, 0,4,2)">Image meta data viewer</a> display <span class='smallCoords'>at row: 2, col: 1, w:4 h: 2</span>
+    </li>
+    <li>
+        <a href="javascript:load3C(0,4,2,2)">3 Color</a> in cell <span class='smallCoords'>at row: 0, col: 4, w:2 h: 2</span>
+    </li>
+    <li>
+        <a href="javascript:loadMoving(2, 4,2,2)">load Moving Images</a> <span class='smallCoords'>at row: 4, col: 1, w:4 h: 2</span>
```
cwang2016: should this be "row: 2, col: 4, w: 2, h: 2" ?

cwang2016:
I tried the following in order 
1. load Moving Table at row: 0, col: 0, w:4 h: 2
2. Image meta data viewer display at row: 2, col: 1, w:4 h: 2
after I added 2, the height of 1 becomes shorter than 2. should they look with the same height?

robyww:
They should be the same height and for me they are. The description is wrong for the image meta data viewer I will change it to be `at row: 2, col: 0, w:4 h: 2`

robyww:
In general the size of the grid is determined by what you add.  If you add only one element then it should not be movable and take up the whole screen.

Also - i am still trying to understand how react-grid-layout works.  It does try to resolve conflict in coordinates.

src/firefly/js/templates/fireflyslate/GridLayoutPanel.jsx
```diff
@@ -0,0 +1,231 @@
...
+        const {gridView,size:{width,height}}= this.props;
+        this.renderedGridView= gridView;
+
+
+        if (isEmpty(gridView) || width<10 || height<10) return <div  style={{flex: '1 1 auto', overflow: 'auto'}}/>;
+
+        const sourceGridDim= getGridDim(gridView);
+        const maxSize= getGridDim(gridView, CV);
+        const small= makeSmallLayout(gridView);
+        const normalLayout= makeNormalLayout(gridView);
```
tgoldina: When non-normal layouts are used?

src/firefly/js/api/ApiViewer.js
```diff
@@ -184,9 +241,9 @@ function buildChartPart(channel,file,dispatch) {
      * @memberof firefly.ApiViewer
      * @public
      */
-    const showHistogram= (histogramOptions) => {
+    const showHistogram= (histogramOptions, viewerId) => {
```
tgoldina: viewerId jsDoc needs to be added to both showHistogram and showXYPlot

src/firefly/js/templates/fireflyslate/GridLayoutPanel.jsx
robyww: I allow for a single column layout when the app is very narrow.

src/firefly/js/api/ApiViewer.js
robyww: added

PR #411, opened 2017-07-14
Heatmap (density) plot in multitrace chart

tgoldina:
https://jira.lsstcorp.org/browse/DM-10832

Added heatmap trace type to multitrace chart with options and server side connection.

Known limitations:
- The only way to recalculate  heatmap is to change table, filter from the plot, or change heatmap parameters. The map is not recalculated on zoom or when axis limits change
- Table highlight is not shown - it's not clear if we should support this connection
colorbar can overlap.
- It's not possible to set from UI which heatmap trace is on top. (Usually the last trace in the data array is the one on top.) We might need a separate ticket for placing active trace (of the same type) on top of the others.
- Colorbar can overlap with the legend, other colorbar ticks. It's not easy to position it perfectly in generic case, when multiple colorbars (and legend) are present, but it is possible to set its length and location in the data array, when you know what exactly will be present on the chart.

### DIFF
--- a/src/firefly/js/charts/ui/options/HeatmapOptions.jsx
+++ b/src/firefly/js/charts/ui/options/HeatmapOptions.jsx
@@ -4,15 +4,16 @@ import {get, isUndefined} from 'lodash';
 import {getChartData} from '../../ChartsCntlr.js';
 import {FieldGroup} from '../../../ui/FieldGroup.jsx';
 
+import {toBoolean} from '../../../util/WebUtil.js';
 import {intValidator} from '../../../util/Validate.js';
 import {ValidationField} from '../../../ui/ValidationField.jsx';
 import {ListBoxInputField} from '../../../ui/ListBoxInputField.jsx';
 import {CheckboxGroupInputField} from '../../../ui/CheckboxGroupInputField.jsx';
 import {SimpleComponent} from '../../../ui/SimpleComponent.jsx';
 import {BasicOptionFields, OptionTopBar, basicFieldReducer, submitChanges} from './BasicOptions.jsx';
-import {addColorbarChanges} from '../../dataTypes/FireflyHeatmap.js';
 import {getColValStats} from '../../TableStatsCntlr.js';
 import {ColumnOrExpression} from '../ColumnOrExpression.jsx';
+
 
 
 const fieldProps = {labelWidth: 62, size: 15};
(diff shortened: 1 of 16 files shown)


src/firefly/js/charts/ui/options/HeatmapOptions.jsx
```diff
@@ -0,0 +1,165 @@
...
+    const dataType = (!tbl_id) ? 'heatmap' : 'fireflyHeatmap';
+    const changes = {
+        [`fireflyData.${activeTrace}.type`] : 'heatmap',
+        [`fireflyData.${activeTrace}.dataType`] : dataType
+    };
+
+    Object.assign(changes, fields);
+
+    // reversescale is boolean
+    changes[`data.${activeTrace}.reversescale`] = get(fields, `data.${activeTrace}.reversescale`, '').includes('true');
```
loitly: there's a `WebUtil.toBoolean` function for this, if you want to use it.

ejoliet:
Great feature!
I've played with it. It looks neat. I have a couple of comments:
- highlighted dot not visible enough depending on the color selected
- when i interact with the data, sometime selection or highlighted dots are not visible or filter is not applied
- No legend/colormap on the plot to indicate the range
- Do we want the background of the tooltip to be the same color as the dot or do we want to fix it?

Thanks.

tgoldina:
In latest commit:
- the default colorbar is added in below the existing, to avoid overlapping.
- color field are only shown if it data.marker.color property is available for the given trace type
- as an attempt to avoid legend and colorbar overlap, legend is switched to the bottom (horizontal configuration) when the default colorbar is added.

Heatmap does not show selected or highlighed from the table, only scatter traces shows them.

The example is a top chart at http://localhost:8080/firefly/demo/ffapi-highlevel-charttest.html

For scatter color maps, we do not display colorbar at the moment, as Emmanuel has noted. There is a ticket (https://jira.lsstcorp.org/browse/DM-11064) to address this issue.

Regarding tooltip background, we'd like to keep as much Plotly behavior as possible. User can override it by giving custom layout or data configuration.

Please, file the ticket for multitrace charts, if you see problems with filter, selection, or highlight. Selected and highlighted points are shown for active trace only, when active trace is a scatter. You can change active trace by using the listbox on the toolbar.

We might want to create a ticket to place an active trace on top of other traces of the same type. (Plotly keeps its own order for the different trace types.) Currently the latest trace is displayed on top, the tooltips will reflect the topmost trace.

PR #413, opened 2017-07-15
IRSA-601: labels change in error bars options

ejoliet:
Labels for error bars option have changed.

X Err -> Error
Sym -> Symm

I took the opportunity to cleanup the code and remove duplicated constant.

Please, confirm change in UI and check that i didn't mess up the code. Thanks.

### DIFF
--- a/src/firefly/js/charts/ui/options/ScatterOptions.jsx
+++ b/src/firefly/js/charts/ui/options/ScatterOptions.jsx
@@ -119,7 +119,7 @@ export function fieldReducer({chartId, activeTrace}) {
             fieldKey: errorMinusFieldKey(activeTrace, 'x'),
             value: get(tablesourceMappings, ['error_x.arrayminus'], ''),
             //tooltip: 'X error',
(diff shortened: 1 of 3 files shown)


src/firefly/js/charts/ui/options/ScatterOptions.jsx
```diff
@@ -119,7 +119,7 @@ export function fieldReducer({chartId, activeTrace}) {
             fieldKey: errorMinusFieldKey(activeTrace, 'x'),
             value: get(tablesourceMappings, ['error_x.arrayminus'], ''),
             //tooltip: 'X error',
-            label: 'X Err\u2193:',
+            label: 'Error\u2193:',
```
tgoldina: We might want to remove tooltip and label properties from error and errorMinus fields here - they have no effect here and are actually defined in the Errors' render (as props to ColumnOrExpression).

wmiipac:
Looks good. approved for merge.

PR #415, opened 2017-07-19
DM-10345: provide management for long running searches

loitly:
https://jira.lsstcorp.org/browse/DM-10345

Normally, when a search is submitted, a table will appears masked with a loading icon.
The masked panel will be replaced with table data when the search results returned.
With this PR, when a search is backgroundable, a 'send to background' button will be added to the bottom of the loading icon.    If pressed, it will close the masked table and send the search to the background.  The job will appear in the Background Monitor. 
Per Xiuqin request, backgroundable is set to true by default.
This feature will be available to all Firefly's derived projects, ie /irsaviewr, /suit.

To test:
http://localhost:8080/firefly/
- Catalogs search with a larger radius
- While loading, click 'send to background'
- If button is not clicked, it will load when results returned.

Also:  this PR also fixes https://jira.lsstcorp.org/browse/DM-11252

### DIFF
--- a/src/firefly/js/api/ApiHighlevelBuild.js
+++ b/src/firefly/js/api/ApiHighlevelBuild.js
@@ -123,6 +123,7 @@ function buildTablePart(llApi) {
      * @prop {string}  tbl_group    the group this table belongs to.  Defaults to 'main'.
      * @prop {number}  pageSize     the starting page size.  Will use the request's pageSize if not given.
      * @prop {boolean} removable    true if this table can be removed from view.  Defaults to true.
+     * @prop {boolean} backgroundable    true if this search can be sent to background.  Defaults to false.
      * @prop {boolean} showUnits    defaults to false
      * @prop {boolean} showFilters  defaults to false
      * @prop {boolean} selectable   defaults to true
--- a/src/firefly/js/api/ApiHighlevelBuild.js
+++ b/src/firefly/js/api/ApiHighlevelBuild.js
@@ -123,7 +123,7 @@ function buildTablePart(llApi) {
      * @prop {string}  tbl_group    the group this table belongs to.  Defaults to 'main'.
      * @prop {number}  pageSize     the starting page size.  Will use the request's pageSize if not given.
      * @prop {boolean} removable    true if this table can be removed from view.  Defaults to true.
-     * @prop {boolean} backgroundable    true if this search can be sent to background.  Defaults to false.
+     * @prop {boolean} backgroundable    true if this search can be sent to background.  Defaults to true.
      * @prop {boolean} showUnits    defaults to false
      * @prop {boolean} showFilters  defaults to false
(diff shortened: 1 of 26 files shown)


src/firefly/js/api/ApiHighlevelBuild.js
```diff
@@ -123,6 +123,7 @@ function buildTablePart(llApi) {
      * @prop {string}  tbl_group    the group this table belongs to.  Defaults to 'main'.
      * @prop {number}  pageSize     the starting page size.  Will use the request's pageSize if not given.
      * @prop {boolean} removable    true if this table can be removed from view.  Defaults to true.
+     * @prop {boolean} backgroundable    true if this search can be sent to background.  Defaults to false.
```
cwang2016: is backgroundable set to be true in default?

cwang2016:
Review completes. All looks great. 
Just some questions, 
- should provide some way to determine if 'backgroundable' or not?  
- From my end, 'Send to backgroud' button seems shown a couple of seconds later than the spinning? is that on purpose? 
- similar question as being raised by @ejoliet that 'abort' the request looks like 'pause', not really 'abort' the result, since the table is displayed quickly if I sent the same search again.

src/firefly/js/api/ApiHighlevelBuild.js
loitly: Good catch.  I made it default to true and did not update this jsdoc.

loitly:
> When having 2 catalog result, one table is focus, and trying to show it again looks like it triggers search again with spinning wheel but no 'send to background' button

'Show results' from background monitor is not backgrounbable by design.  It should be fast, since it's already completed.

> I've set an email but didn't get it (do i need smtp locally configured?)

Email notification is not implemented.  It was not a feature prior to conversion.  Also, I think it could be annoying if an email is sent on every search.  

> if i abort the job, the server logs says that the search has been cancel but later on the job ends and logs shows that it has downloaded the table effectively on the server: is this on purpose to reuse the search cache?

Yes, we remove the job, but did not kill the task.  In most cases, the task is a service call or a database query.  There no api to cancel it, yet.

> From my end, 'Send to backgroud' button seems shown a couple of seconds later than the spinning? is that on purpose?

Yes, there is a one second delay by design.

ejoliet:
>'Show results' from background monitor is not backgrounbable by design. It should be fast, since it's already completed.

Is not coming back for me, it gets stuck. Plus, it looks like is redoing the search, why not showing the table as active, why it goes around the flux flow and refresh image, table and chart?

>Email notification is not implemented. It was not a feature prior to conversion. Also, I think it could be annoying if an email is sent on every search.

If it's not implemented, please, remove the input field form the UI, is misleading.

>Yes, we remove the job, but did not kill the task. In most cases, the task is a service call or a database query. There no api to cancel it, yet.

Understand.

PR #423, opened 2017-07-25
IRSA-581: Expose to public Firefly API table selection info.

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-581

Exposes table selection information to public Firefly API.
Test:  http://localhost:8080/firefly/demo/ffapi-highlevel-test.html
To the right of the 2 table panels, there's a 'show selected' button.
Once clicked, it will display selected 'crval1' and 'crval2' values of the table to the left(http://we..)
Use view source to see code sample.
Here is the added function:
```
/**
 * returns an array of all the values for a columns
 * @param {string} tbl_id
 * @param {string[]} columnNames  defaults to all columns
 * @return {Promise.<TableModel>}
 * @func getSelectedData
 * @public
 * @memberof firefly.util.table
 */
export function getSelectedData(tbl_id, columnNames=[])


```


loitly:
> is there a way i can get the filtered table based on all the columns instead of a range of columns?

Yes, if columnNames is not given, it will return all columns.

PR #425, opened 2017-07-27
DM-11376: backgrounding job bug

loitly:
- on fail search, mask panel remained without error messages.
- API should have backgroundable set to false, since background monitor may not be available.

https://jira.lsstcorp.org/browse/DM-11376
use ticket to duplicate a search which will produce error.
use http://localhost:8080/firefly/demo/ffapi-highlevel-test.html to see that api searches are not backgroundable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ There are several branches the this repository.  Here are the ones that you shou
  -  [Java 1.8] http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html
     This is the version we are using to compile and run Firefly.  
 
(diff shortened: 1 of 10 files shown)


loitly:
After reviewing the problem, it's clear that we need to treat background searches differently then background packages.  For now, I suggest we revert searches back to not backgroundable by default and only enable the ones we see fit.  I'll enable backgrounding for catalog and top level searches in suit.  What do you think, @xiuqin ?

xiuqin:
I agree. Please create a ticket for future design discussion to treat those two types of backgrounding hob differently.

Xiuqin

loitly:
@ejoliet, I've made changes to this PR.  Please review it again at your convenience.

PR #428, opened 2017-07-28
IRSA-613: fixed: WCS match not working right for image that are very far apart

robyww:
_To test:_

Show any image with M16 and any Image NGC 2264 and WCS match

### DIFF
--- a/src/firefly/js/visualize/CsysConverter.js
+++ b/src/firefly/js/visualize/CsysConverter.js
@@ -356,7 +356,7 @@ export class CysConverter {
 
     getDevicePtCoordsOptimize(wpt, retPt) {
         const success= this.getScreenCoordsOptimize(wpt,retPt);
-        if (!success) return false;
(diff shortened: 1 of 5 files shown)


ejoliet:
In this branch, when i tested using your example and apply WCS match: one image goes out of the frame ('recenter' is displayed) and the other one stays. I guess it's the right answer for images whcih field is very far away. 
Once the image goes away and i unselect the WCS match feature, the image doesn't come back to previous position. Is that the accepted behaviour? Otherwise, everything looks ok. 
Good to go.
Thanks!

ejoliet:
BTW, i've found that if i rotate the image and then do and undo target match, once i deselect it the image goes back to initial rotation. I think it should stay.
Test case: bring moving images from demo page, then rotate to north, apply target match and uncheck it, images go back to initial position. This is not the same as WCS match, once deselected, the images stays on their position. I'm not sure what is the behavior agreed or if there a better way to walk the user thru a best flow.

PR #429, opened 2017-07-28
DM-10520: fixed "restore to defaults" changing the active image

robyww:


cwang2016:
Review completes. works well.

PR #431, opened 2017-08-01
Dm 11387 Add ability to limit what will appears as background jobs

loitly:
https://jira.lsstcorp.org/browse/DM-11387

Background jobs from one page(client app) may not be applicable in another page.  This PR add the ability to limit what appears as background jobs.

Use /firefly/firefly.html and /firefly/firefly-dev.html to test that they do share background searches.
User /irsaviewer/ and /irsaviewer/timeseries to test that they are setup to not share background jobs.


loitly:
I've push a new commit which fixes several issues related to backgrounding.
 - added a checkbox to enable email notification
 - sync the checkbox and email field between download dialog and background monitor
 - remove 'resend email' feature which is confusing
 - when email notification is enabled, an email will be sent for each job when it's completed.

PR #433, opened 2017-08-03
DM-11349 Time Series data display issues regarding overlay's position and overlay's rendering

cwang2016:
This development fixes the following issues:
1. The overlay position is not the same as the position of the selected source from LSST WISE search result. 
   PADC WISE source names ra/dec columns differently from what IRSA WISE source does. 
2. The overlay (like the PointSelection symbol by turning on 'lock by click') is not shown on the plot as using the cursor to go down the table to highlight the row. 
  All kinds of overlays, such the one for  'Select Area', 'Distance Tool', 'Marker', 'Footprint', 'Region', 'NorthUpCompass', 'Point Selection', 'Web Grid', are fixed to be rendered as the selected row of the table is changed up and down. 

Test
for issue 1, start suit: localhost:8080/suit, 
for issue 2, you may just start from  firefly: localhost:800/firefly/ts.html and upload a wise table. 

issue 1: from catalog search panel, select WISE_00 -> AllWISE Source Catalog
                                             target m81 -> search
on tri-view page, click 'view time series: xxxxxx' button
move to Time Series Tool: Viewer
- for issue 1: check the target overlay position in each plot and the source position shown in the table. 

issue 2: on Time Series Tool: Viewer, 
- select any overlay, ex. Point selection by turning on 'Lock by click',  Distance tool,, marker, or Region, etc,  and move the cursor down to select the row of the table and see the overlay is shown on all plots.


xiuqin:
The "lock by click" point still does now show up in the image as I move cursor. 
To make the test easier, I only selected to display one image at a time.

PR #439, opened 2017-08-11
DM-11569 Plotly 3d bug fixes

tgoldina:
https://jira.lsstcorp.org/browse/DM-11569

Bug fixes, related to scatter3d (and 3d plots in general):
- 2d axes should not show
- options should be basic (not scatter) and should not include options for 2d XY
- drag mode buttons should be those that are supported for 3d (temporary buttons are used for turntable and orbital rotations)
- selection button should not show unless chart supports selection
- zoom to original should work

To test, use http://localhost:8080/firefly/demo/ffapi-slate-test2.html


cwang2016:
As looking at the chart cases in ffapi-slate-test2.html, the x, y for the chart can be re-stated as an expression in 'trace' definition (like x: "tables:wiseCatTbl,w1mpro-w2mpro") for  dataType 'fireflyScatter', 'fireflyHeatmap', or 'fireflyHistogram'.  As for 'scatter3d' chart case, x, y, z doesn't work as an expression. Does that work as designed?

cwang2016:
Review completes. Works for 3d.

PR #444, opened 2017-08-18
IRSA-670: adding NED search

ejoliet:
I've added a NED panel and a processor to postprocess the result VOTable.
The panel has been added to the main component used in IRSA: CatalogSelectView.jsx

To test: go to irsaviewer, bring catalog search and go to NED tab. Enter favorite extragalactic object. The result should be a table with a object name column clickable that goes to NED with more details. 

Please, have a look and let me know.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/persistence/NedSearch.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/persistence/NedSearch.java
@@ -5,6 +5,7 @@ package edu.caltech.ipac.firefly.server.persistence;
 
 import edu.caltech.ipac.astro.IpacTableWriter;
 import edu.caltech.ipac.firefly.data.TableServerRequest;
+import edu.caltech.ipac.firefly.server.query.DataAccessException;
 import edu.caltech.ipac.firefly.server.query.ParamDoc;
 import edu.caltech.ipac.firefly.server.query.SearchProcessorImpl;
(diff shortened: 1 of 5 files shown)


ejoliet:
@loitly to investigate: although the UI result is ok, the log server shows error. When i'm debugging it, it seems that the search triggers a secondary search that create a statement based on main columns (RA,DEC) (????) here edu/caltech/ipac/firefly/server/query/IpacTablePartProcessor.java:414:
```
String ic = request.getParam(TableServerRequest.INCL_COLUMNS);
        if (decimateInfo == null && !StringUtils.isEmpty(ic) && !ic.equals("ALL")) {
...
```
And that triggers a secondary pass on the postProcessData which fails on the resulting second table because it only contains 2 columns (RA,DEC). 
Worth investigate.

ejoliet:
Ok, I've fixed the problems. It appears that the following requests are from scatter plot and statistic so i've added a way to avoid redoing post-processing when sub-sequence calls. Let me know if it make sense. Thanks!

ejoliet:
I've also made the VO search backgroundable.

ejoliet:
I see an issue: the object name column is not filterable. How can i make a column with html (links) that i can filter? ( @loitly )

ejoliet:
As @loitly reminded me, if a column has HTML format, is not sortable/filterable. Solution: add new column with link to NED. 
I will merge this and open a new ticket for adding a column.

PR #445, opened 2017-08-21
IRSA-688:IRSAViewer xyplot settings grid resets ymin

lznakano:
Fixed the axis range bug when restyle to grid


lznakano:
Fixed, please check it again.  Thanks!

PR #448, opened 2017-08-23
IRSA_672: improve image lable for plot title inline

ejoliet:
Improve image title label by changing the line height (css). 
This is also solving the problem that in smaller screen, the zoom level wasn't shown. 

Check that it doesn't break anything.


robyww:
@ejoliet try it now.  I think it is like we talk about.

ejoliet:
Looks great! Thanks for the improvement! Will merge now.

PR #449, opened 2017-08-24
IRSA-701: fixed wcs match getting out of aligement when multi image viewer goes from single to multi

robyww:
_to test_: 
ticket description: https://caltech-ipac.atlassian.net/browse/IRSA-701
also- confirm time series stays in alignment which which between 1-3-5-7

_Note 8-25-17_: fixed server more bugs:

 - found more much in wcs match and target match   
 - Target match being lost in certain cases   
 - Selection showing image bounds in unrotated case   
 - Expanded view not sizing to fit when more plots are added   
 - In certain cases deleting a plot was causing an exception


ejoliet:
I've tested it in tri-view and is working fine now. It's fixed.
 
Now, i've been testing in LC app the "target match", and i've found that somehow the "target match" is lost if i zoom in an image and then scroll through rows on the table. The images are not centered/aligned. This is a new bug discovered, it happens in dev too. A workaround is to uncheck and check back the "target match" feature. 
(A side note: LC app doesn't have single and multi grid mode)

Unless you think is related to this fix, you can merge into `rc`.

(I will raise a new ticket to be fixed in dev later on for LC target match when images are zoomed in.)

Thank you!

PR #453, opened 2017-09-06
DM-11547: dockerize apache https proxy.

loitly:
- use letsencrypt for public site, and self-signed for development
- add support for authentication via mod_auth_openidc
- expand build script to exposes configurable global variables to javascript

**self-signed https proxy locally.**
cd ./firefly; docker/proxy-dev/proxyctl.sh update
this will create and run the apache proxy in docker.  access it on port 80 or 443.  i.e.  http://localhost/

**javascript global variables**
to make any configuration property a javascript global variable, append '__$' to it.  i.e.  __$help.base.url

### DIFF
--- a/docker/proxy-dev/proxyctl.sh
+++ b/docker/proxy-dev/proxyctl.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+case "$1" in
+  start)
+        docker start proxy-dev
+        ;;
+  stop)
+        docker stop proxy-dev
+        ;;
+  shell)
+        docker exec -it proxy-dev /bin/bash
+        ;;
+  clean)
+        docker rmi $(docker images --filter "dangling=true" -q --no-trunc)
+        ;;
+  update)
+        cd /hydra/cm/firefly
+        gradle proxyDev:dockerImage
+        docker stop proxy-dev
+        docker container rm proxy-dev
+
+        docker run -d \
+                   -p 80:80 \
+                   -p 443:443 \
+                   -e "docker_host=`ifconfig | grep "inet " | grep -Fv 127.0.0.1 | awk '{print $2}'  | tail -1`" \
+                   --name proxy-dev ipac/proxy-dev
+
+#           -e "DOMAINS=`hostname`" \
+#           -e "WEBMASTER_MAIL=loi@ipac.caltech.edu" \
+#           -e "STAGING=proxy" \
+
+        ;;
+  *)
+        echo $"Usage: proxyctl.sh [start|stop|shell|clean|update]"
+        exit
+esac
+
--- a/docker/proxy-dev/proxyctl.sh
+++ b/docker/proxy-dev/proxyctl.sh
@@ -1,4 +1,41 @@
-#!/bin/sh
+#!/bin/bash
+
+
+http_port="80"
+https_port="443"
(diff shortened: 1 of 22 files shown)


robyww:
Loi and I discussed better ways to do global inserted variables.  It would work by building a global objects that includes such things as MODULE_NAME, SCRIPT_NAME, help.base.url and build one object and insert only that as global with webpack.  We think this would work. I created a separate ticket DM-11856.

docker/proxy/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM birgerk/apache-letsencrypt
+
+
+#RUN apt-get update && \
+#    apt-get -f stretch install libapache2-mod-auth-openidc && \
+#    rm -rf /var/lib/apt/lists/*
+
```
robyww: you should install the tools you need as part of the build. I don't think it will make much difference in the build time or the overall size

docker/proxy/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+#RUN ln -s /usr/lib/apache2/modules/mod_auth_openidc.so modules/mod_auth_openidc.so
+COPY ./others/*.conf /etc/apache2/conf-enabled/
+
+RUN   a2enmod proxy; \
+      a2enmod proxy_http; \
+      a2enmod proxy_wstunnel
+
+EXPOSE 80 443
+
+ENTRYPOINT ["httpd-foreground"]
```
robyww: proxy should take parameters to to set which the backend it running so that it is not required to be on 8080.  They way we can start two proxy instances in front of two firelfy servers.

docker/proxy/proxyctl.sh
```diff
@@ -0,0 +1,33 @@
...
+        docker exec -it proxy /bin/bash
+        ;;
+  update)
+        docker stop proxy
+        docker container rm proxy
+
+        docker pull ipac/proxy
+        docker run -d \
+                   -p 80:80 \
+                   -p 443:443 \
```
robyww: it you has more parameters to the docker containers the this file will have to deal with them

docker/proxy-dev/proxyctl.sh
```diff
@@ -0,0 +1,36 @@
...
+        ;;
+  shell)
+        docker exec -it proxy-dev /bin/bash
+        ;;
+  clean)
+        docker rmi $(docker images --filter "dangling=true" -q --no-trunc)
+        ;;
+  update)
+        cd /hydra/cm/firefly
+        gradle proxyDev:dockerImage
```
robyww: i don't have /hydra/cm/firefly. I think you should use the following>
```
DIR=$(dirname "${0}")
cd $DIR/../../../firefly
```

loitly:
I've just pushed a commit that address all of the issues identified in your review.  Please take a quick look and see if it addresses your concerns.  
You should be able to start one up out-of-the-box even if you do not have missing properties identified by tatiana.
Thanks.

```
./docker/proxy-dev/proxyctl.sh

Usage: proxyctl.sh [start|stop|shell|clean|update]
  optional parameters for update only:
    -http|--http_port <number>
    -https|--https_port <number>
    -host|--docker_host <x.x.x.x>  : ip address of the host running firefly
    -firefly|--firefly_port <number>

```

PR #454, opened 2017-09-06
DM-11477 Make plot.ly plotting API to handle Firefly unrecognized charts types cleanly

cwang2016:
This development includes:
- render plotly charts by giving any plotly data and layout JSON description. 
- put Firefly like tool bar for each plotly chart either recognized or unrecognized by Firefly. The tool bar rendering is classified into the following groups, 
1. basic tool bar with buttons for 
   download chart as PNG + show/edit filter + chart options and tools + Expand panel (for Pie) 
2.  1 + buttons for 
     Zoom to the original range + Zoom + Pan (for Histogram, bar, box)
3.  2 + buttons for  select (for each trace with x, y data directly mapped to table  numerical columns, like scatter, histogram2d, histogram2dcontour, heatmap)
4.  2 + buttons for orbital rotation + turntable rotation
     (for 3D-like charts, such as scatter3d, surface, mesh3d) 

- update the chart options and tools side bar content for Firefly unrecognized charts. 
   update the logic for XY options, XY range, marker color, etc. 
- fixed the issue that the mouse actions on the side bar may inaccurately move the entire grid unit of that chart. 

Test:
run demo/ffapi-slate-test3.html to get the following chart samples: 
histogram, surface, bar, pie, histogram2dcontour and box. 
The chart samples are made for showing purpose with no significant meanings.

### DIFF
--- a/src/firefly/js/charts/ui/PlotlyToolbar.jsx
+++ b/src/firefly/js/charts/ui/PlotlyToolbar.jsx
@@ -1,5 +1,5 @@
 import './ChartPanel.css';
-import React, {PureComponent} from 'react';
+import React from 'react';
 import {get, isEmpty} from 'lodash';
 
 import {dispatchChartUpdate, dispatchChartFilterSelection, dispatchChartSelect, getChartData, dispatchSetActiveTrace, dispatchChartExpanded} from '../ChartsCntlr.js';
--- a/src/firefly/js/charts/ui/PlotlyToolbar.jsx
+++ b/src/firefly/js/charts/ui/PlotlyToolbar.jsx
@@ -4,7 +4,7 @@ import {get, isEmpty} from 'lodash';
 
 import {dispatchChartUpdate, dispatchChartFilterSelection, dispatchChartSelect, getChartData, dispatchSetActiveTrace, dispatchChartExpanded} from '../ChartsCntlr.js';
 import {SimpleComponent} from '../../ui/SimpleComponent.jsx';
-import {getTblById, clearFilters, getColumnIdx} from '../../tables/TableUtil.js';
+import {getTblById, clearFilters, getColumnIdx, getColumnType} from '../../tables/TableUtil.js';
 import {dispatchSetLayoutMode, LO_MODE, LO_VIEW} from '../../core/LayoutCntlr.js';
 import {downloadChart} from './PlotlyWrapper.jsx';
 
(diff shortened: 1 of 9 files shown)


src/firefly/js/charts/ui/PlotlyToolbar.jsx
```diff
@@ -44,6 +45,38 @@ export class ScatterToolbar extends SimpleComponent {
     }
 }
 
+function isSelectable(tbl_id, chartId, type) {
```
tgoldina: I don't quite understand how we decide that selection box should be present. We might be creating unnecessary complexity here. What if x and y are expressions? I think we should not allow selection box for "unknown chart types".

cwang2016: Yes, we need to determine 
- if showing the selection box for  'unknown chart types'? 
   I had asked the original design intention if showing XY options in case x, y data is defined for 'unknown chart types', and the reply is 'the intention was to exclude XY options from the charts that do not have XY axes', so should XY options be excluded if the chart which is 'unknown type' but with x, y data?
- If not excluded, selection box is shown in case XY is defined and mapped to table columns directly (per discussion with Trey). Selection box could be rendered if x, y is expressions later once we add new columns into the table for the expression. 

This is thought that makes the code so far.

src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -236,10 +243,10 @@ const now = Date.now();
                         Plotly.Plots.resize(this.div);
                         break;
                     case RenderType.UPDATE:
-                        Plotly.update(this.div, data, layout);
+                        Plotly.update(this.div, data, cloneDeep(layout));
```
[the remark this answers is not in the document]
cwang2016: the layout object gets changed after calling Plotly function for bar chart case and that make the logic confliction in submitChanges of BasicOptions. This happens while calling Plotly.newPlot.  We could talk in more detail about this solution.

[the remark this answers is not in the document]
robyww: Cindy added  the `cloneDeep` because plotly was adding properties to the layout object.  It does not treat it as immutable data.  If this is a problem the the `optimize` then we have a difficult problem. Who can tell us for sure that it is a problem with `optimize`?

src/firefly/js/charts/ui/options/BasicOptions.jsx
```diff
@@ -62,11 +64,16 @@ function getOptions(a, layout) {
...
 
-function hasNoXY(type) {
-    return type.endsWith('3d') || ['pie','surface'].includes(type);
+/*
+ * check if the trace is not 3d-like chart or pie and has x and y defined
+*/
+function hasNoXY(type, tablesource) {
+    if (type.endsWith('3d') || ['pie', 'surface'].includes(type)) return true;
+
+    return (!get(tablesource, ['mappings', 'x']) || !get(tablesource, ['mappings', 'y']));
```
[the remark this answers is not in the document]
cwang2016: I guess this is a question related to if showing XY options for unknown type chart. 

for example, histogram2dcontour has data x, y defined. 
pie has no data x, y defined, (it has labels and values)

Please give more thoughts if there is any missing.

src/firefly/js/charts/ui/options/BasicOptions.jsx
```diff
@@ -86,13 +93,18 @@ export class BasicOptions extends SimpleComponent {
         const tbl_id = get(tablesource, 'tbl_id');
         const type = get(data, `${activeTrace}.type`, 'scatter');
         const noColor = !hasMarkerColor(type);
-        const noXY = hasNoXY(type);
+        const noXY = hasNoXY(type, tablesource);
+        const xColType = noXY ? '' : getColumnType(getTblById(tbl_id), get(tablesource, ['mappings', 'x'], ''));
+        const yColType = noXY ? '' : getColumnType(getTblById(tbl_id), get(tablesource, ['mappings', 'y'], ''));
+        const xNoLog = type.includes('histogram') ? true : undefined;          // histogram2d or histogram2dcontour
```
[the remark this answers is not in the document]
cwang2016: for some cases if the log is checked, the plotly rendering run into some problems. 
- if the column is 'string' type 
- histogram2d, histrogram2dcontour. 

so the chart type and table column type for x & y are considered to be checked, but not sure if this check is needed in case no XY options is shown in the side bar for unknown type chart.

src/firefly/js/charts/ui/PlotlyToolbar.jsx
tgoldina: The problem is that firefly enhanced charts might be creating the x and y arrays using custom logic bypassing mappings. May be we can check fireflyData.type here: for fireflyHeatmap we do want to support area selection?

Also you can use getColValidator in ColumnOrExpression.jsx to check for column expressions. TableStatsCntlr.getColValStats(tblId) will return table statistics.

cwang2016: Please check ffapi-slate-test2.html, the selection area is not changed for heatmap and scatter cases. (the firefly chart type implemented is not changed). In test2 example, originally scatter chart shows no selection box, and heatmap charts shows the selection box.

cwang2016:
Per the comments, please help clarify the following specs.: 
- if rendering selection box for unknown type chart?
- what is the logic for rendering the selection box?
- if showing XY options in chart options and tools side bar for unknown types? 
- what is the logic for rendering each XY options? (grid, reverse, top/right, log)
  and xmin/xmax, ymin/ymax 
   (thought: if x, y is related to 'string' or 'char' type column, log and min/max range could be excluded)

robyww:
Testing Issues:

- **Bar Chart:** If I make any changes to the chart then the plot goes away.
- **Pie Chart:** zoom, pan, and 1x does not seem to do anything, maybe it should not be on the toolbar
- I can still the entire xyplot from the side-bar.

cwang2016:
@robyww 
do you see zoom, pan in pie chart? there is no zoom, pan and 1x in the tool bar 
for pie chart.

src/firefly/js/charts/ui/PlotlyToolbar.jsx
cwang2016: The area selection issue for  heatmap in ffapi-highlevel-charttest.html is solved by add table columns info into the state for PlotlyToolBar for updating the tool bar when the table loading is complete.

cwang2016:
Add some updates: 
- fix the issue regarding selection area in the toolbar for heatmap chart --  some timing issue between the table loading and tool bar rendering. 
- exclude XY options in the chart options and tools for 'bar' chart. 
- remove deepClone while passing 'layout' to Plotly function calls.

PR #458, opened 2017-09-14
DM-11603 Make column expression to work for all plotly charts 

cwang2016:
This development includes
- a generic search processor to handle column expression which is  related to at least on numeric column or one 'text' column from the searched table.  
- a generic chart data handler to create table source entries to get plotly chart data from the server when the trace data is defined on table columns, like x: 'tables:w1mpro+w2mpro'. The data related keyword is not limited to be like 'x' or 'y' only in the JSON description. It varies based on the chart type, ex. pie chart has data defined on 'labels' and 'values'. 

Test: 
try  to run ffpai-slate-test3.html in which the data of some charts is defined in column expression style.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/XYGenericProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/XYGenericProcessor.java
@@ -15,7 +15,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * @author tatianag
+ * @author c.w.
  */
 @SearchProcessorImpl(id = "XYGeneric")
 public class XYGenericProcessor extends IpacTablePartProcessor {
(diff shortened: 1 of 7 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/XYGenericProcessor.java
```diff
@@ -0,0 +1,210 @@
...
+import edu.caltech.ipac.firefly.data.Param;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author tatianag
+ */
```
tgoldina: wrong author :)

cwang2016:
- fix the chart select and highlight issue for scatter type chart
- move the shared code from XYWithErrorsProcessor to the new processor handling column expression. 
- place a  plotly 'scatter' chart example with x, y, error_x and error_y in column expression into ffapi-slate-test2.html

PR #459, opened 2017-09-14
DM-11601 showChart API changes

tgoldina:
https://jira.lsstcorp.org/browse/DM-11601

- Renamed API function showPlot to showChart and removed fireflyData and fireflyLayout from its parameters.
- Firefly specific options are now under firefly property of data and layout, trace type can be fireflyScatter, fireflyHistogram, and fireflyHeatmap for Firefly extended charts.
- Updated addChart action creator to take care of this change and split plotly and firefly info on the fly.
- Added a workaround for hover text appearing without background in Firefox.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -288,7 +288,7 @@ export function getOptionsUI(chartId) {
     const {data, fireflyData, activeTrace=0} = getChartData(chartId);
     const type = get(data, [activeTrace, 'type'], 'scatter');
     const dataType = get(fireflyData, [activeTrace, 'dataType'], '');
-    if (isScatter2d(type)) {
+    if (dataType === 'fireflyScatter' || isScatter2d(type)) {
         return ScatterOptions;
     } else if (type === 'histogram') {
         return HistogramOptions;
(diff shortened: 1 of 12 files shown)


src/firefly/js/charts/ChartUtil.js
```diff
@@ -288,7 +288,7 @@ export function getOptionsUI(chartId) {
     const {data, fireflyData, activeTrace=0} = getChartData(chartId);
     const type = get(data, [activeTrace, 'type'], 'scatter');
     const dataType = get(fireflyData, [activeTrace, 'dataType'], '');
-    if (isScatter2d(type)) {
+    if (dataType === 'fireflyScatter' || isScatter2d(type)) {
```
cwang2016: in here, either 'fireflyScatter' or scatter2d returns ScatterOptions. 
Question, for function getToolBarUI() next to this function, isn't 'fireflyScatter' treated the same as 'isScatter2d'? 
Also there is no 'selection box'  shown up for 'fireflyScatter' chart.

tgoldina: Thank you for finding this issue. Indeed, I made a change that for firefly enhanced charts plotly trace type is not assigned until the data are retrieved. This bug should be fixed in the latest commit.

tgoldina:
Fixed the bug Cindy found, when scatter chart load resets table highlight to 0.

cwang2016:
reviews completes. works well.

PR #460, opened 2017-09-15
DM-5792: Support artifacts on all images that have them (currently wise and 2mass)

robyww:
- Image FileRetrievers can add artifact to RelatedData
  - WISE and 2MASS supported
  - Artifact drawing layer uses lazy loading
  - added layer title matching for visibility, color, and symbol changing
  - firefly: IbeQueryArtifact replaces ife: FinderChartQueryArtifact
  - Also added support for getting adding RelatedData artifacts in IbeFileRetrieve

_The big concept:_

- When an image plot is request is sent to the server it not only returns the image data but also how to retrieve "related data". 
- Sometime this related data are image mask in other cases it is artifact data.
- What related data is return is query specific.  This way the client does not have to know the details of the project.
- A RelatedData object can be turned into a server search request to retrieve it when the user request to see it.
- There are so far 3 types of related data: mask, tabular (artifacts), and image (not yet implemented).
- The mask related data was implemented months ago.  This pull request implements the tabular related data.

_to test:_
 - bring up the firefly or irsa viewer, or finder chart, or time series.
 - when you have a wise or 2mass layer up you should see the artifact drawing layers on the dialog.  They are off by default.
  - if you have more than one image such as several wise or several 2mass images. Checking and unchecking artifact turn all of them of the same type on and off.

### DIFF
--- a/src/firefly/js/drawingLayers/Artifact.js
+++ b/src/firefly/js/drawingLayers/Artifact.js
@@ -32,7 +32,7 @@ const defSymbolMap= {
     diff_spikes: {color: 'orange', symbol: DrawSymbol.DOT, size: 1},
     halos : {color: 'yellow', symbol: DrawSymbol.SQUARE, size: 3},
     ghost : {color: 'pink', symbol: DrawSymbol.DIAMOND, size: 3},
-    latents : {color: 'green', symbol: DrawSymbol.X, size: 3},
+    latents : {color: 'cyan', symbol: DrawSymbol.X, size: 3},
     unknown  : {color: 'blue', symbol: DrawSymbol.X, size: 3},
 };
 
--- a/src/firefly/js/drawingLayers/Artifact.js
+++ b/src/firefly/js/drawingLayers/Artifact.js
@@ -1,0 +1,198 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+
+import {take} from 'redux-saga/effects';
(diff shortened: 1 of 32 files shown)


src/firefly/js/drawingLayers/Artifact.js
```diff
@@ -0,0 +1,198 @@
...
+const factoryDef= makeFactoryDef(TYPE_ID,creator,null, getLayerChanges,null,getUIComponent);
+export default {factoryDef, TYPE_ID}; // every draw layer must default export with factoryDef and TYPE_ID
+
+const defSymbolMap= {
+    pers_arti  : {color: 'orange', symbol: DrawSymbol.CROSS, size: 3},
+    glint_arti : {color: 'purple', symbol: DrawSymbol.DIAMOND, size: 3},
+    diff_spikes: {color: 'orange', symbol: DrawSymbol.DOT, size: 1},
+    halos : {color: 'yellow', symbol: DrawSymbol.SQUARE, size: 3},
+    ghost : {color: 'pink', symbol: DrawSymbol.DIAMOND, size: 3},
+    latents : {color: 'green', symbol: DrawSymbol.X, size: 3},
```
cwang2016: suggestion: green might be difficult for some viewer, how about 'cyan' or some color neither green nor red.

robyww: I was just matching the color from the old UI.  It is a good suggestion.  I will change it.

PR #464, opened 2017-09-25
DM-11604 Support showing the active chart trace

cwang2016:
The development includes: 
- reorder the traces before plotly rendering such that the active chart is rendered on top of others. 
- assign the name and color relevant values to the trace in case those are not assigned initially. 
- create 2 demo apps, one is to render charts with multiple traces of the same type (ffapi-slate-test4.html) and the orher is to render charts with traces of mixed types (ffapi-slate-test5.html). 

test: 
run demo/ffapi-slate-test4.html to test the selection of active trace 
run  demo/ffapi-slate-test5.html to test the selection of active trace on the chart with mixed types. 
note: the chart rendering is based on the selection of active trace and the following rendering priority among different chart types (observation from the samples): 
pie -> scatter -> fireflyHistogram(bar) -> histogram -> histogram2dcontour/histogram2d -> heatmap

some issue: 
- the scatter chart in ffapi-slate-test4.html includes 3 trace with the same data, however, sometime the rendering of all traces is not overlapped, need some investigation about this issue. 
- the inactive trace for the chart with histogram2dcontour(or histogram2d) type is rendered above the active one (main chart part not the colorbar)  while using plotly 'restyle'.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -504,7 +504,12 @@ function getTraceTSEntries({chartDataType, traceTS, chartId, traceNum}) {
 }
 
 // does the default depend on the chart type?
-export function applyDefaults(chartData={}, resetColor = false) {
+/**
+ * set default value for layout and data
+ * @param chartData
+ * @param resetColor reset color generator for default color assignment of the chart
+ */
+export function applyDefaults(chartData={}, resetColor = true) {
     //const chartType = get(chartData, ['data', '0', 'type']);
     //const noXYAxis = chartType && (chartType === 'pie');
 
@@ -566,17 +571,19 @@ export function applyDefaults(chartData={}, resetColor = false) {
 }
 
 
-// color-blind friendly colors
-export const TRACE_COLORS = [  '#333333', '#ff3333', '#00ccff', '#336600',
-                        '#9900cc', '#ff9933', '#009999', '#66ff33',
-                        '#cc9999', '#b22424', '#008fb2', '#244700',
-                        '#6b008f', '#b26b24', '#006b6b', '#47b224', '8F6B6B'];
+// plotly default color (items 0-7) + color-blind friendly colors
+export const TRACE_COLORS = [  '#1f77b4', '#2ca02c', '#d62728', '#9467bd',
+                               '#8c564b', '#e377c2', '#7f7f7f', '#17becf',
+                               '#333333', '#ff3333', '#00ccff', '#336600',
+                               '#9900cc', '#ff9933', '#009999', '#66ff33',
+                               '#cc9999', '#b22424', '#008fb2', '#244700',
+                               '#6b008f', '#b26b24', '#006b6b', '#47b224', '8F6B6B'];
 export const TRACE_COLORSCALE = [  'Bluered', 'Blues', 'Earth', 'Electric', 'Greens',
                                 'Greys', 'Hot', 'Jet', 'Picnic', 'Portland', 'Rainbow',
                                 'RdBu', 'Reds', 'Viridis', 'YlGnBu', 'YlOrRd' ];
 
-export function *traceColorGenerator(colorList, init = 0) {
-    let nextIdx = init;
+export function *traceColorGenerator(colorList) {
+    let nextIdx = -1;
 
     while (true) {
         const result = yield (nextIdx === -1) ? '' : colorList[nextIdx%colorList.length];
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -7,7 +7,7 @@
  * Utilities related to charts
  * Created by tatianag on 3/17/16.
  */
-import {get, uniqueId, isUndefined, omitBy, zip, isEmpty, range, set, isObject, pick, cloneDeep, merge} from 'lodash';
+import {get, uniqueId, isUndefined, omitBy, zip, isEmpty, range, set, isObject, pick, cloneDeep, merge, isNil, has} from 'lodash';
 import shallowequal from 'shallowequal';
 
 import {flux} from '../Firefly.js';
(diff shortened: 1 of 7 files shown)


src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -63,7 +65,15 @@ export class PlotlyChartArea extends PureComponent {
...
-        let pdata = data.map((e) => Object.assign({}, e)); // create shallow copy of data elements to avoid sharing x,y,z arrays
+
+        // put the active trace after all inactive traces
+        let pdata = data.reduce((rdata, e, idx) =>  {
+            (idx !== activeTrace) && rdata.push(e);
+            return rdata;
+        }, []);
+
+        pdata.push(data[activeTrace]);
+        //let pdata = data.map((e) => Object.assign({}, e)); // create shallow copy of data elements to avoid sharing x,y,z arrays
```
tgoldina: I think this piece of code should not be commented. It is making sure expanded and unexpanded charts are not sharing the data.

To see the bug, open a heatmap in expanded mode, go to options, change bin size to 10 x 10, apply. The chart is not changing.  The reason is the expanded chart and unexpanded chart underneath share the data, and the change is first applied to unexpanded chart. Then when expanded chart optimize is checking for updates, the data appear updated.

Now the bug is back. I should have added better explanation in the comment.

src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -199,7 +199,8 @@ export class PlotlyWrapper extends Component {
 
             if (!useScatterGL) {
                 // when using SVG, it's actually faster to redraw then to do multiple updates
-                if (renderType === RenderType.RESTYLE_AND_RELAYOUT || dataUpdate.length > 1) {
+                if (renderType === RenderType.RESTYLE_AND_RELAYOUT || dataUpdate.length > 1 ||
+                    (data.find((d) => get(d, 'type').includes('histogram2d')))) {
```
tgoldina: we need a comment here to explain the workaround or issue with histogram 2d

src/firefly/js/charts/ChartUtil.js
```diff
@@ -545,16 +551,50 @@ export function applyDefaults(chartData={}) {
...
+        type && Object.entries(setDefaultColor(d, type)).forEach(([k, v]) => set(d, k, v));
+    });
 }
 
+
 // color-blind friendly colors
-const TRACE_COLORS = [  '#333333', '#ff3333', '#00ccff','#336600',
-                        '#9900cc', '#ff9933', '#009999', '#66ff33', '#cc9999',
-                        '#333333', '#b22424', '#008fb2', '#244700',
+export const TRACE_COLORS = [  '#333333', '#ff3333', '#00ccff', '#336600',
```
tgoldina: We might want to use plotly default colors first (without orange), skipping orange and yellow-green, because they are close to our highlight and select colors. What do you think? 


```
'#1f77b4 ',  // muted blue
    '#ff7f0e ',  // safety orange
    '#2ca02c ',  // cooked asparagus green
    '#d62728 ',  // brick red
    '#9467bd ',  // muted purple
    '#8c564b ',  // chestnut brown
    '#e377c2 ',  // raspberry yogurt pink
    '#7f7f7f ',  // middle gray
    '#bcbd22 ',  // curry yellow-green
    '#17becf '   // blue-teal
```

src/firefly/js/charts/ChartUtil.js
```diff
@@ -327,7 +327,9 @@ export function newTraceFrom(data, selIndexes, newTraceProps) {
     Object.assign(sdata, {showlegend: false, type: get(data, 'type', 'scatter'), mode: 'markers'});
 
     // the rowIdx doesn't exist for generic plotly chart case
-    if (!get(sdata, 'firefly.rowIdx') && get(sdata, 'x.length', 0) !== 0) {
+    if (isScatter2d(get(data, 'type', '')) &&
+        !get(sdata, 'firefly.rowIdx') &&
```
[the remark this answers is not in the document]
tgoldina: The intention of this was to save the row indexes of the selected rows to the SELECTED trace. We could have used isUndefined here.

For a generic scatter trace, chart point index matches the row index in the table, but for selected points it's no longer true, so we need to preserve these indexes somewhere to correctly handle click events on selected points.

src/firefly/js/charts/ui/PlotlyChartArea.jsx
cwang2016: The code is modified to collect the trace data in shallow copy style.

src/firefly/js/charts/ui/PlotlyWrapper.jsx
cwang2016: done

src/firefly/js/charts/ChartUtil.js
cwang2016: the plotly default colors are applied first.

cwang2016:
- Add a map to identify the trace number for the trace data in the store according to curveNumber returned from plotly chart click and select events. This resolves the issue of point selection from the active trace of 'scatter' type. 
- Reset button will not change the assigned color to the trace which has no color assigned initially.
-

cwang2016:
per discussion with @tgoldina regarding the click and select event on plotly scatter chart, the development is updated as follows, 
- click event on plotly scatter chart:  if the click occurs on inactive trace, then no change to the highlight on the chart, otherwise, the clicked point is highlighted. 
- select event on plotly scatter chart: if the point enclosed by select area  belongs to active trace, that point  will be collected into SELECTED trace.

src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -131,18 +131,30 @@ function calculateChartSize(widthPx, heightPx, xyratio, stretch) {
...
         // we should have active trace, its related selected, and its highlight traces on top
-        const {activeTrace, curveNumberMap} = getChartData(chartId);
+        const {activeTrace=0, curveNumberMap} = getChartData(chartId);
         const curveNumber = get(evData.points, `${0}.curveNumber`);
         const highlighted = get(evData.points, `${0}.pointNumber`);
         const curveName = get(evData.points, `${0}.data.name`);
-        if (curveNumberMap && activeTrace === curveNumberMap[curveNumber]) {
+
+        const traceNum = curveNumber >= curveNumberMap.length ? curveNumber :
+                         curveNumberMap.find((tNum, idx) => idx === curveNumber);
```
tgoldina: why not curveNumberMap[curveNumber]?

PR #465, opened 2017-09-28
Irsa 814 813 add ibe2 api and SEIP/MSX examples

ejoliet:
This FF (branch) implementation is the one used by Finderchart to search and display SEIP. Those are the first steps o add more ATLAS dataset.

The code has a new Atlas ibe datasource and example in Test searches (localhost:8080/firefly/firefly-dev.html).

I've also added a test, under test/astro/ibe.

Checkout the branch and check that i didn't break any image fits viewer related features.

Caveats: The code still miss the 3-color implementation. And the part related to fits download dialog.

In order to test Finderchat, checkout the corresponding branch in IFE and build it. You should see SEIP checkbox added as dataset to be searched. (the code has also MSX but is not enabled, AKARI is coming soon but will do it in another ticket. it will follow the same logic as SEIP and ATLAS/IBE API, requiring to set 'dataset', 'table' and 'filter').

If you have any question, please let me know.

Thanks for testing and reviewing this.

### DIFF
--- a/config/app.config
+++ b/config/app.config
@@ -61,14 +61,12 @@ environments {
         ehcache.multicast.port = 4015
         visualize.fits.Security= false
         ehcache.multicast.ttl = 0
-        atlas.ibe.host    = "irsadev.ipac.caltech.edu"
     }
     dev {
         BuildType = "Development"
         ehcache.multicast.port = "5015"
         lsst.dataAccess.uri = "http://localhost:8661/db/v0/query?"
         lsst.dataAccess.db = "smm_bremerton"
-	atlas.ibe.host    = "irsadev.ipac.caltech.edu"
     }
     test {
         BuildType = "Beta"
--- a/config/app.config
+++ b/config/app.config
@@ -61,12 +61,14 @@ environments {
         ehcache.multicast.port = 4015
         visualize.fits.Security= false
         ehcache.multicast.ttl = 0
+        atlas.ibe.host    = "irsadev.ipac.caltech.edu"
     }
     dev {
         BuildType = "Development"
         ehcache.multicast.port = "5015"
         lsst.dataAccess.uri = "http://localhost:8661/db/v0/query?"
         lsst.dataAccess.db = "smm_bremerton"
+	atlas.ibe.host    = "irsadev.ipac.caltech.edu"
     }
     test {
         BuildType = "Beta"
--- a/config/app.config
+++ b/config/app.config
@@ -48,6 +48,7 @@ irsa.gator.service.periodogram.url = "https://irsa.ipac.caltech.edu/cgi-bin/peri
 
 wise.ibe.host       = "irsa.ipac.caltech.edu/ibe"
 twomass.ibe.host    = "irsa.ipac.caltech.edu/ibe"
+atlas.ibe.host    = "irsa.ipac.caltech.edu"
 
 lsst.dataAccess.uri = ''
 lsst.dataAccess.db = ''
--- a/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
@@ -5,6 +5,9 @@ package edu.caltech.ipac.astro.ibe;
 
 import edu.caltech.ipac.util.StringUtils;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -21,6 +24,14 @@ public class BaseIbeDataSource implements IbeDataSource {
     private String tableName;
     private boolean useFileSystem = false;
     private String baseFilesystemPath;
+    public static final String USER_TARGET_WORLD_PT = "UserTargetWorldPt";
+    public static final String POS = "POS";
+    public static final String REF_BY = "refby";
+    public static final String INTERSECT = "INTERSECT";
+    public static final String SIZE = "SIZE";
+    public static final String MCEN = "mcen";
+    public static final String COLUMNS = "columns";
+    public static final String WHERE = "where";
 
     public BaseIbeDataSource() {
     }
@@ -90,7 +101,118 @@ public class BaseIbeDataSource implements IbeDataSource {
         return baseFilesystemPath;
     }
 
+    @Override
+    public String getDataUrl(IbeDataParam param) {
+        String dataUrl = this.getIbeHost() + "/data/" +
+                this.getMission() + "/" + this.getDataset() +
+                "/" + this.getTableName() + "/" ;
+
+        String fpath = param.getFilePath();
+
+        String url = dataUrl+ fpath;
+        return url;
+    }
+
+    @Override
+    public String getMetaDataUrl() {
+        return getSearchUrl()+ "?FORMAT=METADATA";
+    }
+
+    @Override
+    public String getQueryUrl(IbeQueryParam param) {
+        return getSearchUrl() + "?" + convertToUrl(param);
+    }
+
+    @Override
+    public Map<String, String> getMulipleQueryParam(IbeQueryParam param) {
+        return asMap(param);
+    }
+
+    @Override
+    public String getSearchUrl() {
+        return this.getIbeHost() + "/search/" +
+                this.getMission() + "/" + this.getDataset() +
+                "/" + this.getTableName();
+    }
+
+    @Override
+    public String getCorners() {
+        return "ra1;dec1;EQ_J2000,ra2;dec2;EQ_J2000,ra3;dec3;EQ_J2000,ra4;dec4;EQ_J2000";
+    }
+
+    @Override
+    public String getCenterCols() {
+        return "crval1;crval2;EQ_J2000";
+    }
+
+    @Override
+    public String[] getColsToHide() {
+        return new String[]{"in_row_id", "in_ra", "in_dec",
+                "crval1", "crval2",
+                "ra1", "ra2", "ra3", "ra4",
+                "dec1", "dec2", "dec3", "dec4"
+        };
+    }
+
     public void setBaseFilesystemPath(String baseFilesystemPath) {
         this.baseFilesystemPath = baseFilesystemPath;
     }
+
+    private Map<String, String> asMap(IbeQueryParam param) {
+        HashMap<String, String> params = new HashMap<>();
+        String s = convertToUrl(param);
+        String[] pp = s.split("&");
+        for (String keyval : pp) {
+            if (!StringUtils.isEmpty(keyval)) {
+                String[] parts = keyval.split("=", 2);
+                if (!StringUtils.isEmpty(parts[0])) {
+                    String v = parts.length > 1 && !StringUtils.isEmpty(parts[1]) ? parts[1].trim() : "";
+                    params.put(parts[0], v);
+                }
+            }
+        }
+        return params;
+    }
+
+
+    private String convertToUrl(IbeQueryParam param) {
+        String s = "";
+        if (param == null) return "";
+
+        if (!StringUtils.isEmpty(param.getRefBy())) {
+            s =addUrlParam(s, REF_BY, param.getRefBy());
+        } else if (!StringUtils.isEmpty(param.getPos())) {
+            s = addUrlParam(s, POS, param.getPos());
+            s = addUrlParam(s, INTERSECT, param.getIntersect());
+            if (param.isMcen()) {
+                s = addUrlParam(s, null, MCEN);
+            } else {
+                s = addUrlParam(s, SIZE, param.getSize());
+            }
+        }
+
+        s = addUrlParam(s, COLUMNS, param.getColumns());
+        s = addUrlParam(s, WHERE, param.getWhere(), true);
+        return s;
+    }
+
+    public static String addUrlParam(String url, String key, Object value) {
+        return addUrlParam(url, key, value, false);
+    }
+
+    public static String addUrlParam(String url, String key, Object value, boolean doEncode) {
+        try {
+            if (!StringUtils.isEmpty(value)) {
+                if (!StringUtils.isEmpty(url)) {
+                    url = url + "&";
+                }
+                value = doEncode ? URLEncoder.encode(value.toString(), "UTF-8") : value;
+                key = StringUtils.isEmpty(key) ? "" : key + "=";
+                url = url + key + value;
+            }
+        } catch (UnsupportedEncodingException e) {
+        }
+        return url;
+    }
+
 }
--- a/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
@@ -24,14 +24,6 @@ public class BaseIbeDataSource implements IbeDataSource {
     private String tableName;
     private boolean useFileSystem = false;
     private String baseFilesystemPath;
-    public static final String USER_TARGET_WORLD_PT = "UserTargetWorldPt";
-    public static final String POS = "POS";
-    public static final String REF_BY = "refby";
-    public static final String INTERSECT = "INTERSECT";
-    public static final String SIZE = "SIZE";
-    public static final String MCEN = "mcen";
-    public static final String COLUMNS = "columns";
-    public static final String WHERE = "where";
 
     public BaseIbeDataSource() {
     }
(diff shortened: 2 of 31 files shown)


robyww:
I Added @loitly as a reviewer to specifically look at all *Ibe*.java files.

src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
```diff
@@ -21,6 +24,14 @@
     private String tableName;
     private boolean useFileSystem = false;
     private String baseFilesystemPath;
+    public static final String USER_TARGET_WORLD_PT = "UserTargetWorldPt";
```
robyww: This should referenced from `ServerPaarams.USER_TARGET_WORLD_PT`

src/firefly/java/edu/caltech/ipac/astro/ibe/datasource/AtlasIbeDataSource.java
```diff
@@ -0,0 +1,395 @@
...
+
+        // source search
+        IbeQueryParam queryParam = new IbeQueryParam();
+
+        // process POS - target search
+        String userTargetWorldPt = queryInfo.get("UserTargetWorldPt");
+        if (userTargetWorldPt != null) {
+            WorldPt pt = WorldPt.parse(userTargetWorldPt);
+            if (pt != null) {
+                pt = Plot.convert(pt, CoordinateSys.EQ_J2000);
```
robyww: glad you remembered to do the convert.
it could be simplified to
```
 WorldPt pt = Plot.convert(WorldPt.parse(userTargetWorldPt));
```
It will not throw an exception, just return a null.

If you want to take out another if block. then:
```
WorldPt pt= Plot.convert(
      WorldPt.parse( queryInfo.get("UserTargetWorldPt")));
if (pt!=null) {
     ....
}
```

src/firefly/java/edu/caltech/ipac/astro/ibe/datasource/AtlasIbeDataSource.java
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /**
+     * Check and convert if needed to single quote value to be used in where=band in (...)
+     *
+     * @param str CSV string
+     * @return CSV single quoted values
+     */
+    private String checkAndConvert2StringSingleQuote(String str) {
+        String res = "";
```
robyww: I am not sure if you did this but I really don't think we should send quoted parameters from the client.  The client should just send a string  and if the irsa server requires it then the search processor should put them on.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/IBEUtils.java
```diff
@@ -129,6 +134,14 @@ public static SortInfo getSortInfo(IbeDataSource source) {
             }
 
 
+        } else if (source instanceof AtlasIbeDataSource) {
+
+                // TODO : explain to me what is used for:
+                sortByCols.put("facility_name", "facility_name, instrument_name, band_name, file_type");
+                //TODO what is 'relatedCols' meant for?
+                // relatedCols = "fname";
```
robyww: did you get your answers to the `todo` questions here?

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/QueryIBE.java
```diff
@@ -124,16 +125,12 @@ public void prepareTableMeta(TableMeta meta, List<DataType> columns, ServerReque
             meta.setAttribute("table", source.getTableName());
             meta.setAttribute("subsize", request.getParam("subsize"));
             meta.setAttribute(MetaConst.DATASET_CONVERTER, source.getMission());
-            meta.setAttribute("ALL_CORNERS", "ra1;dec1;EQ_J2000,ra2;dec2;EQ_J2000,ra3;dec3;EQ_J2000,ra4;dec4;EQ_J2000");
-            meta.setAttribute("CENTER_COLUMN", "crval1;crval2;EQ_J2000");
+            meta.setAttribute("ALL_CORNERS", source.getCorners());
```
robyww: This is better than before.

src/firefly/java/edu/caltech/ipac/visualize/net/AtlasImageGetter.java
```diff
@@ -0,0 +1,99 @@
...
+                DataObject row = data.get(0);
+                Map<String, String> dataMap = IpacTableUtil.asMap(row);
+                IbeDataParam dataParam = ibeSource.makeDataParam(dataMap);
+
+                dataParam.setCutout(true, params.getRaJ2000String() + "," + params.getDecJ2000String(), sizeStr);
+                dataParam.setDoZip(true);
+                FileInfo result = ibe.getData(dataParam, null);
+                return new File(result.getInternalFilename());
+            } else {
+                throw new FailedRequestException("Area not covered, images #"+data.values().size());
```
robyww: can you test this exception. This message will go area the way to the user.  Should it just say `Area not covered`?

src/firefly/java/edu/caltech/ipac/visualize/net/IrsaImageParams.java
```diff
@@ -37,6 +37,9 @@ public String getUniqueString() {
              case IRIS :
                        retval= "iris-" + super.toString() + _band + _size;
                        break;
+             case ATLAS:
+                 retval = "atlas-"+ super.toString();
```
robyww: **Really Important!**

i don't think string is unique.  It is a cache key and a file name in the working directory.

I think `AtlasParams` needs override `getUniqueString`. needs to user

some think like:
```
return "atlas-" +_band+"-"+xtraFilter+"-"+_instrument +"-"+String.format(Locale.US,"%8.6f%8.6f",_ra,_dec)
```

src/firefly/js/metaConvert/AtlasRequestList.js
```diff
@@ -0,0 +1,87 @@
...
+        const req= svcBuilder(plotId, reqKey, title, rowNum, extraParams);
+        const subsize = get(table, 'request.subsize');
+        if (subsize && subsize>0) {
+            const {UserTargetWorldPt, sizeUnit} = get(table, 'request', {});
+            if (UserTargetWorldPt && sizeUnit) {
+                const wp = parseWorldPt(UserTargetWorldPt);
+                // cutout is requested when in_ra, in_dec, and subsize are set (see AtlasIbeDataSource)
+                req.setParam('center', `${wp.getLon()},${wp.getLat()}`); // degrees assumed if no unit
+                req.setParam('in_ra', `${wp.getLon()}`);
+                req.setParam('in_dec', `${wp.getLat()}`);
```
robyww: This appears to work but something does not make sense?  The `AtlasIbeDataSource` want to user `UserTargetWorldPt` but you are passing `in_ra` and `in_dec`.  I think you need to call `req.setWorldPt`

Also I think you want to build the WebPlotRequest based on the row not no the original table search request.

src/firefly/js/metaConvert/AtlasRequestList.js
```diff
@@ -0,0 +1,87 @@
...
+
+    if (includeStandard) {
+
+        retval.standard= [
+            builder('spitzer-irac1','ibe_file_retrieve', `${band}`, row, {band:`${band}`, fname, in_ra, in_dec}),
+            // builder('spitzer-irac2','ibe_file_retrieve', 'IRAC 2', row, {band:'IRAC2'}),
+            // builder('spitzer-irac3','ibe_file_retrieve', 'IRAC 3', row, {band:'IRAC3'}),
+            // builder('spitzer-irac4','ibe_file_retrieve', 'IRAC 4', row, {band:'IRAC4'}),
+            // builder('spitzer-mips24','ibe_file_retrieve', 'MIPS 1', row, {band:'MIPS24'})
+        ];
```
robyww: Can you get more one image here?

src/firefly/js/ui/TestQueriesPanel.jsx
```diff
@@ -593,6 +640,24 @@ function do2Mass(request) {
...
+    console.log('atlas', request);
+    const reqParams = makeTblRequest('ibe_processor', 'ATLAS-' + request[ServerParams.USER_TARGET_WORLD_PT],
+        {
+            [ServerParams.USER_TARGET_WORLD_PT]: request[ServerParams.USER_TARGET_WORLD_PT],
+            mission: 'atlas',
+            //mcenter:true, returned 1 image
+            // TODO Not yet fully working, see AtlasRequestList
+            ds:request.ds1,
+            band: request.band,
+            subsize:0.05,
```
robyww: do you get more rows with a bigger size?

src/firefly/js/visualize/ui/ImageSelectPanelProp.js
```diff
@@ -28,6 +28,35 @@ export const initPanelCatalogs = [
         'range': {'min': 1, 'max': 12.5, 'unit': 'deg'},
         'size': 5
     },
+    /*
```
robyww: I don't see atlas showing up in the ImageSelectPanel is this work completed?

src/firefly/java/edu/caltech/ipac/astro/ibe/datasource/AtlasIbeDataSource.java
ejoliet: That's exatcly why i've added this method ;-) The client send regular string (no quotes), but the IBE API needs single quote around values to express the where clause.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/IBEUtils.java
ejoliet: i remember you told me that the related columns are used for artifacts and such, right? The sort info i guess is to be used in table view if query search table is displayed. Isn't it?
I'll clean up the TODOs with the rest of other cleanups.

src/firefly/java/edu/caltech/ipac/visualize/net/AtlasImageGetter.java
ejoliet: the problem is that the data can come back with more than 1 row and that doesn't mean that the area is not covered, it means that we don't know which one to pick among the result of this search. If you have a better message, I'm open to change it also. I've found the problem while implementing the IB2 API, depending on the search you can get more than one answer until we get the 'principal' column which will determine the main image to be displayed.

src/firefly/java/edu/caltech/ipac/visualize/net/IrsaImageParams.java
ejoliet: I know, i discover this logic while implementing, the unique string is used heavily.
But i overwrite it, is less apparent but see AtlasImageParams.java. I've added it there.

src/firefly/js/metaConvert/AtlasRequestList.js
ejoliet: yes, in_ra, in_dec is from the row. I took this code directly from the wiseRequestList and i'm sure is not complete. See below. Sorry.

src/firefly/js/metaConvert/AtlasRequestList.js
ejoliet: I've stopped on one image, somehow i couldn't figure it out why it didn't work. It might now since i've fixed many things in the meantime. This code was added at the beginning, like 3 commits before.

src/firefly/js/ui/TestQueriesPanel.jsx
ejoliet: no, subsize is the parameter used for cutout. Number of rows is the resultof the 'band' request, becaue it will end up in the where clause 'band in (...)' and will return in principle max of 5 rows (5 bands) if 'mcen' implementation from Walter is fixed. Currently it isn't fully working as IBE1 ([IRSA-747](https://caltech-ipac.atlassian.net/browse/IRSA-747?focusedCommentId=23640&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23640)).

src/firefly/js/visualize/ui/ImageSelectPanelProp.js
ejoliet: no, i disabled it because it need more work to show it on that panel and it will be changed in IRSAViewer by the new layout. If you want, you can see it in "Test searches" for single image (see comment on the AtlasRequestList.js code).

src/firefly/java/edu/caltech/ipac/astro/ibe/datasource/AtlasIbeDataSource.java
robyww: oh, good!

src/firefly/java/edu/caltech/ipac/visualize/net/IrsaImageParams.java
robyww: looks better, as long is it produces a reproducible  unique string.

src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
```diff
@@ -151,43 +138,41 @@ public File createDataFilePath(IbeDataParam param) throws IOException {
...
-            qstr= addUrlParam(qstr, CUT_CENTER, param.getCenter());
-            qstr= addUrlParam(qstr, CUT_SIZE, param.getSize());
-            qstr= addUrlParam(qstr, "gzip", param.isDoZip());
+            qstr = addUrlParam(qstr, CUT_CENTER, param.getCenter());
+            qstr = addUrlParam(qstr, CUT_SIZE, param.getSize());
+            qstr = addUrlParam(qstr, "gzip", param.isDoZip());
         }
 
-        url = url + (qstr.length() > 0 ? "?" + qstr : "");
+        url += (qstr.length() > 0 ? "?" + qstr : "");
```
loitly: I understand what you are doing now.  Because IBE2 API changes, and some dataset uses v1 while other use v2, you moved the logic to the IbeDataSource.  I can't say I agree with this.  As is, you have IBE's API logic in 3 places.  v1 is in BaseIbeDataSource,  v2 is in AtlasDataSource, and what's common between them is in IBE.  You should have made IBE handle v1 as well as v2 and have the IbeDataSource identify which version they use.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/IBEUtils.java
loitly: wow.. this is messy.  this logic should have been kept in IbeDataSource instead of a util function with a case statement.

config/app.config
```diff
@@ -60,12 +61,14 @@ environments {
         ehcache.multicast.port = 4015
         visualize.fits.Security= false
         ehcache.multicast.ttl = 0
+        atlas.ibe.host    = "irsadev.ipac.caltech.edu"
```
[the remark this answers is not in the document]
ejoliet: Will fix that.

src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
```diff
@@ -21,6 +24,14 @@
...
+    public static final String POS = "POS";
+    public static final String REF_BY = "refby";
+    public static final String INTERSECT = "INTERSECT";
+    public static final String SIZE = "SIZE";
+    public static final String MCEN = "mcen";
+    public static final String COLUMNS = "columns";
+    public static final String WHERE = "where";
```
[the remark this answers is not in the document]
ejoliet: Will move this to the interface.

src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
ejoliet: Yes. There should be a second phase where we deprecate IBE1 and use only IBE2 (what i call for now Atlas) but it was not plan to do it now because it has much more impacts and need attention from everybody and changes in the server and client. I postponed deliberately this phase for next year, specially also because IBE2 API is still not fully working and metadata search is changing as we speak. differences from v1 to v2 is not only the api url, it is also the way we getting the files from the result, it's different from datasets: WISE, PTF and 2MASS is one group, ATLAS is another group. I'll raise ticket to work on the refactor and move away from IBE1 when we get time. We should also be implementing a TAP processor, SIA2 and SSA processors.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/IBEUtils.java
ejoliet: Same mess as before, didn't clean this up and i won't for now, i've used the similar code path and implementation as others ibe datasources on purpose.

PR #467, opened 2017-09-29
RSA-794: Panel 4: image set panel selection

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-794

To test:
- http://localhost:8080/firefly/firefly-dev.html
- then Image Concept

Looking for feedbacks.

### DIFF
--- a/src/firefly/js/ui/ImageSelect.jsx
+++ b/src/firefly/js/ui/ImageSelect.jsx
@@ -1,0 +1,248 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {PureComponent} from 'react';
+import PropTypes from 'prop-types';
+import {uniqBy, get, countBy} from 'lodash';
+
+
+import {CheckboxGroupInputField} from './CheckboxGroupInputField.jsx';
+import {CollapsiblePanel} from '../ui/panel/CollapsiblePanel.jsx';
+import FieldGroupUtils, {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
+import {dispatchComponentStateChange} from '../core/ComponentCntlr.js';
+
+import './ImageSelect.css';
+
+
+export class ImageSelect extends PureComponent {
+
+    constructor(props) {
+        super(props);
+        this.state= {lastMod:new Date().getTime(), filteredImageData:undefined};   // can be open, close, or auto.
+    }
+
+    render() {
+        const {style, imageMasterData, groupKey} = this.props;
+        var {filteredImageData=imageMasterData} = this.state;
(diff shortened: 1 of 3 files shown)


ejoliet:
WOW! This is great! It has the filters already!! The UI is neat. 
I think it has all the ingredients that have been asked and more.
I really like the add-on expand/collapse all!
There are probably some details that arise by playing with filters. We probably need to add select/unselect all on each filter (or even fancier that can come later like 'limit to', 'exclude') but is pretty complete for now. 
I was looking at the waveband filter but it looks that it doesn't work and some label are 'o micron'? I think those values/labels will be fixed/adjusted with master table and Lijun's branch, don;t you think? Let's see what other opinions are about UI flow and if that make sense at all or is there something wrong. I don't see anything wrong so far!
Great job and fast!

ejoliet:
I see already something that will need to change and i think you mentioned it before: grouping by mission (and filter) but the result main panel is grouped by dataset which is one level down.
Mission: WISE or Spitzer
|_
    dataset: Level3 All sky or SEIP
        |_    instrument: N/A or IRAC
                       |_ bands: 1,2,3,4 OR IRAC1, 2, 3, 4 and MIPS24

src/firefly/js/ui/NewImageSearchPanel.jsx
```diff
@@ -22,8 +22,7 @@ import WebPlotRequest, {WPConst} from '../visualize/WebPlotRequest.js';
 import {dispatchPlotImage} from '../visualize/ImagePlotCntlr.js';
 import {imageMasterData} from '../data/ImageMasterData.js';
```
ejoliet: @loitly Currently this is a constant, i guess this has to be changed to call server side and return it, right?

ejoliet: Nevermind, i've found the call: AllImageSearchConfig.js.

src/firefly/js/ui/ImageSelect.jsx
```diff
@@ -0,0 +1,248 @@
...
+            <div style={{padding: '0 10px'}}>
+                {selections}
+            </div>
+        </div>
+
+    );
+}
+
+const toFilterOptions = (a) => a.map ( (d) => ({label: `${d.label}  (${d.count})`, value: d.name}));
+const toCatSummary = (master, key, desc) => Object.entries(countBy(master, (d) => `${get(d, key)};${get(d, desc)}`))
```
ejoliet: This looks awesomely generic! 
You might know already and this is still in progress as we define the master table but the count (number that appears in parenthesis on the right-hand side of the mission label from the filter) should be the sum of the number of dataset each mission contains, not the number of bands or instruments. Example: in the current state of the json you are using, 2mass should have currently '2', because only LGA and AllSky surveys are present. Right?

PR #470, opened 2017-10-03
DM-12089 1M row table issues, remove fireflyScatter

tgoldina:
https://jira.lsstcorp.org/browse/DM-12089

The test html is attached to the ticket. 

1. Fixed a bug, which was preventing large table column mapping into data arrays. 
(There is a practical limit of the number of arguments a JavaScript function can accept, see `https://stackoverflow.com/questions/22747068/is-there-a-max-number-of-arguments-javascript-functions-can-accept`)

2. Added support for hoverinfo='skip' When no hover layer is present, tooltips are not displayed and point click and select are not supported. The chart is displayed with no highlight or select layer and no point select option.

3. Removed fireflyScatter chart type. Plotly scatter plot is by default enhanced with fireflyScatter features (improved tooltips, etc.)


src/firefly/js/charts/dataTypes/FireflyGenericData.js
```diff
@@ -72,3 +75,142 @@ function fetchData(chartId, traceNum, tablesource) {
...
+ * Firefly added scatter changes
+ * @param {Object} p - parameters
+ * @param p.changes existing changes
+ * @param p.chartId chart id
+ * @param p.traceNum
+ * @param p.tablesource
+ * @param p.tableModel
+ */
+function addScatterChanges({changes, chartId, traceNum, tablesource, tableModel}) {
+
```
loitly: Are we applying the changes only when it's not defined, or always?

tgoldina: The enhancements are applied to scatter-like plots only. We apply firefly tooltips when hoverinfo is either not defined or is equal to 'text'. We apply labels to axes when they are not defined and there is only one trace.

PR #474, opened 2017-10-10
Dm 10833 multitrace charts in tri-view

tgoldina:
https://jira.lsstcorp.org/browse/DM-10833
https://jira.lsstcorp.org/browse/DM-9072

To use multi-trace chart architecture, add `charts: {chartEngine: 'plotly', multitrace: true}` to window.firefly.app.options. For testing purposes, firefly.html and firefly/demo/ffapi-highlevel-charttest.html have these settings.

This pull request includes:
- Supporting default multi-trace charts in triview
- The default chart parameters can be passed via META_INFO of the table. Other than that the default chart behavior is the same as before per Xiuqin's request.
- Creating multi-trace charts from Charts dropdown
- Multi-trace charts error handling, loading indicator, deletable behavior, help_id
- Bug fixes: 
   1. When trace color and alpha are changed, change all color attributes (marker, line, error bars, etc). 
   2. Make sure the default color is reset after color map option is undone. (Color map option is making marker.color an array, the default color is lost.)
   3. Heatmap reversescale option was not always observed
   4. Use scatter-gl and non-gl consistently. The two do not work well together.
- Reworked ChartsContainer, which now supports tbl_group, and will display a default chart for active table in that group. ChartsContainer is the owner of the sagas, updating chart viewer items and creating the default charts. (The hope is to get rid of ChartsSync and its master sagas.

The test case for tbl_group tracking is added to ffapi-highlevel-charttest

I am sure something fell through the cracks. I am looking forward to your feedback.

Important changes:
html/demo/ffapi-highlevel-charttest.html
ApiHighlevelBuild.js
ChartsContainer
ChartSelectDropdown
BasicOptions


loitly:
On resize, the top of the chart may be chopped off, especially when it's being expanded vertically.

tgoldina:
@loitly I have fixed the issues found so far, except the resize issue, which I can not reproduce. (I have reduced the top and bottom margins to match the previous tri-view look and to better use the container space. Without the title, the margin on the top is 10px.

robyww:
UI Notes:
 - If we can add a trace we need to be able to remove it.
 - The charts panel needs to choose between adding a trace and adding a new chart.
 -  Selected does not seem to work with two traces.  _to repeat:_ put two traces up both ra/dec and select and click on select.
  - image overlay should clear when a filter with no data found (this is probably not your problem)
  - would be nice to have a warning dialog with there are two traces and the user attempts to filter the non-active trace. This will result in a "no data found" and probably not what the user wants.

xiuqin:
when I changed the X/Y to different columns, the range of XY plot did not change (stayed at ra/dec range). The ranges need to be reset so the proper plot will show up.

xiuqin:
Here are the suggestions after discussing with Trey:

- Change the slide-down menu associated with gear icon to the dialog associated with clicking on "Charts".
- Modify the dialog with the following:

1.  following the image dialog, give options "new chart", "add to existing chart" (it will be like adding a data series). When it comes up first without existing chart, "New chart" should be the only option.
2. Change button label "ADD" to "OK" to cover both options.
3. For "add to existing chart", the trace field could be filled with the default label as "add series" currently does.
4. Use the same color schema as Image and Catalog dialogs, adding ? for help.

tgoldina:
I think that removing chart trace and modifying charts dropdown possibly replacing current trace settings should be the separate tickets. The last one in particular might need to be clarified. 

I agree that chart options are getting too crowded and hard to navigate. Also, we have no way to create new charts from API. Replacing settings on the chart side with a dialog would give us more space and we'll be able to add charts from API. (And it would be consistent with how we deal with images.)

So, the proposed dialog would have up to 3 options:

Modify active trace, Add new trace, Add new chart

I. First, we need to understand **which chart viewer these options will be referring to**.

The first two (modify active trace, add new trace) would be referring to the active chart of the default chart viewer (if triggered from charts dropdown menu) or to the current chart's viewer if triggered from the chart toolbar. 

The last one (create new) would be adding a chart to the default viewer, if triggered from charts dropdown menu and to the chart's viewer, if triggered from the chart toolbar. 

II. Second, **'Add new trace' and 'Add new chart' options only make sense if there is an active table**.

In Firefly viewer scenario, an active table is the currently active table of the table main group. 

How do we know what the active table is when a chart is created from an API, and different traces may possibly reference different tables?

We can decide that the active table is the table referenced in the current active trace of the chart.
 
Or we can decide, that if chart viewer has tbl_group property, the active table is the one currently active in this group. 

If we can not find active table, we don't display 'Add new trace' and 'Add new chart' options.

OK, it looks like a good plan, but I would still put it in a separate ticket. :)

tgoldina:
The following bugs found in review are fixed:
- Selected does not seem to work with two traces. to repeat: put two traces up both ra/dec and select and click on select.
- Sticky range, which caused new trace not to be visible after the original was panned or zoomed.
(Now we clear the axis range in chart layout whenever the new x or y field is submitted via the chart options.) 

I have merged this branch into dev to avoid merging issues in the future. Also, having it in dev would help finding bugs faster and getting more suggestions. You can always switch back by setting `window.firefly.app.optionscharts.multitrace` to `false`. 

I'll move all outstanding suggestions into a separate ticket.

PR #475, opened 2017-10-12
DM-12208: Check if the document is loaded before rendering firefly

robyww:
_to test:_

Load firefly.  It should work.


src/firefly/js/Firefly.js
```diff
@@ -150,11 +151,28 @@ function bootstrap(options, viewer, props) {
                 dispatchUpdateLayoutInfo({disableDefaultDropDown:true});
             }
         }
+
         if (viewer) {
-            ReactDOM.render(React.createElement(viewer, props),
-                document.getElementById(props.div));
+            if (window.document.readyState==='complete' || window.document.readyState==='interactive') {
```
loitly: Wouldn't it be more straight forward if you were checking for 'loading' instead?
```
if (loading) 
 - add load listening
```

robyww: Your right would look cleaner. Maybe it is because I am just not that familiar with `readyState` but I did it this way because it calls out that states I care about, therefore a little more self documenting.   I think I will leave it this way for now.

PR #478, opened 2017-10-12
IRSA-641 Extend the saving/uploading feature to include the choice of…

wmiipac:
I have reverted most of the changes in FileUpload and FileUploadViewPanel (these were written for the current image search specifically). So it seems to be incomplete.
for the image select upload and time series viewer upload panel, the upload workspace needs to be connected to the file picker. There is no selection logic applied for now.
To view the extended save/upload panel:
1. run an image search,  from image toolbar, click save icon
2.    in firefly/image search panel
3.    in firefly/catalog search load catalog panel
4.   in firefly/Upload panel
5.   time series viewer
6.   added a File Upload dialog in firefly-dev.html page, for testing development of file pick only

### DIFF
--- a/src/firefly/js/core/ReduxFlux.js
+++ b/src/firefly/js/core/ReduxFlux.js
@@ -61,6 +61,9 @@ import RegionPlot from '../drawingLayers/RegionPlot.js';
 import MarkerTool from '../drawingLayers/MarkerTool.js';
 import FootprintTool from '../drawingLayers/FootprintTool.js';
 import {showExampleDialog} from '../ui/ExampleDialog.jsx';
+import {showFileUploadDialog} from '../ui/FileUploadDialog.jsx';
+import {DownloadOptionPanel} from '../ui/DownloadDialog.jsx';
+
 
 //==============
 // import Perf from 'react-addons-perf';
@@ -157,6 +160,16 @@ actionCreators.set(ChartsCntlr.CHART_OPTIONS_UPDATE, ChartsCntlr.makeChartOption
 
 actionCreators.set('exampleDialog', (rawAction) => {
     showExampleDialog();
+    return rawAction;
+});
+
+actionCreators.set('fileuploadDialog', (rawAction) => {
+    showFileUploadDialog();
+    return rawAction;
+});
+
+actionCreators.set('downloadDialog', (rawAction) => {
+    DownloadOptionPanel();
     return rawAction;
 });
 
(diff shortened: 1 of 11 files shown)


src/firefly/js/core/ReduxFlux.js
```diff
@@ -160,6 +163,16 @@ actionCreators.set('exampleDialog', (rawAction) => {
     return rawAction;
 });
 
+actionCreators.set('fileuploadDialog', (rawAction) => {
+    showFileUploadDialog();
+    return rawAction;
+});
+
+actionCreators.set('downloadDialog', (rawAction) => {
+    DownloadOptionPanel();
```
loitly: showExampleDialog was done early on and was not cleaned up properly.  You should not have to register these dialogs as an action creator.  The action is SHOW_DIALOG in ComponentCntlr.  The added lines in this file should be removed.

src/firefly/js/ui/FileUpload.jsx
```diff
@@ -13,9 +13,9 @@ const UL_URL = `${getRootURL()}sticky/CmdSrv?${ServerParams.COMMAND}=${ServerPar
 
 
 function FileUploadView({fileType, isLoading, label, valid, wrapperStyle,  message, onChange, value, labelWidth,
-                         innerStyle, isFromURL, onUrlAnalysis, fileNameStyle}) {
-    var style = !isFromURL ? Object.assign({color: 'transparent', border: 'none', background: 'none'}, innerStyle) : (innerStyle || {});
-    var fileName = (!isFromURL && value) ?  value.split(/(\\|\/)/g).pop() : 'No file chosen';
+                         innerStyle, isFromURL, isFromWsURL, isFromFile, uploadSrc, onUrlAnalysis, fileNameStyle}) {
```
loitly: uploadSrc added but not used.

ejoliet:
Another question:
- Is there also a download/upload demo html page using API where i can trigger the dialog option for choosing local or workspace files? That would definitively be helpful.

cwang2016:
file upload from local file system and from url on both firefly/image search panel and  firefly/catalog search load catalog panel have some running issues which is mainly caused by:
- there are two props, isFromFile and  isFromWsURL, added to 'FileUpload' component in order to denote if the upload is from file, url or workspace, these newly defined props should be added onto 'FileUpload' instance used by 'FITS file' under image search and 'Load Catalog' under catalog search.

src/firefly/js/ui/FileUpload.jsx
```diff
@@ -139,7 +163,7 @@ function handleChange(ev, fireValueChange, type, fileAnalysis) {
     });
 }
 
-function makeDoUpload(file, type, isFromURL, fileAnalysis) {
+function makeDoUpload(file, type, isFromURL, isFromWsURL, fileAnalysis) {
```
cwang2016: parameter 'isFromWsURL' is introduced into this function 'makeDoUpload', please check if every function call on 'makeDoUpload' in this file has add this newly added parameter.

PR #481, opened 2017-10-17
IRSA-924: fix issues found in new image search

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-924

- add select/deselect all at the dataproduct(project) level
- fix mission filter now that mission info is available
- make count returns the number of dataproducts instead of dataset(images)
- make CollapsiblePanel's header takes a string or a component.
- make ImageSelect requires a way to register changeListener
- add option to FormPanel to allow unmounted fields in the request
- set NewImageSearchPanel to take unmounted fields
- fix firefly-dev.html not showing images after embedded db merge
- fix NewSearchPanel not submitting search after previous PR merge.

To test:
http://localhost:8080/firefly/firefly-dev.html
- Image Concept
- enter position, select dataset(s), then click search.

### DIFF
--- a/src/firefly/html/firefly-dev.html
+++ b/src/firefly/html/firefly-dev.html
@@ -13,7 +13,6 @@
         window.firefly = {
             app: {
                 template: 'FireflyViewer',
-                views: 'tables',
                 options : {
                     MenuItemKeys: {maskOverlay:true},
                 }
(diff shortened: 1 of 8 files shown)


ejoliet:
Not priority/urgent, but I will also raise separate ticket to add a help anchor for tooltip and link to mission/dataset info

PR #482, opened 2017-10-19
IRSA-870: added PDF, HTML, PNG and FITS files download for seip and akari

ejoliet:
I've added the files download for SEIP (and AKARI when IRSA-820 is done on client side) on server for Finderchart. There is same branch name in IFE.

Build from there Finderchart, and test searching for SEIP/WISE, try to download pdf, html png and fits.

I've also fixed a rounding problem for small size (<1MB) displayed in background monitor status.

Let me know.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/visualize/WebPlotRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/visualize/WebPlotRequest.java
@@ -30,8 +30,9 @@ import java.util.List;
  */
 public class WebPlotRequest extends ServerRequest {
 
+    // TODO this is actually coupled with edu.caltech.ipac.firefly.data.FinderChartRequestUtil.ImageSet.ImageSet and so we need to add imageset here although SEIP, AKARI are using ATLAS services.
     public enum
-            ServiceType {IRIS, ATLAS, ISSA, DSS, SDSS, TWOMASS, MSX, DSS_OR_IRIS, WISE, NONE}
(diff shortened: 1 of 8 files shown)


ejoliet:
Should be good also to test PDF export from Finderchart API.

src/firefly/java/edu/caltech/ipac/firefly/visualize/WebPlotRequest.java
```diff
@@ -377,6 +378,29 @@ public static WebPlotRequest makeWiseRequest(WorldPt wp,
         return req;
     }
 
+    /**
+     * @param wp
+     * @param survey for atlas, survey is in form of 'schema.table'
+     * @param band SEIP exmaple 'irac1'
+     * @param filter for SEIP, it should loo like type=science and fname like %.mosaic.fits
```
loitly: I can't find where "type=science and.." is specified.  Is this something that's given via meta, or it's something we need to send along as part of IBE's api?  My concern is where this logic is kept.

ejoliet: yes, it will be given from the master table but for FC3 we need to set it in the code. And yes, IBE needs it as part of the query (where clause), different datasets have different filter.

Once we have the framework in place for IV new image search working, and decide to bring the same component to FC, we will need to refactor the FC server code so it can handle generic web plot request and download file from reading the same master table.

PR #483, opened 2017-10-20
IRSA-811: Download PDF goes directly to background monitor

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-811
Changes included in this PR.
- create framework for submitting download tasks with masking and 'send to background' feature
- adapt new finderchart to the newly created framework.
- introduce 'watchers' concept to MasterSaga
- use web.xml from firefly by default
- use web-fragment.xml to add additional settings at the project level

Notes:  more changes are in ife's repo under the same branch.

To test:  
- build finderchart
- single or multi position search
- click on 'PDF' or 'download'.


loitly:
@robyww, per our conversation, I've switched the new FinderChart to use the "send to background" option feature instead of automatically doing it.  You can take a look to see how you like it.
Also, I've added 'watcher' to MasterSaga.  Watchers are long lived sagas with the hooks to cancel.  You can also view it as action listeners.  See MasterSaga for changes.

src/firefly/js/core/MasterSaga.js
```diff
@@ -16,18 +21,99 @@ export function dispatchAddSaga(saga, params={}) {
...
+        switch (action.type) {
+            case ADD_SAGA: {
+                const {getState, dispatch}= flux.getRedux();
+                const {saga,params}= action.payload;
+                if (typeof saga === 'function') yield fork( saga, params, dispatch, getState);
+                break;
+            }
+            case ADD_WATCHER: {
+                const {getState, dispatch}= flux.getRedux();
+                const {id, actions, callback}= action.payload;
```
robyww: I think a watcher should be able to take a params object like add saga.  It might not be as important since many watch would probably closures but there is not reason not to have it.

src/firefly/js/core/MasterSaga.js
```diff
@@ -16,18 +21,99 @@ export function dispatchAddSaga(saga, params={}) {
...
\ No newline at end of file
+}
+
+
+function createSaga({id, actions=[], callback}) {
+    const saga = function* () {
+        while (true) {
+            const action= yield take(actions);
+            try {
+                callback && callback(action);
```
robyww: This watcher could be simplify and mostly replace the way we use sagas if we called the callback with following parametes:

1. action
1. cancelSelf
1. params
1. dispatch
1. getState

i.e.

`callback(action, ()=> dispatch({ type: CANCEL_WATCHER, payload: {id}}), params, dispatch, getState)`

PR #489, opened 2017-11-01
IRSA-975: NED searching seems not to work

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-975
- column names must be enclosed in double-quotes
- remove caching done at the search processor’s level
- introduce SharedDbProcessor: results from same processor share the same database
- convert IBE’s template into a search processor
- convert XYGenericProcessor into an EmbeddedDbProcessor
- use firefly’s config file where applicable.
- limit log message to 2k characters

To Test:
- build firefly
- Catalogs -> NED -> m16 -> 1.2 deg
Should get a tri-view with decimated chart and image plot.
What caused the problem was the column names.  They contains spaces, parentheses, etc
This ticket should fix all of that.  There is no restricting on column name now.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
@@ -28,7 +28,7 @@ import static edu.caltech.ipac.firefly.data.TableServerRequest.INCL_COLUMNS;
  * @version $Id: DbInstance.java,v 1.3 2012/03/15 20:35:40 loi Exp $
  */
 abstract public class BaseDbAdapter implements DbAdapter {
-    private static long MAX_IDLE_TIME = 1000 * 60 * 5;      // will be cleaned up if idle more than 5 minutes.
+    private static long MAX_IDLE_TIME = 1000 * 60 * 15;      // will be purged up if idle more than 5 minutes.
     private static Map<String, EmbeddedDbInstance> dbInstances = new HashMap<>();
     private static Logger.LoggerImpl LOGGER = Logger.getLogger();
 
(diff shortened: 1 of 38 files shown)


loitly:
@tgoldina , I've just made another commit to fix the problem you found.  It involved FireflyHeatMap.js returning a bad mappings object on re-mount.  Can you take a look to see if the change is acceptable?  Thanks.

src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
```diff
@@ -28,7 +28,7 @@
  * @version $Id: DbInstance.java,v 1.3 2012/03/15 20:35:40 loi Exp $
  */
 abstract public class BaseDbAdapter implements DbAdapter {
-    private static long MAX_IDLE_TIME = 1000 * 60 * 5;      // will be cleaned up if idle more than 5 minutes.
+    private static long MAX_IDLE_TIME = 1000 * 60 * 15;      // will be purged up if idle more than 5 minutes.
```
tgoldina: the comment here does not match the number any more

src/firefly/js/visualize/saga/CatalogWatcher.js
```diff
@@ -114,7 +114,7 @@ function handleCatalogUpdate(tbl_id) {
     const params= {
         startIdx : 0,
         pageSize : MAX_ROW,
-        inclCols : `${columns.lonCol},${columns.latCol},ROW_IDX`
+        inclCols : `"${columns.lonCol}","${columns.latCol}",ROW_IDX`        // column names should be in quotes
```
tgoldina: Only the unknown columns should be in quotes. The columns that we know (like ROW_IDX) are safe, right?

From tis point of view we did not have to add quotes to columns like Peak, Period, etc.

loitly: Actually, all column names should be in quotes.  But because ROW_IDX is all-caps single word, it does not need to be in quotes.  But, I should quotes them as well for consistency.

PR #494, opened 2017-11-10
IRSA-909-927: fixed 2 issues

robyww:
- toolbar is now correct with only one images
  - for not images found- fixed exception and shows correct display

_note:_
you must checkout the same branch in ife to test.


robyww:
Ok. I will look at that as well.

robyww:
Ok, I added the fixes.  I think the expanded should be as expected.

PR #496, opened 2017-11-11
IRSA-793 IRSA-798-add image type option to main panel add image source panel option

lznakano:
Add source and type options
   Changed the RadioGroupInputField and RadioGroupInputFieldView to allow a customized label style
  Combine two .css file to one as NewImageSearch.css

Note:  
  For three color images, only irsa works.  You have to check both missionId and band.  More work is needed to get three color images work.

### DIFF
--- a/src/firefly/js/ui/ImageSelect.jsx
+++ b/src/firefly/js/ui/ImageSelect.jsx
@@ -13,7 +13,7 @@ import FieldGroupUtils, {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
 import {dispatchComponentStateChange} from '../core/ComponentCntlr.js';
 import {updateSet} from '../util/WebUtil.js';
 
-import './ImageSelect.css';
+import './NewImageSearchPanel.css';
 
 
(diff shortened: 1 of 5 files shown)


src/firefly/js/ui/ImageSelect.jsx
```diff
@@ -13,7 +13,7 @@ import FieldGroupUtils, {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
 import {dispatchComponentStateChange} from '../core/ComponentCntlr.js';
 import {updateSet} from '../util/WebUtil.js';
 
-import './ImageSelect.css';
+import './NewImageSearchPanel.css';
```
loitly: Not duplicating the style is a good idea, but in this case, most of the style in ImageSelect.css is for ImageSelect only.  You should have just moved the relevant style from ImageSelect to NewImageSearchPanel.css.

lznakano:
Thanks @Loi for reviewing it.   OK, I will do what you suggested in the dev.

PR #498, opened 2017-11-14
IRSA-944: FinderChart fail to recognize when search catalog is set to 'no'

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-944
Also, fixed working mask not showing during download.

Additional changes in ife under same branch.
To test:
- build finderchart
- select 'no'  for 'search corresponding catalog'.


ejoliet:
Works great! Thanks for fixing both problem the catalog option and the background monitor!

PR #500, opened 2017-11-14
DM-12268 Fix fits image rendering problem since latest FITS library update

cwang2016:
The development fixes the image rendering by changing the way to clone the fits header. 

test: 
go to localhost:8080/firefly
do image search on m31, WISE image, '4 band level1' and band 2. 
or 
do upload, image search on the following sample, 
- firefly_test_data/FileUpload-samples/fits/imagetable/hpacs1342250905_00hps3d_00.fits, extension 1. 

issue: 
there is some issue regarding how nom.tam.fits library parses the fits header when the keyword value ending in '&' and immediately followed by 'CONTINUE' line. The string after the CONTINUE keyword is ignored if keyword 'LONGSTRN' is defined after the keyword with long string values per the parsing method of current nom.tam.fits library used, therefore the keyword value of long string case may be displayed inaccurately. (i.e. it ends with '&' and without containing the string after CONTINUE keyword).


src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1727,18 +1717,21 @@ public void freeResources() {
...
         Cursor iter = header.iterator();
-        String cards[] = new String[header.getNumberOfCards()];
-        int i = 0;
+        Header clonedHeader = new Header();
+
         while (iter.hasNext()) {
             HeaderCard card = (HeaderCard) iter.next();
-            cards[i] = card.toString();
-            i++;
+            clonedHeader.addLine(card);
```
robyww: I think we should make a copy.

`clonedHeader.addLine(card.copy());`
_or_
`clonedHeader.addLine(new HeaderCard(card.getKey(), card.getValue(), card.getComment()));`

PR #501, opened 2017-11-17
IRSA-791: Create new image search panel

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-791

You may select multiple images when 'Single channel image' type is selected.
In '3-color' mode, you can only select one image for each color.  You may use any combination of colors and not necessarily all 3.  Select 'None' for #2(image source) to exclude that color.

NOTE:  There is a 3-color display bug in dev.  I showed it to @robyww, and he agreed that it's probably a drawing issue and not this panel selection's problem.

To Test:
- goto http://localhost:8080/firefly/firefly-dev.html
- 'Image Concept'

This panel should work under any condition and in all browsers except for the 3-color bug mentioned above.  You should report all anomalies.


robyww:
Initial UI Notes:

 - The select all should say something like 'MSX - all' or 'DSS - all'
 - when I plot 3 images and return and select 3 other images, I get 6, because the previous 3 are still selected
 - if I filter on 2MASS, leave and come back, 2 mass is still filter but the filter is hidden behind the more. I is confusing
 - The panel fills too crowded. It needs more room. At least needs to be wider.
 - I need feedback somewhere on how many image are selected.
 - The select data set section should take all the room available to it. Currently it is a fixed size.  This probably means adding more layout options to the FormPanel.

src/firefly/js/ui/ImageSearchPanelV2.jsx
```diff
@@ -0,0 +1,432 @@
...
+
+function makeFitsWebRequest(request, color='') {
+
+    var fits = get(request, 'fileUpload');
+
+    const params = {
+        [WPConst.PLOT_ID] : rootId+rootIdCnt,
+        [WPConst.FILE] : fits,
+        [WPConst.PLOT_GROUP_ID] : 'multiImageGroup',
+        [WPConst.GROUP_LOCKED] : 'true'
```
robyww: `RequestType` is normally implied but when you add workspace you will have to set it.
`[WPConst.TYPE] : RequestType.WORKSPACE`

src/firefly/js/ui/ImageSearchPanelV2.jsx
```diff
@@ -0,0 +1,432 @@
...
+    return WebPlotRequest.makeFromObj(inReq);
+}
+
+function makeURLWebRequest(request) {
+
+    var url = get(request, 'txURL');
+    const params = {
+        [WPConst.PLOT_ID] : rootId+rootIdCnt,
+        [WPConst.URLKEY] : url,
+        [WPConst.PLOT_GROUP_ID] : 'multiImageGroup',
```
robyww: plot group will have to eventually come from a property, PLOT_ID as well in a coverage case when you are just replacing.

src/firefly/js/ui/ImageSearchPanelV2.jsx
```diff
@@ -0,0 +1,432 @@
...
+                        {label: '3-color', value: 'threeColor'}]}
+                fieldKey='imageType'
+            />
+        </FieldGroup>
+    );
+}
+
+function ImageSource({groupKey, imageMasterData, multiSelect}) {
+    const isThreeColor = getFieldVal(FG_KEYS.image_type, 'imageType') === 'threeColor';
+    const options = [   {label: 'IRSA', value: 'irsa'},
```
robyww: ISRA needs to be set-able from a property,  I would prefer it say `Archive` and the property sets it to `IRSA`

loitly:
I added a commit to address some of the feedbacks from this PR.
The ones not addressed are in a new ticket I've raised: https://caltech-ipac.atlassian.net/browse/IRSA-1070
Here are what was fixed:
- moved ImageSearchPanel to visualize/ui
- renamed IRSA to Archive and made it set-able
- allow PLOT_ID and PLOT_GROUP_ID to be passed in
- increase vertical size to lessen clutter
- fix 3-color retaining prior selection
- add better 3-color error messages

PR #505, opened 2017-11-29
DM-12551: HiPS: Make the zoom give more feedback and other optimizations 

robyww:
- better zoom feedback: zoom draws existing tiles when zooming
 - improved caching
 - loading allsky when switching repositories
 - feedback when switching repositories
 - many drawing optimizations
 - text can now be drawn with canvas
 - reorder level 2 all sky to make a level 2 for faster drawing
 - should work (slowly) if there is not allsky image
 - failed projection conversion now returns null instead of throwing exception
 - moved aladinProj to js/externalSource


tgoldina:
First observation:
http://localhost:8080/firefly/firefly-dev.html -> Test Searches -> HiPS -> ngc 13
Clicking on 'zoom in' has no effect until I try 'zoom out' or 'zoom to fit' (needs to be fixed)
Initial HiPS level is 2 (which is expected)

tgoldina:
A few more observations:

http://localhost:8080/firefly/firefly-dev.html -> Test Searches -> HiPS -> ngc 13
I have zoomed in to level 12 and  switching Options to something else. The level is not changing, only the relevant tiles seem to be retrieved, but the mask message says "Retrieving AllSky"

Tooltip for Options:
'Choose a differ HiPS survey' -> 'Choose a different HiPS survey'

In Layers dialog: 'HiPS' -> 'HiPS Grid'  

It've very nice to be able to select grid behavior. I think switching from Auto or Match to Lock should preserve current grid level. I would assume user wants to lock currently used grid level.
(At the moment, when switching from Auto to Lock first time, the level is automatically set to 3. Later it remembers the latest selection.)

PR #507, opened 2017-11-29
IRSA-748: Replace image search with new ImageSearchPanelV2

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-748

Also includes:
- https://caltech-ipac.atlassian.net/browse/IRSA-955
- https://caltech-ipac.atlassian.net/browse/IRSA-956
- fix TargetPanel resolution in API demo pages
- fix API showImage and showCoverage not setting the right viewer's option

Build firefly and irsaviewer on this branch
Anywhere image search/load is available, the new ImageSearchPanel will appears.
This includes API, Gator, and elsewhere in firefly.

When updating current image, then panel allow for only 1 image to be selected.
Otherwise, you can select more than one, and have them all loaded at once.

@cwang2016 , will you help check to see if any features missing in this new panel.
@robyww , will you do the same.  Also, you want to look over the small changes I made to image API.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -70,7 +70,7 @@
     "react/no-did-update-set-state": 1,
     "react/no-direct-mutation-state": 1,
     "react/no-unknown-property": 1,
-    "react/prop-types": [1, {"ignore": ["children"]}],
+    "react/prop-types": [1, {"ignore": ["children"], "skipUndeclared": true}],
     "react/react-in-jsx-scope": 1,
     "react/self-closing-comp": 1,
     "react/wrap-multilines": 1,
--- a/src/firefly/html/firefly-dev.html
+++ b/src/firefly/html/firefly-dev.html
@@ -15,7 +15,8 @@
                 template: 'FireflyViewer',
                 options : {
                     MenuItemKeys: {maskOverlay:true},
-                    workspace : {showOptions: true}
+                    workspace : {showOptions: true},
+                    imageMasterSourcesOrder: ['WISE', '2MASS', 'Spitzer']
                 }
             },
         };
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/VisServerCommands.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/VisServerCommands.java
@@ -126,7 +126,7 @@ public class VisServerCommands {
             }
 
 
-            data.put("success", true);
+//            data.put("success", true);
 
             JSONArray wrapperAry= new JSONArray();
             obj.put("data", data);
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotCreator.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotCreator.java
@@ -111,6 +111,7 @@ public class ImagePlotCreator {
                 if (state.isThreeColor()) {
                     plot.setThreeColorBand(state.isBandVisible(band) ? readInfo.getFitsRead() :null,
                             band,frGroup);
+                    stretchBand(band,state,plot,frGroup);
                 }
                 if (readInfo.getModFileWriter()!=null) fileWriterMap.put(band,readInfo.getModFileWriter());
                 first= false;
@@ -209,13 +210,19 @@ public class ImagePlotCreator {
             FitsCacher.addFitsReadToCache(retval.getTargetFile(), new FitsRead[]{tmpFR});
         }
 
-        RangeValues rv= state.getRangeValues(readInfo.getBand());
+        stretchBand(band,state, plot,frGroup);;
+        return retval;
+    }
+
+    private static void stretchBand(Band band, PlotState state, ImagePlot plot, ActiveFitsReadGroup frGroup) {
+        RangeValues rv= state.getRangeValues(band);
+        HistogramOps histOps= plot.getHistogramOps(band,frGroup);
         if (rv==null) {
             rv= FitsRead.getDefaultFutureStretch();
-            state.setRangeValues(rv, readInfo.getBand());
+            state.setRangeValues(rv, band);
         }
         histOps.recomputeStretch(rv, true);
-        return retval;
+
     }
 
     private static void initPlotTitle(PlotState state,
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/imagesources/ImageMasterData.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/imagesources/ImageMasterData.java
@@ -10,9 +10,14 @@ import edu.caltech.ipac.firefly.server.util.Logger;
 import org.json.simple.JSONArray;
 import org.json.simple.JSONObject;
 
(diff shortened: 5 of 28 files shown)


cwang2016:
please check if this is a bug, 
- select single channel image -> archive -> target: m31 -> SDSS -> u
   the image is shown as Failed - SDSS Area not covered. 
- back to image search panel, unckeck SDSS and check WISE (at this time SDSS/u seems still stored, but not seen) -> click search
result: no error message popped up to ask for selecting at least an image and the image is still shown as 'Failed - SDSS Area not covered. '
Q: should SDSS/u be unchecked before SDSS is unchecked?

cwang2016:
Is this as design? 
'Archive' is the default selection for 'Single channel image' and 'None' is the default selection for 3-color red, green and blue?

cwang2016:
Is this working as designed?  There is  no 'replace image' (vs. create new image)  from the image search?

cwang2016:
A very minor question, do you want to add ':' to "Choose Radius'  like 'Name or Position: '.

cwang2016:
This is regarding the image rendering, no title (only zoom number)  is shown at the left corner if the image is selected  from 'Upload' and image file.

ejoliet:
> select single channel image -> archive -> target: m31 -> SDSS -> u
the image is shown as Failed - SDSS Area not covered.
Not bug, try m81 ;-)

cwang2016:
This is regarding 
select single channel image -> archive -> target: m31 -> SDSS -> u

Sorry I didn't state clearly. I didn't mean ' Failed - SDSS Area not covered.' is a bug.  What I meant is when you go back to deselect 'SDSS' and then select WISE or others (without selecting an image), the search is still repeatedly done on SDSS since the same error message is displayed.

robyww:
Initial Observations:

-  file upload and url should use the WebPlotRequest factory methods: `makeFilePlotRequest` and `makeURLPlotRequest`. This will fixed the title issues the @cwang2016 saw.
- Workspace should be added.
- When workspace is added it needs to check app options: `workspace.showOptions`
- We need to implement the sort order parameter in `ImageMasterData.java` and `GetMasterImageData` on the server side. This must be done we can't hard code  the order. 
- The client should specify the sort order in app options in the html file.  This allows IRSA and LSST to get the correct images on top.
- I still really think we need to clear the image options when the dialog is raise.  It is just too easy to bring up multiple copies of the same image.
- I think it would look a little less cramped it the `.ImageSearch__section` has `padding: 6px 2px 4px 2px`, `margin: 2px 4px 2px 4px`  and `width: calc(100% - 8px);`

loitly:
> 'Archive' is the default selection for 'Single channel image' and 'None' is the default selection for 3-color red, green and blue?
This was my decision.  I wanted the user to purposely select which color they want to create.  If they choose just 1 or 2, they should not be prompt with validation errors, then needing to go back and deselect what they did not select in the first place.  If this is not desire, I have no problem changing it to whatever the desire behavior should be.  @ejoliet @robyww

loitly:
> Is this working as designed? There is no 'replace image' (vs. create new image) from the image search?

Yes, it's default to add.  We will replace only when add is not allowed.

loitly:
> select single channel image -> archive -> target: m31 -> SDSS -> u

This is normal, because we do not clear previous results.  Therefore, WISE was added to the previously results leading you to think that 2 images was added on your second try when in fact only WISE was added.

ejoliet:
>file upload and url should use the WebPlotRequest factory methods: makeFilePlotRequest and makeURLPlotRequest. This will fixed the title issues the @cwang2016 saw.

Good!

>    Workspace should be added.    When workspace is added it needs to check app options: workspace.showOptions

Agree.
    
> We need to implement the sort order parameter in ImageMasterData.java and GetMasterImageData on the server side. This must be done we can't hard code the order.
The client should specify the sort order in app options in the html file. This allows IRSA and LSST to get the correct images on top.

I vote for postponing unless it can be done in this branch.

> I still really think we need to clear the image options when the dialog is raise. It is just too easy to bring up multiple copies of the same image.

Until we have a way to see what have been selected, i would suggest to keep the selected checkboxes, otherwise there is no other way currently to see what i've selected. Right?

loitly:
In response to @ejoliet comment, I've already done most of the suggested changes in this PR.
I still need to add workspace and selected/filtered indicators.

So far, these are fixed:
-  file upload and url should use the WebPlotRequest factory methods: `makeFilePlotRequest` and `makeURLPlotRequest`. This will fixed the title issues the @cwang2016 saw.
- We need to implement the sort order parameter in `ImageMasterData.java` and `GetMasterImageData` on the server side. This must be done we can't hard code  the order. 
- The client should specify the sort order in app options in the html file.  This allows IRSA and LSST to get the correct images on top.
- I still really think we need to clear the image options when the dialog is raise.  It is just too easy to bring up multiple copies of the same image.
- I think it would look a little less cramped it the `.ImageSearch__section` has `padding: 6px 2px 4px 2px`, `margin: 2px 4px 2px 4px`  and `width: calc(100% - 8px);`
-A very minor question, do you want to add ':' to "Choose Radius' like 'Name or Position: '.

loitly:
@cwang2016 @robyww @ejoliet 

I've pushed a new commit to deal with the above requests.
I still have `workspace` to do tomorrow.
But since we're on a tight schedule, if you can review the new UI elements to see if it's acceptable while I work on `workspace`.   It would be helpful.

Here are the things I added:
- use WebPlotRequest factory methods
- implement ImageMasterData sortOrder
- clear inputs and cosmetics changes
- add selected indicators in filter and selection panels
- added clear filters and clear selections

If you go to firefly-dev.html, you'll get the new sorted feature added using ['WISE', '2MASS', 'Spitzer'] as the sort param.

I, without thinking, rebase my branch with dev.  You'll have to remove your local branch to get my new changes.  Sorry.

ejoliet:
Thanks. 
I'm happy to see the introduction of "clear selection". Selected indicators is a cool workaround! Like it! Nice idea!

I've tried to filter MSX, SDSS OR DSS but it fails with "No data match these criteria
".

I'm not sure that we still want to clear the selection of the form. It seems that now when i go back to the form, it also cleared the search target. I don't think is an improvement. but anyway, we don;t have time for that so ship it as it is, i'm sure others will have different opinion and this shouldn't stop from merging.

The idea of the order for IRSA (Archive) was that it was coming from the table otherwise Vandana can't control it with fine granularity but again, is fine for now. We will have to see how we adjust to it.

robyww:
Starting to look at it... The three color does not seem  to work anymore.  When I click search it does not do anything.
So far everything else looks good.  I will keep testing.

loitly:
Latest commit includes workspace option for Image Source.  This PR is complete and ready to merge once approve.
NOTE:
You need to edit `/hydra/server/tomcat/webapps/firefly/WEB-INF/config/app.prop` and set `ignore.auth=false` for workspace to work.

cwang2016:
The latest commit works well. Here a small concern, workspace upload related. The title shown on the image is the filename at the server instead of the original filename from the workspace, not like what Fits file upload we did on local file system.

PR #512, opened 2017-12-07
DM-12940 Fix the error while adding marker or footprint to the plot without image

cwang2016:
The development fixes the error when adding marker or footprint to the drawing layer containing the plot without image, 
- skip adding the marker or footprint to the plot that is attached to the marker/footprint drawing layer and has no image. 
- disable the buttons on the toolbar in case the selected plot containing no image. 

test case 1:
go to firefly do image search, 
target: m31, 
select all images under  'SDSS' and 'WISE All-Sky Altas'

result: No image found for SDSS, and  buttons on the tool bar for those plots are disabled. 

test case 2:
go to finderchart do multiple positions search. 
upload file 'forirsa_ref1.tbl' as attached in https://jira.lsstcorp.org/browse/DM-12940
Marker tool can be added to all plots with images with no error.

### DIFF
--- a/src/firefly/js/drawingLayers/FootprintTool.js
+++ b/src/firefly/js/drawingLayers/FootprintTool.js
@@ -8,13 +8,13 @@ import {makeDrawingDef, TextLocation} from '../visualize/draw/DrawingDef.js';
 import DrawLayer, {DataTypes, ColorChangeType}  from '../visualize/draw/DrawLayer.js';
 import {MouseState} from '../visualize/VisMouseSync.js';
 import CsysConverter from '../visualize/CsysConverter.js';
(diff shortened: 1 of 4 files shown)


ejoliet:
Is this related to IRSA-1131?

cwang2016:
@ejoliet  this one solves the issue as you reported 'Marker tool is broken in FinderChart'. 
I couldn't reproduce the issue stated in IRSA-1131 by using this update. I will check it if those two issues are relevant.

cwang2016:
This update mainly solves the issue when adding marker or footprint into the plot with no image.  Issue of IRSA-1131 is relevant and please check if that is solved, too.

PR #516, opened 2017-12-12
DM-12949 column expression issues in charts

tgoldina:
- fixed filter from chart when column expressions are used
- use db to calculate column expression values for general charts
- allow same columns to be used for x and y

What is not done:
- I might need to update DecimationProcessor and HistogramProcessor as a part of this PR.
- Expression validation is still done using old Expression parser. Column expression input field tooltip is updated with the operators and functions that are common for both java/js expressions and h2 db. A separate ticket will address column expression validation issues. 
- XYGenericProcessor and XYWithErrorProcessor can not be removed while we are using old single-trace chart architecture. After we finally switch to multi-trace charts, we should remove the code supporting old charts.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -395,7 +395,9 @@ export function getDataChangesForMappings({tableModel, mappings, traceNum}) {
 
     if (mappings) {
         Object.entries(mappings).forEach(([k,v]) => {
-            changes[`data.${traceNum}.${k}`] = getDataVal(v);
+            // using plotly attribute path (key in the mappings object) as a column name
+            // this makes it possible to use the same column as x and y, for example
+            changes[`data.${traceNum}.${k}`] = getDataVal(k);
         });
     }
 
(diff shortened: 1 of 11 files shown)


src/firefly/js/charts/ChartUtil.js
```diff
@@ -652,6 +654,34 @@ export function applyDefaults(chartData={}, resetColor = true) {
     });
 }
 
+/**
+ * Convert column expression to h2 database syntax
```
loitly: change `h2` to `internal`, or `hsqldb`.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -652,6 +654,34 @@ export function applyDefaults(chartData={}, resetColor = true) {
...
+export function formatColExpr(colOrExpr, colNames) {
+    const expr = new Expression(colOrExpr, colNames);
+    if (expr.isValid()) {
+        // quote columns, assuming column names are alpha-numeric
+        expr.getParsedVariables().forEach( (v) => {
+            const re = new RegExp('([^A-Za-z\d_"]|^)('+v+')([^A-Za-z\d_"]|$)', 'g');
+            colOrExpr = colOrExpr.replace(re, '$1"$2"$3'); // add quotes
+        });
+
+        // substitute functions ceil, lg, ln with the functions h2 understands
```
loitly: If the decision is to support these functions as is, then the indirection should happen in DbAdapter.  This is where you would convert publicly supported functions into database specific syntax.

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -354,7 +355,7 @@ function chartUpdate(action) {
 
 
         const {data, fireflyData} = Object.entries(changes)
-                             .filter(([k,v]) => (k.startsWith('data') || k.startsWith('fireflyData')))
+                             .filter(([k,]) => (k.startsWith('data') || k.startsWith('fireflyData')))
```
loitly: `([k,])` this is great.  This is good compromise between unused variable warning and an indication that more parameters are available for use if needed.

tgoldina: In the latest commit:
- Fixed filtering issues for expressions with alpha-numeric columns. Unfortunately, if a column name contains space, filter will not work the second time, because filter parser is using space as a delimiter. You can check it on a NED table in irsawebdev1.
- Updated histogram request to get the expression calculation in the database.
- Fixed selection bug in heatmap I have introduced when checking that user is selecting active trace.

I can not reproduce column dialog issue reliably.

Not sure, I can move lg->log10 to DbAdopter at the moment, I need to get Expression parser to handle quoted tokens, to make sure "lg(" is not part of the column name. 

I'll create tickets for outstanding issues.

PR #517, opened 2017-12-15
DM-12555 query HiPS server to get available HiPS

cwang2016:
The development includes,
- send query to HiPS server to get public HiPS
   per information from [http://aladin.u-strasbg.fr/hips/list](url) 
   the available HiPS images can be retrieved by sending the query like
  [http://alasky.unistra.fr/MocServer/query?hips_service_url=*&dataproduct_type=!catalog&dataproduct_type=!cube&get=record](url)
- show the list of the found public HiPS  in a table 

test:
 - go firefly/firefly-dev.html
 - select 'Test Searches', 
 - select 'HiPS', a table containing available HiPS is shown. 
 - highlight the HiPS to be searched, click 'search'
 - on the tri-view page, click 'Find HiPS', and get a popup with a table containing all available HiPS 
 - highlight the HiPS for another search.


robyww:
UI Notes:

- It is work exactly as we imaged. That's great.
- This is more than the scope of the ticket but I think you should go ahead and do it. I think there needs to be a check box the says `popular`. When click it should restrict the list to those in `hipsSURVEYS` object in `HipsUtils.js`
- The search needs to fields need to be turned on for the table.  I want to be able to type DSS and see all the DSS HiPS
- Title should be wider and URL more narrow.

src/firefly/js/ui/HiPSViewer.jsx
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
```
robyww: HiPSViewer is not the right name of this file.  How about `HiPSSurveyListDisplay.jsx`.  HiPSViewer sounds like it is the viewer not the list of surveys.

src/firefly/js/visualize/HiPSCntlr.js
```diff
@@ -0,0 +1,193 @@
+import {flux} from '../Firefly.js';
```
robyww: This is good.  I had not considered keeping the hips list in the store.  This is good.

src/firefly/js/ui/HiPSViewer.jsx
xiuqin: We do need to make the popular list settable by different applications, even the same application at different deployment sites. I can see SUIT portal at NCSA and Chile setting different sites due to network.

cwang2016:
- fix client table filter processing issue (processRequest() in TableUtil.js & creatorComparator()  in FilterInfo.js)
- add 'popular' options for showing HiPS popular surveys out of all available surveys, the popular list is currently hard coded. It will be made to be settable in the future. 
- create component containing a check box for showing popular surveys and the table with information of available survey images.

PR #520, opened 2017-12-19
IRSA-1287: fixed: Firefly Lock by Click feature is broken

robyww:


ejoliet:
Update: 
(As i said Firefly build is ok and issue is fixed so that PR is good to go.)
I've had to merge rc into your branch to get stuff that didn't have and failed to compile in IFE, and then i could test the fix in FC/IFE and i confirm is fixed. But i will need to check why dev branch doesn't have all the thing from RC. That will be easy to merge back anyway.

PR #522, opened 2017-12-21
IRSA-1284:Finder Chart example does not work

lznakano:
Fixed the Finderchat example bugs.  No IFE changes.


lznakano:
If you look at the examples in firefly catalog search and image search (in dev branch), none of them accept ',' right now. If you copy and paste the example (46.53, -0.251 gal), the validation will fail.  The TargetPanel.jsx validates the input, if there is a comma, it returns false.

xiuqin:
There are two wrong examples - with "," - in the "Select Target" input panel. We should fix it too.

lznakano:
@ejoliet  I tested all the examples in this branch.  Can I merge it now?

lznakano:
The GWT image search for wise does accept ','.  So the example works fine in the GWT version of the search.

PR #526, opened 2018-01-09
IRSA-1290: Finder Chart URL construction feature broken

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-1290

Allow finderchart UI to be launched by a URL.  The method used to construct the URL is described in the ticket above.

Changes are in firefly as well as ife.


ejoliet:
I've also noticed that catalog are mostly empty. And i think why but i just wonder what is the radius search value passed to the lanucher? 
I'm guessing it is the default one (5" that i see in the form), right? 
If that is the case, instead of adding an attribute to the json request (which i don't think we had that radius search available in the old url construction api) we could just set a better default radius search per catalog. I will ask Vandana for that input.

PR #529, opened 2018-01-12
DM-6929 Support circular selection in addition to current rectangular selection

cwang2016:
This development includes,
- adding circular selection support for area selection. 
- change select area icon on the tool bar from on/off button to a dropdown button. The dropdown list supports the selection for 'rectangular selection', 'elliptical section' (circular area) and 'None'. 
- add stats calculation for circular area selection on both rotated and non-rotated images. 
- fix stats calculation for rectangular area selection on rotated image case.
- update 'select catalog' and 'filter catalog' functions for circular area selection. 

Test:
1. Do image selection: 
  Click 'select area' icon to get selection dropdown list, 
  Select either 'rectangular selection' or 'elliptical selection' and draw a selected area by choosing a point over the image and dragging the mouse across the image and select 'None' to remove the selected area. 
   Select one of the extension buttons for 'crop', 'stats', 'zoom', or 'recenter' while either  'rectangular selection' or 'elliptical selection' is selected. 

2. Repeat the steps by rotating the image first. 

3. Do catalog selection next:
   repeat the 'select area' steps done as  image selection, and select one of the extension buttons for 'crop', 'stats', 'select mark data', 'filter', 'zoom' or 'recenter' while either 'rectangular selection' or 'elliptical selection' is selected.


src/firefly/js/visualize/ui/SelectAreaDropDownView.jsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export function SelectAreaDropDownView({plotView:pv, allPlots, dropDownCB, crtSelection}) {
+    var enabled = !!pv;
+    let sep = 1;
+
+    const ddCB = dropDownCB ? dropDownCB : null;
+
+    const selectAreaCommands = () => {
+        return [selectAreas.rect, selectAreas.circle, selectAreas.noselect].reduce((prev, s) => {
```
robyww: why use reduce and not map?

src/firefly/js/visualize/ui/VisToolbarView.jsx
```diff
@@ -155,10 +155,19 @@ VisToolbarViewWrapper.propTypes= {
 export class VisToolbarView extends PureComponent {
     constructor(props) {
         super(props);
+
+        this.selectCallBack = this.selectCallBack.bind(this);
+        const dl = getDrawLayerByType(getDlAry(), SelectArea.TYPE_ID);
```
robyww: VisToolbarView is not suppose to have any state.  The class above VisToolbar does keep state.  However, that does not work either because there is the case where I delete the select drawing layer from the LayerControl menu.  I that case there is a bug, because the toolbar show selection is on but it is not.  The toolbar does not go back to selection off.

src/firefly/js/visualize/VisUtil.js
```diff
@@ -593,13 +594,65 @@ export function getBoundingBox(ptAry) {
...
+        const pt0= cc.getDeviceCoords(selection.pt0);
+        const pt1= cc.getDeviceCoords(selection.pt1);
+        if (!pt0 || !pt1) return selectedList;
+
+        const c_x = (pt0.x + pt1.x)/2;
+        const c_y = (pt0.y + pt1.y)/2;
+        const r1_2= Math.pow(Math.abs(pt0.x-pt1.x)/2, 2);
+        const r2_2= Math.pow(Math.abs(pt0.y-pt1.y)/2, 2);
+
+        const containsEllipse = (x, y) => {
```
robyww: can you make `containsEllipse()` a stand alone function like `contains()`.  I know you will have to pass more parameters.

src/firefly/js/visualize/ui/SelectAreaDropDownView.jsx
cwang2016: I used 'reduce' since two tags are pushed into the result array in each iteration, it would be nice if 'map' can make it.

src/firefly/js/visualize/VisUtil.js
cwang2016: will make it

cwang2016:
from @robyww:
When It is is on then clicking on select turns it off. If It is off then clicking will give you get a drop down (just like now) but with only the two options.

It sounds like making a component which is kind of combination of <SimpleLayerOnOffButton> and <DropdownToolbarButton>.

cwang2016:
From @robyww:
When It is is on then clicking on select turns it off. If It is off then clicking will give you get a drop down (just like now) but with only the two options.

It sounds like to make a component which is a combination of SimpleLayerOnOffButton and DropdownToolbarbutton and includes parameter 'typeId' like what SimpleLayerOnOffButton does, then no need to use state in VisToolbarView. (yes, there should be state in VisToobarView).

cwang2016:
From @robyww:
When It is is on then clicking on select turns it off. If It is off then clicking will give you get a drop down (just like now) but with only the two options.

It sounds like to make a component which is a combination of SimpleLayerOnOffButton and DropdownToolbarbutton and includes parameter 'typeId' like what SimpleLayerOnOffButton does, then no need to use state in VisToolbarView. (yes, there should be state in VisToobarView).

PR #530, opened 2018-01-16
DM-12553: Allow automatic switching between FITS and HiPS

robyww:
- Switch between image and FITS automatically
 - Show image outline drawing layer when switching to HiPS from image
 - Added support for HiPS cube
 - Refactor ZoomUtil code to be ZoomUtil and ZoomTask
 - Refactor ImagePlotCntlr to break out more reducer code into HandlePlotAdmin
 - Refactor PlotImageTask to break out PlotHipsTask

_To Test_:
 - FITS/HiPS switching: bring at catalogs up, zoom out and the image will turning into a HiPS
 - Cube test: bring a hiPS in that is a cube, you should see cube switch controls in the context toolbar.  Cube switching might be slow depend on the CDS server.


src/firefly/js/visualize/PlotViewUtil.js
```diff
@@ -584,6 +585,44 @@ export function getCenterOfProjection(plot) {
...
+ * @param {number} [scrollY] optional scrollY, if not defined use plotView.scrollY
+ * @return {ImagePt} the center point
+ */
+export function findCurrentCenterPoint(plotView,scrollX,scrollY) {
+    const plot= primePlot(plotView);
+    if (!plot) return null;
+    const {viewDim}= plotView;
+
+    let cc;
+    if (!isUndefined(scrollX) && isUndefined(!scrollY)) {
```
cwang2016: not sure what "!scrollY" inside isUndefined means? do you mean !isUndefined(scrollY) ?

robyww: You right.  This is an old function that I just moved.  After doing some searching I found that the optional parameters were never passed in any of the usages.  That is why it was working.

robyww: You right. This is an old function that I just moved. After doing some searching I found that the optional parameters were never passed in any of the usages. That is why it was working.

cwang2016:
works well. 
find one issue and I can help fix that in another ticket if needed. 
 - do catalog search, and zoom the image to get HiPS. 
 - click on the 'Find HiPS Plot', there is no response -- a popup window showing HiPS list is expected to see.

cwang2016:
One more finding, not sure if this is an issue:
do catalog search on m31, cone size: 300 arcsec.
Zoom out the image to get HiPS image and image outline layer 1, 
zoom in the image back to get FITS image, and zoom out the image to get HiPS again and image outline layer 2, 
the outline shown on layer 1 looks closer to the catalog points than that on layer 2.

PR #531, opened 2018-01-17
IRSA-1009 gwt tpv version added

ejoliet:
I've pushed the branch where i've implemented the TPV projection in firefly: IRSA-1009-gwt-add-tpv

This is to be merged in GWT_RC, so in order to test it you can build IRSAViewer with this branch in firefly and 'gwt_rc' in IFE. I didn't touch IFE so no new branch.

I have attached to the ticket IRSA-1009 2 identical PTF images, one with TPV projection set and the other is the default SIP/TAN.
(See 'CTYPE1', 'CYTPE2' in fits header for more detail)

So far i can see the readout information and compare the data with both images and it looks ok.

The overlays also work meaning that Forward and Reverse transformation looks ok.

Somehow, rotation doesn't seems to work and look similar to the problem you solved for MSX (https://jira.lsstcorp.org/browse/DM-9581).

Feature that requires image rotation such as WCS, 'north-up', image rotation and 'restore to default settings' after rotation are all broken.

Please have a look. If you see a fix good, if not we can talk to @robyww for more ideas on what could be the cause. In the meantime I will migrate the code to JavaScript in dev branch where rotation is happening on the client side.

Let me know if you have problem to build and to test the branch.


lznakano:
I was able to reproduce the rotation issue and the cause is the same as previously I detected in FitsRead.java (DM-9851).  But it will take some thought and time to fix it since there are so many code changes in dev to achieve what I did.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
```diff
@@ -1544,6 +1545,8 @@ private static double getFluxFromFitsFile(File f,
                 if (clientFitsHeader ==null) {
                     throw new IOException("Can't read file, ClientFitsHeader is null");
                 }
+                //If test needed to debug,uncomment line below to test tpv for x,y IRSA-1009
+                //testTpv(f,ipt);
```
loitly: Alternatively, you can do:
`if (ServerContext.DEBUG_MODE) testTpv(f,ipt);`
That way, method testTpv doesn't need to be commented out.  In debugger, you can call it even when DEBUG_MODE is set to false.

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -69,7 +69,7 @@
     private final BasicHDU hdu;
     private float[] float1d;
     private Fits fits;
-    private ImageHeader imageHeader;
+    private static ImageHeader imageHeader;
```
loitly: Will it still be thread-safe after it's changed to static?

src/firefly/java/edu/caltech/ipac/visualize/plot/projection/TpvProjection.java
```diff
@@ -0,0 +1,321 @@
...
+ *
+ * In the paper above, intermediate coordinates (u,v) are renamed here fsamp,fline
+ * to follow same nomenclature as in @{@link GnomonicProjection}
+ * @author ejoliet
+ */
+public class TpvProjection {
+
+
+    static double dtr = Projection.dtr;
+    static double rtd = Projection.rtd;
```
loitly: `dtr` and `rtd` should be final or it can be replaced by Projection.{dtr,rtd} respectively.  It was misleading, thinking values could changed.

ejoliet: Good one.

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
ejoliet: Good point.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
ejoliet: Nice tip!

ejoliet:
I will take a look to @loitly feedback ASAP.

lznakano:
With the same changes in DM-9581, it only saved that issue about loading MSX image.  This branch introduced bugs that lead 'north-up', image rotation, 'restore to default settings' after rotation  broken.

src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
lznakano: The ImageHeader is an instance variable and should not changed to static.  Remove the static, north-up, return to default, rotation all work fine.  The only issue is the wcs match when the image is MSX.  This issue was fixed in the dev.  But it is an existing issue in gwt_rc.

PR #532, opened 2018-01-17
IRSA-1105:Herschel tool firefly component won't save png or reg

lznakano:
Fixed the footprintAry bug

There is no IFE changes.

To test it, 
1. Build irsa and Herschel in bacchus1 and set hydra sever to your localhost name 
2. Check out this branch and build firefly.


lznakano:
While trying to merge this branch, I found that the bug was fixed in the dev.  I talked to Cindy and she told me that when she was adding a new circle marker and found this bug.  Her fix and mine are similar.  So there is no need to merge now.  I tested the dev version and the saving PNG and REG for Herschel image works the same as my branch.  I can close this branch.

PR #533, opened 2018-01-19
Dm 12598 adding tpv projection

ejoliet:
TPV projection is missing from Firefly and is needed for ZTF images in particular.

I’ve tested and compare projections between images from PTF, one with SIP and one with TPV because PTF images contained both coefficients in FITS files.

Only problem noticed is the compass is not rendered correctly after rotation or north-up.

Please have a look.


ejoliet:
For testing, see files attached in IRSA-1009. Typically, you can take any PTF image and change SIP CTYPE to TPV.

src/firefly/js/visualize/projection/TpvProjection.js
```diff
@@ -0,0 +1,257 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+
+import {DtoR,RtoD, acosd,asind,atan2d,atand,sind,cosd,tand} from './ProjectionUtil.js';
+import {makeProjectionPt,makeImagePt} from '../Point.js';
+
```
robyww: File seems fine to me.  There is some stuff that is not used from ProjectionUtil.js
and some warning about let and const that could be fixed.

robyww:
Never mind about the clean up comments.  I have to clean something up anyway so I just did it.

ejoliet:
I've pushed java server part and updated files used for testing. Note that i've also added a branch in /firefly_test_data dm-12598-tpv if you need to test. This will need to be merged back to dev once this PR is merged.

stargaser:
Ah, I thought about this more. Indeed we _expect_ differences in the Gaia overlays between SIP and TPV, at a very small level. For SIP we provide a backwards polynomial that is an approximation to the pixels-to-sky one. The iterative backwards (sky-to-pixels) method for TPV should be better. So tiny differences are expected.

Good to go from my point of view!! Great work.

PR #534, opened 2018-01-20
DM-13241 Fix client table filtering issues on 'LIKE' and 'IN' conditions

cwang2016:
this development includes, 
- fix client table filtering by using 'like' and 'in' expression styles. 
- convert 'like' condition into regular expression match for client table filtering. 

test:
use firefly-dev.html
select tab 'Test Searches' and then tab 'HiPS'
uncheck 'Popular HiPS' to get more table content, this is a client table. 
try filtering examples: 
type: in (cube), in ('cube') or in ("cube")
         like cube, like 'cube' or like "cube"
title: like 2mass
url: like spit%.caltech or spit\%.caltech

### DIFF
--- a/src/firefly/js/tables/FilterInfo.js
+++ b/src/firefly/js/tables/FilterInfo.js
@@ -4,7 +4,8 @@
 
 import {getColumnIdx, getColumn, isNumericType, getTblById} from './TableUtil.js';
 import {Expression} from '../util/expr/Expression.js';
-import {isUndefined, get} from 'lodash';
+import {isUndefined, get, isArray} from 'lodash';
+import {showInfoPopup} from '../ui/PopupUtil.jsx';
 
 const cond_regex = new RegExp('(!=|>=|<=|<|>|=|like|in|is|is not)?\\s*(.+)');
 const cond_only_regex = new RegExp('^' + cond_regex.source, 'i');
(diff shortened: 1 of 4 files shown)


cwang2016:
Here are some updates on auto-correction function regarding table filtering:

the table filter condition string syntax is assumed to be compliant to that of SQL where clause, like
<op> filter_string   ; op means the operator, optional, and it is >, <, >=, <=, =, !=, like or in
                               ; if op is ‘in’, filter_string contains a list of strings separated by ‘,’ and it may or 
                               ;  may not be enclosed in braces, ‘()’, such as `in ‘a, b, c’,` or in `` (’a’, ‘b’, ‘c’).

The filter string will be validated and auto-corrected per the syntax of SQL ‘where’ clause.

The rules of auto-correction of filter string: 
case 1: no operator is specified or operator is 'like' (for both numeric and text column)
         - Treat it as operator ‘like’. 
         - if the string is enclosed in single quotes, no correction applies to that string.   
            Ex: like ‘abc’ , or ‘abc’ => like ‘abc’
         - if the  string is not enclosed in single quotes:
           1. replace %, _, or | with \%, \_ or \\. (escape the wildcard and escape character)
           2. enclose the string in '%' and then in single quotes.
           Ex. abc% => like '%abc\%%' 
 
case 2: operator is 'in',
         -  enclose each value in single quotes in case the single quotes are missing (for text column only)
         - enclose the entire string in braces, (), if the braces are missing (for both text and numeric columns)
            Ex. in a, b  => in ('a', 'b')     // for text column
            Ex. in a, b  => in (a, b)         // for numeric column

case 3: other operators, >, >=, <, <=, =, !=
         -  enclose the string in single quotes if single quotes are missing (for text Column only)
             Ex.  =abc => ='abc'  // for text column
             Ex.  =abc => =abc    // numeric column

Examples of filter string for either client or non-client table,
1.  if searching for a row with a numeric value like 12.34,
     the following filter string  can find that row successfully
      <like>   12.34
      <like>   ‘%12.34%’
      <like>   ‘12.34E0’
       in (12.34, 12)
       = 12.34
       = ’12.34’
       = 12.34E0     
       = ’12.34E0’
2.  if searching a row with a text string like (abc\\%_)
     the following filter string can find that row successfully
      <like> (
      <like> ‘%(%’
      <like> %
      <like> ‘%\%%’
      <like> _)
      <like> ‘%\_)%’
      <like> \\
      <like> ‘%\\\\%’
      in  (‘(abc\\%_)’)
      in  ((abc\\%_),  ‘aaa’)
      = (abc\\%_)
      = '(abc\\%_)'

After the auto-correction process, the filter condition from non-client table will be sent to server for SQL query, and the filter condition from client table will be converted to regular expression for table filtering purpose.

loitly:
Tested the latest changes on client and server tables, and it works great.  I really appreciate the thoroughness of your test-cases.

PR #535, opened 2018-01-23
DM-13309 Update PDAC Portal for WISE single-epoch source data

tgoldina:
https://jira.lsstcorp.org/browse/DM-13309


tgoldina:
Regarding 'Failed to load table. No data is found in the search range.' - this is a workaround for empty result set `{"result": {"table": {"metadata": {"elements": []}, "data": []}}}`, that gets returned by DAX, when no matching data are found. DAX is aware of this issue.

PR #541, opened 2018-01-31
DM-12864 Coverage and chart rendering on Tri-View when a catalog without position information is searched

tgoldina:
https://jira.lsstcorp.org/browse/DM-12864

If the table with no positional info is the first rendered in firefly, no coverage or chart area will show.
If other tables are added, the coverage and chart areas will not disappear, when changing the active table. This is intentional behavior to avoid confusing layout changes.

An example of a table with no position info WISE -> AllWISE Inventory Table 

Other bugs fixed:
- non-returning search
- removed chart hanging in the store
- empty result area with no drop down, when all tables are deleted
- missing layout updates

### DIFF
--- a/src/firefly/js/tables/TableInfoUtil.js
+++ b/src/firefly/js/tables/TableInfoUtil.js
@@ -4,7 +4,7 @@
 
 import {MetaConst} from '../data/MetaConst.js';
 import {CoordinateSys} from '../visualize/CoordSys.js';
-import {getColumnIdx} from './TableUtil.js';
+import {getColumn, getColumnIdx} from './TableUtil.js';
 
 
 export const DEF_CORNER_COLS= ['ra1;dec1', 'ra2;dec2', 'ra3;dec3', 'ra4;dec4'];
@@ -37,9 +37,10 @@ const makeCoordColAry= (cAry, table) => cAry.map( (c) => makeCoordCol(c,table)).
  *
  * @param def
  * @param {TableModel} table - used to look up column information
- * @return {CoordColsDescription}
+ * @return {CoordColsDescription|null}
  */
 function makeCoordCol(def, table) {
+    if (!def) return null;
     const s = def.split(';');
     if (s.length!== 3 && s.length!==2) return null;
     const s0Idx= getColumnIdx(table,s[0]);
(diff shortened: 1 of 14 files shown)


src/firefly/js/tables/TableInfoUtil.js
```diff
@@ -97,10 +98,12 @@ export function getCenterColumns(table) {
...
-    if (colList.includes('ra') && colList.includes('dec')) return 'ra;dec;EQ_J2000';
-    if (colList.includes('lon') && colList.includes('lat')) return 'lon;lat;EQ_J2000';
-    return 'ra;dec;EQ_J2000';
+    // ex. return 'ra;dec;EQ_J2000' or 'lon;lat;EQ_J2000'
+    const guess = (lon, lat) => {
+        const lonCol = getColumn(table, lon, true);
+        const latCol = getColumn(table, lat, true);
+        if (lonCol && latCol) return `${lonCol.name};${latCol.name};EQ_J2000`;
+    };
+    return guess('ra','dec') || guess('lon', 'lat');
```
cwang2016: I guess 'ra','dec', 'lon'&'lat' guess applied to most cases, right? or could we make the 'guess' configurable based on the mission? just a thought.

src/firefly/js/visualize/saga/CoverageWatcher.js
```diff
@@ -212,49 +212,56 @@ function removeCoverage(tbl_id, decimatedTables) {
...
-    if (table.totalRows>10000) {
-        const cenCol= options.getCenterColumns(table);
-        params.decimate=  serializeDecimateInfo(cenCol.lonCol, cenCol.latCol, 10000);
-        req = makeTableFunctionRequest(table.request, 'DecimateTable', 'coverage',
-            {decimate: serializeDecimateInfo(cenCol.lonCol, cenCol.latCol, 10000), pageSize: MAX_ROW});
-    }
+    try {
+        const table = getTblById(tbl_id);
+        if (!table) return null;
+        if (!options.hasCoverageData(options, table)) return null;
```
cwang2016: minor thing, instead of passing itself to the function defined in 'options', I guess you could access the property from 'options' by using 'this'.

src/firefly/js/tables/TableInfoUtil.js
tgoldina: I think, this is mainly for uploaded tables, as the mission-related catalogs have "CatalogCoordColumns" set in meta. I did not change this method, only made it to work with case independent column names, as was the original intention.

cwang2016:
some finding from the test, 
- for the case of WISE -> AllWISE Inventory Table, there is no column name marked on the axis of the chart, is that as design?

cwang2016:
one more finding, 
while doing search like wise -> neowise reactication database -> neowise-R known solar system object possible ..., the result shows an empty table at the bottome with 'No Data Found' words inside the table, and an empty coverage on top of empty table. no chart. Is this as design?

tgoldina:
For histogram, indeed, we do not provide the default labels in multi-trace chart. I think, we should.

And I'll add a check for no data before putting on coverage tab.

cwang2016:
review completes. work well as the spec.

PR #542, opened 2018-02-05
DM-13273: upgrade react to 16 and upgrade other packages

robyww:
Upgrade react to 16 and upgrade other packages  

  - removed `chartEngine` app option, and references to `highcharts`
  - removed `react-component-resizable`, all code now using `react-sizeme`
  - removed `react-addons-perf`, it is not longer supports for 16
  - fixed a couple of files that had warning issues
  - new version of `validator` now only takes strings, fixed any code using it
  - removed ImageViewerView because it was no longer necessary with SizeMe
  - a little code clean up if I was already working in the file.

_To Test:_

 - Delete your node_modules directory.
 - Choose area of the code you know about.
 - Make sure it behaves like before.
 - Pay attention to any UI subtleties: resizing, updating correctly, layout, etc


tgoldina:
Unable to open column selection table (looking glass icon) from the chart options:
```
Uncaught TypeError: Cannot assign to read only property 'minWidth' of object '#<Object>'
     at popupForm (ColSelectView.jsx:93)
    at showColSelectPopup (ColSelectView.jsx:76)
    at onClick (ColumnOrExpression.jsx:121)
```

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -11,7 +11,6 @@ import {getMenu, isAppReady, dispatchSetMenu, dispatchOnAppReady} from '../../co
 import {dispatchHideDropDown, getLayouInfo, SHOW_DROPDOWN} from '../../core/LayoutCntlr.js';
 import {lcManager, LC} from './LcManager.js';
 import {LcResult} from './LcResult.jsx';
-import {LcPeriod} from './LcPeriod.jsx';
```
cwang2016: LcPeriod.jsx was made on highchart, should this file be removed?

cwang2016:
review completes. I did some test on image search, catalog search, file upload, time series tool, and HiPS search except charts options,  running is smooth, don't find any problem.

src/firefly/js/templates/lightcurve/LcViewer.jsx
robyww: yes, I will remove it

robyww: @cwang2016 I don't see this at line 14 I think you were looking at the dev side of the diff.

src/firefly/js/ui/FieldGroup.jsx
```diff
@@ -30,9 +56,18 @@ export class FieldGroup extends PureComponent {
     }
 
     componentWillMount() {
-        const {groupKey:wrapperGroupKey}= this.context;
-        var {groupKey, reducerFunc, keepState, initValues, actionTypes}= this.props;
-        dispatchMountFieldGroup(groupKey, true, keepState, initValues, reducerFunc, actionTypes, wrapperGroupKey);
+        const {groupKey}= this.props;
+        const groupState= getFieldGroupState(groupKey);
+        if (groupState && groupState.mounted) {
```
tgoldina: Should we also check for !keepMounted here? Looks like we never use this property, but if it is true, we will never dispatch unmount, hence the callback will never execute.

src/firefly/js/charts/ui/ColSelectView.jsx
```diff
@@ -90,9 +91,9 @@ export function hideColSelectPopup() {
 
 function popupForm(tableModel, onColSelected,buttonText,popupId, minWidth) {
     const tblId = tableModel.tbl_id;
-    popupPanelResizableStyle.minWidth = Math.min(minWidth, 560);
+    const style= clone(popupPanelResizableStyle, {minWidth: Math.min(minWidth, 560)});
```
tgoldina: Frozen style objects could have interesting side effects. We did not see a warning in React 15, because it's always one component at a time that was using it.

src/firefly/js/templates/lightcurve/LcViewer.jsx
cwang2016: yes, the line is gone in this PR. Just asked to remove file LcPeriod.jsx.

PR #543, opened 2018-02-05
DM-13326 Fixed issues on circular selection per feedback from IRSA tester

cwang2016:
The development is made per ticket of 
https://jira.lsstcorp.org/browse/DM-13326

- remove the select area layer after the selected area is cropped or zoomed (to fit selected area)
- for circular selection, a circular region in red is shown on the top of cropped or zoomed (to fit selected area) image in order to indicate the selected (elliptical) area. This is done by using ImageOutline drawing layer, and ImageOutline drawing layer is updated to be generic to contain any Firefly defined drawing object. 
- the image is re-centered at the center of the selected area in case 'recenter image to selected area' is clicked after 'select an area'. 
- Fix the select area help text shown in overlay display menu. 

test:
do image search on Firefly, 
click 'select area' button and make 'Elliptical selection' and make a selection
try 'crop', 'zoom to fit selected area' or 'recenter image to select area' respectively. 
try 'restore to the defaults' to resume the image 
check the title and help line for select area in 'overlay display' menu. 

repeat the test on rotated image.

### DIFF
--- a/src/firefly/js/drawingLayers/ImageOutline.js
+++ b/src/firefly/js/drawingLayers/ImageOutline.js
@@ -100,14 +100,18 @@ function computeDrawData(drawLayer, plotId) {
     return computeDrawobj(drawLayer, plot);
 }
 
-/*
-drawObj could be either statically set by dl.drawObj or dynamically updated in attribute
- PlotAttribute.OUTLINEIMAGE_DRAWOBJ or PlotAttribute.OUTLINEIMAGE_BOUNDS
 
- title could be either statically set by dl.text or dynamically updated in attribute PlotAttribute.OUTLINEIMAGE_TITLE
- text location could either statically set by dl.textLoc or dynamically updated in attribute PlotAttribute.OUTLINEIMAGE_TITLELOC
-
- statically set is one time set.
+/**
+ * create drawing object and title object based on the statical or dynamical settings.
+ * drawObj could be either statically set by dl.drawObj or dynamically set by plot attribute,
+ * i.e. PlotAttribute.OUTLINEIMAGE_DRAWOBJ or PlotAttribute.OUTLINEIMAGE_BOUNDS
+ * title could be either statically set by dl.text or dynamically set by attribute
+ * PlotAttribute.OUTLINEIMAGE_TITLE
+ * text location could be either statically set by dl.textLoc or dynamically set by attribute
+ * PlotAttribute.OUTLINEIMAGE_TITLELOC
+ * @param dl
+ * @param plot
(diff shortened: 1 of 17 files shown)


cwang2016:
some issue is found regarding image crop on selected area. 
- should image crop be applied to all plots if no image is unlocked? the current code may crop only one image when no image is unlocked.

xiuqin:
We should make a separate ticket for crop all the images locked, which will affect rectangle selection too. 

The "Zoom to fit selected area" action should behave the same as " zoom the image to fit into visible space" except that the image is the selected area. This affects the rectangle selection zoom too. It is brought out because the ellipse shown in the image made it obvious. Please talk to me if you need more information.

cwang2016:
More development are added, 
.  fix the issues related to the following operations, 
   - 'select area' , then 'crop' or 'zoom to selected area' and then 'restore to defaults', 'restore to defaults' will  remove the overlay related to 'select area'.
  - 'unlock image' , then 'select area' and then 'crop' or 'zoom to selected area' or 'stats' or 'recenter' and then 'restore to defaults'. 
- fix the catalog selection issue reported in DM-13337
  https://jira.lsstcorp.org/browse/DM-13337
   After the catalog overlap is filtered or selected, the 'select area' layer is removed and only 'clear filter' or 'unselected' button is shown on context tool bar. The catalog overlay points can be selected freely.

cwang2016:
some updates: 
-  'zoom to fit selected area' put the selected image area to fit into the visible space of the viewer 
- add an outline layer on top of the 'zoom to fit selected area' or cropped image to indicate the selected area for either rectangular or elliptical selection.

robyww:
I am seeing a problem when the images are not grouped.  It may have been there forever. We should talk about how to fix it.

1. plot several images
1. unlock
1. click on select but don't select.
1. click on another image
1. click on select
1. select an area on that image.
1. you will see selection happen on both but only one will show selection options.

src/firefly/js/drawingLayers/ImageOutline.js
```diff
@@ -88,23 +97,77 @@ function computeDrawData(drawLayer, plotId) {
     const plot= primePlot(visRoot(),plotId);
     if (!plot) return [];
 
-    const wpAry= plot.attributes[PlotAttribute.LAST_IMAGE_BOUNDS];
-    const lastTitle= plot.attributes[PlotAttribute.LAST_IMAGE_TITLE];
+    return computeDrawobj(drawLayer, plot);
+}
+
+/*
```
robyww: this comment should be in javadoc style

PR #544, opened 2018-02-07
IRSA-1169: fixed FinderChart image jumping bug

robyww:
you need to checkout ife branch irsa-1169


robyww:
the icon was a cleanup meant for anther PR but it does not matter, they will both be merged in the next 2 days.

PR #545, opened 2018-02-07
DM-11457: Selected table rows should be kept after sorting

loitly:
https://jira.lsstcorp.org/browse/DM-11457

As described in the ticket, selection info should be kept after sorting and filtering.
In the case of filtering, when a selected row is removed due to a filter, it will be removed from selection info as well.  This is to ensure that the selection info is valid for the current view.

For example:
- select all, apply a filter.
- all rows are still selected.
- clear filters
- only the previously selected rows are selected.

Also in this ticket, `null` values are sorted to top for ascending and to the bottom on descending.

This ticket can be tested on any table.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
@@ -21,6 +21,7 @@ public class TableServerRequest extends ServerRequest implements Serializable, D
     public static final String TBL_FILE_PATH = "tblFilePath";       // this meta if exists contains source of the data
     public static final String TBL_FILE_TYPE = "tblFileType";       // this meta if exists contains storage type, ipac, h2, sqlite, etc
     public static final String RESULTSET_ID = "resultSetID";        // this meta if exists contains the ID of the resultset returned.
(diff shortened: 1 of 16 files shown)


loitly:
@ejoliet, I've just pushed a commit to fix the problem you saw in irsaviewer.  It was a bug in table support.  However, it was masked by plot.ly chart's code.  That's why it was not noticeable in firefly and only in irsaviewer.  Please check again.

tgoldina:
I was wrong, it does not work in irsaviewer, indeed. If I select from the table, I don't even see the selection in the chart or catalog overlay. Interestingly, if I select from the chart, selection stays through sort and filter.

Also, I am getting an SQL exception when sorting a table after 30 min have passed.

cwang2016:
Here are some cases I found, 
- catalog search, and do 'select all' on the table, 'select area' to cover partial points on the coverage and select 'filter' from context tool bar, some of the filtered points on the table become un-selected, is this as design? 
- the sorting on client table doesn't work correctly in case there is null or undefined data.

loitly:
> catalog search, and do 'select all' on the table, 'select area' to cover partial points on the coverage and select 'filter' from context tool bar, some of the filtered points on the table become un-selected, is this as design?

It's a bug.  I've fixed it.

PR #546, opened 2018-02-10
IRSA-1299: enable full size display

ejoliet:
This PR is allowing the user to display full size images. Please have a look and let me know if it doesn't break other image searches.


src/firefly/js/visualize/ui/ImageSearchPanelV2.jsx
```diff
@@ -306,12 +306,12 @@ function SelectArchive({groupKey,  imageMasterData, multiSelect}) {
                                      initialState={{
                                              unit: 'arcsec',
                                              labelWidth : 0,
-                                             nullAllowed: false,
+                                             nullAllowed: true,
```
robyww: I think we need to do more than `nullAllowed`. We probably should have a radio box that has 2 options `full image` or `cutout`

PR #547, opened 2018-02-12
DM-13265 Saga error handling

tgoldina:
https://jira.lsstcorp.org/browse/DM-13265

- Use spawn when using dispatchAddSaga. This prevents the unhandled exceptions in one saga to cancel all the siblings.
- Use fork with dispatchAddActionWatcher, because we are catching the unhandled exceptions. If an exception occurs in the callback, it won't cancel the saga.
- Added documentation for actionWatcherCallback
- Fixed a bug preventing remote charts on the test page 
- converted some dispatchAddSaga to dispatchAddActionWatcher, including coverage and image metadata watchers and firfely.utl.addActionListener API method. 

Test case:  http://localhost:8080/firefly/demo/ffapi-highlevel-test.html  Open console to view the output, "Start selection extensions", "Track mouth readout". Before, you would get an error when switching readout from one image to the other, which would cancel both listeners. Now, a single error in user defined readout listener does not prevent the following calls to succeed.

### DIFF
--- a/src/firefly/js/api/ApiBuild.js
+++ b/src/firefly/js/api/ApiBuild.js
@@ -16,7 +16,7 @@ import * as AppDataCntlr from '../core/AppDataCntlr.js';
 import * as DrawLayerCntlr from '../visualize/DrawLayerCntlr.js';
 import {ApiExpandedView} from './ApiExpandedView.jsx';
 import {dispatchAddCell, dispatchRemoveCell, dispatchEnableSpecialViewer} from '../core/LayoutCntlr.js';
-import {dispatchAddSaga} from '../core/MasterSaga.js';
+import {dispatchAddSaga, dispatchAddActionWatcher} from '../core/MasterSaga.js';
 
 // Parts of the lowlevel api
 import * as ApiUtil from './ApiUtil.js';
@@ -174,7 +174,7 @@ export function buildLowlevelAPI() {
         findDispatch(AppDataCntlr),
         findDispatch(DrawLayerCntlr),
         {dispatchAddCell, dispatchRemoveCell, dispatchEnableSpecialViewer},
-        {dispatchAddSaga}
+        {dispatchAddSaga, dispatchAddActionWatcher, }
     );
 
     const ui= {
--- a/src/firefly/js/api/ApiUtil.js
+++ b/src/firefly/js/api/ApiUtil.js
@@ -5,11 +5,10 @@
 
 import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
-import {get,isEmpty} from 'lodash';
-import {isElement,isString} from 'lodash';
+import {get,isArray,isElement,isEmpty,isString} from 'lodash';
 import {logErrorWithPrefix} from '../util/WebUtil.js';
-import {dispatchAddSaga} from '../core/MasterSaga.js';
-import {take,race,call} from 'redux-saga/effects';
+import {dispatchAddActionWatcher, dispatchCancelActionWatcher} from '../core/MasterSaga.js';
+import {uniqueID} from '../util/WebUtil';
 
 // NOTE 
 // NOTE 
@@ -94,10 +93,10 @@ export function unrenderDOM(div) {
 
 /**
  * @summary Add a listener to any action type
- * @param {string} actionType a string or and array of strings. Each string is an action constant from firefly.action.type
- * @param {function} callBack the call back will be call with two parameters: action object and state object
+ * @param {string} actionType a string or an array of strings. Each string is an action constant from firefly.action.type
+ * @param {function} callBack the call back will be called with three parameters: action object, state object, and extraData
  *                 If it returns true the listener will be removed.
- * @param {object} extraData, an object with data to send to the callback, can be anything
+ * @param {object} extraData an object with data to send to the callback, can be anything
  *
  * @return {function} a function that will remove the listener
  * @public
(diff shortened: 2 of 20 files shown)


src/firefly/js/api/ApiBuild.js
```diff
@@ -174,7 +174,7 @@ export function buildLowlevelAPI() {
         findDispatch(AppDataCntlr),
         findDispatch(DrawLayerCntlr),
         {dispatchAddCell, dispatchRemoveCell, dispatchEnableSpecialViewer},
-        {dispatchAddSaga}
+        {dispatchAddSaga, dispatchAddActionWatcher, }
```
loitly: This is a good idea.  Much easier for an API user to understand as oppose to generator.

src/firefly/js/api/ApiUtil.js
```diff
@@ -94,54 +93,33 @@ export function unrenderDOM(div) {
...
  *                 If it returns true the listener will be removed.
- * @param {object} extraData, an object with data to send to the callback, can be anything
+ * @param {object} extraData an object with data to send to the callback, can be anything
  *
  * @return {function} a function that will remove the listener
  * @public
  * @func addActionListener
  * @memberof firefly.util
  */
 export function addActionListener(actionType,callBack, extraData) {
```
loitly: I wonder if this is needed anymore.  This was created to shield API users from saga.  dispatchAddActionWatcher is doing the same.  Maybe it should be deprecated?

src/firefly/js/core/MasterSaga.js
```diff
@@ -23,13 +23,23 @@ export function dispatchAddSaga(saga, params={}) {
     flux.process({ type: ADD_SAGA, payload: { saga,params}});
 }
 
+/**
+ * Action watcher callback.
+ * @callback actionWatcherCallback
+ * @param {Action} action - the triggering action
+ * @param {function} cancelSelf - a function to cancel this watcher
+ * @param {object} params – params passed through dispatchAddActionWatcher or (if not undefined) the last returned value from the watcher callback
```
loitly: `params – params`
I think the '-' there is actually an em-dash or long-dash.  It does not show up in my 'US-ASCI' setting.
I am not sure if it'll show up in jsdoc correctly.

src/firefly/js/api/ApiUtil.js
tgoldina: We might want to keep it for backward compatibility and simplicity. We don't have to explain flux dispatch and getStore parameters. We did not add saga or watcher dispatchers to the javadoc, this is our internal implementation of action listeners, that we advertise in public API.

robyww: I am pretty sure IRSA is already using this.  We should keep it at least until they give us suggestions (Jeff is planning to) about how we could improve it.

src/firefly/js/api/ApiUtil.js
```diff
@@ -94,54 +93,33 @@ export function unrenderDOM(div) {
...
-    if (!actionType && !callBack) return;
-    var stopListening= false;
-    while (!stopListening) {
-        var raceWinner = yield race({
-            action: take(actionType),
-            cancel: call(() => cancelPromise)
-        });
-        stopListening= raceWinner.action ? callBack(raceWinner.action,getState(),extraData) : true;
+    if (!isArray(actionType) && !isString(actionType)) {
+        throw 'actionType must be a string or an array of strings';
```
robyww: Wow! This cleans up the code so much! Everything is much more simple. very nice.

src/firefly/js/core/AppDataCntlr.js
```diff
@@ -305,20 +305,21 @@ function updateAppData(appData) {
 }
 
 /**
- * This saga watches for app_data.isReady.  
- * When that happens, it will execute the given callback with the current state. 
- * @param {function} callback
+ * Action watcher callback. Watches for app_data.isReady.
+ * When app data are loaded, it will execute the callback passed with params with the current state.
+ * @callback actionWatcherCallback
```
robyww: I am glad you documented this.

src/firefly/js/core/MasterSaga.js
```diff
@@ -74,7 +84,10 @@ export function* masterSaga() {
             case ADD_SAGA: {
                 const {getState, dispatch}= flux.getRedux();
                 const {saga,params}= action.payload;
-                if (typeof saga === 'function') yield fork( saga, params, dispatch, getState);
+                // with fork every exception will bubble up from the child to the parent:
+                // an unhandled exception in one saga will cancel all sibling sagas
+                // with spawn, only the saga with the unhandled error will be cancelled
+                if (typeof saga === 'function') yield spawn(saga, params, dispatch, getState);
```
robyww: I know this was here before but I has actually bitten me.  We should `console.log` an error. Also why not us lodash `isFunction`?

Another question: I though  we had decided to use `fork` so we could catch it and log an error? Will spawn log and error?

src/firefly/js/core/MasterSaga.js
```diff
@@ -118,11 +131,20 @@ export function* masterSaga() {
...
+        // loop exits when saga is cancelled
         while (true) {
-            const action= yield take(actions);
+            const action = yield take(actions);
             try {
-                prevParams= callback(action, cancelSelf, prevParams, dispatch, getState) || params;
+                // the same callback can return modified parameters or undefined
+                // when undefined is returned use previous parameters
+                returnedParams = callback(action, cancelSelf, prevParams, dispatch, getState);
+                if (!isUndefined(returnedParams)) {
```
robyww: This is a good catch over the way I did it.

src/firefly/js/tables/TablesCntlr.js
```diff
@@ -600,6 +600,17 @@ function asyncFetch(request, hlRowIdx, dispatch, tbl_id) {
     });
 }
 
+/**
+ * @callback actionWatcherCallback
```
robyww: can we put his line on all our watchers?  Maybe you already did it.  This is a great way to identify them.  I have been concerned about quickly recognizing a watcher.

src/firefly/js/core/MasterSaga.js
tgoldina: We decided that we need to use fork for action watchers, because the parent is catching errors. 

In this case, the individual sagas are responsible for catching and logging errors. The unhandled errors are caught by middleware. I have tested that with spawn the errors are still printed to console, but do not cancel the siblings. The only saga that is cancelled is the one that caused the unhandled error.

PR #548, opened 2018-02-13
DM-13538: Add a Jenkins job to build and deploy Firefly's app to a local k8s cluster

loitly:
https://jira.lsstcorp.org/browse/DM-13538

I've also committed a couple of files directly into `ife-env:master`.

http://irsawebdev5:8100/view/active/
login -> click on `play` icon left of `k8s_firefly`
image

Enter a branch or commit hash
Enter a version label.  This will be used as part of the URL to access this deployment in k8s.  Defaults to BRANCH input.

After build successfully completes, goto:  https://irsawebdev9.ipac.caltech.edu/<VERSION_INPUT_FROM_ABOVE>/firefly/

You can do the same for `k8s_irsaviewer`.  Replace `firefly` with `irsaviewer` in the URL above.

NOTE: Only context.xml was modified for this PR.  The other 2 is code refactoring.

### DIFF
--- a/docker/base/context.xml
+++ b/docker/base/context.xml
@@ -16,7 +16,7 @@
   limitations under the License.
 -->
 <!-- The contents of this file will be loaded for each web application -->
-<Context>
+<Context sessionCookiePath="/">
 
     <!-- Default set of monitored resources -->
     <WatchedResource>WEB-INF/web.xml</WatchedResource>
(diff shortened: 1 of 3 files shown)


loitly:
Useful kubectl commands:

`kubectl get all`
- List all of the object in the `default` namespace.  This is where we deploy our applications.  So, it's a good view of what's out there.

`kubectl delete deploy,svc,ing <VERSION>`
- Completely remove this version from k8s added by Jenkins.  Jenkins added a deployment, a service and an ingress entry.  This command remove all of them with the given `VERSION` label.

`kubectl logs <pod name> [-n <name space>] [-c <container name if more than one>]`
- Display the logs  of the given pod

`kubectl get pods -o wide [-n <name space> | --all-namespaces]`
- Display pods in a more detailed view.

`kubectl exec -ti <pod-name> [-n <name space>] -- /bin/bash`
- Enter the container's bash shell

https://kubernetes.io/docs/reference/kubectl/cheatsheet/
Look at this for a quick reference/guide to the most commonly used commands.

PR #551, opened 2018-02-20
IRSA-1032: Error message is not user friendly anymore

loitly:
Also fixed IRSA-782: Update Error Message in Time Series Tool

https://caltech-ipac.atlassian.net/browse/IRSA-1032
https://caltech-ipac.atlassian.net/browse/IRSA-782

http://localhost:8080/firefly/ts.html, or
http://localhost:8080/irsaviewer/timeseries

Try different variation of tables and mission to make sure that error message matches the condition of your input.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
@@ -10,6 +10,7 @@ import edu.caltech.ipac.firefly.util.DataSetParser;
 import edu.caltech.ipac.util.DataGroup;
 import edu.caltech.ipac.util.DataType;
 import edu.caltech.ipac.util.StringUtils;
+import org.eclipse.jetty.util.StringUtil;
 import org.json.simple.JSONObject;
 
 import java.io.IOException;
@@ -47,7 +48,9 @@ public class JsonTableUtil {
         }
 
         JSONObject tableModel = new JSONObject();
-        if (meta.contains(TableServerRequest.TBL_ID)) {
+        if (!StringUtils.isEmpty(request.getTblId())) {
+            tableModel.put("tbl_id",  request.getTblId());
+        } else if (meta.contains(TableServerRequest.TBL_ID)) {
             tableModel.put("tbl_id",  meta.getAttribute(TableServerRequest.TBL_ID).getValue());
         }
         tableModel.put("title", page.getData().getTitle());
@@ -69,6 +72,10 @@ public class JsonTableUtil {
         if (request != null ){
             tableModel.put("request", toJsonTableRequest(request));
         }
+        if (!StringUtils.isEmpty(page.getErrorMsg())) {
+            tableModel.put("error", page.getErrorMsg());
+        }
+
         return tableModel;
     }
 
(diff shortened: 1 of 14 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
```diff
@@ -10,6 +10,7 @@
 import edu.caltech.ipac.util.DataGroup;
 import edu.caltech.ipac.util.DataType;
 import edu.caltech.ipac.util.StringUtils;
+import org.eclipse.jetty.util.StringUtil;
```
ejoliet: Are you sure you are not using the StringUtils from Firefly instead?

src/firefly/js/tables/ui/TablePanel.jsx
```diff
@@ -313,16 +313,15 @@ function Loading({showTitle, tbl_id, title, removable, bgStatus}) {
...
-        const origRequest = omit(cloneDeep(request), 'filters', 'sortInfo');
-        dispatchTableSearch(origRequest);
+        dispatchTableSearch(JSON.parse(prevReq));
     };
     return (
         <div className='TablePanel__error'>
+            <div><b>Table Load Error:</b></div>
             <div>{message}</div>
-            {canReset && <button type='button' className='button std' onClick={reloadTable}>Reload</button>}
+            {prevReq && <button type='button' className='button std' onClick={reloadTable}>Back</button>}
```
ejoliet: What is the 'back' button doing? Don't you mean 'OK' or some user reading confirmation?

src/firefly/js/templates/lightcurve/ptf/PTFMissionOptions.js
```diff
@@ -84,7 +84,7 @@ export function isValidPTFTable() {
     }
     else {
          const errorMsg = `The uploaded table is not valid. The PTF option requires pid.
-                        Please select one of the other upload options`;
+                        Please select the "Other" upload option for tables that do not meet these requirements.`;
```
ejoliet: I think we want to keep the other options as a text and not indicating the option as alternative. The previous text was good, no need to change unless i'm missing something else.
Same for WISE case below.

ejoliet: See IRSA-782 for the exact text.

src/firefly/java/edu/caltech/ipac/firefly/server/util/ipactable/JsonTableUtil.java
loitly: Good eyes.  It was inserted as by IntelliJ auto-complete feature.  Not used.  But, I'll remove it.

src/firefly/js/tables/ui/TablePanel.jsx
loitly: This error usually shows up when you've entered a bad filter or sort condition.  The 'Back' button is to put the table back to the state prior to this bad operation.  If you have a better word, let me know.  I have no problem changing it.

src/firefly/js/templates/lightcurve/ptf/PTFMissionOptions.js
loitly: Okay, now I am confused.  The exact text in the ticket is what I put here, which is not what they want.  What they want is what we have already.  I'll just revert the change.

PR #554, opened 2018-02-26
Dm 13276 webpack3

tgoldina:
https://jira.lsstcorp.org/browse/DM-13276

- Modified buildScript/webpack.config.js according to https://webpack.js.org/guides/migrating/
- Updated the related libraries (webpack 3.x babel-loader 7.x | babel 6.x)
- Updated Babel to use [preset env](https://babeljs.io/docs/plugins/preset-env/) for the two latest versions of Chrome, Safari, Firefox, and Edge. The result is that the ES6 code will be transformed into ES5 only if a feature is not supported by one of the browsers. This results in smaller file size.

    The size of firefly-dev.js
```
1. when transpiled to ES5, using default modules:
uncompressed: 9.68MB 
compressed: 3.5M

2. when ES6 is used as an output 
(only the features unsupported by the target browsers are transpiled):
uncompressed: 8.91MB
compressed: 3.2MB
```
 
- Using UglifyJS v3 (https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/master), which supports ES6 modules.

    Webpack is relying on UglifyJsPlugin to remove the unused code.

    If we keep transpiling code into es5 with Babel, we can use UglifyJS v2 (https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/version-0.4) included with webpack3. This version does not understand ES6 modules: which means that if our transpiled code contains ES6 modules, the unused modules will not be removed. 

- devMode task is now producing a single file with source map embedded in generated bundle. ('--cheap-module-eval-source-map'). The advantage is that the first compilation takes under 30s (vs. 40s with source-map option now), and the following compilations 2s (vs. 10s now) on my machine.
With webpack1, we had a separate file for the source map. This change is due to the change of behavior in webpack3: `-d` is a shortcut for `--debug --devtool --cheap-module-eval-source-map --output-pathinfo`

    Webpack official explanation of source map options:  https://webpack.js.org/configuration/devtool/
    A more comprehensive but perhaps outdated guide is [here](https://lorefnon.me/2016/12/03/on-webpack-and-source-map-integration.html#cheap-module-eval-source-map).

- Updated eslint and the related plugins

- Updated firefly_loader.js to use anonymous self-calling function.


buildScript/webpack.config.js
```diff
@@ -106,54 +107,88 @@ export default function makeWebpackConfig(config) {
...
+            query: {
+                // later presets run before earlier for each AST node
+                // use 'es2015', {modules: false}] for es5 with es6 modules
+                presets: [
+                    ['env', {
+                        targets: {
+                            browsers: ['safari >= 10', 'chrome >= 63', 'firefox >= 57', 'edge >= 15']
+                        },
+                        debug: true,
+                        modules: PROD ? false : undefined}], // AMD modules produce smaller uncompressed size
```
robyww: @tgoldina and I talked about always using always using ES6 modules even in development.  That way we don't have any unexpected results.

buildScript/gwt.gincl
```diff
@@ -167,7 +167,7 @@ task devMode (dependsOn: loadConfig) {
   outputs.upToDateWhen { false }
 
   doLast {
-    def res = project.ext.NODE 'yarn', 'run', 'dev', "--", "--baseWarName=${webapp.baseWarName}", "--app-name=${project.ext["app-name"]}"
+    def res = project.ext.NODE 'yarn', 'run', 'dev', "--", "--env.baseWarName=${webapp.baseWarName}", "--env.app-name=${project.ext["app-name"]}"
```
robyww: I agree with @tgoldina that this whole file should be cleaned up.  She is planning to do that.

buildScript/webpack.config.js
```diff
@@ -106,54 +107,88 @@ export default function makeWebpackConfig(config) {
...
+            loader: 'babel-loader',
+            query: {
+                // later presets run before earlier for each AST node
+                // use 'es2015', {modules: false}] for es5 with es6 modules
+                presets: [
+                    ['env', {
+                        targets: {
+                            browsers: ['safari >= 10', 'chrome >= 63', 'firefox >= 57', 'edge >= 15']
+                        },
+                        debug: true,
```
robyww: We should keep debug to be `true`, at least for `!PROD`. I like seeing the supported browsers and which transforms are being chosen.

buildScript/webpack.config.js
```diff
@@ -106,54 +107,88 @@ export default function makeWebpackConfig(config) {
...
-            new webpack.optimize.OccurrenceOrderPlugin(),
-            new webpack.optimize.DedupePlugin(),
-            new webpack.optimize.UglifyJsPlugin({
-                compress : {
-                    warnings  : false,
-                    unused    : true,
-                    dead_code : true
+            // if using the latest uglifyjs-webpack-plugin, based on UglifyJSv3, understanding es6 modules
+            // this uglifier also minimizes the code, no LoaderOptionsPlugin is needed
+            new UglifyJsPlugin({
```
robyww: i agree that using uglify 3 is the right decision when  it will be standard in webpack 4

package.json
```diff
@@ -49,15 +50,15 @@
...
     "chai": "^3.5.0",
     "css-loader": "^0.19.0",
-    "eslint": "^3.4.0",
-    "eslint-loader": "^1.5.0",
-    "eslint-plugin-jsdoc": "^2.3.1",
-    "eslint-plugin-jsx-a11y": "^0.6.2",
-    "eslint-plugin-react": "^6.2.0",
-    "extract-text-webpack-plugin": "^0.8.0",
+    "eslint": "^4.3.0",
+    "eslint-loader": "^1.9.0",
```
robyww: I am happy you did these updates.

.eslintrc
```diff
@@ -65,6 +65,7 @@
     "react/jsx-no-undef": 1,
     "react/jsx-uses-react": 1,
     "react/jsx-uses-vars": 1,
+    "react/jsx-wrap-multilines": 1,
```
robyww: Is this one new? what does it do?

buildScript/loadScript.js
```diff
@@ -2,37 +2,25 @@
...
-    scriptTag.type = 'text/javascript';
+        scriptTag.type = 'text/javascript';
 
-    const url = getScriptURL(loader) + script;
-    if (scriptTag.readyState){  //IE
-        scriptTag.onreadystatechange = function(){
-            if (['loaded', 'complete'].includes(scriptTag.readyState)) {
-                scriptTag.onreadystatechange = null;
-                callback && callback();
+        const getScriptURL = (loader) => {
```
robyww: I saw this originally bug forgot to comment.  I don't think you should do this a a arrow function.  Since it is not precompiled it should be in ES5.
```
var getScriptURL =  function(loader) {
````

File: buildScript/loadScript.js [hunk 8]
```diff
@@ -2,37 +2,25 @@
...
-    } else {  //Others
-        scriptTag.onload = function(){
-            callback && callback();
-        };
-    }
 
-    scriptTag.src = url;
-    document.getElementsByTagName('head')[0].appendChild(scriptTag);
-}
+        const url = getScriptURL(loader) + script;
```
robyww: _more comments_ 
 - `const` to `var`
 - the `loader` is already available in the closure.

buildScript/loadScript.js
```diff
@@ -2,37 +2,25 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-
-function loadScript(loader, script, callback){
+(function(loader, script){
     var scriptTag = document.createElement('script');
-    scriptTag.type = 'text/javascript';
+        scriptTag.type = 'text/javascript';
```
robyww: _one more issue-_  all the code is indented more than line 6.

buildScript/loadScript.js
```diff
@@ -2,37 +2,25 @@
...
-        };
-    }
 
-    scriptTag.src = url;
-    document.getElementsByTagName('head')[0].appendChild(scriptTag);
-}
+        const url = getScriptURL(loader) + script;
+        scriptTag.src = url;
+        document.getElementsByTagName('head')[0].appendChild(scriptTag);
+})(/*params*/);
```
robyww: often for substitution strings we use all caps.
```
})(/*PARAMS_HERE*/);
```
so it is real obvious. If you do this remember to change `webpack.config.js` as well

wmiipac:
Follow up, I did test on firefox: 58, Chrome: 63, all looks good.
On Safari :9.1.3, the apps (irsaviewer, timeserise, finderchart) would not load, they all came up with a blank page. Don't know if this should be a concern.

tgoldina:
Since I have modified babel presets, we'll need to change webpack.config.js for irsa and irsaviewer, and suit. `require('babel-core/register')({presets: ['es2015', 'react', 'stage-2']});` should become `require('babel-core/register')({presets: ['env', 'react', 'stage-3']})`

I'd like Loi to take a look at it first.

PR #556, opened 2018-03-02
DM-13698:  Small build tweeks, removed fftools dir

robyww:
_note- _ there were very view files changed, the high number is bacause I removed fftools

  - Added webpack-visualizer-plugin
  - now in dev mode, http://localhost:8080/firefly/package-stats.html is created
  - added useBuiltIns : true to webpack.config.js
  - useBuiltIns allows us to have the same env effect with polyfills
  - a little code clean up to removed unused modules
  - resulting build is about 8%-10% little smaller after zipping
  - removed fftools directory, no longer used

_to test package-stats.html_: `gradle -Penv=dev clean :fire:bAD`, then load the above URL


buildScript/webpack.config.js
```diff
@@ -150,13 +155,18 @@ export default function makeWebpackConfig(config) {
...
-                        modules: false}], // preserve application module style - in our case es6 modules
-                    'react', 'stage-3'],
+                    ['env',
+                        {
+                            targets: {
+                                browsers: ['safari >= 9', 'chrome >= 62', 'firefox >= 56', 'edge >= 14']
+                            },
+                            debug: !PROD,
+                            modules: false,  // preserve application module style - in our case es6 modules
+                            useBuiltIns : true
```
tgoldina: The [docs](https://babeljs.io/docs/plugins/preset-env/#options) says
`This option enables a new plugin that replaces the statement import "babel-polyfill" or require("babel-polyfill") with individual requires for babel-polyfill based on environment.` 

We still have `import 'babel-polyfill'` in Firefly.js

tgoldina:
I guess, I have misunderstood the docs. We still need `import 'babel-polyfill'`.

robyww:
I think all the `useBuiltIns` does is to convert the `import 'babel-polyfill'` line to a set of lines.  One `import` per polyfill.

So yes, I think you need that line.

PR #560, opened 2018-03-14
IRSA-1581: 2mass image request fix when survey was implicitly value asky

ejoliet:
I've updated the function that make the 2mass request image from the client side. This was a change needed after IRSA-1447.
Please, check that 2mass image are back in Finder Chart (same branch name in IFE) and when coverage 2mass image is chosen.

I saw already a problem that i couldn't figure it out in Safari. Chrome is fine.

### DIFF
--- a/src/firefly/js/visualize/WebPlotRequest.js
+++ b/src/firefly/js/visualize/WebPlotRequest.js
@@ -350,13 +350,15 @@ export class WebPlotRequest extends ServerRequest {
     /**
      *
      * @param wp
-     * @param {string} survey  must be one of 'j','h','k'
+     * @param {string} survey  must be one of 'asky', 'askyw', 'sx', 'sxw', 'cal'
+     * @param {string} band  must be one of 'j','h','k'
      * @param sizeInDeg less then .138 degrees (500 arcsec)
      * @return {WebPlotRequest}
      */
-    static make2MASSRequest(wp, survey, sizeInDeg) {
+    static make2MASSRequest(wp, survey, band, sizeInDeg) {
         const req= this.makePlotServiceReq(ServiceType.TWOMASS, wp, survey, sizeInDeg);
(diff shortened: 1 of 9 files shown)


src/firefly/js/visualize/WebPlotRequest.js
```diff
@@ -350,13 +350,15 @@ export class WebPlotRequest extends ServerRequest {
...
      *
      * @param wp
-     * @param {string} survey  must be one of 'j','h','k'
+     * @param {string} survey  must be one of 'asky', 'askyw', 'sx', 'sxw', 'cal'
+     * @param {string} band  must be one of 'j','h','k'
      * @param sizeInDeg less then .138 degrees (500 arcsec)
      * @return {WebPlotRequest}
      */
-    static make2MASSRequest(wp, survey, sizeInDeg) {
+    static make2MASSRequest(wp, survey, band, sizeInDeg) {
```
robyww: There is a equivalent method in WebPlotRequest.**java**  that needs to be changed as well.  I think it is used by the FinderChart API.

src/firefly/js/visualize/ui/ImageSelectPanelResult.js
```diff
@@ -122,7 +122,7 @@ function imagePlotOnSurvey(crtCatalogId, request) {
             break;
 
         case TWOMASS:
-            wpr = WebPlotRequest.make2MASSRequest(wp, survey, sizeInDeg);
+            wpr = WebPlotRequest.make2MASSRequest(wp, survey, band, sizeInDeg);
```
robyww: This is fine, just worth noting,  I think `ImageSelectPanelResult.js` is deprecated.

src/firefly/js/visualize/WebPlotRequest.js
ejoliet: Thanks. I will double check in IFE but i think i refactor also the code there.

PR #561, opened 2018-03-15
DM-13038 Column expressions for column names that are not alphanumeric

tgoldina:
https://jira.lsstcorp.org/browse/DM-13038

This ticket takes care of the client-side validation for column expressions with columns that are not alphanumeric. Non-alphanumeric column names must be quoted. To test use a NED table, and try to plot a scatter, using column expressions.

https://irsawebdev9.ipac.caltech.edu/dm-13038/firefly

Database supports quoted column expressions, however heatmap and histogram processors do not at the moment. (DM-13036 will address server side.)

Also made a db function for lg (log10) instead of substituting log10 for lg.

### DIFF
--- a/src/firefly/js/charts/dataTypes/FireflyGenericData.js
+++ b/src/firefly/js/charts/dataTypes/FireflyGenericData.js
@@ -1,7 +1,7 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import {get, isArray, uniqueId} from 'lodash';
+import {get, isArray, truncate, uniqueId} from 'lodash';
 import {COL_TYPE, getTblById, getColumns, getColumn, doFetchTable} from '../../tables/TableUtil.js';
 import {cloneRequest, MAX_ROW} from '../../tables/TableRequestUtil.js';
 import {dispatchChartUpdate, dispatchError, getChartData} from '../ChartsCntlr.js';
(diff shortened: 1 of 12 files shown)


src/firefly/js/charts/dataTypes/FireflyGenericData.js
```diff
@@ -101,10 +101,11 @@ function addScatterChanges({changes, chartId, traceNum, tablesource, tableModel}
...
 
-    const xLabel = get(mappings, 'x'); // default axis label for the first trace
-    const yLabel = get(mappings, 'y'); // default axis label for the first trace
-    const xTipLabel = xLabel.length > 20 ? 'x' : xLabel;
-    const yTipLabel = yLabel.length > 20 ? 'y' : yLabel;
+    // default axes labels for the first trace (remove surrounding quotes, if any)
+    const xLabel = get(mappings, 'x').replace(/^"(.+)"$/, '$1');
+    const yLabel = get(mappings, 'y').replace(/^"(.+)"$/, '$1');
+    const xTipLabel = xLabel.length > 20 ? xLabel.substring(0,18)+'...' : xLabel;
+    const yTipLabel = yLabel.length > 20 ? yLabel.substring(0,18)+'...' : yLabel;
```
loitly: lodash's truncate works the same.

src/firefly/js/util/expr/Variable.js
```diff
@@ -6,20 +6,23 @@
...
-     let result = variables.get(name);
-     if (!result) {
-         result = new Variable(name);
-         variables.set(name, result);
-     }
-     return result;
- }
+    if (!/^[A-Za-z\d_]+$/.test(name) && !name.startsWith('"')) {
+        name = `"${name}"`;
+    }
```
loitly: I think you should create a function `quoteNonAlphaNum()`.. or something like it in ChartUtil to put his logic in there.  It's not that the logic is hard, it's because it'll be easier to use `find usage` when you need to search for places where you had to do this.

PR #558, opened 2018-03-08
DM-13624 Create HiPS survey loading framework 

cwang2016:
The development includes, 
- Implement hips sources framework with search processor to search and combine available HiPS data from various sources, including CDS MocServer, IRSA, LSST, and others. 
- In current release, only CDS and IRSA sources provide the list of HiPS. For CDS, Firefly gets list of HiPS by sending query as shown in http://aladin.u-strasbg.fr/hips/list. For IRSA, Firefly prepares a file, irsa-hips-master-table.csv, to define the available HiPS.  
- Based on the implemented framework, a HiPS master list search query is sent to the server and returned with a HiPS master table containing the information of all available HiPS from various sources. 
- In this release, a filter which forms a table containing popular HiPS by modifying HiPS master table  is added. In near future(https://jira.lsstcorp.org/browse/DM-13172), filter for making popular HiPS list will be made to be configurable by the application. 

test:
- do catalog search, and click ‘Find HiPS Plot’ to get HiPS master table or popular table. 
- do image search by providing URL or clicking one HiPS from the HiPS master or popular table. The HiPS table highlight which reflects the image of the active plot is updated dynamically as the active plot is changed among multiple images.

PR #562, opened 2018-03-16
DM-13625: Do final work with coverage and clean up of HiPS UI issues

robyww:
- coverage from API
 - coverage update issues
 - coverage works with image, hips, allsky image
 - experimenting with title
 - fixed some locking issues
 - added hips to image match button - image do simple match to HiPS
 - clean up HealpixIndex.js file
 - app options allows for setting the coverage hips and image
 - dialogs can take root element to determine z-index, this caused some refactoring

### DIFF
--- a/src/firefly/html/demo/ffapi-coverage-test.html
+++ b/src/firefly/html/demo/ffapi-coverage-test.html
@@ -81,7 +81,8 @@
                         useAllSkyFitsForPlus180 : true,
                         // fovDegFallOver: .02,
                         fovDegFallOver: .5, // 720 arcsec
-                        hipsSourceURL : 'http://alasky.u-strasbg.fr/DSS/DSSColor', // url
+                         hipsSourceURL : 'AllWISE_COLOR',
+                        // hipsSourceURL : 'http://alasky.u-strasbg.fr/DSS/DSSColor', // url
                         // hipsSourceURL : 'http://alasky.u-strasbg.fr/Planets/CDS_P_Earth_BlueMarble/', // url
                         imageSourceParams: {
                             Service : 'WISE',
--- a/src/firefly/html/demo/ffapi-coverage-test.html
+++ b/src/firefly/html/demo/ffapi-coverage-test.html
@@ -71,13 +71,23 @@
 
             firefly.showCoverage('coverageHere',
                     {
-                        gRIdOn:true,
+                        gridOn:true,
                         symbol: {aTable: 'X', bTable: 'arrow'},
                         symbolSize: {aTable: 3, bTable: 12},
                         color: {aTable: 'blue', bTable: 'yellow'},
-//                        overlayPosition : '149.9;68.8;EQ_J2000',
-                        OVErlayPosition : '149.08;68.739;EQ_J2000',
-                        useHiPS: true
+                        overlayPosition : '149.08;68.739;EQ_J2000',
+
+                        useHiPS: true,
+                        useAllSkyFitsForPlus180 : true,
+                        // fovDegFallOver: .02,
+                        fovDegFallOver: .5, // 720 arcsec
+                        hipsSourceURL : 'http://alasky.u-strasbg.fr/DSS/DSSColor', // url
+                        // hipsSourceURL : 'http://alasky.u-strasbg.fr/Planets/CDS_P_Earth_BlueMarble/', // url
+                        imageSourceParams: {
+                            Service : 'WISE',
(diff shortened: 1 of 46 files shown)


cwang2016:
I don't see other issue except the above mentioned problem.

robyww:
What problem?  I think your comment may have gotten lost.

src/firefly/html/demo/ffapi-coverage-test.html
```diff
@@ -71,13 +71,23 @@
...
+                        overlayPosition : '149.08;68.739;EQ_J2000',
+
+                        useHiPS: true,
+                        useAllSkyFitsForPlus180 : true,
+                        // fovDegFallOver: .02,
+                        fovDegFallOver: .5, // 720 arcsec
+                        hipsSourceURL : 'http://alasky.u-strasbg.fr/DSS/DSSColor', // url
+                        // hipsSourceURL : 'http://alasky.u-strasbg.fr/Planets/CDS_P_Earth_BlueMarble/', // url
+                        imageSourceParams: {
+                            Service : 'WISE',
```
ejoliet: is this json attribute (`imageSourceParams`) defining the fits image dataset when zooming in/out on coverage?

robyww: > and i'm not sure if it was working previously but footprint overlay doesn't work on hips maps, so either we need to raise a ticket ot fix it or disable the feature when hips map is active.

That is being worked on in
https://github.com/Caltech-IPAC/firefly/pull/558

ejoliet:
Another 2 issues: 
- the table filter doesn't update the catalog overlays on api mode.
- locking hips map recenter them, we need to make sure that we don;t miss another use case which could be looking at one hips on one field, and another field on other hips but be able to apply same displacement when moving around without been the center of the image in sync. For example, that's how is currently the zoom feature. if images have different zoom and one lock them, their zoom stayed , their are no reset to be equal.

src/firefly/html/demo/ffapi-coverage-test.html
ejoliet: Great!

robyww:
> and i'm not sure if it was working previously but footprint overlay doesn't work on hips maps, so either we need to raise a ticket ot fix it or disable the feature when hips map is active.

That is being worked on in
https://github.com/Caltech-IPAC/firefly/pull/558

src/firefly/html/demo/ffapi-coverage-test.html
robyww: Image is for zooming in and hips is for zooming out.  When come first is based on how big the search area is and `fovDegFallOver`

ejoliet: I guess that’s where we will need in the future some logic to decide on which image to show related to hips map - see IRSA-1586. Currently is fixed to WISE, correct?

robyww: From Gator's perspective, I think it should provide the parameters.  When loading catalogs directly in Firefly, I guess we could have some logic to guess.  However, that is not what we have done in the pass.  It would be very easy to do some basic guessing. 
>  wise on wise, 2 mass on 2 mass, x on x, anything else on wise.
But as I said firefly is not done that before.

src/firefly/js/visualize/WebPlot.js
```diff
@@ -260,22 +261,22 @@ export const PlotAttribute= {
...
  * @param asOverlay
  * @param imageCoordSys
  * @return {WebPlot}
  */
-function makePlotTemplate(plotId, type, asOverlay, imageCoordSys) {
+function makePlotTemplate(plotId, plotType, asOverlay, imageCoordSys) {
     return {
         plotId,
-        type,
+        plotType,
```
cwang2016: in file 'CsysConverter.js' constructor, there is line like 'this.type = plot.type'. Should the same change be applied to that line? I found that marker rendering is invisible and the problem is because there is no 'plotType' key in CsysConverter object.

robyww: yes! thank you for catching that.  The was a very hard refactor.

robyww:
> the table filter doesn't update the catalog overlays on api mode.

It does for me with.  
http://localhost:8080/firefly/demo/ffapi-coverage-test.html

I suspect you might have be testing in with ffapi-highlevel-test.html, not all the tables are connected to coverage one that one. ffapi-coverage-test.html is closer to what gator will use.

PR #564, opened 2018-03-20
IRSA-1596: Saved table contains hidden columns, i.e.  ROW_IDX and ROW_NUM

loitly:
https://caltech-ipac.atlassian.net/browse/IRSA-1596
Also:
- remove DataSetParser as part of ongoing code cleanup.
- fix IRSA-1601

To test:
- build firefly 
- catalog search
- save table by clicking on the 'diskette' icon

You should not see hidden columns like ROW_IDX and ROW_NUM anymore.


loitly:
> I think the refactor need to be also done in IFE, i couldn't compile irsaviewer.

The errors are from firefly.  There are no changes in `ife` for this PR.
Sorry, I fail to redo my test after rebasing from `dev`.  This latest commit should fixed it.  Please test again.

ejoliet:
I've raised a new ticket to consider the improvement of the table header which i think has lost some content that was useful and used by users in the past before migration and need to be reviewed and changed for future release (see IRSA-1602).

loitly:
I don't think we ever had a `WHERE` header in our output table.  We used to have a commend entry `created from statement: `.  It is only a comment and was not meant as a way to trace back to the original data.  
It can get pretty complicated if we want to include meta information to record the operations done to the original table.

PR #565, opened 2018-03-22
DM-13797 remove HiPS grid on and support to IRSA HiPS list

cwang2016:
The development includes:
- Remove HiPS grid icon from the toolbar, Add HiPS grid layer with the checkbox turned off (the layer is hidden by default) when a new HiPS is added. 
- Add support for calling IRSA hips list,
    . Update App config file by adding information related to IRSA HiPS map service 
    . Each item in the HiPS list has unique value for HiPS property creator_did. 
    . Extract title value from property creator_id for IRSA HiPS list

Note: please checkout ife with the same branch name 'DM-13797-rmHiPSGridIconIrsaHiPSList'. 
 
test:
do HiPS image search. 
For this PR, HiPS map list is formed by sending request to IRSA HiPS map service, 
https://irsadev.ipac.caltech.edu/data/hips/list

### DIFF
--- a/config/app.config
+++ b/config/app.config
@@ -48,7 +48,7 @@ irsa.gator.service.periodogram.url = "https://irsa.ipac.caltech.edu/cgi-bin/peri
 // IRSA hips map access: from "file" or from "url" && hips url
 irsa.hips.list.source="url"
 irsa.hips.masterUrl="https://irsadev.ipac.caltech.edu/data/hips/list"
-irsa.hips.url.base="irsa.ipac.caltech.edu"
+
 /*  ----------------------------------------------------------------------- */
 
 wise.ibe.host       = "irsa.ipac.caltech.edu/ibe"
@@ -66,19 +66,16 @@ environments {
         ehcache.multicast.port = 4015
         visualize.fits.Security= false
         ehcache.multicast.ttl = 0
-        irsa.hips.url.base="irsadev.ipac.caltech.edu"
     }
     dev {
         BuildType = "Development"
         ehcache.multicast.port = "5015"
         lsst.dataAccess.uri = "http://localhost:8661/db/v0/query?"
         lsst.dataAccess.db = "smm_bremerton"
-        irsa.hips.url.base="irsadev.ipac.caltech.edu"
     }
     test {
         BuildType = "Beta"
         ehcache.multicast.port = "6015"
-        irsa.hips.url.base="irsadev.ipac.caltech.edu"
     }
     ops_int {
         ehcache.multicast.port = "7515"
--- a/config/app.config
+++ b/config/app.config
@@ -45,6 +45,10 @@ irsa.gator.hostname     = "https://irsa.ipac.caltech.edu"
 most.host = "https://irsasearchops.ipac.caltech.edu/cgi-bin/MOST/nph-most"
 // IRSA Periodogram API
 irsa.gator.service.periodogram.url = "https://irsa.ipac.caltech.edu/cgi-bin/periodogram/nph-periodogram_api"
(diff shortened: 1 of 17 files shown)


config/app.config
```diff
@@ -45,6 +45,10 @@ irsa.gator.hostname     = "https://irsa.ipac.caltech.edu"
 most.host = "https://irsasearchops.ipac.caltech.edu/cgi-bin/MOST/nph-most"
 // IRSA Periodogram API
 irsa.gator.service.periodogram.url = "https://irsa.ipac.caltech.edu/cgi-bin/periodogram/nph-periodogram_api"
+// IRSA hips map access: from "file" or from "url" && hips url
+irsa.hips.list.source="url"
+irsa.hips.masterUrl="https://irsadev.ipac.caltech.edu/data/hips/list"
+irsa.hips.url.base="irsa.ipac.caltech.edu"
```
loitly: I noticed protocol is not included.  What is the logic for choosing the protocol?
I find it's more straight forward to use either relative or full url, including protocol and ports as the base

config/app.config
```diff
@@ -45,6 +45,10 @@ irsa.gator.hostname     = "https://irsa.ipac.caltech.edu"
 most.host = "https://irsasearchops.ipac.caltech.edu/cgi-bin/MOST/nph-most"
 // IRSA Periodogram API
 irsa.gator.service.periodogram.url = "https://irsa.ipac.caltech.edu/cgi-bin/periodogram/nph-periodogram_api"
+// IRSA hips map access: from "file" or from "url" && hips url
+irsa.hips.list.source="url"
+irsa.hips.masterUrl="https://irsadev.ipac.caltech.edu/data/hips/list"
```
loitly: Properties should defaults to OPS.  Also, when we open-source firefly, we had to remove all reference of internal machine names.  I think that's still a policy.  @robyww @xiuqin ?

src/firefly/js/visualize/task/PlotHipsTask.js
```diff
@@ -188,6 +191,13 @@ export function makePlotHiPSAction(rawAction) {
     };
 }
 
+function addHiPSGridLayer() {
```
robyww: This confused me a little until I understood what was going on. You might want to call this function `createHiPSGridLayer` since you actually add it somewhere else.

config/app.config
robyww: Yes, that is what we want.

config/app.config
robyww: Maybe there is another way to do this instead of passing build properties. The three of us should talk about that.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/hips/IrsaHiPSListSource.java
```diff
@@ -47,8 +80,117 @@
...
+        if (listFile.getResponseCode() >= 400) {
+            String err = LSSTQuery.getErrorMessageFromFile(file);
+            throw new DataAccessException("[HiPS_CDS] " + (err == null ? listFile.getResponseCodeMsg() : err));
+        }
+
+        return getListDataFrom(file, paramsMap, source);
+
+    }
+
+    private List<HiPSMasterListEntry> getListDataFrom(File f, Map<String, String> keyMap, String source)
```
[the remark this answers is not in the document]
cwang2016: yes, they are not exactly the same between cds and irsa, but can be made to be one routine used for both.

config/app.config
cwang2016: remove IRSA HiPS map url base from the config files per discussion.

PR #572, opened 2018-03-31
IRSA-1632 workspace upload option broken

ejoliet:
I've fixed the setState problem where we used it and in the FilePicker.
Please, check the features where workspace is involved. This is broken currently in dev and this PR fixed all those problem related.

I'm building IRSAViewer with this fix will appear soon for testing in https://caltech-ipac.slack.com/messages/C9SDXRAHH


cwang2016:
It works well.

PR #573, opened 2018-04-01
DM-7991: handling upper limits

tgoldina:
https://irsawebdev9.ipac.caltech.edu/dm-7991/firefly/

https://jira.lsstcorp.org/browse/DM-7991

To test, use firefly/demo/ffapi-sed.html 

(Note the use of chart options there: `{singleTraceUI: true, upperLimitUI: true}` - the default values for both options are false.)

- Upper limit column or expression is stored in fireflyData[traceNum].yMax, using the same notation as for other mapped columns.
- Upper Limits are represented by annotations (arrows pointing down).
- Since Plotly stores annotations in the layout, trace annotations needed to be implemented.

Per trace annotations are stored in fireflyData, and are constructed when the data arrive. 

The annotations in plotly layout are constructed from the original annotations (moved to fireflyLayout.annotations) and trace annotations before chart is rendered. The assumption is that we won't have trace annotations when the number of points in the trace is large.

When upper limit annotations are present, reversing y axis or changing its type to 'log' requires reloading data and recreating annotations. Currently the data are reloaded only for the active trace (when the options are modified).

### DIFF
--- a/docs/firefly-api-overview.md
+++ b/docs/firefly-api-overview.md
@@ -94,7 +94,7 @@ Firefly React components can be found in `firefly.ui` module.
 | ---------- | ---- | ----------- |
 | div | string or Object | a div element or a string id of the div element |
 
-<br>
+
 *Example:* 
 ```js
 const props = {
@@ -188,8 +188,8 @@ The FITS viewer can take many, many possible parameters.  Some parameters contro
 
 For the details of FITS plotting parameters see: [fits-plotting-parameters.md](fits-plotting-parameters.md)
  
-<br>
-*Examples:* 
+
+*Examples:*
 ```js
 firefly.setGlobalImageDef({
     ZoomType  : 'TO_WIDTH'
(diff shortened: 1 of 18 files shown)


ejoliet:
@tgoldina when i zoom in, i don't see the visual for the upper limit. Is it expected?

tgoldina:
Thank you for pointing it out, it's a bug. I'll fix it: if upper limit is set, filter should tale into account upper limit column.

tgoldina:
I have addressed the review comments: 
- Filter should preserve upper limits
- The distinct symbol (dash) in addition to arrow is now used for upper limits with no measurement value. (If the measurement value is present upper limit will have no dash at the end of arrow.)
- Reverse axis should reverse upper limit arrows in all traces.
- The example html is updated to avoid unsupported expressions in the options.

@loitly please check it again. The updated deployment is at
    https://irsawebdev9.ipac.caltech.edu/dm-7991/firefly/demo/ffapi-sed.html

loitly:
The changes look great.  Good to go.

PR #575, opened 2018-04-03
DM-13780: fixed image recenter problem

robyww:
-  fixed image recenter problem
  - also fixed the max zoom message show zoom level with HiPS

_to test:_

 - recenter: bring in 4 images, then bring image a hips.  The image should stay centered.
 - zoom: bring in a HiPS and zoom it all the way.  The message should not show a zoom level.


cwang2016:
Works well.

PR #576, opened 2018-04-03
DM-13989: FOV and cube issues

robyww:
Do the following:
 - Cube: fixed the bugs related to moving between HiPS cubes and regular hips
 - Cube: honor the data_cube_* properties and show information about the 3rd dimension
 - Use: `value = crval3 + (cubeIdx - crpix3 ) * cdelt3`
 - Make field of view and target optional
 - Use the HiPS properties if they are not specified.
 - If neither are specified show 180 degrees at 0,0

_To Test:_
 
 - bring up a cube and switch to a regular images, there should be no issues
 - bring up a cube, by the controller should be information about the 3rd dimension
 - bring a up HiPS without enter fov or target, (DSS is a good example). It the properties are specified it will set the FOV and the target

### DIFF
--- a/src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
+++ b/src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
@@ -5,9 +5,9 @@
 import React, {PureComponent} from 'react';
 import numeral from 'numeral';
 import PropTypes from 'prop-types';
-import {flux} from '../../Firefly.js';
 import {isEmpty, get} from 'lodash';
-import {primePlot,isMultiImageFitsWithSameArea, getPlotViewById} from '../PlotViewUtil.js';
+import {primePlot,isMultiImageFitsWithSameArea, getPlotViewById,
+                getDrawLayersByType, isDrawLayerAttached } from '../PlotViewUtil.js';
 import {findScrollPtToCenterImagePt} from '../reducer/PlotView.js';
 import {CysConverter} from '../CsysConverter.js';
 import {PlotAttribute,isHiPS, isImage} from '../WebPlot.js';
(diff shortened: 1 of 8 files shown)


src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -656,3 +657,16 @@ MultiImageControllerView.propTypes= {
...
+    if (!isHiPS(plot)) return '';
+    const {hipsProperties}= plot;
+    const {data_cube_crpix3, data_cube_crval3, data_cube_cdelt3, data_cube_bunit3=''}= hipsProperties;
+    if (!data_cube_crpix3 || !data_cube_crval3 || !data_cube_cdelt3) return ''
+    const crpix3= Number(data_cube_crpix3);
+    const crval3= Number(data_cube_crval3);
+    const cdelt3= Number(data_cube_cdelt3);
+    if (isNaN(crpix3) || isNaN(crval3) || isNaN(cdelt3)) return '';
+    const value = crval3 + ( plot.cubeIdx - crpix3 ) * cdelt3;
+    return `${numeral(value).format('0.00000')} ${data_cube_bunit3}`
```
cwang2016: for cube case like 'CALIFA DR2 V500...', 'data_cube_bunit3' is undefined, and the display shows nothing. 
for cube case like  'HI4PI HI Survey (individual cubes)', 'data_cube_bunit3' is defined as 'null', and the display shows 'null'.  Does 'null' mean some unit in default? or could the display also show nothing instead of 'null'?

cwang2016:
Works well except the minor concern added in previous comment.

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
robyww: This is because `data_cube_bunit3` is the string "null".  I think we are doing the right think by the standard.  However, I think I will just look for the string 'null' and ignore it.

PR #577, opened 2018-04-04
DM-12973 add showImageOrHiPS to API and update demo app

cwang2016:
The development 
- adds 'showImageOrHiPS' into Firefly API. 
- for showImageOrHiPS, either HiPS or regular image will be rendered based on certain criteria including that HiPS request will take the priority in case image request doesn't have target position specified. 

test: 
start firefly/demo/ffapi-hips-test.html

### DIFF
--- a/src/firefly/js/visualize/task/PlotHipsTask.js
+++ b/src/firefly/js/visualize/task/PlotHipsTask.js
@@ -254,6 +254,7 @@ export function makeImageOrHiPSAction(rawAction) {
         const hipsRequest= ensureWPR(payload.hipsRequest);
         const imageRequest= ensureWPR(payload.imageRequest);
         const allSkyRequest= ensureWPR(payload.allSkyRequest);
+
         if (!validateHipsAndImage(imageRequest, hipsRequest, payload.fovDegFallOver)) return;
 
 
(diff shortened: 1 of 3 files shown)


src/firefly/js/visualize/task/PlotHipsTask.js
```diff
@@ -254,14 +254,15 @@ export function makeImageOrHiPSAction(rawAction) {
         const hipsRequest= ensureWPR(payload.hipsRequest);
         const imageRequest= ensureWPR(payload.imageRequest);
         const allSkyRequest= ensureWPR(payload.allSkyRequest);
+
         if (!validateHipsAndImage(imageRequest, hipsRequest, payload.fovDegFallOver)) return;
 
 
         const {plotId, fovDegFallOver, pvOptions, attributes, plotAllSkyFirst=false}= payload;
         const viewerId= determineViewerId(payload.viewerId, plotId);
         const size= getSizeInDeg(imageRequest, hipsRequest);
```
robyww: We talked about if neither specify size then use image could also be false.

cwang2016: ok. will fix it.

PR #579, opened 2018-04-05
DM-13866: Optimize server memory usage

loitly:
https://jira.lsstcorp.org/browse/DM-13866

Update status page to include more details on EmbeddedDB.
http://localhost:8080/firefly/admin/status

Add cache eviction strategy for VIS_SHARE_MEM
        
        - cached data expires after 60 minute of inactivity
        - force eviction check once every 1 minute.

Add cleanup strategy for EmbeddedDB

        There are two stages of clean-up; compact(remove all temp tables), then shutdown(remove from memory)
        - Compact DB once it has idled longer than CLEANUP_INTVL
        - Shutdown DB once it has expired; idle longer than MAX_IDLE_TIME
        - Shutdown DB based on LRU(Least Recently Used) once the total rows have exceeded MAX_MEMORY_ROWS

        Current settings:
          - CLEANUP_INTVL:  1 minutes
          - MAX_IDLE_TIME: 15 minutes
          - MAX_MEMORY_ROWS:  250k rows for every 1GB of max heap, between the range of 1-10 millions.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
@@ -22,6 +22,8 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Date: Jul 17, 2008
@@ -38,6 +40,14 @@ public class EhcacheProvider implements Cache.Provider {
     private static HashMap<String, Boolean> fileListenersReg = new HashMap<String, Boolean>();
     private static HashMap<String, Boolean> logListenersReg = new HashMap<String, Boolean>();
     private static long curConfModTime = 0;
+
+    /*
+      CLEANUP POLICY:
+        - cached data expires after 60 minute of inactivity
+        - force eviction check once every 1 minute.
+     */
+    private static int DEF_TTI_SEC       = 60*60;     // default vis.shared maximum time to idle in seconds before evicted
+    private static int EVICT_CHECK_INTVL = 1;         // interval in minutes to check for expired vis.shared cache items
 
     static {
 
@@ -91,6 +101,13 @@ public class EhcacheProvider implements Cache.Provider {
                     sharedMemSize =  String.format("%dM", (int)(Runtime.getRuntime().maxMemory() * pctVisSharedMemSize/1024/1024));
                 }
                 sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getCacheConfiguration().setMaxBytesLocalHeap(sharedMemSize);
+
+                // setup cleanup task
+                int ttiSecs = AppProperties.getIntProperty("vis.shared.tti.secs", DEF_TTI_SEC);  // defaults to expire after 60 mins of inactivity.
+                sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getCacheConfiguration().setTimeToIdleSeconds(ttiSecs);
+                Executors.newSingleThreadScheduledExecutor()
+                        .scheduleAtFixedRate( () -> sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getKeysWithExpiryCheck()        // this forces eviction
+                                              , EVICT_CHECK_INTVL, EVICT_CHECK_INTVL, TimeUnit.MINUTES);      // check every n minutes
             }
             _log.info("shared cache manager config file: " + sharedConfig);
         }
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/DbAdapter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/DbAdapter.java
@@ -33,12 +33,16 @@ public interface DbAdapter {
         Current settings:
           - CLEANUP_INTVL:  1 minutes
           - MAX_IDLE_TIME: 15 minutes
-          - MAX_MEMORY_ROWS:  250k rows for every 1GB of max heap, between the range of 1-10 millions.
+          - MAX_MEMORY_ROWS:  250k rows for every 1GB of max heap, between the range of 250k to 10 millions.
      */
     long MAX_IDLE_TIME  = 1000 * 60 * 15;   // will shutdown database if idle more than 15 minutes.
     int  CLEANUP_INTVL  = 1000 * 60;        // check every 1 minutes
-    long MAX_MEMORY_ROWS = Math.min(10000000, Math.max(1000000, Runtime.getRuntime().maxMemory()/1024/1024/1024 * 250000));
+    static long maxMemRows() {
+        long availMem = Runtime.getRuntime().maxMemory() - Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
+        return Math.min(10000000, Math.max(1000000, availMem/1024/1024/1024 * 250000));
+    }
 
+    EmbeddedDbStats getRuntimeStats();
 
     /**
      * @return the name of this database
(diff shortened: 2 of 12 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
```diff
@@ -119,11 +105,19 @@ private static void displayCacheInfo(PrintWriter writer, CacheManager cm) {
...
+        Collections.unmodifiableCollection(DbAdapter.getAdapter().getDbInstances().values()).stream()
                     .sorted((db1, db2) -> Long.compare(db2.getLastAccessed(), db1.getLastAccessed()))
-                    .forEach((db) -> writer.println(String.format("\tidled: %2$tM:%2$tS %s", db.getDbFile().getPath(), System.currentTimeMillis() - db.getLastAccessed())));
+                    .forEach((db) -> writer.println(String.format("%4$tM'%4$tS\"  %5$tM'%5$tS\"  %6d  %8d  %s",
+                                                        db.getTblCount(),
+                                                        db.getRowCount(),
+                                                        db.getDbFile().getPath(),
+                                                        System.currentTimeMillis() - db.getLastAccessed(),
+                                                        System.currentTimeMillis() - db.getCreated()
+                    )));
```
tgoldina: I am staring at this println and do not see how the order of the arguments matches the format or the header line above. :) Oh, `4$` and `5$` stand for the argument indices! 

Anyway, I would use ':' to separate minutes and seconds. I think the notation you used is for angles, not for time.

Also, where would I see this printout? How do I check the db status?

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbAdapter.java
```diff
@@ -20,6 +21,25 @@
...
+
+        Current settings:
+          - CLEANUP_INTVL:  1 minutes
+          - MAX_IDLE_TIME: 15 minutes
+          - MAX_MEMORY_ROWS:  250k rows for every 1GB of max heap, between the range of 1-10 millions.
+     */
+    long MAX_IDLE_TIME  = 1000 * 60 * 15;   // will shutdown database if idle more than 15 minutes.
+    int  CLEANUP_INTVL  = 1000 * 60;        // check every 1 minutes
+    long MAX_MEMORY_ROWS = Math.min(10000000, Math.max(1000000, Runtime.getRuntime().maxMemory()/1024/1024/1024 * 250000));
+
```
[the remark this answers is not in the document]
loitly: So yes, I'll have to make this a server's attribute instead of an application one, similar to VIS_MEM_SHARE_MEM.  However, I think it's should be a fixed value rather than on available memory.

Status page link is posted in PR.  http://localhost:8080/firefly/admin/status

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
loitly: Status page link is posted in PR.  http://localhost:8080/firefly/admin/status

I want to denote duration, not time.  How do I say it's been idled for 3 minutes and 20 seconds?

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbAdapter.java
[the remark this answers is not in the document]
loitly: I've changed my mind.  The simplest solution is to set MAX_MEMORY_ROWS based on available memory.  That way, it does not matter how many applications are deployed in the same jvm.  I'll use the same 250k/GB of memory but based on available size instead.  This performance optimization.  It should work even if nothing were in memory.  What do you think?

src/firefly/java/edu/caltech/ipac/firefly/server/cache/EhcacheProvider.java
```diff
@@ -91,6 +101,13 @@
                     sharedMemSize =  String.format("%dM", (int)(Runtime.getRuntime().maxMemory() * pctVisSharedMemSize/1024/1024));
                 }
                 sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getCacheConfiguration().setMaxBytesLocalHeap(sharedMemSize);
+
+                // setup cleanup task
+                int ttiSecs = AppProperties.getIntProperty("vis.shared.tti.secs", DEF_TTI_SEC);  // defaults to expire after 60 mins of inactivity.
+                sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getCacheConfiguration().setTimeToIdleSeconds(ttiSecs);
+                Executors.newSingleThreadScheduledExecutor()
+                        .scheduleAtFixedRate( () -> sharedManager.getCache(Cache.TYPE_VIS_SHARED_MEM).getKeysWithExpiryCheck()        // this forces eviction
+                                              , EVICT_CHECK_INTVL, EVICT_CHECK_INTVL, TimeUnit.MINUTES);      // check every n minutes
```
robyww: This is almost scary, how easy it is now to start a background thread that runs forever.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbAdapter.java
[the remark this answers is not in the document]
robyww: I think that is fine, I like basing it on  available memory

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
```diff
@@ -119,11 +105,19 @@ private static void displayCacheInfo(PrintWriter writer, CacheManager cm) {
     private static void showDatabaseStatus(PrintWriter writer) {
         writer.println("DATABASE INFORMATION");
         writer.println("--------------------");
-        writer.println("Open: " + BaseDbAdapter.getDbInstances().size());
-        writer.println("Details: idle time is in (mm:ss)");
-        Collections.unmodifiableCollection(BaseDbAdapter.getDbInstances().values()).stream()
+        writer.println(String.format("Open: %d    CHECK_INTVL: %ds    MAX_ROWS: %d    MAX_IDLE: %dm",
+                        DbAdapter.getAdapter().getDbInstances().size(), DbAdapter.CLEANUP_INTVL/1000, DbAdapter.MAX_MEMORY_ROWS, DbAdapter.MAX_IDLE_TIME/1000/60));
+        writer.println("Idled   Age     Tables  Rows      File Path");
+        writer.println("------  ------  ------  --------  ---------");
```
robyww: Is it possible to do row and columns? They way is would be easier to estimate table size

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
tgoldina: I am not an expert in English notation. :) I just don't remember seeing ' and '' for time anywhere. I was curious though to google it up. Looks like this notation was used for duration (elapsed time) before, but it's hardly used now:
https://english.stackexchange.com/questions/114205/english-notation-for-hour-minutes-and-seconds 
https://www.dailywritingtips.com/how-to-refer-to-time/

tgoldina:
One more thing. 

It would be useful to keep the highest total "Rows" - we calculate this number anyway on every cleaning. We can store the max value. 

If we can get "Columns", "Rows"x"Columns is another useful metrics for statistics, because it gives the number of table cells and is better estimate for table size. 

I would suggest to keep max values for both: sum("Rows") and sum("Rows"*"Columns").

ejoliet:
@loitly Is this ticket IRSA-999 related to this PR?

loitly:
> @loitly Is this ticket IRSA-999 related to this PR?

No, it is not.

loitly:
@tgoldina , added many of your suggestions if you want to take a look.
@ejoliet , added ticket: https://jira.ipac.caltech.edu/browse/IRSA-999 to this PR as well.

PR #580, opened 2018-04-06
DM-14009: Made the parameter passing smaller and then general cleanup.

robyww:
this size of the data passed is a lot smaller especially for three color. Before it was about 6000 characters not it is about 3000.

  - Then there is some general cleanup.  This is hard to test.  Everything should be working like before.
 - Also fixed a couple of small HiPS issues that came up.

_to test:_ make a plot and bring up network in the developer tools.  Notice that the size of `FireFly_ImageDownload` is smaller then before.


ejoliet:
Thank you!
I've tested that color images are working smoothly now but someone should check if other fixes in this PR affect other features in HiPS (?).

PR #583, opened 2018-04-10
DM-13872: Title of layer-selection dialog should change with change of image

tgoldina:
https://jira.lsstcorp.org/browse/DM-13872

Deployment: https://irsawebdev9.ipac.caltech.edu/dm-13872/irsaviewer/

To test: using Images drop-down load a few images, open layers dialog, change active image, observe dialog title change.


tgoldina:
@robyww Sure, let' talk about SimpleComponent. Just for statistics: currently we have 50 other components that use the same approach as SimpleComponent (`flux.addListener`) to connect to the store. SimpleComponent is extended 19 times.

PR #591, opened 2018-04-13
DM-13037: Second filter fails on column names with spaces (NED)

loitly:
https://jira.lsstcorp.org/browse/DM-13037

See ticket for details on how to reproduce.

This can be tested with this demo page as well.
http://localhost:8080/firefly/demo/ffapi-sed.html


cwang2016:
review completes and it works well.

PR #592, opened 2018-04-13
DM-14058 Make HiPS selection back to the popular/merged list style

cwang2016:
The development includes, 
- make a single checkbox with default label like  'IRSA Featured' . If the checkbox is checked,  It means HiPS featured search. If the checkbox is unchecked. it means merged list search. 
- The initial value of the checkbox is 'checked'. 
- The application can have the following parameters configured in context of HiPS list search, 
    . hipsSources, ex: hipsSources: 'irsa,cds'     
        -- string, all source(s) for the merged list. The sources in the list are sorted by the order in the merged list. 
    . defHipsSources, ex:  defHipsSources: {source: 'irsa', label: 'IRSA Featured'}  
        --  array or an object, default source(s) for featured search. 
             Both the source name and the label for the checkbox are defined. 
    . mergedListPriority:, ex: mergedListPriority: 'irsa'
        -- string, defining sources that take higher priority for the merged list. It may contain single source or multiple sources from highest priority to lowest.  
        --  If two HIPS  are with the same ivoid, the HiPS from the source which is not listed in the defined string or with lower priority will be removed from the merged list. 
        -- no HiPS will be removed from the merged list if not defined. 

PS. please check IFE for the same branch name, DM-14058-HiPSPopularMergedList

Test: 
- start HiPS search from firefly.html or firefly-dev.html 
- check or uncheck 'IRSA Featured' checkbox
- do HiPS search from HiPS list or 'Url' (entering Url or IVOID) 
- click 'HiPS change' to get HiPS list popup, check or uncheck  'IRSA Featured' checkbox, or change HiPS search.


ejoliet:
Very strange I've built irsaviewer and can't see the table nor the check box. After minutes, it came back without checkbox. To test here:
https://irsawebdev9.ipac.caltech.edu/dm-14058/irsaviewer/

On Firefly is ok:
https://irsawebdev9.ipac.caltech.edu/dm-14058/firefly/ 

I tried the IVO ID in url and that also works! It little bit confusing not knowing where the map is coming from when i put the IVOA-ID but that's a different problem.

About IRSAViewer problem, my guess is that we need to update properties in IFE?

Can you, @robyww or @loitly check what is happening with IRSAViewer? 

Thanks.

robyww:
@cwang2016 and I talked about putting `hipsSearch` back into `hips` and she will modify `IrsaViewer.js`

cwang2016:
@ejoliet the HiPS configuration update is pushed into IFE, DM-14058-HiPSPopularMergedList. 
Please help review it and I will merge it after the review is done.

PR #594, opened 2018-04-18
IRSA-1598 and IRSA-1015

lznakano:
There is the branch in ife with the same branch name. 

 In this pull request, the above two tickets were implemented:

- Add workspace option to download dialog for packaging images

- Improve Finder Chart workspace access error message

### DIFF
--- a/src/firefly/js/visualize/WorkspaceCntlr.js
+++ b/src/firefly/js/visualize/WorkspaceCntlr.js
@@ -258,6 +258,26 @@ export function getWorkspaceStatus() {
     return isEmpty(getWorkspaceList()) && get(flux.getState(), [WORKSPACE_PATH, 'status']);
 }
 
+export function getWorkspaceErrorMsg(){
+    const statusMsg = getWorkspaceStatus();
+    //The status code and corresponding messages are in ResponseMessage.java
+    var errorMsg;
+    switch (statusMsg){
+        case 'Unauthorized':
+            errorMsg = 'You have not logged in';
+            break;
+        case 'Forbidden':
+            errorMsg = 'You do not have access right';
+            break;
+        case 'Not Found':
+            errorMsg = 'The request resources can not be found';
+            break;
+        default:
+            errorMsg = statusMsg;
+            break;
+    }
+    return errorMsg;
+}
 export function isAccessWorkspace() {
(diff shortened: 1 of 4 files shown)


lznakano:
URL: https://irsawebdev9.ipac.caltech.edu/irsa-1015-irsa-1598/applications/finderchart/

src/firefly/js/visualize/WorkspaceCntlr.js
```diff
@@ -258,6 +258,26 @@ export function getWorkspaceStatus() {
     return isEmpty(getWorkspaceList()) && get(flux.getState(), [WORKSPACE_PATH, 'status']);
 }
 
+export function getWorkspaceErrorMsg(){
+    const statusMsg = getWorkspaceStatus();
+    //The status code and corresponding messages are in ResponseMessage.java
+    var errorMsg;
+    switch (statusMsg){
+        case 'Unauthorized':
```
ejoliet: Isn't more generic to use a status code than the status message that can change anytime?

src/firefly/js/visualize/WorkspaceCntlr.js
```diff
@@ -258,6 +258,26 @@ export function getWorkspaceStatus() {
...
+            errorMsg = 'You have not logged in';
+            break;
```
ejoliet: From the ticket IRSA-1598, Vandana mentioned another sentence which i think is little bit more useful for the user: "Please click 'Login' in the upper right corner of the window"

I would add it to the current one so it becomes:
`You are not logged in. Please click 'Login' in the upper right corner of the window`

src/firefly/js/visualize/WorkspaceCntlr.js
lznakano: I planned to change the server code to pass the status code instead of the status message.  But I was worried the impact to others.  Since only the login message is matter, so I gave up my first plan.  But if you want it, I can do so.

src/firefly/js/visualize/WorkspaceCntlr.js
lznakano: Changed the message as what  it was in the ticket.

lznakano:
I implemented IRSA-1731 in the same pull request.  The server now returns the status code to the client and then the client uses the status code to display the error message.

@ejoliet Please review it again.  Thanks!

PR #595, opened 2018-04-20
IRSA-318 Time Series Tool should not put lower limit on periods. set …

wmiipac:
As requested by user in the "Set Period" panel, the desired behavior is to:
  1.  keep default minimum value of the slider as 0.001
  2.  set the period value validation should be just larger than 0, i.e. user can enter period >0 and set lower limit of the slider >0.


ejoliet:
@wmiipac Can you build irsaviewer in kubernetes (via Jenkins) and paste the url for Luisa to test?

lrebull:
OK, as far as I can tell, this build does what was asked of it. We have the whole 5000 point limit, which is a different thing entirely. But as far as I can tell, this meets the requirements.

ejoliet:
Luisa, just to clarify, this change is not for the release anyway. This will appear in dev.

PR #597, opened 2018-04-21
IRSA-1709: Catalog search intermittently fail to load results

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-1709
IRSA-1709: Catalog search intermittently fail to load results

Best way to reproduce this is to use Safari's '_private window_'.  Once firefly is loaded, do a catalog search with a larger radius search.  This will put the search into a 'backgroundable' mode, but the results will fail to return.

This is an issue when there are multiple connections to the server without a `usrkey`(cookie).
The logic is to create a `usrkey` when one is not defined.  In this scenario, a websocket connection
is given a different usrkey then subsequence calls, causing server to client communication failure.

Solution is to create a `usrkey` if one is not found.  Also, establish a websocket connection first 
before doing anything else.  This ensure that the client is getting all of the messages from the
server during startup.

Another notable change in this PR is the reorganization of what we call startup parameter.
In the old way, we define a firefly.app object which contains properties used during startup(bootstrap).
firefly.app may contains an 'options' property which is used to customize a variety of Firefly's behaviors.
Also, there are places where 'options' are defined directly under firefly, i.e. '`firefly.options`'

In this PR, bootstrap will take 2 parameters; props and options.
- props(AppProps) is defined as firefly.app
- options(FireflyOptions) is defined as firefly.options

This will be used consistently everywhere, API as well as the application HTML page.

The defaults for these 2 parameters are defined in `Firefly.js`.  It should also provides basic documentation on its usage.. (just skeleton, to be filled in later)
There are 2 places where the defaults can be overridden; at the application entry point or at window.firefly.
Properties defined in window.firefly will override entry point, and what's defined in entry point will override Firefly.js's defaults.
Parameters are deep(recursively) merged, unless the value is an array, in which case, it will replace.

Also in this PR:
- remove old GWT code
- refactor Table code to use a generic 'Send to background' panel

Changes are in firefly, ife, and suit under the same branch name.
Reviewers, please help to ensure that the proper options/props are defined for every applications.  I've removed many duplicate declarations and fear that I may have remove one too many.


src/firefly/js/Firefly.js
```diff
@@ -65,35 +86,81 @@ export const Templates = {
...
-    imageMasterSourcesOrder: '',
+    imageMasterSourcesOrder: undefined,
     workspace : { showOptions: false},
-    hips : {useForCoverage: true, useForImageSearch: true,
-            hipsSources: 'all',
-            defHipsSources: {source: 'irsa', label: 'Featured'},
-            mergedListPriority: 'irsa'}
+
+    charts: {
+        // TODO: need to define all options with defaults here.
```
tgoldina: maxRowsForScatter: 5000, // maximum table rows, for which scatter plot can be created
multitrace: true,  // deprecated - by default we use multi-trace architecture
singleTraceUI: false, // by default we support multi-trace in UI
ui: {HistogramOptions: {fixedAlgorithm: undefined}} // by default we allow both "uniform binning" and "bayesian blocks"

lznakano:
I checked the Finderchart in this pull request.  The single and multiple search work fine.  But download is hanging.  It showed the "send to background" but never got result back.

src/firefly/html/firefly.html
```diff
@@ -12,33 +12,7 @@
...
-                    coverage : { // example of using DSS and wise combination for coverage (not that anyone would want to combination)
-                        hipsSourceURL : 'http://alasky.u-strasbg.fr/DSS/DSSColor', // url
-                        imageSourceParams: {
-                            Service : 'WISE',
-                            SurveyKey: '1b',
-                            SurveyKeyBand: '4'
-                        }
-                    }
-                }
+                template: 'FireflyViewer'     // set to app mode, instead of api.
```
robyww: Loi an I discuss moving all the defined options in firefly.html to firefly-dev.html.  This way firefly-dev.html can become an example of how to set options.  They will defined in firefly-dev.html even if they are not overriding the default.  Also we will add more and some documentation there as well.  This example will be very valuable of NED and new developers that are less familiar with options.

src/firefly/html/slate.html
```diff
@@ -17,14 +17,14 @@
...
-                    hips : {useForCoverage: true, useForImageSearch: true},
-                    charts: {multitrace: true}
-                }
+                ],
+            },
+            options : {
+                MenuItemKeys: {maskOverlay:true},
+                catalogSpacialOp: 'polygonWhenPlotExist',
+                disableDefaultDropDown: true,
+                hips : {useForCoverage: true, useForImageSearch: true},
```
robyww: hips should be removed since it will pickup the default. I think hips is defaulted now.

PR #607, opened 2018-05-11
DM-13382: In TablePanel, allow option to show data type

loitly:
https://jira.lsstcorp.org/browse/DM-13382

Add option to show/hide column's data type.  Option is added to option panel and exposed through API as well.  
I took the liberty of adding some style to the data type because it looks pretty bad without it.  
If you don't like it, I have not problem changing it to something people agree to.

A couple of things you should be aware of.
- Sorting is triggered by clicking anywhere on the column header, not just the column's label.
- Cursor's is change to a 'pointer' icon when hover over the column's header.

Deployed to k8s:
https://irsawebdev9.ipac.caltech.edu/dm-13382/firefly/

### DIFF
--- a/src/firefly/js/api/ApiHighlevelBuild.js
+++ b/src/firefly/js/api/ApiHighlevelBuild.js
@@ -124,14 +124,22 @@ function buildTablePart(llApi) {
      * @prop {boolean} removable    true if this table can be removed from view.  Defaults to true.
      * @prop {boolean} backgroundable    true if this search can be sent to background.  Defaults to false.
      * @prop {boolean} showUnits    defaults to false
+     * @prop {boolean} showTypes    defaults to false
      * @prop {boolean} showFilters  defaults to false
      * @prop {boolean} selectable   defaults to true
      * @prop {boolean} expandable   defaults to true
      * @prop {boolean} showToolbar  defaults to true
+     * @prop {boolean} showTitle    defaults to true
+     * @prop {boolean} showPaging   defaults to true
+     * @prop {boolean} showSave     defaults to true
+     * @prop {boolean} showOptionButton    defaults to true
+     * @prop {boolean} showFilterButton    defaults to true
      * @prop {boolean} border       defaults to true
+     * @prop {boolean} help_id      link to help if applicable
      * @prop {function[]}  leftButtons   an array of functions that returns a button-like component laid out on the left side of this table header.  Function will be called with table's state.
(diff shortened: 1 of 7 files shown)


ejoliet:
I actually found a problem. When filters are shown and enter in one, the column get sorted unexpectedly. You might want to check. It doesn't happen in dev or test.

loitly:
> I actually found a problem. When filters are shown and enter in one, the column get sorted unexpectedly. You might want to check. It doesn't happen in dev or test.

This is fixed and redeployed to k8s.

ejoliet:
@loitly Expected feedback from @vandesai1 this Friday.

vandesai1:
I was hoping to get more scientists to weigh in on this, but they are all busy. So here are my comments: 

- In order to display the data types (they are off by default), you have to click on the gears icon and select "Show Types" to turn it on. This is not clear enough. "Show Data Types" would be more clear. If that is too long to fit in the dialog, then consolidate all instances of "Show", so it would be "Show: Units Types Filters", which check boxes in front of "Units", "Types, and "Filters".

- Note that the above suggestion of consolidation also changes the order, which should be in the same order as the information is displayed. Currently, it is column name, data type, units, filters. I think it should be column name, units, data type, filters. I realize that data type comes above units in IPAC tables, but it's not the most sensible presentation of the data.

- The data types should not be red. That makes them see very important. They aren't more important than the column name. It's ok if they remain italicized.

loitly:
@vandesai1, I've made changes according to your suggestions.  You can see it here: https://irsawebdev9.ipac.caltech.edu/dm-13382/firefly/.

vandesai1:
@loitly, in the dialog where you control the column header, I think it should say "Data types" instead of just "Types". Also, I think units should be on by default.

loitly:
> Also, I think units should be on by default.
Are you sure about this?  Many tables do not have units.  Leaving it on by default will create blank spaces in the header for these tables.

vandesai1:
Do you have an example of tables without units, so I can take a look?

xiuqin:
Unit display turn on/off by default should be captured in separate ticket. We need to go through other types of table data. I would say for this ticket, lets concentrate on data type display. And I think by default it should be turned off.

ejoliet:
If CCB agree on having unit displayed by default and this PR is still open, i think we should take the opportunity to check this off the list.
Otherwise, raise ticket and assign to me. I will do it right away after CCB agrees.

loitly:
Updated label from 'Types' to 'Data Types'.

PR #609, opened 2018-05-17
IRSA-1785: Add skeleton of SOFIA app to IFE

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-1785
https://jira.ipac.caltech.edu/browse/IRSA-1837
https://jira.ipac.caltech.edu/browse/IRSA-1878
https://jira.ipac.caltech.edu/browse/IRSA-1880
https://jira.ipac.caltech.edu/browse/IRSA-1897
https://jira.ipac.caltech.edu/browse/IRSA-1898

Changes also in IFE under the same branch.
Deployed to k8s at: https://irsawebdev9.ipac.caltech.edu/irsa-1785/applications/sofia/

To build:  cd ./ife; gradle :sofia:war

Skeleton created with:
- Sofia.jsx (main app)
- Sofia.css (stylesheet)
- SofiaConfig.js  (config, constant, etc)
- SofiaManager.js
- SofiaSearchForm.jsx
- SofiaBasicView.jsx
- SofiaQuery
   - query wise TAP interface
- SofiaImageRetrieve
  - based on selected row's ra/dec, retrieved wise most-centered image from IBE.


ejoliet:
Also, maybe @robyww can help to debug the issue related to the image expand mode.

loitly:
@ejoliet , @wmiipac 
I've added another commit to address the additional features requested.  It's pushed out to k8s now.
However, the `Sent to Viewer` will not work because the link to irsaviewer is set to /irsaviewer, which will not work in this k8s environment.  It should work locally and elsewhere in irsa.

ejoliet:
I guess you have left also this for after merging?

> - we would need also to have a way to set the order and the active tab such that AOR table is the first one and highlighted. (this is also related to this issue in FC IRSA-1249 and related issues)

loitly:
Added another commit that fixed irsa ticket [IRSA-1897](https://jira.ipac.caltech.edu/browse/IRSA-1897).
Also, make irsaviewer path relative so that it can be deployed to k8s at a subpath.

loitly:
Added another commit to work on irsa-1898: Make possible to order tables result tabs.

ejoliet:
@loitly I've checked that IRSA-1898 and IRSA-1897 looks good in the latest commit. Thanks!

src/firefly/js/tables/tables-typedefs.jsdoc
```diff
@@ -134,3 +134,30 @@
...
+ * @prop {boolean} [showFilters=false]
+ * @prop {boolean} [showToolbar=true]
+ * @prop {boolean} [showTitle=true]
+ * @prop {boolean} [showPaging=true]
+ * @prop {boolean} [showSave=true]
+ * @prop {boolean} [showOptionButton=true]
+ * @prop {boolean} [showFilterButton=true]
+ * @prop {function[]}  [leftButtons]   an array of functions that returns a button-like component laid out on the left side of this table header.
+ * @prop {function[]}  [rightButtons]  an array of functions that returns a button-like component laid out on the right side of this table header.
+ */
```
ejoliet: Always good to see doc improvements! +1

src/firefly/js/tables/TablesCntlr.js
```diff
@@ -357,7 +357,7 @@ function tblResultsAdded(action) {
             var {tbl_id, title, options={}, tbl_ui_id} = action.payload;
 
             title = title || tbl_id;
-            options = Object.assign({tbl_group: 'main', removable: true}, options);
+            options = Object.assign({tbl_group: 'main', removable: true, setAsActive:true}, options);
```
ejoliet: That would mean that in order to activate a particular table, i need to pass on the attribute 'setAsActive:false' to any other table request?

loitly: I would not put it that way, because there's an action to set a table active.

This added option is useful in the case when you know whether or not this table should be active by default when added.

src/firefly/js/api/ApiViewer.js
```diff
@@ -117,6 +117,19 @@ export function getViewer(channel,file=defaultViewerFile) {
...
+ * @param {object} p            parmeter
+ * @param {String} [p.channel]  default to current connected channel.
+ * @param {String} [p.file]     url path, or the html file to load.  Defaults to blank(index of the app).
+ * @param {String} [p.iframeId]
+ */
+export function getRemoteViewer({channel, file, iframeId}) {
+    const win = iframeId ? document.getElementById(iframeId).contentWindow : window;
+    const getViewer = get(win, 'firefly.getViewer');
+    return getViewer && getViewer(channel, file);
+}
```
robyww: A few comments:

- Where is loadRemoteApi?  Has it been written yet?
- What is the difference between the name getViewer and getRemoteViewer? They are both really the same.  They both want to talk to a viewer
- right now, `getRemoteViewer` will act like `getViewer` if channel and file are only passed.
- could the iframe id and the channel id be the same?  In other words, we always name an iframe by this new channel we a using to talk? They we would not need a iframeId parameter. It could check using channel.
- Down the road if we removed the iframe we could not touch the api.
- I do like the fact that the rest of the codes does not need to know it is talking to an iframe 
- I think:
    - `getRemoteViewer` should be merged with `getViewer` and called `getViewer`
    -  take 2 parameters as it does now
     - have coded added to check it an iframe with the channel name exist
     - it it does, then operate as `getRemoveViewer` otherwise operate as `getViewer`

loitly: IMO, this is how getViewer should be:
`getViewer({channel, file, baseUrl})`
- baseUrl is where you can locate firefly_loader.js, i.e /irsaviewer, /applications/finderchart, etc
  - use baseUrl as the frame-id, if iframe is used.
  - we should not tie channel to baseUrl because it's possible to have multiple channels. 
- it should be async, either via promise(preferred but not fully supported) or a callback.
  - this allow for lazy-loading and does not require loading the api separately.
- the single parameter with properties is to allow for more parameters in the future without fear of breaking the api.

If we want to do the above and still be backward compatible, we can make `getRemoteViewer` the new api and deprecate `getViewer`.  `getViewer` will call `getRemoteViewer` with immediate return for a `synch` behavior.

robyww: I guess you could do that.  That might even work better if the future if we make part of it npm installable.  But, you could since you could still use `getViewer`, If it is passed an object just use the new behavior, if a string use the old.

robyww: promises are support if we load firefly because we  bring them in.

robyww: OK, I am having second thoughts about `getViewer()` returning a promise. I think your original two function approach is better.  The reason is that for 90% of the use cases it makes `getViewer()` more complex. I think we need to make it simple for the normal use case.

loitly:
@robyww , I've made another commit to address some of your concerns.
- getViewer remains as is.  getRemoteViewer introduced to support loading `viewer` from a different application.
- both are synchronous call because browser popup will block if it's asynchronous.
- iframe implementation reference is completely hidden
Please review again.

PR #612, opened 2018-05-22
DM-14516 make ds9 region file parsing case insensitive

cwang2016:
This development includes: 
- make the case-insensitive parsing on ds9 regions to be case insensitive on both server and front ends. 

test: 
- do image search on firefly on target (0,0, j2000) and upload footprint file from 'Load DS9 Region File' popup.   
- go to firefly/demo/ffapi-footprint-test.html, and add footprint layer from the footprint tool page.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/RegionFactory.java
+++ b/src/firefly/java/edu/caltech/ipac/util/RegionFactory.java
@@ -165,6 +165,7 @@ public class RegionFactory {
                     }
                 }
 
+                region_type = region_type.toLowerCase();
                 boolean isWorldCoord = isWorldCoords(coordSys);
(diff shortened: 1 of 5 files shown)


src/firefly/java/edu/caltech/ipac/util/RegionFactory.java
```diff
@@ -568,15 +569,19 @@ private static RegionValue convertToRegionValue(String radius_string, ValueType
                     unit= RegionValue.Unit.ARCMIN;
                     break;
                 case 'd':      // degrees
+                case 'D':
```
robyww: why not convert `unit_char` to lower as well.

`Character.toLowerCase()` static method in the java.lang.Character package.

PR #616, opened 2018-06-04
IRSA-1437:Finder Chart misinterprets non-standard coordinates

lznakano:
Fixed the bug for incorrectly parsing the target "10 10 10 +10 10 10".

 There is no change in IFE.

Please test this build:
URL: https://irsawebdev9.ipac.caltech.edu/irsa-1437/applications/finderchart/

### DIFF
--- a/src/firefly/js/util/PositionParser.js
+++ b/src/firefly/js/util/PositionParser.js
@@ -229,7 +229,8 @@ var makePositionParser = function(helper) {
                     dec = numericList[1];
 
                 } else if (numericList.length>2) {
-                    idx=0;
+                    throw 'Invalid input';
+                   /* idx=0;
                     for (i=0; (i<numericList.length); i++) {
                         item= numericList[i];
                         if ((idx++)*2<numericList.length) {
@@ -237,7 +238,7 @@ var makePositionParser = function(helper) {
                         } else {
                             dec += (item + ' ');
                         }
-                    }
+                    }*/
                 } else if (numericList.length===1) {
                     ra= tokenAry[0];
                     if (tokenAry.length>1) {


ejoliet:
Thanks for fixing this. I've clicked on the link to test it but the validation of "10 10 10 +10 10 10" fails. Are you sure the branch has been build correctly?

lznakano:
@xiuqin Yes, the Firefly treated "10 10 10 +10 10 10" as "10h00m00s 10d00m00s".  If there are numbers separated by space, firefly split to half and half, the first half is RA and second half is DEC.  The helpdesk ticket complained about it because "10 10 10 +10 10 10"  was interpreted in correctly.
My fix is reject such non-standard input.  The "10 10 10 +10 10 10" is entered, validation should warn
the users the wrong input.

lznakano:
@ejoliet  The failure is the behavior I just changed to.  The ticket complained that "10 10 10 +10 10 10" was not interpreted correctly.  The best way for such input is to reject it since it is not standard.  Nobody can read users' mind.

xiuqin:
There should be a CCB decision about this behavior change.
Please don’t merge yet.

Xiuqin

loitly:
The ticket says.. `10 10 10 +10 10 10` is read as `10h00m00s 10d00m00s`, which is no longer true.
In ops /irsaviewer, when I enter `10 10 10 +10 10 10`, I get `10h10m10s 10d10m10s` which is consistent with IRSA's Radar and Gator.

Now, I don't know why in this instance of FinderChart, it became validation error.  
This is not a fix.  It should be what is currently in /irsaviewer ops.

ejoliet:
@lznakano there is nothing to be fixed as It turns out that the issue was fixed in dev/ops (see IRSAViewer) but not in FC because we didn't release it. The idea has always been the same: if i enter a coordinates that looks like HMS/DMS withoiut the letters, then it should expand and interpret it as it is. 10h10m10s,etc... which is the case in dev/ops. 
Please, chech that this is the case in irsadev/applications/Finderchart for future release of FC (hopefully sooner than later).

If test is positive, then nothing to be merged. Thanks.

lznakano:
The current Finderchart interpreted the input as expected.  This pull request is closed now.

PR #617, opened 2018-06-06
DM-14125 Remove unused chart code and server side expressions, update docs

tgoldina:
https://jira.lsstcorp.org/browse/DM-14125
k8s deployed: https://irsawebdev9.ipac.caltech.edu/dm-14125/irsaviewer/

- Removed pre-multitrace charts code
- Removed server-side expression library
- Updated Firefly API Overview docs
- Fixed a filtering bug (parseInput in FilterInfo was removing only the outermost quotes, which did not work for column expressions, like `"w1mpro"-"w2mpro"`, resulting in a failure when another filter was applied.)
- Fixed column expression quoting bug  in ChartUtil's `formatColExpr` method: the second variable was not quoted in the expressions like `w1mpro*w1mpro`
- Uncommented previously failing queries in LSSTDbServTest, since DM-14617 is completed.

Apart from bug fixes, everything should work as before.


loitly:
chart select-filter fail on table with 'like' filters.
- catalog search > m51, 100 arcsec
- filter table on column designation using string '329'  -> like '%329%'
- select an area in chart
- click 'filter' icon
- fail with 'Column not found: DESIGNATION'

PR #619, opened 2018-06-08
DM-14694 fix pixel coordinate to be compatible with FITS convention

cwang2016:
This development fixes the image pixel coordinate per FITS convention and using WCSA,  i.e. the center of the lower-left pixel is (1, 1).   The updates include, 
- the coordinate conversion between image point and world point, image point and screen pixel. 
- the coordinate range of image pixel starting from [0.5, 0.5], (i.e. the lower-left corner of pixel xy0) and the center coordinate of image pixels. 
- adjustment on image coordinates for image crop request sent to the server

test: 
do image search, 
- check the image coordinate at the edge
- crop the image and check the image pixel coordinate starting from 0.5, 0.5
- upload a region file with content like the following for image case, target: m45, wise allwise w1, and check the image pixel of the regions, 
  
   image;polygon 119 25 140 195 120 150 #color=cyan
   image;circle 196 182 2i # color=orange

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
@@ -471,6 +471,11 @@ public class VisServerOps {
             Band bands[] = state.getBands();
             WebPlotRequest cropRequest[] = new WebPlotRequest[bands.length];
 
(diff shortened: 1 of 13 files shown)


cwang2016:
regarding the constant name, xy0Fits, how about renaming it to be like wcsaFits?

robyww:
I think we will close this pull request but keep the branch around for reference for awhile.

PR #621, opened 2018-06-08
DM-14694: Changed readout so that it will show the FITS standard Image point

robyww:
DM-14694: Changed readout so that it will show the FITS standard Image point

- image pixel show as FITS standard
- includes option to show the lsst image point
- make convert image pixel between fits convention and internal image style while reading in and write out the region description with image coordinates.

### DIFF
--- a/src/firefly/js/visualize/CsysConverter.js
+++ b/src/firefly/js/visualize/CsysConverter.js
@@ -256,7 +256,7 @@ export class CysConverter {
         return makeImagePt(imPt.x-.5, imPt.y-.5);
     }
 
-    convertFitsStandardImagePtToInternalImage(pt) {
+    static convertFitsStandardImagePtToInternalImage(pt) {
         return makeImagePt(pt.x-.5, pt.y-.5);
     }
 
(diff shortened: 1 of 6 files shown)


cwang2016:
It is good to have fits image pixel and zero based image pixel into the coordinate options.
Review completes

src/firefly/js/visualize/region/RegionDescription.js
```diff
@@ -153,8 +154,7 @@ export function startRegionDes(regionType, cc, ptAry,
             sys = Point.IM_PT;
             des = `image;${DS9RegionName[regionType.key]}`;
             ptAry.forEach((pt) => {
-                var imgPt = cc.getImageCoords(pt);
-
+                const imgPt = cc.getFitsStandardImagePtFromInternal(cc.getImageCoords(pt)); // screen
```
robyww: you can just say `cc.getFitsStandardImagePtFromInternal(pt)` and it should work.

src/firefly/js/visualize/CsysConverter.js
```diff
@@ -256,7 +256,7 @@ export class CysConverter {
         return makeImagePt(imPt.x-.5, imPt.y-.5);
     }
 
-    convertFitsStandardImagePtToInternalImage(pt) {
+    static convertFitsStandardImagePtToInternalImage(pt) {
```
robyww: good catch

PR #622, opened 2018-06-11
DM-14759: support XY0, default it for slate.html

robyww:
_to test:_
bring the api with slate.html or just load http://localhost:8080/firefly/slate.html.

`O based image point` should be the default.  XY0 support should work.

### DIFF
--- a/src/firefly/js/visualize/CsysConverter.js
+++ b/src/firefly/js/visualize/CsysConverter.js
@@ -337,7 +337,7 @@ export class CysConverter {
     makeIPtFromxZeroImPt(pt) {
         if (!pt) return null;
         const {ltv1,ltv2}= CysConverter.getLtv(this.projection.header);
-        return makeZeroBasedImagePt(pt.x+.5+ltv1, pt.y+.5+ltv2);
+        return makeImagePt(pt.x+.5+ltv1, pt.y+.5+ltv2);
     }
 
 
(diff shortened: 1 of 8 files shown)


src/firefly/js/visualize/CsysConverter.js
```diff
@@ -295,8 +327,17 @@ export class CysConverter {
...
+    }
+
+    static makeIPtFromFitsImPt(pt) {
+        return pt ? makeImagePt(pt.x-.5, pt.y-.5) : null;
+    }
+
+    makeIPtFromxZeroImPt(pt) {
+        if (!pt) return null;
+        const {ltv1,ltv2}= CysConverter.getLtv(this.projection.header);
+        return makeZeroBasedImagePt(pt.x+.5+ltv1, pt.y+.5+ltv2);
```
cwang2016: does here make {type:ZERO_BASED_IM_PT} or {type:IM_PT}?  should that be the second one instead of the first one?

src/firefly/js/visualize/CsysConverter.js
```diff
@@ -295,8 +327,17 @@ export class CysConverter {
...
-        if (!iwPt) return null;
-        return makeImagePt(iwPt.x, iwPt.y);
+        return iwPt ? makeImagePt(iwPt.x, iwPt.y) : null;
+    }
+
+    static makeIPtFromFitsImPt(pt) {
+        return pt ? makeImagePt(pt.x-.5, pt.y-.5) : null;
+    }
+
+    makeIPtFromxZeroImPt(pt) {
```
cwang2016: could this be static?

src/firefly/js/visualize/CsysConverter.js
robyww: good catch, thanks.

src/firefly/js/visualize/CsysConverter.js
robyww: no, it references `this.projection.header`

PR #625, opened 2018-06-13
DM-14780: power law and asinh stretch bug

tgoldina:
Ticket: https://jira.lsstcorp.org/browse/DM-14780
Data values in lower and upper range are not mapped correctly for power law and asinh stretch

To test: use https://irsawebdev9.ipac.caltech.edu/dm-14780/firefly/
- bring an image, 
- open Color Stretch dialog, 
- modify Upper and Lower range
- refresh
- test with Power Law Gamma and Asinh stretch types

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
@@ -7,7 +7,6 @@ import nom.tam.fits.BasicHDU;
 import nom.tam.fits.Fits;
 import nom.tam.fits.FitsException;
 import nom.tam.fits.HeaderCardException;
-import nom.tam.fits.FitsFactory;
 import nom.tam.fits.Header;
 import nom.tam.fits.HeaderCard;
 import nom.tam.fits.ImageData;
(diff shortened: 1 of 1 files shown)


src/firefly/java/edu/caltech/ipac/visualize/plot/FitsRead.java
```diff
@@ -1140,11 +1147,17 @@ private static double getPowerLawGammaStretchedPixelValue(double x, double gamma
...
+     * @param maxFlux
+     * @param minFlux
      * @param beta
      * @return
      */
     private static double  getASinhStretchedPixelValue(double flux, double maxFlux, double minFlux, double beta)  {
 
+        if (flux <= minFlux) { return 0d; }
+        if (flux >= maxFlux) { return 254d; }
+
```
robyww: why do our other stretch type not purposely have to do this code as well?

tgoldina: other algorithms compute discrete values, all of which are in the range, and return the index of a discrete value, which is closest to the data value

PR #626, opened 2018-06-14
Dm 14029 improve table parsing

loitly:
https://jira.lsstcorp.org/browse/DM-14029
k8s: https://irsawebdev9.ipac.caltech.edu/dm-14029/firefly/

Main objective is to make DataGroup into a generic tabular data model instead of an ipac-table based model.  This ticket did the followings:
- refactor code to separate ipac-table constraints from DataGroup
- repackage table related classes.  It was scattered all over the places.
- improve ipac table write performance.

Please help test to make sure everything is still working as it should.

Notes:
To run the one test that measure the performance of ipac table read/wrte (50k rows).
`$ gradle -Pcategory=edu.caltech.ipac.TestCategory\$Perf :firefly:test -x jsTest`


loitly:
@robyww @tgoldina , I've just made another commit to make use of column-base data model.  This allows DataGroup to use primitive types where applicable.  This significantly reduces in-memory usage.  Will you review it when you get a chance?  Thanks.

These commands work in both `dev` and this PR branch.
To run all the tests:
`gradle -Pcategory=edu.caltech.ipac.TestCategory\$Perf :firefly:test -x jsTest`

To run just perfTestMidSize: 
`gradle -Pcategory=edu.caltech.ipac.TestCategory\$Perf :firefly:test -x jsTest --tests *perfTestMidSize`

tgoldina:
Using https://irsawebdev9.ipac.caltech.edu/dm-14029/firefly/. Load src.fits FITS binary table attached to https://jira.lsstcorp.org/browse/DM-14227 - looks like all values in coord_ra and coord_dec columns are NONE, which is not right. Compare to the same table loaded to irsawebdev5.ipac.caltech.edu:8080/firefly

tgoldina:
Another bug shows up when creating histogram for a column containing
negative and positive numbers.
Test case: Catalog search m31, radius 500 arcsec. Create histogram for
"w4snr" - no chart (the validation of histogram data fails) - it looks like
the histogram was created with unsorted data, even though in
FireflyHistogram we pass sortInfo with the search request.

2018-06-21 15:51 GMT-07:00 Tatiana Goldina <notifications@github.com>:

loitly:
@tgoldina , the `NONE` bug is fixed.  A new version is deployed to k8s.
The histogram bug above is most likely due to a bad build.  It's not appearing in the new deployed version.  Please check again.

tgoldina:
The histogram bug was probably due to floating point operations resulting in slight differences between previous bin edge and the following bin edge. I've removed this difference.

With this change I can not reproduce this bug any more.

PR #632, opened 2018-06-29
DM-14778: fix asinh stretch

tgoldina:
https://jira.lsstcorp.org/browse/DM-14778

- Improved algorithm to calculate asinh stretch using Q parameter
- Added slider for Q parameter (with image refresh on slider change pause)
- Autofill range option (for non z-scale case) for cases when full data range can be used. This option works well for small images, but might not be appropriate for large images. 

Deployment to test: https://irsawebdev9.ipac.caltech.edu/dm-14778/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/Zscale.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/Zscale.java
@@ -291,7 +291,7 @@ public class Zscale
 	if (npix <= 0)
 	    return new FitLineRetval(1, 0.0F, 0.0F);
 	else if (npix == 1) {
-		    return new FitLineRetval(1, data[1], 0.0F);
+	    return new FitLineRetval(1, data[1], 0.0F);
 	} else
 	    xscale = 2.0 / (npix - 1);
 
@@ -384,14 +384,14 @@ public class Zscale
 		break;
 	}
 
-	/* Transform the line coefficients back to the X range [1:npix]. */
-	float zstart = (float) (z0 - dz);
-	float  zslope = (float) (dz * xscale);
-        if (Math.abs(zslope) < 0.001)
-            zslope = (float) (o_dz * xscale);
+	  /* Transform the line coefficients back to the X range [1:npix]. */
+	  float zstart = (float) (z0 - dz);
+	  float  zslope = (float) (dz * xscale);
+	  if (Math.abs(zslope) < 1e-10)
+		  zslope = (float) (o_dz * xscale);
 
 
-	return new FitLineRetval(ngoodpix, zstart, zslope);
+	  return new FitLineRetval(ngoodpix, zstart, zslope);
 }
 
 
--- a/src/firefly/js/visualize/ui/ColorBandPanel.jsx
+++ b/src/firefly/js/visualize/ui/ColorBandPanel.jsx
@@ -5,12 +5,13 @@
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import numeral from 'numeral';
+import {debounce, get} from 'lodash';
 
 import {ValidationField} from '../../ui/ValidationField.jsx';
 import {ListBoxInputField} from '../../ui/ListBoxInputField.jsx';
 import {CheckboxGroupInputField} from '../../ui/CheckboxGroupInputField.jsx';
-import {callGetColorHistogram, callGetBeta} from '../../rpc/PlotServicesJson.js';
-import {dispatchValueChange} from '../../fieldGroup/FieldGroupCntlr.js';
+import {RangeSlider} from '../../ui/RangeSlider.jsx';
+import {callGetColorHistogram} from '../../rpc/PlotServicesJson.js';
 import {encodeServerUrl} from '../../util/WebUtil.js';
 import {formatFlux} from '../VisUtil.js';
 import {getRootURL} from '../../util/BrowserUtil.js';
@@ -20,7 +21,11 @@ import {
     STRETCH_LINEAR, STRETCH_LOG, STRETCH_LOGLOG, STRETCH_EQUAL,
     STRETCH_SQUARED, STRETCH_SQRT, STRETCH_ASINH, STRETCH_POWERLAW_GAMMA} from '../RangeValues.js';
 
-
+import {getFieldGroupResults} from '../../fieldGroup/FieldGroupUtils.js';
+import {dispatchStretchChange, visRoot} from '../ImagePlotCntlr.js';
+import {getActivePlotView} from '../PlotViewUtil.js';
+import {makeSerializedRv} from './ColorDialog.jsx';
+import {toBoolean} from '../../util/WebUtil.js';
 
 
 const LABEL_WIDTH= 105;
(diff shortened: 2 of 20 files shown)


robyww:
Are you going to take out references to and computation of beta? It adds overhead and if we are not using it we should take it out.

robyww:
Under the stretch pull down.  We wanted to add a 'Z Scale asinh Stretch' option

robyww:
When is use asinh in the dialog and check `autofill for full range` stretch. Only the upper range goes to data the lower stays at % 0.

robyww:
We had talked about adding a help line at the bottom of the panel

src/firefly/js/visualize/ui/ColorBandPanel.jsx
```diff
@@ -370,16 +344,83 @@ function renderGamma(fields) {
     );
 }
 
-function renderAsinH(fields) {
-    var {zscale}= fields;
-    var range= (zscale.value==='zscale') ? renderZscale() : getUpperAndLowerFields();
+const asinhSliderMarks = {
+ 0: '0', 5: '5', 10: '10', 15: '15', 20: '20', 25: 'InF'
```
stargaser: My opinion is that the label 'Inf' should be '25'. The text above the slider says that Q = Inf is log stretch and I think that conveys to a user that large values are a log stretch. I think 25 is not close enough to a very large value for us to call this 'Inf'.

I could be talked out of this though so maybe it is better to leave this as-is and get opinions from more people.

stargaser: My opinion is that the label 'Inf' should be '25'. The text above the slider says that Q = Inf is log stretch and I think that conveys to a user that large values are a log stretch. I think 25 is not close enough to a very large value for us to call this 'Inf'.

I could be talked out of this though so maybe it is better to leave this as-is and get opinions from more people.

src/firefly/js/visualize/ui/ColorPanelReducer.js
```diff
@@ -252,18 +282,17 @@ var getFieldInit= function() {
...
 
-        beta: {
-            fieldKey: 'beta',
-            value: NaN,
-            validator: Validate.isPositiveFiniteNumber.bind(null, 'Beta'),
-            tooltip: 'an arbitrary "softness"',
-            label: 'Beta:'
+        asinhQ: {
+            fieldKey: 'asinhQ',
+            validator: Validate.floatRange.bind(null, 0, 30, 4, 'Q'),
```
stargaser: I suggest to change the 30 to 100 or even 1000. The slider prevents a user from going over 25 right now. 

This limit only applies to the UI and not to values passed through the JS API, right?

stargaser: I suggest to change the 30 to 100 or even 1000. The slider prevents a user from going over 25 right now. 

This limit only applies to the UI and not to values passed through the JS API, right?

tgoldina:
David and Trey, thank you for your comments.

I agree with David that 'Autofill for full range' gives questionable
benefits. It does work nicely with certain histograms. Finding better lower
limit would help improve the results, but at the moment I don't know how to
choose it. That's why lower percentage is set to 0 to include all data.

If we decide we don't want it, it's easy to remove.

There are no restrictions on values when passed from API. I tried to follow
astropy implementation, forgot to change the comment.

In UI, I can set InF value (above 24.8) to 1000.

Regarding Trey's comments.

Percentage 0 for lower range with autofill is intended, as explained above.

Help line is above the slider in light gray, matching histogram help. It
looks bad under the slider.

I thought I have removed all refs to beta, they must have sneaked in during
rebase. Will check and remove.

I have missed asinh z-stretch convenience item in menu. Will add.

After merge, I will change beta to asinh_q in firefly_client.

Did I miss something?

stargaser:
If when the slider selection results in Q > 24.8 at Inf, you then set it to 1000, I think that would be great.

I think "Z Scale Asinh Stretch" could replace "Z Scale Log-Log Stretch" on the front menu, if you don't want to make the menu longer. I would even argue to replace "Z Scale Log Stretch" with "Z Scale Asinh Stretch", and replace "Z Scale Log-Log Stretch" with a 1% to 80% Asinh stretch...but maybe it is just better to add options for now, and subtract others later.

xiuqin:
The menu change has bigger impact. I went through the list again.
The top three convenient stretches never change,  and boom ones changes according to the type of stretch selected in the dialog ignoring  Z-scale check-box.

I think we should leave them alone and adding the best default asinh for now, and ask a wider audience for input later. 

Xiuqin

robyww:
@tgoldina I did a rebase and a force push of this branch.  DM-15157 did a major refactor of the server side so I went ahead and made sure you changes would work without conflict. I made a copy of the branch before the rebase  dm-14778_asinh-stretch-saved.  I think I got all your changes in but you might want to check and talk to me about it.

stargaser:
@tgoldina here are some comments from discussions with @vandesai1. Generally the asinh works well. For this pull request these items in the slider box ought to be changed:

- Allow setting Q=0 in the slider (while really making it 1.e-10 since we can't exactly have zero).
- Make the resolution of the slider 0.1 instead of 0.2
- The text "Q = 0 results in linear, large Q = in log shape" is still a problem, since changing the stretch to log gives something completely different from large Q. (I, David, suggest "Q=0 for linear stretch; increase Q to make brighter features visible").

The following comments don't have to be fixed right now. We could get feedback from more users and have more discussion about the interface.

Here is a summary of other comments pertaining to asinh, which don't need to be fixed for this PR.
- The slider can be difficult to use and can be ergonomically taxing. It would be nice to be able to use the right arrow key.
- Lack of a text input box for Q makes it harder to quickly input a value the user likes.
- Depending on how one gets to asinh, Q sometimes does not have a default value, and at other times it is unclear how it is set and whether old values are used for new images.
- Why is there a Refresh button when changing the slider causes an update?

Other comments not specific to asinh which also need not be fixed for this PR but should be kept in mind.
- When hovering inside the histogram, index, Size and Mean Value are displayed. Better would be index, Number, and Bin Center.
- Instead of DataMin and DataMax, better to have spaces, Data Min and Data Max.
- Lower range and Upper range are confusing. Better would be Lower limit and Upper limit.
- When changing from % to Data, one expects the Data value to be the 1% or 99% value, but instead Min and Max are filled in.
- Changing the stretch type doesn't automatically update, nor does changing data parameters, but changing Q does. (Up to now we've always relied on the Refresh button before doing anything.)

src/firefly/java/edu/caltech/ipac/visualize/plot/Zscale.java
```diff
@@ -387,7 +387,7 @@ else if (npix == 1) {
 	/* Transform the line coefficients back to the X range [1:npix]. */
 	float zstart = (float) (z0 - dz);
 	float  zslope = (float) (dz * xscale);
-        if (Math.abs(zslope) < 0.001)
+        if (Math.abs(zslope) < 1e-7)
```
stargaser: I would feel better if we made this 1e-10.

tgoldina:
I have addressed the following comments before merging:
- Allow setting Q=0 in the slider (while really making it 1.e-10 since we can't exactly have zero).
-  Make the resolution of the slider 0.1 instead of 0.2
- The text "Q = 0 results in linear, large Q = in log shape" is still a problem, since changing the stretch to log gives something completely different from large Q. (I, David, suggest "Q=0 for linear stretch; increase Q to make brighter features visible").
- Use 1e-10 instead of 1e-7 when checking zslope in Zscale.java 
- Instead of DataMin and DataMax, better to have spaces, Data Min and Data Max

I'll make a separate ticket for other suggestions.

PR #639, opened 2018-08-06
DM-15215: support an initial scroll position for the images

robyww:
Loan an image with an initial position

_to test:_

This must be tested from python using the slate.html entry point.  I loaded the `firefly_client/demo/basic-demo` jupyter notebook.

I modified the two cells the first to start slate.html

`fc = FireflyClient('http://127.0.0.1:8080/firefly', html_file='slate.html')`

then later when I loaded an image.

`status = fc.show_fits(file_on_server=imval, plot_id="wise-cutout", title='WISE Cuteout', InitialCenterPosition='5;5;PIXEL', InitZoomLevel='5', ZoomType='STANDARD')`

The `InitialCenterPosition` should be in the middle of the screen.


cwang2016:
code looks good. just a question, if 'INITIAL_CENTER_POSITION' is for image pixel, like '5,5,PIXEL', is that treated as zero based or fits image point or else?

robyww:
It is just for PIXEL is firefly image PIXEL.  In a different pull request we can add more.  I just don't think I can it completely tested this week.

PR #640, opened 2018-08-08
DM-15179: Table: Do not send ROW_IDX and ROW_NUM columns when not requested

loitly:
https://jira.lsstcorp.org/browse/DM-15179

Do not send these 2 columns when it's not requested. 
It should still send them when all(*) is requested.

Use debugger to validate that the table sent back contains only the columns requested.


cwang2016:
Works well. Thanks.

PR #643, opened 2018-08-10
IRSA-2067: FileInfoProcessor need to take into consideration login status when caching

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-2067

To test, using irsadev Gator, download a light curve table.
-> login as someone with ztf access -> click example link -> To Time Series Tool -> Download Light Curve Table

Goto https://irsawebdev9.ipac.caltech.edu/irsa-2067/irsaviewer/timeseries
You should be logged in as the same user.
-> upload previously downloaded file -> select ztf -> upload 
-> navigate through table rows to load protected images -> logout
-> as guest user -> upload same file -> select ztf -> upload 
the images are no longer loading because they are protected and cannot be view by user without privileges

Doing the same test in IRSA ops will reveal the security breach.

@robyww, when you have time next week, it would be nice if you can take a look at the changes to ensure that it does not impact anything else.

Also, you may want to take a look at the current implementation of edu.caltech.ipac.visualize.net.AnyUrlParams#getUniqueString
It's used as a cache key as well as filename.  When favoring for a shorter filename, it deceases the uniqueness of the cache key.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/net/AnyUrlParams.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/net/AnyUrlParams.java
@@ -47,8 +47,10 @@ public class AnyUrlParams extends BaseNetParams {
             securCookie= "-"+ cookies.get(_securityCookie);
         }
 
-        String baseKey= loginName+ securCookie +"-"+ fileStr;
-        int originalHashCode = (_url.getHost()+ baseKey).hashCode();
+        // since this string is limited to MAX_LENGTH, having loginName in the baseKey is not ideal
+        // loginName can be long.  it'll be used in hashcode calculation instead.
+        String baseKey = securCookie +"-"+ fileStr;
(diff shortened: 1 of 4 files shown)


src/firefly/java/edu/caltech/ipac/visualize/net/AnyUrlParams.java
```diff
@@ -47,8 +47,10 @@ public String getUniqueString() {
             securCookie= "-"+ cookies.get(_securityCookie);
         }
 
-        String baseKey= loginName+ securCookie +"-"+ fileStr;
-        int originalHashCode = (_url.getHost()+ baseKey).hashCode();
+        // since this string is limited to MAX_LENGTH, having loginName in the baseKey is not ideal
+        // loginName can be long.  it'll be used in hashcode calculation instead.
+        String baseKey = securCookie +"-"+ fileStr;
```
robyww: Makes sense.

PR #644, opened 2018-08-10
Irsa 2051 an observation data field should be added

deannaji:
Kindly review my changes.
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-2051
Goto: https://irsawebdev9.ipac.caltech.edu/irsa-2051/applications/sofia/

Note: This ticket includes work on IFE also.

### DIFF
--- a/src/firefly/js/util/Validate.js
+++ b/src/firefly/js/util/Validate.js
@@ -88,6 +88,29 @@ export const validateEmail = function(description,valStr) {
     return retval;
 };
 
+export const validateDate = function(description, valStr){
+    let retval ={
+        valid: true,
+        message: ''
+    };
+    if(valStr){
+        try {
+            let dateObj = new Date(valStr);
+            //Extracting the "date" string from the date object, in yyyy-mm-dd format.
+            let parsedDate = dateObj.toISOString().substr(0, dateObj.toISOString().indexOf('T'));
+            if(parsedDate !== valStr){
+                retval.valid = false;
+                retval.message = description + ': must be a valid date formatted as yyyy-mm-dd';
+            }
+        }
+        catch (e) {
+            retval.valid = false;
+            retval.message = description + ': Unrecognized entry. date should follow the format yyyy-mm-dd';
+        }
+    }
+    return retval;
+};
+
 export const validateUrl = function(description,valStr) {
     var retval = {
         valid: true,
@@ -175,11 +198,16 @@ export const urlValidator = function(description) {
 export const emailValidator = function(description) {
     return (val) => validateEmail(description, val);
 };
+
+export const dateValidator = function(description) {
+    return (val) => validateDate(description, val);
+};
+
 /*--------------------------------------------------------------------------------------------------------------------*/
 
 
 
 var Validate = {
-    validateEmail, validateUrl, intRange, floatRange, isFloat, isInt,isPositiveFiniteNumber
+    validateEmail, validateUrl, intRange, floatRange, isFloat, isInt,isPositiveFiniteNumber, validateDate
 };
 export default Validate;


loitly:
General comments not related to this ticket. 
- should mention in PR that additional changes are in git:ife
- should squash the 3 commits in to one

> import {ValidationField} from '../../../../../firefly/src/firefly/js/ui/ValidationField';
use 'firefly' alias instead of relative path when referencing firefly directory.
i.e: import {ValidationField} from 'firefly/ui/ValidationField';

deannaji:
Understood, thanks!

loitly:
From UI Test:
- search fail if no dates were given
- from 9/17/2017 to 9/17/2017  returns no data.  There are data for that date.  If set 'to' to 9/28, you will see it.
- Safari: rendered badly. Date field not displayed.  Most likely not supported.

deannaji:
For the UI Test:
- The search is failing if no date supplied, it is a problem with the API.
- For the dates, the only date provided by the API now is 9/21/2017, so the only way to see data (for now) is to provide a date range includes 9/21/2017, or to set both the 'from' and 'to' to 9/21/2017.
- For the HTML5 Calendar, yes it is not supported by Safari.

ejoliet:
Please, Dean, replace the date widget with input text. Thanks.

--
Emmanuel Joliet
Office: 170
Phone:  (626) 395-1489

deannaji:
For the UI Test: 
The search is failing if no date supplied, it is a problem with the API. For the dates, the only date provided by the API now is 9/21/2017, so the only way to see data (for now) is to provide a date range includes 9/21/2017. For the HTML5 Calendar, yes it is not supported by Safari.

deannaji:
@loitly , new changes pushed, kindly review.

robyww:
Concerning, html 5 support. It does not appear that safari yet has plans to support it.  However, until we find a widget we could support both.  My sense it it would be pretty easy.  Checking that browser and setting the input type.  The html 5 calendar appears to integrate easily with little code change.  We can use the same validation either way.

src/firefly/js/ui/ValidationField.jsx
```diff
@@ -1,13 +1,19 @@
...
 import {fieldGroupConnector} from './FieldGroupConnector.jsx';
 
 
 function onChange(ev, store, fireValueChange) {
     var value = ev.target.value;
+
+    //html5 validation, using validity property object. If it is marked invalid, then no need to sent it to validator.js
+    if(! get(ev, 'target.validity.valid')) {
+        fireValueChange({value, message: get(ev, 'target.validationMessage','Invalid date entered') , valid:false});
+        return;
```
robyww: The should be no mention of date here.  `ValidationField` works for many types not just date.

src/firefly/js/util/Validate.js
```diff
@@ -100,6 +100,32 @@ export const validateUrl = function(description,valStr) {
...
+export const validateDate = function(description, valStr){
+    var retval = {
+        valid: true,
+        message: ''
+    };
+
+    if(valStr && valStr !== '') {
+        try {
+            var dateObj = new Date(valStr),
+                parsedDate = dateObj.toISOString().substr(0, dateObj.toISOString().indexOf('T'))//Extracting the "date" string from the date object, in yyyy-mm-dd format.
```
robyww: Don't use commas. This should be two statements.  Add a semicolon. ESlint should catch this. 
Do you have eslint set up in you IntelliJ?

It there are reason why you did not use the validate NPM package here? did it not do the job?

src/firefly/js/util/Validate.js
```diff
@@ -100,6 +100,32 @@ export const validateUrl = function(description,valStr) {
     return retval;
 };
 
+export const validateDate = function(description, valStr){
+    var retval = {
+        valid: true,
```
robyww: use `const` and `let` instead of `var`. This file predates our ES6 conversion but any new code should use `let` instead of `var`.

src/firefly/js/util/Validate.js
```diff
@@ -175,11 +201,16 @@ export const urlValidator = function(description) {
 export const emailValidator = function(description) {
     return (val) => validateEmail(description, val);
 };
+
+export const dateValidator = function (description) {
```
robyww: good.

src/firefly/js/util/Validate.js
deannaji: - Fixed.
- I don't have the ESlint activated. I will do that.
- The main problem with NPM validate is its not catching entries like for example '02/31/2018'
So, I'm simply throwing the entered date into a JS date obj. and check it after, if the value of the date object is valid or not.

src/firefly/js/util/Validate.js
deannaji: Done!

src/firefly/js/util/Validate.js
deannaji: Thank you Trey for your feedback!

src/firefly/js/util/Validate.js
robyww: Great.  Glad you have a good reason.  Setting up ESlint is important. Any coding standard we do have are defined by how we set up ESlint.

src/firefly/js/ui/ValidationField.jsx
```diff
@@ -1,13 +1,19 @@
...
+import {has, get} from 'lodash';
 import {InputFieldView} from './InputFieldView.jsx';
 import {fieldGroupConnector} from './FieldGroupConnector.jsx';
 
 
 function onChange(ev, store, fireValueChange) {
     var value = ev.target.value;
+
+    //html5 validation, using validity property object. If it is marked invalid, then no need to sent it to validator.js
+    if(! get(ev, 'target.validity.valid')) {
```
robyww: One more question. I thought I asked this before but must have forgot.  Why is this `if` block of code even necessary? It the validation flow now working right? I think it is fine for other types of input data, what if different about date?

deannaji: This if is checking if the input field is marked invalid by html5 validation, if so then no need to curry on with our validator, and simply return an invalid message back to the user.

robyww: ok, makes sense. thanks.

robyww:
dont' forget to `rebase -i` and squash the commits.

src/firefly/js/ui/ValidationField.jsx
robyww: So maybe I need to understand more.  We have to have validity as the state of the field or we can validate a whole FieldGroup.  i guess I am going to have to really study this then to understand the impact.  Please don't merge until I can.

deannaji: I used this link for the validity property:
https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation

robyww: That is not wrong at all, it is just we need to validation state to be in our store.  We cannot be dependent on UI for state.

deannaji: I see. So far I'm not using it at all, since I'm using a plain input field (with no type='date'). I was using the validity object when I was using the html5 calendar widget. So I can remove it.

robyww: ok, i guess you should.  I would like it all to work right with the html calendar widget eventually, but I going to need a chunk of time to think though it which I can't do while I am at LSST2018 conf.

deannaji: Are you ok with the validateDate function in validator.js. If so then I will remove the code on the validationField.jsx (the html5 validity stuff), and curry on.

robyww: That's fine.

PR #656, opened 2018-09-06
DM-15219: chart bug fixes, found in NED 

tgoldina:
https://jira.lsstcorp.org/browse/DM-15219
https://jira.lsstcorp.org/browse/DM-14886
Test deployment: https://irsawebdev9.ipac.caltech.edu/dm-15219/firefly/

This PR addresses multiple chart and filter related bugs, found in NED:
- Fixed filter bug, described in DM-15148: error when chart is filtered after another table filter is applied (for example on Units)
- Quoted columns should be considered valid in chart options 
- Error type in chart options should match the data
- When error is empty, no error bar should not be displayed (0 length bar was displayed before) 
- FilterEditor should allow filters update without clearing expression-based filters.
- Allow filter update by changing free-form filter in FilterEditor

To test, use http://localhost:8080/firefly/demo/ffapi-sed.html
- apply filters from table and then chart and vice versa
- verify that chart options show Asym errors for the first chart, and Sym errors for the second
- change the chart via options (apply grid, for example) - there should be no validation errors in the pre-populated fields, like "Frequency".
- check that points without uncertainty appear as dots
- try modifying or removing filters in FilterEditor

### DIFF
--- a/src/firefly/html/demo/ffapi-sed.html
+++ b/src/firefly/html/demo/ffapi-sed.html
@@ -50,10 +50,7 @@
                 null,  // alt_source
                 {filters: '"NED Units" like \'Jy\''} // options
             );
-
-            // while waiting for NED to provide us a numeric column for upper limit and uncertainty
-            // using the db expression to generate upper limit column:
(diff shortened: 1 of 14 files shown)


tgoldina:
The last commit addresses https://jira.lsstcorp.org/browse/DM-14886 (after filtering a decimated data set from an image, the following filters fail).

PR #658, opened 2018-09-10
DM-15693: bug applying field reducer function

tgoldina:
https://jira.lsstcorp.org/browse/DM-15693

- Fixed the bug preventing the full update of radio group input field options via reducer function
- Updated ExampleDialog to show how field state can be modified on an action.

To test
https://irsawebdev9.ipac.caltech.edu/dm-15693/firefly/firefly-dev.html,
Change the options in Master Radio Group, observe the changes in Dependent Radio Group.

### DIFF
--- a/src/firefly/js/fieldGroup/FieldGroupCntlr.js
+++ b/src/firefly/js/fieldGroup/FieldGroupCntlr.js
@@ -228,12 +228,12 @@ function multiValueChangeActionCreator(rawAction) {
 
 
 export function* watchForRelatedActions(params, dispatcher, getState ) {
-    var action=  yield take();
-    var state;
+    let action=  yield take();
+    let state;
     while(true) {
         state= getState()[FIELD_GROUP_KEY];
         Object.keys(state).forEach((groupKey)=> {
-            var fg= state[groupKey];
+            const fg= state[groupKey];
             if (fg.mounted && fg.actionTypes.includes(action.type)) {
                 dispatcher({type:RELATED_ACTION, payload:{fieldGroup:fg, originalAction:action}});
             }
@@ -262,7 +262,7 @@ function reducer(state={}, action={}) {
 
     if (!action.payload || !action.type) return state;
 
-    var retState= state;
+    let retState= state;
     switch (action.type) {
         case INIT_FIELD_GROUP  :
             retState= initFieldGroup(state, action);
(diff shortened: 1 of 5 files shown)


src/firefly/js/fieldGroup/FieldGroupCntlr.js
```diff
@@ -386,8 +386,7 @@ const updateFieldGroupMount= function(state,action) {
 const fireFieldsReducer= function(fg, action) {
     // return  fg.reducerFunc ? fg.reducerFunc(revalidateFields(fg.fields), action) : fg.fields;
     if (fg.reducerFunc ) {
-        const newFields= fg.reducerFunc(revalidateFields(fg.fields), action);
-        return smartMerge(fg.fields,newFields);
+        return fg.reducerFunc(revalidateFields(fg.fields), action);
```
loitly: With this change, we are reliant on the reducerFunc to do the right thing, which returns the original `fields` if nothing has changed, or returns a new Immutable-styled updated `fields` object where there are changes.
That said, I found couple of places where this may break the existing behavior.
In firefly:
 src/firefly/js/ui/CatalogSearchMethodType.jsx#searchMethodTypeReducer:437
This function returns new object regardless whether or not there were changes.  

In irsa-ife:
src/irsa/js/finderchart/FinderchartDownloadDialog.jsx#FinderChartOptionDLReducer:380
This function returns new object regardless whether or not there were changes.

loitly:
The changes make sense.  However, I spotted a couple of places where reducerFunc where returning new object regardless whether or not changes were made.  The old code would detect this, and return the original fields.  This change will return the 'new' fields even though nothing has changed.
This may cause a redundant render.  I am only basing this off reading the code.  What do you think?

robyww:
It should not return a new object if there are no changes.

tgoldina:
I've updated fireFieldsReducer function to use old field state object unless there are changes in it. (Using lodash isEqual to compare old and new state for each field.)

src/firefly/js/fieldGroup/FieldGroupUtils.js
```diff
@@ -173,19 +173,24 @@ const bindToStore= function(groupKey, stateUpdaterFunc) {
...
+    const newfields= {};   //clone(fields);
+    let hasChanged = false;
     Object.keys(fields).forEach( (key) => {
         const f= fields[key];
         if (f.validator && !isFunction(f.value) && f.value && !f.value.then) {
             // const {valid,message} = f.validator(f.value);
-            newfields[key]= smartMerge(f,f.validator(f.value));
             // newfields[key]= (valid!==f.valid || message!==f.message) ? clone(f,{valid,message}) : f;
+            newfields[key]= smartMerge(f,f.validator(f.value));
+            if (newfields[key] !== f) {hasChanged = true; }
```
robyww: picky: why use brackets?

src/firefly/js/fieldGroup/FieldGroupCntlr.js
```diff
@@ -383,18 +382,44 @@ const updateFieldGroupMount= function(state,action) {
...
+}
 
-const valueChange= function(state,action) {
+/**
+ * Returns old fields unless some attribute has changed.
+ * @param oldFields
+ * @param newFields
+ * @returns {*}
+ */
+function smartReplace(oldFields, newFields) {
```
robyww: I looked over `smartReplace ` closely. It looks good.

PR #659, opened 2018-09-17
DM-15145: convert 'not connected' message to an icon and make it optional

loitly:
https://jira.lsstcorp.org/browse/DM-15145
k8s link: https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/

The original design was too intrusive.  In this rewrite, it will look for a div with a `fireflyWarn` ID to attach the warning to.  If one does not exists, it will do nothing.
This way, users of Firefly can decide how/where to display the warning and whether they want to display it at all.
On our existing Firefly's apps, it will appear as an icon on the upper-right corner of the screen.  
For testing purposes, I enabled the warning in the coverage-test demo page as well. 
You'll find it here:  https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-coverage-test.html

An easy way to test is to turn off WIFI or remove LAN cable.


loitly:
@xiuqin @ejoliet, can you take a look to see how you feel about the new offline indicator.  is it now not noticeable as Tatiana commented?
https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/
https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-coverage-test.html

loitly:
Base on reviewer's comments, I've added several variations of the warning message.
Using a custom 'data-decor' attribute, user of firefly can specify the version the message.
Currently, there are 3 versions:  small, medium, and full.  See src/firefly/js/ui/LostConnection.jsx:28 for more details.

Firefly app is using the full-centered version.  @ejoliet @xiuqin , you two should take a look to see if it makes sense.
Below are the variations of the message in different use-case.
full-centered:  https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/
small-left: https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-coverage-test.html
medium-left: https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-2images.html
full-centered:  https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-table-test.html

tgoldina:
Loi, thank you for making the behavior flexible.

I can add smaller or larger icon anywhere on the page and style it as I like: 
external page

At the same time, lost connection warning is quite prominent in the app:
app

We just have to remember to document this behavior in our API docs, otherwise no one will use it.

ejoliet:
On chrome (latest), i've put it in offline mode, but i don't see anything. Is it working un any browser, any version?

xiuqin:
All samples work for me in Chrome onMac. It took a few seconds for the warning to show up after I disconnect network.

loitly:
> Is it working un any browser, any version?

It should.  I've tried it on all 3 of my browsers.  Like Xiuqin said, it could take up to 10 sec for it to trigger.

ejoliet:
Ok, yes, i can see it now. Dev offline mode didn't work, i had to test without cable.

ejoliet:
BTW, the small one is tricky to see, not very telling, specially if you have a big screen and not really paying attention to the right hand corner. I would really prefer to see a warning whenever a user trigger an action anywhere anytime on anything instead of polling. But i guess this is radically different.

loitly:
@robyww , added another commit to support multiple warnings applicable throughout the the session of firefly and not just at document load.

For usage, see /hydra/cm/firefly/src/firefly/js/ui/LostConnection.jsx
https://irsawebdev9.ipac.caltech.edu/dm-15145/firefly/demo/ffapi-table-test.html
has usage of multiple warnings.

loitly:
> Maybe I am being picky but fireflyWarn now seems to easily mistaken. I think it should be able little longer, such as fireflyServerDownWarn.

I was thinking the same at the start of this ticket.  But, I did not want to suggest a different name, since it was specifically mentioned in the ticket.

@xiuqin , if it's not a big deal, maybe we can name it something else.. like 'fireflyLostConnWarn'.

xiuqin:
Using the long name is better. It was my first attempt to give a name. 
I support the longer one “fireflyLostConnWarn”.

Xiuqin

PR #660, opened 2018-09-20
DM-15791: Creates framework for an async SearchProcessor

loitly:
https://jira.lsstcorp.org/browse/DM-15791

- create a rough implementation of AsyncTapQuery
- refactor EmbeddedDbProcessor
- refactor HttpServices

Please review code.  There's no UI changes.  Pay closer attention to new `AsyncSearchProcessor`, `AsyncJob`, and the roughly implemented `AsyncTapQuery`

Added a simple test:  AsyncTapQueryTest.java
` gradle :fi:test --tests *AsyncTapQueryTest`

Note:  `testError` is ignored.  async TAP error returns a voTable with only INFO tags.  We are unable to parse the returned table.  Not sure if starlink is capable of parsing it.  Also, I think IRSA is returning a 'bad' error doc.  If so, either they fix it, or we'll need to have special cases for it.

Small changes to irsa-ife:  https://github.com/IPAC-SW/irsa-ife/pull/8

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/network/HttpServices.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/network/HttpServices.java
@@ -19,6 +19,7 @@ import org.apache.commons.httpclient.methods.multipart.StringPart;
 import org.apache.commons.httpclient.params.HttpConnectionManagerParams;
 import org.apache.commons.httpclient.cookie.CookiePolicy;
 import org.apache.http.HttpHeaders;
+import org.apache.xpath.operations.Bool;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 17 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/network/HttpServices.java
```diff
@@ -210,50 +198,88 @@ public static int executeMethod(HttpMethod method, HttpServiceInput input, Outpu
...
+    public static class Status {
+        private String errMsg;
+        private int statusCode;
+
+        public Status(int statusCode, String errMsg) {
+            this.errMsg = errMsg;
+            this.statusCode = statusCode;
+        }
+
+        public boolean isError() { return statusCode < 200 || statusCode >= 300; }
```
tgoldina: I like that we now pass Status. I think it's OK to treat 3xx codes (redirects) as an error. If HttpServices is not set to follow redirect, it's up to the user to deal with it.

PR #662, opened 2018-09-25
DM-15490 Modify behavior of 2D-histogram (heatmap) fallback

tgoldina:
Ticket: https://jira.lsstcorp.org/browse/DM-15490
Test deployment: https://irsawebdev9.ipac.caltech.edu/dm-15490/firefly/

- Modified Firefly behavior of scatter charts to move between scatter and heatmap representation depending on whether the number of table rows exceeds the threshold stored in charts.maxRowsForScatter application option. 
- Introduced custom (non-plotly) color scales, see Colorscale.js so that the color of the first 8 traces does not change dramatically, when switching between scatter and heatmap. 
- Fixed a  bug, preventing TableFunctionProcessor from refetching the table data after session timeout has passed.
- Fixed bugs related to trace removal.


loitly:
existing bug: (uncaught error)
- bring up a scatter plot
- goto option -> Add New Chart
- x: ra, y: dec, y-error: 1
- switch to 'Add New Trace' radio option
app crashed.  only way to recover is to reload.

Would you like me to create ticket?

PR #664, opened 2018-09-28
IRSA-2155: added IRAS/IRIS to image search panel

ejoliet:
This PR is adding IRAS/IRIS to the image search panel.

Please test here with 'm31'
https://irsawebdev9.ipac.caltech.edu/irsa-2155-iris/irsaviewer/

Thanks.


lznakano:
I tested m31.  For Spitzer, only SEIP has images and rest has none.  I compared with the same search in dev and result is the same.  I did not see any issue at all.  You can merge.

PR #666, opened 2018-10-03
Add table column filter using a multiple choice UI

loitly:
The option is available when the number of unique values is less than 15.
This feature is only available for columns with string, char, int, long, or boolean data type.

ticket : https://jira.ipac.caltech.edu/browse/IRSA-1940
test deploy: https://irsawebdev9.ipac.caltech.edu/irsa-1940/firefly/

Also in this PR:
- fixed https://jira.lsstcorp.org/browse/DM-15765
- fixed https://jira.ipac.caltech.edu/browse/IRSA-2151
- added support for a 'drop-down' UI
- fixed `FmtDisp`
- ensure onFireflyLoad is not called more than once

To Test, load catalog smaller catalog so there's more chance of columns with a choice UI.

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -270,8 +270,9 @@
 .btn-close:after {
     content: '×'; /* UTF-8 symbol */
     left: 50%;
+    top: 50%;
     position: absolute;
-    transform: translate(-50%, 0);
+    transform: translate(-50%, -50%);
     font-weight: bold;
 }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/BackgroundStatus.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/BackgroundStatus.java
@@ -24,7 +24,6 @@ import java.util.*;
  * @author Trey Roby
  */
 public class BackgroundStatus implements Serializable {
-    public static final String BG_STATUS_ACTION = "background.bgStatus";
     public static final String PARAM_SEP = "<<BGSEP>>";
     public static final String KW_VAL_SEP = "==>>";
     public final static String NO_ID = "WARNING:_UNKNOWN_PACKAGE_ID";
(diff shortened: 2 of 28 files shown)


xiuqin:
[coverage-4corners-notworking.tbl.txt](https://github.com/Caltech-IPAC/firefly/files/2448540/coverage-4corners-notworking.tbl.txt)
Bug report: 

1. uploaded the attached table
2. filter on column configuration, 
3. select first one of the two values, apply, 
4. select second one, apply again. 
Error occured. 

Suggestions:

1.  If there is only one value, I don't think there is need for filter.
2. The font for the filter list seems bigger than the font for the table display. It does not look comfortable to me.

loitly:
@xiuqin , thank you for the review.  I fixed the bugs you reported.
In regard to `one value`, I thought that it was useful still to show single values, because in the case where there are thousand of rows, it'll tell the user that there's only one value available without having to page through all of the rows.

I also encounter discrepancies in handling `null` values.  It's not easy to fit `null` value into the `where cname IN (1,2,3,null)` syntax.  You have to say `where cname IN (1,2,3) or cname is NULL`.
But, we don't allow `or` condition in column filter.  For now, I remove `null` values from the choices as a simple solution.  You can still filter the `null` values by using the text box, i.e.  `is not null`

Also, `15` is an arbitrary number.  I can change it to anything.  It used to be 10 I think.  You should think about if `15` a good cut off point.

robyww:
_UI comment/question_: If there is only one of a value in column, as in @xiuqin example (target_name), should we show the the multiple choice UI. Does it add any value?

xiuqin:
I had the same question and Loi answered it. I agree with him that it does give more information to user if the table is very big.

loitly:
done.
- decrease arrow icon from 8px to 6px
- close drop-down on outside click events
- change text from 'Apply' to 'filter'

robyww:
Does it close when they click filter?

loitly:
> Does it close when they click filter?

It was already doing that before.  If it's not closing, then it's a bug.  I'll keep an eye on it now that it's merged.

PR #667, opened 2018-10-03
DM-16012: table load failures

tgoldina:
Fixed failing table loads when a table contains quoted column name or a control characters in a cell.

https://jira.lsstcorp.org/browse/DM-16012 Firefly fails to load a table with '\n' and '\r' characters in a cell
https://jira.lsstcorp.org/browse/DM-15982 Firefly fails to load a table with a quoted column

Test deploy: https://irsawebdev9.ipac.caltech.edu/dm-16012/firefly

The solution was to remove quotes from column names and to escape/unescape control characters when writing and reading IPAC table to/from file. 

HSQLDB does not accept quotes in table column names. Even though DataGroup can handle quoted columns, we are removing quotes from column name in  DataType constructor rather than during DataGroup ingest into db, because it could result in inconsistent column names before and after ingest, which is undesirable.

The loading can be verified with a simple tap browser:
https://irsawebdev9.ipac.caltech.edu/dm-16012/firefly/demo/ffapi-tap-test.html
(Compare with nightly dev build)

It was previously failing on http://gea.esac.esa.int/tap-server/tap and http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/tap

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/DataType.java
+++ b/src/firefly/java/edu/caltech/ipac/table/DataType.java
@@ -11,6 +11,9 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 
+import static org.apache.commons.lang.StringEscapeUtils.escapeJava;
+import static org.apache.commons.lang.StringEscapeUtils.unescapeJava;
+
 public class DataType implements Serializable, Cloneable {
 
     public enum Visibility {show, hide, hidden};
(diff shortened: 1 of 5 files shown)


src/firefly/java/edu/caltech/ipac/table/DataType.java
```diff
@@ -286,6 +303,16 @@ public boolean isKnownType() {
         );
     }
 
+    /**
+     * Convert a string into an object
+     * Java unescaping is applied to all strings
+     * @param s string representation of an object
+     * @return an object
+     */
+    public Object convertStringToDataUnescaped(String s) {
```
loitly: IMO: The naming is a bit confusing.  The `Unescaped` is an internal process that may be confusing for people who do not know that you use `convertStringToDataUnescaped` with `formatDataEscaped`.

I suggest overloading the `formatData` and `convertStringToData` with an additional flag `useEscape` or something similar to indicate that escape/unescape is applied to the data.. i.e

```
public String formatData(Object value, boolean useEscape)
public Object convertStringToData(String s, boolean useEscape)
```

PR #668, opened 2018-10-08
DM-15826 Improvements on reading VOTable in Firefly 

cwang2016:
This PR includes Improvement on reading VOTable by processing more elements and the attributes from TABLE and FIELD elements  in votable file, 

- Using Starlink votable-aware DOM parser to traverse the xml elements in the votable file and collect more details from TABLE element including info. from TABLE's attributes (ID, name, ucd, utype, ref) and from TABLE's children elements (DESCRIPTION, LINK, GROUP, PARAM, INFO, LINK, FIELD). 
- collect more details from FIELD element including info. from FIELD's attributes such as  ID, name, datatype, unit, precision, width, ref, ucd, utype  and from FIELD's children elements including DESCRIPTION, LINK, and VALUES. 
- The JAVA class implementation regarding the table (DataGroup) and table column (DatType)  are updated accordingly. 

Test:
Run firefly and upload votable example (ex. firefly_test_data/FileUpload-samples/VOTable/tabledata/multiTables_Ned_test.xml)  from upload dialog.

Result of  example '/multiTables_Ned_test.xml': 
- 826 tables are shown in the summary table
- Highlight the first table with index 0, then table name from attribute ‘name’,  table description from child element DESCRIPTION, and table info from child element INFO, PARAM, LINK, GROUP  of that TABLE are shown at the right side of the summary table.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -99,7 +99,7 @@ public class AsyncTapQuery extends AsyncSearchProcessor {
                 DataGroup errorTbl = results[0];
                 return errorTbl.getAttribute("QUERY STATUS");
             } catch (IOException e) {
-                e.printStackTrace();
+                logger.error(e);
             }
             return "";
         }
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -14,6 +14,7 @@ import edu.caltech.ipac.util.StringUtils;
 import org.apache.commons.httpclient.Header;
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 
 @SearchProcessorImpl(id = AsyncTapQuery.ID, params = {
         @ParamDoc(name = "serviceUrl", desc = "base TAP url endpoint excluding '/async'"),
@@ -93,9 +94,14 @@ public class AsyncTapQuery extends AsyncSearchProcessor {
         }
 
         public String getErrorMsg() {
-            DataGroup[] results = VoTableReader.voToDataGroups(baseJobUrl + "/error");
-            DataGroup errorTbl = results[0];
-            return errorTbl.getAttribute("QUERY STATUS");
+            try {
+                DataGroup[] results = VoTableReader.voToDataGroups(baseJobUrl + "/error");
+                DataGroup errorTbl = results[0];
+                return errorTbl.getAttribute("QUERY STATUS");
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+            return "";
         }
 
         public long getTimeout() {
--- a/src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
@@ -9,6 +9,7 @@ import edu.caltech.ipac.table.DataType;
 import edu.caltech.ipac.table.IpacTableUtil;
 import edu.caltech.ipac.table.GroupInfo;
 import edu.caltech.ipac.table.LinkInfo;
+import edu.caltech.ipac.table.TableMeta;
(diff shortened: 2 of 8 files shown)


src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
```diff
@@ -72,6 +83,288 @@
...
+        VOElement[] voLinkAry = fieldEl.getChildrenByName("LINK");
+        VOElement[] links = new LinkElement[voLinkAry.length];
+        System.arraycopy(voLinkAry, 0, links, 0, voLinkAry.length);
+
+        return links;
+    }
+
+    // convert <GROUP>s under <TABLE> to a list of GroupInfo and add it to the associatede table (a DataGroup object)
+    private static List<GroupInfo> makeGroupInfosFromTable(TableElement tableEl, DataGroup dg) {
+        VOElement[] groupAry = getGroupsFromTable(tableEl);
```
loitly: This is a good example of an encapsulation entry-point.  GroupInfo is a known entity to Firefly and collecting GroupInfos from a votable is an identifiable task.
I would argue that `tableEl.getChildrenByName("GROUP");` is more informative then `getGroupsFromTable(tableEl);`.

ejoliet:
A general question: is the reader reads in the data in memory or only parse the header to get the ucds, etc.?

cwang2016:
regarding the question from @ejoliet, starlink provides the following storage policy for read in the votable file, 
PREFER_MEMORY: Stores table data in memory
PREFER_DISK: Generally attempts to store data in a temporary disk file
ADAPTIVE: Stores table data in memory for relatively small tables, and in a temporary disk file for larger ones.
SIDEWAYS: Generally attempts to store data in temporary disk files using column-oriented storage.
DISCARD: Metadata is retained, but the rows are simply thrown away.

"ADAPTIVE" is the default setting in Starlink.  For Firefly file upload & analysis, I intended to use the policy "DISCARD", however, if that,  there is no way to get the number of rows which is reported in the upload summary unless the TABLE element has attribute 'nrows'. So "ADAPTIVE" is currently used for either file upload and table search.

src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
```diff
@@ -672,10 +697,10 @@ private static DataGroup convertToDataGroup(TableElement tableEl, StarTable tabl
...
         if (tableEl != null) {
             // attribute ID, ref, ucd, utype from TABLE
-            dg.setID((tableEl != null) ? getElementAttribute(tableEl, ID) : null);
-            dg.setRef((tableEl != null) ? getElementAttribute(tableEl, REF) : null);
-            dg.setUCD((tableEl != null) ? getElementAttribute(tableEl, UCD) : null);
-            dg.setUType((tableEl != null) ? getElementAttribute(tableEl, UTYPE) : null);
+            dg.addAttribute(ID,  getElementAttribute(tableEl, ID));
+            dg.addAttribute(REF, getElementAttribute(tableEl, REF));
+            dg.addAttribute(UCD, getElementAttribute(tableEl, UCD));
+            dg.addAttribute(UTYPE, getElementAttribute(tableEl, UTYPE));
```
loitly: `dg.addAttribute(UCD` should be `dg.addAttribute(TableMeta.UCD`
because we do not want to couple our table model with votable.  FITS table may also have UCD and you don't want to have your code looks like this
`dg.addAttribute(VoTableReader.UCD, ucd)`
Of course, this comment applies to all of the above constants including the DESC below.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -93,9 +94,14 @@ public Phase getPhase() {
...
         public String getErrorMsg() {
-            DataGroup[] results = VoTableReader.voToDataGroups(baseJobUrl + "/error");
-            DataGroup errorTbl = results[0];
-            return errorTbl.getAttribute("QUERY STATUS");
+            try {
+                DataGroup[] results = VoTableReader.voToDataGroups(baseJobUrl + "/error");
+                DataGroup errorTbl = results[0];
+                return errorTbl.getAttribute("QUERY STATUS");
+            } catch (IOException e) {
+                e.printStackTrace();
```
loitly: Please use logger.error(e) instead

PR #670, opened 2018-10-10
IRSA-2152: Firelfy table filter fails when underscore '_' is in filter

loitly:
ticket: https://jira.ipac.caltech.edu/browse/IRSA-2152
test deployment: https://irsawebdev9.ipac.caltech.edu/irsa-2152/applications/sofia/

Ticket provide steps to reproduce bug.  Please use that to confirm that bug is fixed.


ejoliet:
Any thoughts?

PR #671, opened 2018-10-10
DM-15725: hue-preserving dialog changes

tgoldina:
https://jira.lsstcorp.org/browse/DM-15725

Test deployment: https://irsawebdev9.ipac.caltech.edu/dm-15725/firefly/

- Added help text and help anchor to hue-preserving 3-color stretch dialog
- Added scaling coefficients for each band (a log slider ranging from 0.1 to 10)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/RangeValues.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/RangeValues.java
@@ -23,10 +23,11 @@ public final class RangeValues implements Cloneable, Serializable {
     public static final int SIGMA      = 92;
 
     private static final double ASINH_Q =  Double.NaN; // if NaN, Q will be estimated based on range;
-    private static final double ASINH_STRETCH = 1;
-    private static final double SCALING_K = 1;
     private static final double GAMMA=2.0;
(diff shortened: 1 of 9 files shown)


tgoldina:
David and Trey, thank you for your comments.  The test branch is rebuilt. Please, take a look.

Hopefully, it's much easier to use, even though for big images the performance might not be great. We can improve on it later.

In the latest commit:
- intensity, bounds, etc. are calculated with the scaled fluxes - this way we can make sure that full color range will be used. (In previous version the scaling coefficients were applied when calculating colors for each band, after intensity was calculated and stretched. This was wrong.)
- layout changes: scaling coefficients on top and expanded, black point values on the bottom and closed
- stretch parameter, while still available via API, is removed from UI. Low and high intensity for asinh stretch algorithm are calculated with z-scale algorithm, if stretch value is not set (NaN). To see the calculated value in stretch dialog, you can set firefly.debugStretch=true.

The performance hit comes from the need to recalculate intensity bounds whenever any of the scaling factors changes. No need to recalculate, when only Q is changing.

robyww:
UI feels easier to work with now.

PR #672, opened 2018-10-12
DM-16135: Add long_string type into DataType

loitly:
https://jira.lsstcorp.org/browse/DM-16135

Please test using VOTable with array types.


cwang2016:
review completes. work well.

PR #674, opened 2018-10-17
IRSA-2165: adding help url in icon next to sub project label

ejoliet:
This PR should be adding an icon next tot the label of the group checkbox representing a sub-project in the image search panel. 
With more time after release we can always move the code and make a proper HelpInfoText or similar but it is probably ok for December release.
Nevertheless, this is the first attempt with one <info> icon found here: https://icons8.com/icon/set/info/all

If you have a better icon, let me know. 

Please review the code and test it so it doesn't break any other CSS where collapsible panel is used.

Thanks!


ejoliet:
yes, i forgot, please test here: https://irsawebdev9.ipac.caltech.edu/irsa-2165/irsaviewer/

robyww:
external-link 2
external-link 1

src/firefly/js/ui/ImageSelect.jsx
```diff
@@ -16,7 +16,7 @@ import {dispatchComponentStateChange} from '../core/ComponentCntlr.js';
 import {updateSet} from '../util/WebUtil.js';
 
 import './ImageSelect.css';
-
+import infoIcon from 'html/images/info-16x16.png';
```
loitly: The icon is fine, except it too small or badly re-rendered.  I can't make out the `i` in the icon.

src/firefly/js/ui/ImageSelect.jsx
```diff
@@ -320,6 +323,12 @@ function Header({project, multiSelect}) {
                 labelWidth={35}
                 wrapperStyle={{whiteSpace: 'normal'}}
             />
+            </div>
+            <div style={{marginLeft: -5}}><div>
+                <a target='_blank' href={href}>
+                    <img
+                     src={infoIcon}/></a>
+            </div></div>
```
loitly: You may want to check to see if `href` has a value before rendering.  I assume there may be some dataset without info link.

src/firefly/js/ui/panel/CollapsiblePanel.css
```diff
@@ -6,7 +6,7 @@
   cursor: pointer;
   border: none;
   background: inherit;
-  display: block;
+  display: flex;
```
loitly: You may want to build Sofia to make sure this does not affect collapsible panel used else where.

gpdf:
Comments on the Kubernetes build that are directly related to this PR:

1. The i-in-a-circle icon is informative and acceptable.  It looks a tiny bit fuzzy to me; I think it does in fact use some grey anti-aliasing, visible when I zoom way in on it.  But this is not an obstacle to releasing it.
1. The info icon should not toggle the disclosure state.  I would suggest limiting that behavior to the disclosure triangle itself.  
1. I would suggest that the mouse pointer be a standard arrow when over the disclosure triangle or the checkboxes, a text-selection cursor when over the dataset names (just as it is, e.g., when over "extragalactic" in the filtering pane), and a hand when it is over the info link.  Using the arrow pointer over disclosure triangles mimics what is done in Finder and is consistent with the Apple HIG, where pointing hand is reserved for links.  The pointing hand should in any event not be shown when the mouse is over blank space in the selection pane.  (These are all relevant to this PR because consistent use of the pointer icons strengthens the user's understanding of what will happen when clicking on the "(i)".)

Other comments:

1. Selection of a dataset or a band within a dataset adds "red text" to a field in the UI under the "Selection:" label.  When the first such piece of text is added the UI opens up space for it, and closes up that space if the last selection is removed.  This causes the rest of the pane to move vertically and I think this is undesirable and surprising.  If someone is trying to check several boxes quickly they will find this less natural if the checkboxes move out from under them.  Can we make the "red text" field a fixed size, since it is in any event limited to a single line?
1. Don't forget, for the future, about IRSA-2179 (tri-state checkboxes for the datasets).

src/firefly/js/ui/panel/CollapsiblePanel.css
ejoliet: Oh, great, yes, i will test SOFIA. Thanks! (Is FinderChart using also collapsible panel too?)

ejoliet:
Thanks all the comments and feedback. I've tested SOFIA collapsible panel and is it's ok. I think the alignment is fixed with Loi's suggestion.

I won't fix anything but item 2. (toggle when clicking on info).
The others comments requires more CCB AND discussion AND involves refactoring and more effort (the way the component is built relies on checkbox behaviour (trigger and css) and so the combination of isOpen(), isToggle(), and cursor CSS etc...) that i won't start/continue changing it.

Please, feel free to make another ticket(s) for future IV releases.

I will push the last commit with those fixes and ask CCB final approval on the change to merge.
Thanks all.

ejoliet:
Update with fix here: https://irsawebdev9.ipac.caltech.edu/irsa-2165/irsaviewer/

Other icon mentioned here: https://irsawebdev9.ipac.caltech.edu/irsa-2165-2/irsaviewer/

PR #677, opened 2018-10-22
DM-16036: Persist added information in DataGroup to database then send them to client

loitly:
https://jira.lsstcorp.org/browse/DM-16036

- Added PARAMs and PARAMrefs to GroupInfo from VoTable.
- Preserve the additional meta during its data flow: VoTable -> DataGroup -> database -> DataGroup -> JSON string -> client
- Added VotableTest for unit testing
- update jsdocs to include the new meta info, `tables-typedefs.jsdoc`

Cleanup:
- remove TableDef from main SearchProcessor flow.  It is now renamed to IpacTableDef, since it's only used for IPAC table.
- delete a couple of files related to ResultSet -> DataGroup translation.  Added a simple, generic conversion routine instead.

NOTE:  Some of the meta did make it to the client when using catalog upload or via 'Upload' menu due to ticket DM-16273.
 
Testing: 
https://irsawebdev9.ipac.caltech.edu/dm-16036/firefly/
- Please help with regression test.  The cleanup works that may break existing code.
- As far as the new meta info, there's no UI added.  It's best to look at the test to see if it adequately cover the requirement.


src/firefly/java/edu/caltech/ipac/util/StringUtils.java
```diff
@@ -63,6 +63,15 @@
         }
     }
 
+    /**
+     * applies the consumer's logic if v is not null or is an empty string
+     * @param v a value to check against
+     * @param f the consumer function to apply if v is not empty
+     */
+    public static <T> void applyIfNotEmpty(T v, Consumer<T> f){
```
tgoldina: This is very useful function! I guess it takes us to move  to Java10 to start using Java 8 features.

cwang2016:
Excellent update for table handling! Expect to see the full detail of votable sent to the client site next.

src/firefly/java/edu/caltech/ipac/util/StringUtils.java
ejoliet: I'm curious, what is exactly the newest feature here compare to jdk7?

loitly: generics is old, since jdk 1.5.  but with lambda in jdk 1.8, it allows you to write a single-method interface like this:

`  applyIfNotEmpty(something.getName(), v -> aMap.put("name", v));`

PR #678, opened 2018-10-23
IRSA-2154 add herschel hhli images search

ejoliet:
This branch added the Herschel datasets and take care of the HHLI cubes making use of a column unused previously 'dataType'. For HHLI is set to 'cube' allowing to display cubes (avoiding cutout).

Luisa will test it but @loitly / @robyww  might want to have a look on the code if i didn't break anything. 

For testing Herschel, you might need to go to Atlas and get a target example.


ejoliet:
I realized my rebase brought other stuff that i don't understand how they ended up here. I will do a fresh branch.

PR #690, opened 2018-11-12
DM-10500: async jobs, table info dialog

tgoldina:
Ticket: https://jira.lsstcorp.org/browse/DM-10500

To test, use "TAP Searches" tab in firefly-dev.html endpoint: 
https://irsawebdev9.ipac.caltech.edu/dm-10500/firefly/firefly-dev.html?__action=layout.showDropDown&visible=true&view=TestTAPSearch

- Added error checking and handling for DALI-style errors in VOTable and errors in async tap queries
- Job description endpoint is passed to the client as a table link
- Added "Table Info" dialog, which currently displays table title and table links
- Each table link is represented by an href and a button to copy this url to the clipboard 

The default for showTableInfoButton is false, the changes should not affect IRSA UI.


ejoliet:
Looks working. I have some issues but it might be running from the conference wifi. apart from some issues, is there a property that control the table info ( (i) button) to appear or not in the table toolbar? i don't think IRSA is ready or have time to talk about designing any UI right now...

tgoldina:
@ejoliet, yes the property is 'tables.showInfoButton' - the info button should not show up in IRSA's apps, I'll fix it.

cwang2016:
the TAP search Async request works well.
Just a little bit confused about the content shown after clicking the table info button. In the beginning I expected to see some information regarding the table itself like what we had from table analysis. Later, I got some understanding about the meaning of the table info content after reading your release comment again.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -71,14 +78,29 @@ public AsyncTapJob(String baseJobUrl) {
             this.baseJobUrl = baseJobUrl;
```
cwang2016: could you please look at line 57, it is like 
HttpServices.postData(asyncTap.baseJobUrl + "/phase", new HttpServiceInput().setParam("PHASE", "RUN"), null);

please check if  "RUN" should be "EXECUTING"? or not?

tgoldina: While "RUN" is not an execution phase, it should be used to "commit the job", if its phase is "PENDING".

According section 5.2 of [TAP 1.0](http://www.ivoa.net/documents/TAP/20100327/REC-TAP-1.0.html) (section 5.1.3 of [TAP 1.1](http://www.ivoa.net/documents/TAP/20181024/PR-TAP-1.1-20181024.html#tth_sEc5.1.3)):

The client runs the job by posting to the phase URL:
```
HTTP POST http://example.com/tap /async/42/phase
PHASE=RUN
```

As for the content of Table Info dialog, it's just the beginning, there will be much more there. The design of this dialog is out of scope of this ticket, but it is logical to handle the link to the job description similarly to how we would handle other table links.

PR #694, opened 2018-11-14
IRSA-2274: export only the original meta

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-2274
test deploy: https://irsawebdev9.ipac.caltech.edu/irsa-2274/firefly/
Additional changes in firefly_test_data under the same branch name; updated test file.

This change is to preserve the original meta information from the source for later use.
One usage is when exporting.  It should only export the original meta, not any that were added by Firefly during processing.
In the future, we should have an option where the original meta information is displayed in the UI.
This behavior applies to all table, including FITS, IPAC, and VOTABLE.

Design Notes:
TableMeta now contains 2 types of meta; attributes and keywords.
`keywords` is a list of the original meta info, which may contains comments and duplicate keys.
`attributes` is a map containing all meta info, both from the source and added ones during processing.

As developers, you should continue to use `setAttribute` when you want to add meta info into the table.

To test:
- Load table from different sources; IRSA catalog, uploaded FITS table, uploaded VOTable, etc.
- Apply several operations to the table, like sort, filter, show/hide some columns
- Save the table.

There should be only the original meta info in the downloaded table.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/JsonTableUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/table/JsonTableUtil.java
@@ -6,6 +6,7 @@ package edu.caltech.ipac.table;
 import edu.caltech.ipac.firefly.data.Param;
 import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.util.StringUtils;
+import netscape.javascript.JSObject;
 import org.json.simple.JSONObject;
 
 import java.io.IOException;
@@ -70,7 +71,7 @@ public class JsonTableUtil {
         tableModel.put("type", guessType(meta));
         tableModel.put("tableData", toJsonTableData(dataGroup));
 
-        tableModel.put("tableMeta", toJsonTableMeta(meta));
+        tableModel.put("allMeta", toJsonTableMeta(meta));
 
         if (dataGroup.getLinkInfos().size() > 0) {
             tableModel.put("links", toJsonLinkInfos(dataGroup.getLinkInfos()));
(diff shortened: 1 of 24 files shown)


tgoldina:
For IPAC table, indeed, only the original attributes are preserved. Nothing added by Firefly.

For VOTable (sample_votable.xml), I see a few extra things:
```
\ID = 
\ref = 
\ucd = 
\utype = 
\
```
Should the empty values can be omitted? They were not in the original table.

loitly:
> Should the empty values can be omitted? They were not in the original table.

You are absolutely right.  I overlooked this.  Will fix.

src/firefly/java/edu/caltech/ipac/table/JsonTableUtil.java
```diff
@@ -167,20 +152,39 @@ public static JSONObject toJsonTableData(DataGroup data) {
...
+     * returns the value of the meta info for the given key
+     * @param jsonTable     tableModel in the form of a JSONObject
+     * @param key
      * @return
      */
-    public static JSONObject toJsonTableMeta(TableMeta tableDef) {
-        JSONObject tmeta = new JSONObject();
-        for (DataGroup.Attribute att : tableDef.getAttributeList()) {
-            tmeta.put(att.getKey(), att.getValue());
+    public static String getMetaFromAllMeta(JSONObject jsonTable, String key) {
```
tgoldina: The name is a bit confusing, but javadoc helps.

ejoliet:
Thanks!
I will have a look later today but when the original table was a VOTable, the ipac format is an export actually. I just wonder if it would be easire instead of parsing and postprocess it,, why not returning the original table if it's not filtered/sorted/changed when user wants to download it? Just a comment.
I'll check later and play around with IRSAViewer and SOFIA.

PR #696, opened 2018-11-20
DM-16475 Recognize ObsCore, some UCDs for doing catalog overlay

cwang2016:
This development provides functions to get table center columns by checking 
- table meta ('CatalogCoordColumns' or 'CENTER_COLUMN')
- UCD and ( 'pos.eq.ra;meta.main'/'pos.eq.dec;meta.main'  or  'pos.eq.ra'/'pos.eq.dec') 
- ObsCore columns (check the existence of mandatory columns)  or
- guessing

test: 
case for checking table meta: 
do catalog search

case for checking UCD, 
run firefly/firefly-dev.html
select tab 'TAP searches'  && TAP service URL for IRSA

case for checking ObsCore & guessing, 
firefly/firefly-dev.html
select tab 'TAP searches' && TAP service URL for CASDA

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -116,7 +116,7 @@ public class AsyncTapQuery extends AsyncSearchProcessor {
                 throw new DataAccessException("Error getting phase from "+baseJobUrl+" "+status.getErrMsg());
             }
             try {
-                return Phase.valueOf(phase.toString().replaceAll("\\p{Cntrl}", ""));
+                return Phase.valueOf(phase.toString().trim());
             } catch (Exception e) {
                 logger.error("Unknown phase \""+phase.toString()+"\" from service "+baseJobUrl);
                 return Phase.UNKNOWN;
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -116,7 +116,7 @@ public class AsyncTapQuery extends AsyncSearchProcessor {
                 throw new DataAccessException("Error getting phase from "+baseJobUrl+" "+status.getErrMsg());
             }
             try {
-                return Phase.valueOf(phase.toString());
+                return Phase.valueOf(phase.toString().replaceAll("\\p{Cntrl}", ""));
             } catch (Exception e) {
                 logger.error("Unknown phase \""+phase.toString()+"\" from service "+baseJobUrl);
                 return Phase.UNKNOWN;
--- a/src/firefly/js/util/VOAnalyzer.js
+++ b/src/firefly/js/util/VOAnalyzer.js
@@ -1,0 +1,337 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import {get, has, isEmpty} from 'lodash';
+import Enum from 'enum';
+import {getColumn, getColumnIdx} from './../tables/TableUtil.js';
+import {MetaConst} from '../data/MetaConst.js';
+import {CoordinateSys} from '../visualize/CoordSys.js';
+
+const UCDCoord = new Enum(['eq', 'ecliptic', 'galactic']);
+
+//const pObs = 'obscore:';
+const ObsTapColName = 0;
+//const UCD = 1;
+//const Utype = 2;
+const mainMeta = 'meta.main';
+const posCol = {[UCDCoord.eq.key]: {ucd: ['pos.eq.ra', 'pos.eq.dec'],
+                                    coord: CoordinateSys.EQ_J2000},
+                [UCDCoord.ecliptic.key]: {ucd: ['pos.ecliptic.lon', 'pos.ecliptic.lat'],
+                                          coord: CoordinateSys.ECL_J2000},
+                [UCDCoord.galactic.key]: {ucd: ['pos.galactic.lon', 'pos.galactic.lat']},
+                                          coord: CoordinateSys.GALACTIC};
+
+
+
+const obsCorePosColumns = ['s_ra', 's_dec'];
+
+const ObsTapMandColumns = [
+    ['dataproduct_type',  'meta.id',                    'ObsDataset.dataProductType'],
+    ['calib_level',       'meta.code;obs.calib',        'ObsDataset.calibLevel'],
+    ['obs_collection',    'meta.id',                    'DataID.collection'],
+    ['obs_id',            'meta.id',                    'DataID.observationID'],
+    ['obs_publisher_did', 'meta.ref.uri;meta.curation', 'Curation.publisherDID'],
+    ['access_url',        'meta.ref.url',               'Access.reference'],
+    ['access_format',     'meta.code.mime',             'Access.format'],
+    ['access_estsize',    'phys.size;meta.file',        'Access.size'],
+    ['target_name',       'meta.id;src',                'Target.name'],
+    ['s_ra',              'pos.eq.ra',                  'Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C1'],
+    ['s_dec',             'pos.eq.dec',                 'Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C2'],
+    ['s_fov',             'phys.angSize;instr.fov',     'Char.SpatialAxis.Coverage.Bounds.Extent.diameter'],
+    ['s_region',          'pos.outline;obs.field',      'Char.SpatialAxis.Coverage.Support.Area'],
+    ['s_resolution',      'pos.angResolution',          'Char.SpatialAxis.Resolution.Refval.value'],
+    ['s_xel1',            'meta.number',                'Char.SpatialAxis.numBins1'],
+    ['s_xel2',            'meta.number',                'Char.SpatialAxis.numBins2'],
+    ['t_min',             'time.start;obs.exposure',    'Char.TimeAxis.Coverage.Bounds.Limits.StartTime'],
+    ['t_max',             'time.end;obs.exposure',      'Char.TimeAxis.Coverage.Bounds.Limits.StopTime'],
+    ['t_exptime',         'time.duration;obs.exposure', 'Char.TimeAxis.Coverage.Support.Extent'],
+    ['t_resolution',      'time.resolution',            'Char.TimeAxis.Resolution.Refval.value'],
+    ['t_xel',             'meta.number',                'Char.TimeAxis.numBins'],
+    ['em_min',            'em.wl;stat.min',             'Char.SpectralAxis.Coverage.Bounds.Limits.LoLimit'],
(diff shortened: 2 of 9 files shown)


cwang2016:
In order to get center columns (or coordinate columns) from the table consistently, the newly updated getCenterColumns is applied to
- CatalogWatcher
- CoverageWatcher
- ChartUtil
- WisePlotRequest

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+        // only one pair
+        if (posPairs[0].length === 1 && posPairs[1].length === 1) {
+            pairs.push([posPairs[0][0], posPairs[1][0]]);
+        } else {
+            const mainColIdx = [];
+
+            // find the leftmost pair with 'main.meta'
+            for (let n = 0; n < 2; n++) {
+                mainColIdx.push(posPairs[n].findIndex((col) => {
+                    return (col.UCD && col.UCD === `${centerColUCDs[n]};${mainMeta}`);
```
tgoldina: _A UCD is a string which contains textual tokens called ‘words’, separated by semicolons(;). A word is composed of ‘atoms’, separated by periods(.)._
We should not rely on the certain order of UCD words when detecting the presence of a particular word in UCD. They can be in any order. We might want to split UCD by semicolon and then check for the presence of a particular word.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+                [UCDCoord.ecliptic.key]: {ucd: ['pos.ecliptic.lon', 'pos.ecliptic.lat'],
+                                          coord: CoordinateSys.ECL_J2000},
+                [UCDCoord.galactic.key]: {ucd: ['pos.galactic.lon', 'pos.galactic.lat']},
+                                          coord: CoordinateSys.GALACTIC};
+
+
+
+const obsCorePosColumns = ['s_ra', 's_dec'];
+
+const ObsTapMandColumns = [
```
tgoldina: Adding an explanation comment here with the link to the standard would be useful.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+import {get, has, isEmpty} from 'lodash';
+import Enum from 'enum';
+import {getColumn, getColumnIdx} from './../tables/TableUtil.js';
+import {MetaConst} from '../data/MetaConst.js';
+import {CoordinateSys} from '../visualize/CoordSys.js';
+
+const UCDCoord = new Enum(['eq', 'ecliptic', 'galactic']);
+
+//const pObs = 'obscore:';
+const ObsTapColName = 0;
```
tgoldina: Minor: it took me some time to trace the meaning of this field. It might be easier, if it's closer to ObsTapMandColumns definition or has an explanation comment.

src/firefly/js/util/VOAnalyzer.js
loitly: Minor:  Although you are referencing ObsTap, we are really checking for any ObsCore compliant table model.  They happen to be the same.  I just wanted to point out that it does not have to come from a Tap service.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -116,7 +116,7 @@ public Phase getPhase() throws DataAccessException {
                 throw new DataAccessException("Error getting phase from "+baseJobUrl+" "+status.getErrMsg());
             }
             try {
-                return Phase.valueOf(phase.toString());
+                return Phase.valueOf(phase.toString().replaceAll("\\p{Cntrl}", ""));
```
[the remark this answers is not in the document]
tgoldina: The failing case had phase ending with "\r\n", but you are right. trim() is better, because it would also remove space.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+    }
+    return tblRecog;
+}
+
+
+export function findTableCenterColumns(table) {
+   const tblRecog = makeTableRecognizer(table);
+
+   return tblRecog && tblRecog.getCenterColumns();
+}
```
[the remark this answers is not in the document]
cwang2016: per discussion with @robyww & @loitly before the implementation, @robyww prefers the file is more related to IVOA various items instead of being limited to table only. We can talk more about this later.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+                this.guessCenterColumns();
+    }
+
+
+    static newInstance(tableModel) {
+        return new TableRecognizer(tableModel);
+    }
+}
+
+const tableRecognizerMap = new Map();
```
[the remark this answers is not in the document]
cwang2016: The computation is light though the  function for getting center columns is repeatedly called while rendering the table and coverage. Since we didn't come into a firm conclusion in the discussion with @robyww & @loitly  about this, so I just kept the map. The key issue is to decide if we want to keep  the table center columns info or not, the implementation is doable if the info is to be kept.

src/firefly/js/util/VOAnalyzer.js
cwang2016: per UCD document, http://www.ivoa.net/documents/REC/UCD/UCDlist-20070402.pdf, the token's position might be fixed as first word or secondary word. I will exam the code again to make sure the 'if condition' has the right test logic.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     * get center columns pair by checking the table meta
+     * @returns {null|{lonCol: *, latCol: *, lonIdx: *, latIdx: *, csys: *}|*}
+     */
+    getCenterColumnsOnMeta() {
+        if (this.centerColumnsInfo) {
+            return this.centerColumnsInfo;
+        }
```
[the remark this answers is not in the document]
cwang2016: such check is based on the case that this function may be called independently, not relying on other method.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,337 @@
...
+
+        const {tableMeta} = this.tableModel || {};
+
+        if (!tableMeta ||
+            (!tableMeta[MetaConst.CATALOG_COORD_COLS] && !tableMeta[MetaConst.CENTER_COLUMN])) {
+            return this.centerColumnsInfo;
```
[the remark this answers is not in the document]
cwang2016: at this point, this.centerColumnInfo is null.

src/firefly/js/util/VOAnalyzer.js
cwang2016: ObsTapMandColumns is defined per ObsTAP.  Currently only the first element of each item is used as  it shows the names of the mandatory columns for ObsCore. I put the info for ObsTAP  here for the possible need of the future.

cwang2016:
VOAnalyzer.js is updated per the center column search algorithm (the flowchart is attached in https://jira.lsstcorp.org/browse/DM-16475) we discussed in the meeting

You may try test from catalog search, file upload, or firefly/firefly-dev.html & TAP Searches.

tgoldina:
Using `utype` also fails. Looks like we are loosing `utype` in columns.
```
<?xml version="1.0"?>
<VOTABLE version="1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xmlns="http://www.ivoa.net/xml/VOTable/v1.2"
 xmlns:stc="http://www.ivoa.net/xml/STC/v1.30" >
  <RESOURCE name="myFavouriteGalaxies">
    <TABLE name="results">

      <FIELD name="c1_center" utype="Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C1" datatype="double"/>
      <FIELD name="c1_error" datatype="double"/>
      <FIELD name="c2_center" utype="Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C2" datatype="double"/>
      <FIELD name="c2_error" datatype="double"/>
      <DATA>
        <TABLEDATA>
        <TR>
          <TD>10.6890</TD><TD>0.0598</TD><TD>287.2001</TD><TD>0.0123</TD>
        </TR>
        <TR>
          <TD>10.6811</TD><TD>0.0698</TD><TD>287.2120</TD><TD>0.0234</TD>
        </TR>
        </TABLEDATA>
      </DATA>
    </TABLE>
  </RESOURCE>
</VOTABLE>
```

src/firefly/js/util/VOAnalyzer.js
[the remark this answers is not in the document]
cwang2016: In current implementation, VOAnalyzer provides function getCenterColumns to get the center columns in which the search goes through several steps, i.e. the execution of each step is done when the previous step doesn't have result. For that, I can set this.centerColumnsInfo once in the constructor. However, my thought it more generic style is that each individual search function could be used independently for some purpose, therefore, this.centerColumnInfo is reset at the beginning of  each search function, just being treated as a variable to record the computation result of that function.

tgoldina:
The changes look good to me. Tested ucd pair `POS_EQ_RA_MAIN`/`POS_EQ_DEC_MAIN` and utype pair `Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C1`/`Char.SpatialAxis.Coverage.Location.Coord.Position2D.Value2.C2` with and without `obscore` prefix. Works as expected.

src/firefly/js/util/VOAnalyzer.js
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+
+    static newInstance(tableModel) {
+        return new TableRecognizer(tableModel);
+    }
+}
+
+export function findTableCenterColumns(table) {
+   const tblRecog = table && get(table, ['tableData', 'columns']) && TableRecognizer.newInstance(table);
```
robyww: is there not a TableUtil function to get the columns? If not you should add one an use it here.

src/firefly/js/visualize/saga/CatalogWatcher.js
```diff
@@ -137,30 +138,19 @@ function handleCatalogUpdate(tbl_id) {
...
+        return;
     }
 
     const {ignoreTables}=  getDlRoot();
     if (ignoreTables.some( (obj) => obj.tableId===tbl_id && obj.drawLayerTypeId===Catalog.TYPE_ID)) {
         return;
     }
 
-    const cenData= tableMeta[MetaConst.CATALOG_COORD_COLS] || tableMeta[MetaConst.CENTER_COLUMN];
+    const columns= findTableCenterColumns(sourceTable);
```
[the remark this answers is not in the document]
cwang2016: good, the comment also reminds me that here it is better to call TableInfoUtil.getCenterColumns.  'todo' is removed.

src/firefly/js/util/VOAnalyzer.js
cwang2016: there was function getCenterColumns in TableInfoUtil.js and in this PR the content of TableInfoUtil.getCenterColumns is replaced by calling the above function, findTableCenterColumns. Portions like catalog watch, coverage watcher, chart util,  lsst footprint or wise plot request all use TableInfoUtil.getCenterColumns to get ra/dec columns.

cwang2016:
Done the test against catalog search, VO catalog search, NED search, SOFIA, and time series to verify that the results are not changed.

PR #697, opened 2018-11-27
IRSA-2270: update format function to include usage of precision and width

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-2270
test deploy: https://irsawebdev9.ipac.caltech.edu/irsa-2270/firefly/

Use the added column's `precision` attribute when formatting data.
Added some docs to describe the policy here: `edu.caltech.ipac.table.DataType#format(java.lang.Object, boolean)`

Changed FitsTableReader to convert FITS's TDISP into our `precision` format.

Added TableTest to test the new format function.  More table related tests will be added here.
`gradle :fi:test -x jsTest --tests TableTest*`

A fun performance test to see how long it takes to perform 10mil format() calls.
`gradle -Pcategory=edu.caltech.ipac.TestCategory\$Perf :fi:test -x jsTest --tests TableTest*`

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/spring/mapper/IpacTableExtractor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/spring/mapper/IpacTableExtractor.java
@@ -185,7 +185,7 @@ public class IpacTableExtractor {
             DataType dt = headers.get(i);
             try {
                 Object obj = rs.getObject(dt.getKeyName());
-                writer.print(" " + dt.format(obj));
+                writer.print(" " + dt.format(obj, true));
             } catch (SQLException e) {
                 LOG.warn(e, "SQLException at col:" + headers.get(i).getKeyName());
                 writer.print(" " + dt.format("#ERROR#"));
--- a/src/firefly/java/edu/caltech/ipac/table/DataObject.java
+++ b/src/firefly/java/edu/caltech/ipac/table/DataObject.java
@@ -76,15 +76,17 @@ public class DataObject implements Serializable, Cloneable {
     }
 
     /**
+     * @param replaceCtrl true to replace control charters.  Use this when writing out into format where multiline row is not supported, like csv and ipac table.
      * @return this row of data as strings, formatted according to its column's info.
      */
-    public String[] getFormattedData() {
+    public String[] getFormattedData(boolean replaceCtrl) {
         return Arrays.stream(group.getDataDefinitions())
-                .map(dt -> getFormattedData(dt)).toArray(String[]::new);
+                .map(dt -> dt.format(getDataElement(dt), replaceCtrl))
+                .toArray(String[]::new);
     }
 
-    public String getFormattedData(DataType dt) {
-        return dt.format(getDataElement(dt));
+    public String[] getFormattedData() {
+        return getFormattedData(false);
     }
 
     public Object getDataElement(DataType fdt) {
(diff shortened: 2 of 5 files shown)


tgoldina:
I noticed that  we've been adding extra zeros in some cases and we still do. For example, when we display `nedserchresults.xml` from `firefly_test_data` repo - many columns have unreasonable number of zeros in the data. Ex. the original Velocity -300 shows up as -300.00000000.

By adding trailing zeroes we are making wrong statement about precision, increasing it from the original. I think, it's better not to print trailing zeros at all, if precision is unknown. 

Also, at some point we've stopped paying attention to the header width. Header is often invisible if there is one digit in a column.

loitly:
In regards to column's width, those columns you mentioned contains `width` attributes in it's votable file.  Therefore, we rendered the column's width as requested.  However, I think you are right.  In the past, we would not render it narrower than the header's width.  But, I am pretty sure it was Jeff who reported that the width's attribute did not work as expected.  Therefore, I 'fixed' it.

As for the precision issue, our VoTableReader was forcing floating-point numbers to have at least 8 decimal places.  I've removed that and stopped guessing the format when precision is not given.
However, since our format policy is to print floating-point numbers with 6 decimal places of precision, that's how it will look after my latest commit.

Gregory is tasked with how to display floating-numbers when precision is not given.  It's a CCB task that's not completed, yet.

tgoldina:
https://irsawebdev9.ipac.caltech.edu/irsa-2270/firefly/firefly-dev.html -> TAP Searches - choose Gaia.

Getting error: Type mismatch(false): expecting class java.lang.Boolean but found class java.lang.String

ejoliet:
After discussing with Loi about this PR, i've asked him to put back the guessing format and precision as it was before keeping the other fixes so it can be merged for RC next week. 

We can't change the format without CCB approval of what the UI looks like. This is too risky in terms of the impact in the UI in IRSA context with such a short notice before a release and AAS.

loitly:
@ejoliet , I've pushed the commit that will revert back to the old defaults.
when precision is not given:
- double in fits file will get "%.9g"
- floating-numbers in votable will go through a 'guess' logic with a minimum of 8 decimal places.

tgoldina:
@loitly, just so you know 8 digits min precision is something we have added after the latest release of IrsaViewer.

tgoldina:
@loitly  you were right. Current IRSA Viewer has 8 digits min precision for votables. 

The latest version will preserve ops bugs with the default precision: votable with float and double columns and values 10.68 and 287.3 will now appear as wrong as it was in ops. We can address it in a future ticket.

ops: 
irsa-ops
irsa-2270:
irsa-2270

sample votable:
```
<?xml version="1.0"?>
<VOTABLE version="1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xmlns="http://www.ivoa.net/xml/VOTable/v1.2"
 xmlns:stc="http://www.ivoa.net/xml/STC/v1.30" >
  <RESOURCE name="test-nums">
    <TABLE name="results">
      <FIELD name="FloatFld"  ID="col1" datatype="float"/>
      <FIELD name="DoubleFld"  ID="col2" datatype="double"/>
      <DATA>
        <TABLEDATA>
        <TR>
          <TD>10.68</TD><TD>10.68</TD>
        </TR>
        <TR>
          <TD>287.3</TD><TD>287.3</TD>
	</TR>
        </TABLEDATA>
      </DATA>
    </TABLE>
  </RESOURCE>
</VOTABLE>
```

PR #705, opened 2018-12-11
create integration test framework

loitly:
k8s: https://irsawebdev9.ipac.caltech.edu/int-test/firefly/

Create an integration test framework to ease the effort of adding new test cases.
PR's goals: 
- Test results that's easy to understand
- Minimize the work needed to create a test
- Tests must be isolated from others.  Minimizes bad results.

To run the tests:
https://irsawebdev9.ipac.caltech.edu/int-test/firefly/test/


loitly:
Although I've addressed some of @robyww concerns personally, I'll summarized it here so others can see as well.

> How do we scale this to 50 or 60 test?

We will separate them into multiple specialized pages.

> Loading multiple firefly's in each iframe, what happens when we are loading 25. Can the browser handle it? Will it take 15 min to load?

I have tried 10.  It's noticeably slower.  But, it's more like under 30s as oppose to 15mins.

> Should be break it into subpages maybe with 5 - 10 test each?

Yes.  It was written with that in mind.  It's trivial to separate them into multiple pages.
I would separate them based on functional domain instead of numbers.  Depending on the test, supporting up to 25 tests on a page should be fine.

> Do we want a way to mark test as failing? Maybe in html and recommit? I don't know, but I can image some cases like this.

I think if your PR causes the failure, it should be fixed before merge.  Otherwise, I would create a ticket so it can be assigned to someone to fix.

> I think the individual iframe's should scroll.

I think scrolling sub frames can make the page overly busy.  But, I don't have a strong opinion on this. 

> Should the test has more than a title area? Such as a description? such as:

No problem with this as well.  I am sure more improvement will come.

ejoliet:
That is neat!! 
My 2 cents after playing with it:

- I think we should have some kind of convention for the file name (or folder names?) as we have in the code such as test related to tables would look like 'test-int-table', for charts, 'test-int-charts'. 
- The title for each test could be more descriptive of what exactly is being tested but i should say that the 'expected' left hand side is helpful. We should answer this question: Is the intent of those pages to be used by developers or also by a user who wants to 'play' and understand the API? We use to have 'demo' pages but this is clearer and also it seems scalable.

We are one step away from automatic testing and jenkins integration!! 
We just need to run a browser embedded with DOM to test it ;-)

PR #706, opened 2018-12-12
DM-15274: file upload bug: the precision of numbers changed

loitly:
https://jira.lsstcorp.org/browse/DM-15274
k8s: https://irsawebdev9.ipac.caltech.edu/dm-15274/firefly/

To test, follow the instructions in the ticket.
This PR did not address all of the problems.  It at least picks up the correct data.
It did not preserve the exact precision.  
This is because we have a default precision when one is not given.
There are other tickets that will address this issue.


xiuqin:
Tested to confirm that the data displayed makes sense now.

PR #708, opened 2018-12-12
DM-13099 set binning value of any axis on decimated plot 

cwang2016:
This development includes, 
- fixing the bug regarding changing binning value on the plot parameter dialog of Heatmap chart
- the numbers (including the defaults for the plot in the beginning)  for x-bins & y-bins fields are  shown in plot parameters setting dialog. 

test:
start catalog search on cone search with large radius and that will result in a table more than 5000 rows. (default setting in Firefly.js), ex: m31 & radius=1600"
open up charts options and tools dialog,  change  the number of xbins or ybins or both and apply the change.

### DIFF
--- a/src/firefly/js/charts/dataTypes/FireflyHeatmap.js
+++ b/src/firefly/js/charts/dataTypes/FireflyHeatmap.js
@@ -15,6 +15,7 @@ import {COL_TYPE, getColumns} from '../../tables/TableUtil.js';
 import {getTraceTSEntries as genericTSGetter} from './FireflyGenericData.js';
 import {DECIMATE_TAG} from '../../tables/Decimate.js';
 
+const DEFBINS = 100;
 /**
  * This function creates table source entries to get firefly scatter and error data from the server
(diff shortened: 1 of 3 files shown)


src/firefly/js/charts/dataTypes/FireflyHeatmap.js
```diff
@@ -26,18 +28,17 @@ import {getTraceTSEntries as genericTSGetter} from './FireflyGenericData.js';
...
-    if (!mappings) return {}; 
+    if (!mappings) return {};
 
     const {fireflyData, fireflyLayout} = getChartData(chartId) || {};
     // server call parameters
     const xbins = get(fireflyData, `${traceNum}.nbins.x`);
     const ybins = get(fireflyData, `${traceNum}.nbins.y`);
-    let maxbins = 10000;
-    let xyratio = get(fireflyLayout, 'xyratio', 1.0);
-    if (xbins && ybins) {
```
tgoldina: I think we were talking into account xyratio that might be specified in the options when calculating the default number of bins. This feature is now lost.

PR #711, opened 2018-12-19
DM-16922 fixing parsing bug on regions with size unit in arcmin or arcsec

cwang2016:
The development fixed the bug in region parser at client side regarding the region size with the unit like  arcmin or arcsec, ex, image;box(100, 150, 1.0', 1.2', 0) # color=red

test:
Go to firefly/demo/ffapi-footprint-test.html or firefly/demo/ffapi-region-test.html 
Add some region description including a size with unit arcmin or arcsec into the text area for creating footprint or region layer and check the overlay.


lznakano:
Worked like a charm!  Great job!!!

PR #714, opened 2018-12-21
DM-14350: Improve the cube and multi-image FITS display controls

robyww:
Improve the cube and multi-image FITS display controls

  - Added individual cube readout display for cases with multiple cubes
   - Show the HDU number if more than one or not the 0 HDU (simple FITS)
   - Added: for HiPS/FITS cubes and multi image FITS text entry field to jump directly to the plane
   - show show the text entry field if more than 8 planes
   - Created a text entry StatedInputField for text input outside of a field group
   - Fixed bug: a plot id not being created correctly for upload and URL
   - add a set of PlotViewUtil functions to get cube and multi-image FITS info
   - cleaned up of unused BandState and PlotState functions

_To Test:_

  - load any of the multi image fits or cubes in `firefly_test_data/multi-ext-fits`
  - observe the control area of the cube / multi image FITS
  - this will also work with HIPS cube

### DIFF
--- a/src/firefly/js/visualize/PlotViewUtil.js
+++ b/src/firefly/js/visualize/PlotViewUtil.js
@@ -542,7 +542,7 @@ export function findPlot(plotView, imagePlotId) {
  * @return {boolean}
  */
 export function isMultiImageFitsWithSameArea(pv) {
-    if (!pv.plotViewCtx.containsMultiImageFits) return false;
+    if (!isMultiImageFits(pv)) return false;
     const plot= primePlot(pv);
     const {dataWidth:w, dataHeight:h} = plot;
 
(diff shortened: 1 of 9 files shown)


src/firefly/js/visualize/PlotViewUtil.js
```diff
@@ -725,5 +725,109 @@ export function getFoV(pv, alternateZoomFactor) {
...
+/**
+ * Get the HDU of this primaryPlot in the FITS file
+ * @param {PlotView} pv
+ * @return {number} the HDU number, single images will always return 0
+ */
+export function getPrimaryPlotHdu(pv) {
+    if (!pv) return 0;
+    const p= primePlot(pv);
+    if (!isImage(p)) return 0;
+    return p ? getHDU(plot) : 0;
```
cwang2016: plot -> p, right?

cwang2016:
Review completes. This really makes the multiple cubes readout much clearer!
Just 2 questions, 
1. for the number in StatedInputField starting from 1 and the number after 'Cube:' starting from 0, is that the way it is? 
2. when an invalid number entered in StatedInputField, I can sense something wrong since the field looks changed. could we also show a brief error message like in tooltip?

src/firefly/js/visualize/PlotViewUtil.js
```diff
@@ -725,5 +725,109 @@ export function getFoV(pv, alternateZoomFactor) {
...
+/**
+ * Count the number of cubes
+ * @param {PlotView} pv
+ * @return {number} the number of cubes, 0 if none
+ */
+export function getNumberOfCubesInPV(pv) {
+    if (!pv || !isImage(primePlot(pv)) ) return 0;
+    return pv.plots.reduce( (total, p, idx) => {
+        if (idx===0) return getImageCubeIdx(p)>=1 ? 1 : 0;
+        return (getHDU(p)!==getHDU(pv.plots[idx-1] && getImageCubeIdx(p)>-1)) ? total+1 : total;
```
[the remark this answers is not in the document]
robyww: your right, thanks.

src/firefly/js/visualize/PlotViewUtil.js
robyww: yes, should be `p`

robyww:
> for the number in StatedInputField starting from 1 and the number after 'Cube:' starting from 0, is that the way it is?

Yes, we want the user to reference cube 1 even though the index is 0.

> when an invalid number entered in StatedInputField, I can sense something wrong since the field looks changed. could we also show a brief error message like in tooltip?

I purposely turning off the error feedback to keep the UI footprint small. In the case I think it is obvious or easily fixed by hitting the arrow.

PR #715, opened 2019-01-07
IRSA-2032:SOFIA: Add footprint to viewer

lznakano:
Add sofia in the menu and then add submenu for forcast
  Add region files
  Add sever side codes to process region file and then create Footprints

URL: https://irsawebdev9.ipac.caltech.edu/irsa-2032/applications/sofia/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisContext.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisContext.java
@@ -114,6 +114,14 @@ public class VisContext {
         footprintMap.put("SPITZER_IRAC36", "footprint/Footprint_SPITZER.reg");
         footprintMap.put("SPITZER_IRAC45", "footprint/Footprint_SPITZER.reg");
         footprintMap.put("WFIRST",      "footprint/Footprint_WFIRST.reg");
+        footprintMap.put("SOFIA",      "footprint/Footprint_SOFIA.reg");
+        footprintMap.put("SOFIA_FORCAST_IMAG",      "footprint/Footprint_SOFIA.reg");
+        footprintMap.put("SOFIA_FIFI-LS_Blue",      "footprint/Footprint_SOFIA.reg");
+        footprintMap.put("SOFIA_FIFI-LS_Red",      "footprint/Footprint_SOFIA.reg");
+        footprintMap.put("SOFIA_FORCAST_GRISMS_A",      "footprint/Footprint_SOFIA.reg");
+        footprintMap.put("SOFIA_FORCAST_GRISMS_B",      "footprint/Footprint_SOFIA.reg");
+
+
     }
 
     static public void addFootprint(String key, String path) {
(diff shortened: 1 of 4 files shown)


lznakano:
@ejoliet 
Can you add your request to the ticket?  I followed the description in the ticket to add the submenu which was more work than just one menu like the others.

ejoliet:
Thanks. Sure. What i meant is that under menu 'FORCAST' there should be only 3 items. FIFI-LS items should be under a different menu 'FIFI-LS'. Please, also expect label changes because @vandesai1 needs to review those labels. The ticket was created in early stages and i expect it will change again.

lznakano:
NP, I will change it as what you described here.  What I would like is to keep the useful information in the ticket so that we can refer it later.  In the pull request, once the pull is closed, it is hard to trace it.

ejoliet:
@lznakano here are the comments from Vandana with the change request that should be made. Others comment in the ticket should be in other tickets for now.
Here are the change request:

- In the menu, I see "FIFI_LS". It should. It be "FIFI-LS".
- In the FIFI-LS submenu, sometimes there is a colon and sometimes there is a space. I don't think the colon is needed, but if it's there, it should be there consistently. Same thing for the FORCAST menu. For FIFI-LS, I don't think the word "Side" is needed. We don't use it on the Search page, and we should try to be consistent.
- I overlaid a FIFI-LS Blue footprint on a HiPS image. I saw a green box with a red cross in it, and a label. I measured one side of the green box with the distance tool. It said ~3 arcsec. In the description of this ticket, it says that the FIFI-LS blue FOV is supposed to be 30 arcsec on a side.
- The default label is "SOFIA FIFI_LS_BL". It should be "SOFIA FIFI-LS Blue".
- If I click to pan the HiPS image, the green box gets separated from the red cross and the label.
- I don't understand the second two FORCAST choices: "FORCAST: Grisms a" and "FORCAST: Grisms b". In the UI, I think these map to Low Resolution Spectroscopy SWC and LWC, but I'm not sure. @vandesai1 / Roberta should be consulted for this one.
- The label for FORCAST is "Footprint: SOFIA FORCAST_IMAG". I don't think we need the underscore. If we are short on space, I think we also don't need the word "Footprint".

lznakano:
This pull request is revised.  Please test at this URL:
   https://irsawebdev9.ipac.caltech.edu/irsa-2032-sofiaFootPrint-revised/applications/sofia/

Previously, I changed the Footprint size to a smaller size so that it can be seen in the image.  In this commit, I used the exactly sizes given in the ticket.  In the ticket IRSA-2032, it listed three choice for FORCAST.  So you will see three choice under the FORCAST menu.

ejoliet:
@lznakano i've built IRSAViewer with this branch and i don't see any SOFIA footprint overlay but i see an error in the browser console. The SOFIA footprint should also be available through other application by inherit the image toolbar footprint as well as the SOFIA GUI. Could you look at it? 

The label looks good in SOFIA app.

lznakano:
@ejoliet 
I rebased both firefly and ife branches and built irsaviewer.  Please test at  URL: https://irsawebdev9.ipac.caltech.edu/irsa-2032/irsaviewer/

PR #716, opened 2019-01-09
DM-15378 Scale the size and synchronize the movement of footprint/marker on both HiPS and fits display

cwang2016:
This development mainly fixes the issues when footprint/marker layer is attached to both HiPS and FITS images: 
- the point distance (how close to the center of the footprint or marker) on screen
pixel coordinate is adjusted per plot  (HiPS plot or FITS plot) that the footprint/marker layer attached to. 
- Scale the size of the marker according to the primary image while a new plot is created. 
   Every new marker is added with fixed screen pixel size no matter if the primary image is HIPS or FITS and the plot created later should have the marker scaled in world coordinate by looking at that in the primary image. 
- fix bugs to show the fake outline box when the outline box surrounding the footprint/marker is out of view. (this may happen when you move the marker/footprint on HiPS image and that causes the appearance of marker/footprint out of FITS image's coverage).  

Here is brief definition of two types of fake outline box as proposed in https://jira.lsstcorp.org/browse/DM-6516,
1. 'center outline box' for  the case that the center of the footprint/marker but no sides of the original outline box are within the plot area
2. 'plotcenter outline box' for the case that both the center of footprint/marker and sides of original  outline box are not within the plot area.
Both fake outline boxes are shown with resize and rotate handles for operating the marker/footprint.
'plotcenter outline box' is shown in solid line, and 'center outline box' is shown in dotted line. 
- Synchronize the movement, rotation and resizing among all types of images. 

test:
case 1: 
start a FITS image on a target, 
start a HiPS image search on the same target
case 2: 
start 2 FITS image search on the same target
case 3:
start 2 HiPS image search on the same target

for each case, 
- pick one image as the primary image, 
- add a new marker or footprint
- operate the marker or footprint (move, resize, rotate) on the primary image. 
- switch to another image as the primary image, operate the marker or footprint

to check:
- when you operate the footprint/marker on HiPS image, if the other image is FITS image, try to resize the marker or move the marker/footprint to be out of the coverage of the FITS image and see the change of outline box on FITS image. 
- when you switch around two images, make sure the footprint/marker on the primary image can be selected and moved smoothly.

### DIFF
--- a/src/firefly/js/drawingLayers/DistanceTool.js
+++ b/src/firefly/js/drawingLayers/DistanceTool.js
@@ -321,7 +321,7 @@ function makeSelectObj(firstPt,currentPt, posAngle,cc) {
     var anyPt1;
     var anyPt2;
     var dist;
-    var world= cc.projection.isSpecified();
+    var world= cc.projection.isSpecified() && cc.projection.isImplemented();
 
     if (world) {
         anyPt1 = cc.getWorldCoords(ptAry[0]);
--- a/src/firefly/js/drawingLayers/FootprintTool.js
+++ b/src/firefly/js/drawingLayers/FootprintTool.js
@@ -13,7 +13,7 @@ import {makeFactoryDef} from '../visualize/draw/DrawLayerFactory.js';
 import {ANGLE_UNIT, OutlineType, getWorldOrImage, findClosestIndex, makeFootprint,
         lengthSizeUnit, updateFootprintDrawobjAngle,
         updateFootprintTranslate, updateFootprintOutline} from '../visualize/draw/MarkerFootprintObj.js';
-import {getCC, cancelTimeoutProcess, initMarkerPos, getPlot,
+import {getCC, cancelTimeoutProcess, initMarkerPos, getPlot, ifUpdateOutline, isPointViewable,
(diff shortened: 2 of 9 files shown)


ejoliet:
@cwang2016 before merging, please make sure that this still works: IRSA-2032. thanks.

cwang2016:
1. Made updates to apply 'marker' layer onto the image  which either has no projection or an undefined projection.
the marker size and movement is scaled on world coordinate for the image which has projection. 
the marker size and movement is scaled on device coordinate (screen pixel) for the image which has no projection. 

the following cases with no projection image (ex. ex: no-projection-images/3emus-2-22_r99_fe55_test.fits) are tested as the following, 
- get image with no projection, get another image with no projection, add a marker.
- get image with no projection. add a marker, get another image with no projection.
- get image with no projection, get a fits image, add a marker (plot display error).
- get image with no projection, get a hips image, add a marker.
- get fits image, get image with no projection, add a marker.
- get hips image, get image with no projection, add a marker.
- get fits image, add a marker, get image with no projection
- get hips image, add a marker, get image with  no projection 

2. turn off footprint layer for the image which either has project or an undefined projection.

PR #720, opened 2019-01-15
DM-16048 Save tables in other formats

cwang2016:
The PR includes development:
- for making the infrastructure to save table in different formats
- for saving table in csv and tsv formats

test
do catalog search
save the table content by selecting the file format and entering a file name. 
upload the save table file and compare the table content with previously catalog search result.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/BaseFileInfoProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/BaseFileInfoProcessor.java
@@ -66,7 +66,7 @@ public abstract class BaseFileInfoProcessor implements SearchProcessor<FileInfo>
         return request.getRequestId() + "-" + id + "-" + StringUtils.toString(request.getParams());
     }
 
-    public FileInfo writeData(OutputStream out, ServerRequest request) throws DataAccessException {
+    public FileInfo writeData(OutputStream out, ServerRequest request, String format) throws DataAccessException {
         /* does not apply.. do nothing */
         return null;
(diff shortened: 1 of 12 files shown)


robyww:
>  I would expect to see it below the name and location fields. 

Below is fine with me.

cwang2016:
@robyww The table save dialog reuses the the same component as that for Image save dialog, DownloadOptionsDialog. If the format options is moved to be below the name and location fields, should the image save dialog be updated similarly?

robyww:
> The table save dialog reuses the the same component as that for Image save dialog, DownloadOptionsDialog. If the format options is moved to be below the name and location fields, should the image save dialog be updated similarly?

I just looked at the image save. Let's not change the UI much and do it the same in both. So keep it as it is.  If we decide to take Loi's suggestion at some later time we can.  It is probably not a big deal either way.

ejoliet:
If this UI will be included in IRSAViewer and other IRSA applications, we need to let Luisa for the documentation update. Also, did someone test CSV from ATLAS with column name with spaces etc?

ejoliet:
Might not be done in this PR but just a comment: also, why not providing the original format? If i do a search with TAP or in SOFIA, the format will be XML. Why not supporting XML too?

robyww:
>  Why not supporting XML too?
We broke her ticket into 3 since it was a big task.
1. do the frame work and implement cvs and tsv
2. do votables
3. do FITS tables

But your right.  When it is all done it should default to the original format. @cwang2016 you might want to keep that in mind.

PR #723, opened 2019-01-16
DM-17265 Bug with Default color scale in heatmap chart

tgoldina:
Bug report: https://jira.lsstcorp.org/browse/DM-17265
Test build: https://irsawebdev9.ipac.caltech.edu/dm-17265_heatmapbug/firefly/

Fixed "Default" color scale bug in heatmap charts. 
Default color scale is chosen from color scale list based on trace number.


cwang2016:
the color scale is fixed. well done.

PR #728, opened 2019-01-23
DM-17214: Create an advanced ADQL query panel

loitly:
https://jira.lsstcorp.org/browse/DM-17214

Create a free-hand ADQL query panel.  

test: https://irsawebdev9.ipac.caltech.edu/dm-17214/firefly/firefly-dev.html   
goto TAP Searches then click ADQL tab

The tables and columns are dynamically fetched.  A couple of things that's not obvious and should be either documented in help or on the panel:
- when clicked on a table, a default `SELECT` statement on that table is generated and place into the text area
- when clicked on a column, the column name plus a comma separator is inserted into the text area at its last active cursor.  This allow the user to easily add several columns by clicking.

There's a popup warning when a query is submitted without a `TOP` or `WHERE` constraint.  This is just a rudimentary sanity check.


src/firefly/js/ui/ValidationField.jsx
```diff
@@ -27,3 +28,4 @@ const propTypes= {
 };
 
 export const ValidationField= fieldGroupConnector(InputFieldView,getProps,propTypes);
+export const ValidationTextArea= fieldGroupConnector(InputAreaFieldView,getProps,propTypes);
```
cwang2016: is 'InputAreaFieldConnected' (in file InputAreaField.jsx) the same as 'ValidationTextArea'?

cwang2016:
Looks great! 
just little thoughts, 
- when selecting a table, could 'SELECT *...' overwrite the existing ADQL statement? or start from a new line? 
- how about a 'clear' button to erase the query box? 
- more detail about ADQL statement or links to ADQL documentation may be included in the help page.

loitly:
Thank for the inputs, they were all good.  I just added a commit that did the following:
DM-17214: Advanced ADQL panel
- refresh schema when serviceUrl changes
- add a `clear` button
- add browser's hint to the panel
- insert default statement on new line
- insert column name, then set focus on the input field so input can continue without delay

The last bullet point, I think it's a better solution.  This allow the user to continue where they left off, either from the `select` clause or the `where` clause.

ejoliet:
I have a comment also about UX after i played with it: maybe finding a good way to tell the user that clicking on the column is actually editing the right-hand side form? Was not intuitive enough to me but i might be wrong user.

Maybe adding a link when you hoover on the column name of the tree? Or just a info bubble to describe the feature and it's usability?

loitly:
@robyww , thanks for the review.  I'll merge this PR as is for now.  We'll gather the additional features into another ticket to be implemented at a later time.  There are other "nice to have" features like, sort by `name`, instead of `index`, custom sample queries for selected project(s), and in general, what to show in the bottom help section.

PR #730, opened 2019-01-28
IRSA-1991-ZTF:ZTF time series: North up not applied

lznakano:
Fixed the ztf image direction issue
   Fixed the ra/dec format issue in IBE.java

   Test URL:  https://irsawebdev9.ipac.caltech.edu/irsa-1991/irsaviewer/ts.html
   Please use the ztf_tbl attached in the ticket IRSA-1991.  This is the most recently table saved from IBE gator searching result.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.net.URLEncoder;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


ejoliet:
Seems that the orientation is correct now but the zoom is not the same as in OPS. Can you check why? Maybe is just a Firefly change that we didn't catch. 
Other than that, that's good to go. Thanks!

lznakano:
I can not compare with ops for ztf.  I checked wise images.  The FOV angle is very different.  In ops, it is 12  minutes and in dev it is 1.8 degree.   I don't have idea why it was changed.  I think finding it out is beyond the scope of this ticket.   I commented out my changes and found that the zoom is the same.  So I can conclude that the zooming issue is not introduced by my fix in this branch.  Can you please approve the change so that I can merge it to the dev?  Thanks!

ejoliet:
Ok, but could make a ticket with the FOV issue? thanks.

PR #731, opened 2019-01-29
IRSA-945:IRSAViewer panner shows blank flux for pixels with NaN values

lznakano:
Display the NaN value instead of blank.

URL to test this PR is here https://irsawebdev9.ipac.caltech.edu/irsa-945/irsaviewer/


lznakano:
@ejoliet 
Good catch, I accidentally commented a break in a case block that leads everything value to NaN.   
Please test it again.  Thanks!

PR #733, opened 2019-02-01
DM-17212 TAP Search Schema and Table selection

tgoldina:
https://jira.lsstcorp.org/browse/DM-17212

Test build: https://irsawebdev9.ipac.caltech.edu/dm-17212_tableselect/firefly/firefly-dev.html?__action=layout.showDropDown&visible=true&view=TestTAPSearch

- Added select widgets for TAP schema and table selection
- The group header shows the number of schemas/tables and has a placeholder button to open an advanced table select popup.


src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
```diff
@@ -368,24 +368,39 @@ private static void getTableElements(VOElement root, List<TableElement>tblAry) {
...
-                    }
-                }
+                checkForErrors(top);
             }
         }
 
         return tableAry;
     }
 
+    private static void checkForErrors(VOElement top) throws DataAccessException {
```
loitly: This looks like a TAP's specific error condition.  I am not sure it should be here.  Your thoughts?

robyww: > There is lots of space. Why not put the label outside the fields:

I encouraged @tgoldina to keep it inside for now because I would like to look at ways to make the whole upper area smaller to leave more rooms for the table on bottom.  Of course we are going to have to figure out how to make the spacial input smaller as well.

To keep inside it will have to stand out more and It cannot scroll, it needs to stay like a title area. Not scrolling is important. If the widget can't do that then we should just go ahead and move it out.

PR #735, opened 2019-02-01
IRSA-2568: Coverage now does a more rigorous check for validate coverage

robyww:
Coverage now does a more rigorous check to validate and show coverage.

  - For box coverage a least 10% of the rows must have non-empty data
  - All empty data rows are thrown out (box or point)
  - Before empty data was creating 0,0 j2000 points.

_To test:_

  - load sofia
  - search m17
  - look at coverage


deannaji:
Looks great, thanks!!

PR #739, opened 2019-02-05
IRSA-27_firefly_should_point_to_IRSA_to_get_DSS_image

deannaji:
Kindly review:
This PR is about getting the DSS images from IRSA.
As part of it, we have also fixed the cut out for images (getting full images, when asking for a cut out). This was introduced in IRSA-1991 PR. 

Jira: https://jira.ipac.caltech.edu/browse/IRSA-27
ife: https://github.com/IPAC-SW/irsa-ife/pull/46
K8s: https://irsawebdev9.ipac.caltech.edu/irsa-27/applications/finderchart/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
@@ -148,8 +148,8 @@ public class IBE {
             qstr = addUrlParam(qstr, "gzip", param.isDoZip());
         }
 
-        //url += (qstr.length() > 0 ? "?" + qstr : "");
-        url += (qstr.length() > 0 ? "?" + URLEncoder.encode(qstr, "UTF-8") : "");
+        url += (qstr.length() > 0 ? "?" + qstr : "");
+        //url += (qstr.length() > 0 ? "?" + URLEncoder.encode(qstr, "UTF-8") : "");
 
         return new URL(url);
     }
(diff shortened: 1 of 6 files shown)


ejoliet:
@deannaji  @robyww although the bug introduced in IRSA-1991 is reverted i think the idea from @lznakano  to fix the problem seen in Time series with ZTF might be still happening then.

ejoliet:
This PR has now a fix version of DSS using IBE. I've fixed a cache issue when the key built based on bands name doesn't contain a band which is the case for getting images from DSS. DSS IBE search rely on filter contraint only , not band name because they are not unique values so in order to distiguish we need a filter on file name. 

This PR shouldn't merged yet, it still need a work for the download feature and the 3 color images.

Also, DSS images from IRSA doesn't include WCS and a ticket was raised (IRSA-2613) so until that get fixed, we can't be using IRSA DSS images.

src/firefly/java/edu/caltech/ipac/astro/ibe/IBE.java
```diff
@@ -148,8 +148,8 @@ public URL createDataUrl(IbeDataParam param) throws IOException {
             qstr = addUrlParam(qstr, "gzip", param.isDoZip());
         }
 
-        //url += (qstr.length() > 0 ? "?" + qstr : "");
-        url += (qstr.length() > 0 ? "?" + URLEncoder.encode(qstr, "UTF-8") : "");
+        url += (qstr.length() > 0 ? "?" + qstr : "");
+        //url += (qstr.length() > 0 ? "?" + URLEncoder.encode(qstr, "UTF-8") : "");
```
robyww: good

loitly: Sorry, I don't mean to bud in, saw this went by my email.  
The bad commented line is already removed from `dev`.  You should remove it from here as well.  There's no reason to leave bad code in the source, even though it's commented out.

robyww: good point. thanks.

ejoliet:
Will close this and reopen when we have a conclusion on getting DSS images from IRSA.

PR #740, opened 2019-02-07
DM-17499 Support error docs in plain text

tgoldina:
Ticket: https://jira.lsstcorp.org/browse/DM-17499

Needs to be built locally for test, because only LSST is producing plain text error docs, and LSST services require VPN.

"AsyncTapQuery" search processor now handles both formats of errors: VOTable and plain text.

Bugs fixed:
- LSST result queries (when using URLDownload with GET method) were not following redirects
- Single char datatype in VOTable was not converted to the right hsqldb type, which was causing table load failure: "Invalid Statement". 

To test:
1. Build locally, use NCSA VPN to run:
http://localhost:8080/firefly/firefly-dev.html?__action=layout.showDropDown&visible=true&view=TestTAPSearch
2. Use ADQL tab to modify ADQL query to make it invalid. Search.
3. Switch between services (sample query should change). Try both valid and invalid queries.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -13,13 +13,12 @@ import edu.caltech.ipac.table.DataGroup;
 import edu.caltech.ipac.table.LinkInfo;
 import edu.caltech.ipac.table.io.VoTableReader;
 import edu.caltech.ipac.util.StringUtils;
-import edu.caltech.ipac.util.download.URLDownload;
 import org.apache.commons.httpclient.Header;
 
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.IOException;
+import java.io.*;
 import java.net.URL;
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -124,32 +124,65 @@ public Phase getPhase() throws DataAccessException {
...
+            String errorUrl = baseJobUrl + "/error";
+
+            Ref<String> err = new Ref<>();
+            HttpServices.Status status = HttpServices.getData(errorUrl, HttpServiceInput.createWithCredential(),
+                    (method -> {
+                        boolean decompress = false, isText = false;
+                        Header encoding = method.getResponseHeader("Content-Encoding");
+                        if (encoding != null && encoding.getValue().contains("gzip")) {
+                            decompress = true;
+                        }
```
loitly: You should not have to be concern with handling of encoding.  This knowledge should be in HttpServices.  
Could you create a common HttpServices.getInputStream() so that this and OutputStreamHandler can use?

PR #741, opened 2019-02-07
DM-17722: upgrade react.js to v16.8

loitly:
https://jira.lsstcorp.org/browse/DM-17722
Test: https://irsawebdev9.ipac.caltech.edu/dm-17722/firefly/

- upgraded React.js to version 16.8
- converted AdvancedADQL to use react hooks which is a big feature of this release

The converted file make use of the new useState and useRef hooks, as well as demonstrate how updated props can be compared directly in the render function.
It serves as a test for the upgrade as well as demonstration of how new features are used.

Here is a complete changelog for react:
https://github.com/facebook/react/blob/master/CHANGELOG.md#1680-february-6-2019

This blog highlight some of the changes in v16, up to v16.7.
https://hackernoon.com/react-16-0-16-3-new-features-for-every-day-use-f397da374acf

Summary of the blog:
- Returning multiple elements from components with fragments (16.2)
  <React.Fragment>
      …
  </React.Fragment>
or
  <>
      …
  </>

- Returning strings and numbers from components
- Cancelling setState() to avoid rerendering
- Avoiding prop drilling with the official context API (16.3)
- Updating state based on props with getDerivedStateFromProps() (16.3)
- Re-rendering function components on props change with React.memo() (16.6)
- Easier access to context in class components with contextType (16.6)


loitly:
Close due to misinterpretation of the ticket.  Instead of upgrading just React and anything else needed, it was meant to upgrade everything possible except for redux.

PR #745, opened 2019-02-11
DM-16777: Cleanup, updates, and documentation of who we hand meta data

robyww:
Cleanup, updates, and documentation of meta data

   - documented all meta in MetaConst.js and MetaConst.java
   - removed or deprecated unnecessary or redundent meta data
   - cleaned up definition of catalog
   - moved more logic into VOAnalyzer
   - refactors some search processors that use meta data
   - Added more testing

_To Test:_

https://irsawebdev9.ipac.caltech.edu/dm-16777-metadata/firefly/

Everything should work the same as before

  - Search a catalog
  - Search a votable
  - Search a wise image (under test searches)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
@@ -15,21 +15,34 @@ package edu.caltech.ipac.firefly.data.table;
  */
 public class MetaConst {
 
+    /**
+     * Meta entry that defines how to compute the center position of table row
+     * Form: lonCol;latCol;CoordinateSys.toString(), eg ra;dec;J2000
+     * @see edu.caltech.ipac.table.TableMeta#setCenterCoordColumns
+     */
     public static final String CENTER_COLUMN = "CENTER_COLUMN";
+
+
+    /**
+     *
+     * Meta entry that defines the corners of a object defined by a table row
+     * like CENTER_COLUMN but each position separated by comma
+     * eg - ra1;dec1;EQ_J2000,ra2;dec2;EQ_J2000,ra3;dec3;EQ_J2000,ra4;dec4;EQ_J2000
+     * @see edu.caltech.ipac.table.TableMeta#setCorners
+     */
     public static final String ALL_CORNERS   = "ALL_CORNERS";
 
-    public static final String WORLD_PT_KEY = "SearchByPosition.worldpt";
-    public static final String RADIUS_KEY = "SearchByPosition.field.radius";
+    /** if defined to any value the the table is a catalog */
     public static final String CATALOG_OVERLAY_TYPE = "CatalogOverlayType";
-    public static final String CATALOG_HINTS = "CatalogHints";
-    public static final String DATA_PRIMARY = "DataPrimary";
 
-    public final static String CATALOG_TARGET_COL_NAME = "CatalogTargetColName";
-    public final static String CATALOG_COORD_COLS = "CatalogCoordColumns";
-    public final static String MOVING_COORD_COLS = "MovingObjCoordColumns";
+    /** @deprecated replaced by IMAGE_SOURCE_ID */
+    @Deprecated
     public final static String DATASET_CONVERTER = "datasetInfoConverterId";
+    
+    /** id string for types of image */
+    public final static String IMAGE_SOURCE_ID = "ImageSourceId";
+    /** id string for types of mission */
+    public static final String MISSION = "mission";
 
-
-    public static final String DEFAULT_COLOR="DEFAULT_COLOR";
 }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
@@ -32,7 +32,11 @@ public class MetaConst {
      */
     public static final String ALL_CORNERS   = "ALL_CORNERS";
 
-    /** if defined to any value the the table is a catalog */
+    /**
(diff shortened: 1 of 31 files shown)


src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
```diff
@@ -15,21 +15,34 @@
...
+     * like CENTER_COLUMN but each position separated by comma
+     * eg - ra1;dec1;EQ_J2000,ra2;dec2;EQ_J2000,ra3;dec3;EQ_J2000,ra4;dec4;EQ_J2000
+     * @see edu.caltech.ipac.table.TableMeta#setCorners
+     */
     public static final String ALL_CORNERS   = "ALL_CORNERS";
 
-    public static final String WORLD_PT_KEY = "SearchByPosition.worldpt";
-    public static final String RADIUS_KEY = "SearchByPosition.field.radius";
+    /** if defined to any value the the table is a catalog */
     public static final String CATALOG_OVERLAY_TYPE = "CatalogOverlayType";
```
loitly: This is not quite true, right?
It looks like this is only used to determine whether `guessing` is allowed when searching for center columns.  As long as there's center columns info, it's a catalog regardless of this setting.

robyww: I do need to make it more clear.  `CatalogOverlayType` is required to even guess if there is not VO information. Also it can be defined to FALSE to disable it being a catalog altogether even with VO information.

The problem is then we could guess it to be a catalog and it could be image meta data with corner information.  Therefore `CatalogOverlayType` is required to treat it as a catalog.

I have spent some time thinking about this and I don't see anyway we can deal with simple tables like cvs, tsv, ipac, or FITS with something telling us how to treat the table.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/QueryIBE.java
```diff
@@ -45,21 +45,20 @@
...
-    public static final String MISSION = "mission";
     public static final String POS_WORLDPT = "UserTargetWorldPt";
     public static final String RADIUS = "radius";
     public static final String MOST_CENTER = "mcenter";
 
 
 
     protected String getFilePrefix(TableServerRequest request) {
-        return request.getParam(MISSION);
+        return request.getParam(MetaConst.MISSION);
```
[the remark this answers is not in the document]
robyww: OK, I will change it.

src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
loitly: Ah.. it makes sense now.  Too bad prop is already used by others.  Otherwise, having a prop like `data_type` to say what this table is('image-meta'), instead of using `CatalogOverlayType=false` to say what it is not would make more sense.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/QueryRTreeInventory.java
```diff
@@ -127,7 +127,7 @@ public void prepareTableMeta(TableMeta meta, List<DataType> columns, ServerReque
                 "ra1", "dec1", "ra2", "dec2", "ra3", "dec3", "ra4", "dec4"
         };
         String dataset = request.getParam("RequestedDataSet");
-        meta.setAttribute(MetaConst.DATASET_CONVERTER, dataset);
+        meta.setAttribute(MetaConst.MISSION, dataset);
```
cwang2016: should here be MetaConst.MISSION or MetaConst.IMAGE_SOURCE_ID?

robyww: should be `IMAGE_SOURCE_ID`, i will fix it.

src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCatalogSearch.java
```diff
@@ -331,7 +331,7 @@ public void prepareTableMeta(TableMeta meta, List<DataType> columns, ServerReque
             }
 
             // only set for image meta table
-            meta.setAttribute(MetaConst.DATASET_CONVERTER,
+            meta.setAttribute(MetaConst.IMAGE_SOURCE_ID,
                     (String)LSSTQuery.getDatasetInfo(catTable, new String[]{MetaConst.DATASET_CONVERTER}));
```
cwang2016: MetaConst.DATASET_CONVERTER -> MetaConst.IMAGE_SOURCE_ID ?

robyww: Yes, we are deprecating `DATASET_CONVERTER` and using `IMAGE_SOURCE_ID`

cwang2016:
thanks for the update and the fixing on VOAnalyzer. Did some test on catalog search, votable upload & LSST footprint, work as expected.

src/firefly/java/edu/caltech/ipac/firefly/server/persistence/QueryRTreeInventory.java
robyww: IMAGE_SOURCE_ID. fixed.

src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTCatalogSearch.java
robyww: fixed.

PR #749, opened 2019-02-20
IRSA-2623: patch: simple big cube with many extension fix

robyww:
patch to rc to fix file with over 1000 images bugb

### DIFF
--- a/config/ehcache.xml
+++ b/config/ehcache.xml
@@ -479,7 +479,7 @@
      The data store in this cache is large.  it will NOT be replicated.
     -->
     <cache name="VISUALIZE"
-           maxElementsInMemory="1000"
+           maxElementsInMemory="30000"
            eternal="false"
            timeToIdleSeconds="7200"
            memoryStoreEvictionPolicy="LRU"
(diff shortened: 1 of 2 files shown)


robyww:
>In Firefly the fix is confirmed but not in SOFIA. I've built the app and I follow the step to reproduce in the ticket. The 'data' tab and the rest of the app got frozen unresponsive. Do you see the same on your end?

I think the bug is fixed. It was a server bug and it is not longer happening. I was able to load the data.  I did see that it took a very long time in the sofia app.  Much longer that in the firefly app.  I think that indicates that the is some sort of extra work happening on the sofia client code, probably in a worker or saga that is doing something for every new image. In that case the fix would be in irsa-ife.

I will go look for it and see what I can find but I recommend we merge this PR.

robyww:
A little more analysis. I think the client side slow down is in parsing the json result. I don't know why it appears slower on for Sofia, I don't think there is a difference. 

This is something we could look at when we try to do more optimization in our cube processing but I don't think we can do anymore in this patch.

I recommend we merge this PR.

ejoliet:
Ok, make sense. Merge ok anytime.

PR #751, opened 2019-02-26
Dm 17330 tap temporal search box content implementation

cwang2016:
The PR includes development to do temporal search for TAP search, 
- a date/time picker is provided to assist the selection of date/time by using momentjs and react-datetime 
- the time can be set in either ISO string or MJD number. 
- a pop-up date/time picker is popped up by clicking the date/time icon  while in ISO mode. This pop-up is a modal dialog. 
- the ADQL where clause statement for temporal search is made on single column at this PR. 

Test:
go to 
https://irsawebdev9.ipac.caltech.edu/dm-17330/firefly/firefly-dev.html
select tab 'TAP searches'
select Schema: caom, table: caom.plane
open Temporal search panel, 
         enter 'time_bounds_lower' to Temporal Columns field.   
         select button 'MJD'
         enter '59898.32901' and '59898.32970' to the right two text fields. 
        (you may click button 'ISO' to see the equivalent date/time for those two MJD numbers, and click the date/time icon to get a pop-up date/time picker)
click 'Populate and edit ADQL' to see the formed ADQL query or click 'search' to get the query results.


src/firefly/js/ui/tap/TableSearchMethods.jsx
```diff
@@ -546,14 +698,87 @@ function tapSearchMethodReducer(columnsModel) {
                                 rFields.polygoncoords = clone(inFields.polygoncoords, {value: v});
                             }
                         }
+                    } else if (fieldKey === TimeFrom || fieldKey === TimeTo) {
```
robyww: switch statements with a lot lof logic like this are hard to read. I think the whole sections under `FieldGroupCntlr.VALUE_CHANGE:` needs to be broken out into 2 or 3 functions. It is just too difficult to determine what is going on.

cwang2016: thanks. updated.

cwang2016:
Per comments from @tgoldina 
- fixed the bug regarding the position columns - empty field is allowed. 
- change "Temporal columns" to be "Temporal column" (this is a tentative fix. the temporal search formula might be updated in the future and it may have more than one column involved for the search).

PR #753, opened 2019-02-28
Irsa 2146 add naifid resolver to firefly

deannaji:
Kindly review:
This ticket is about adding a Naifid Panel to firefly.

Jira: https://jira.ipac.caltech.edu/browse/IRSA-2146
K8s: https://irsawebdev9.ipac.caltech.edu/irsa-2146/firefly/

Note: For testing, please: 
- Go to (https://irsawebdev9.ipac.caltech.edu/irsa-2146/firefly//firefly-dev.html)
- Click on 'Test Searches' button (on the top menu), then click on the 'NAIF-ID' tab (the last tab). 
- Or, click on 'Examples Js Dialog' button.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
@@ -41,7 +41,7 @@ public class ResolveServerCommands {
             }
         }
 
-        public boolean getCanCreateJson() { return true; }
+        public boolean getCanCreateJson() { return false; }
     }
 
     public static class ResolveNaifidName extends ServCommand {
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
@@ -4,12 +4,15 @@
 package edu.caltech.ipac.firefly.server.rpc;
 
 
+import edu.caltech.ipac.astro.net.HorizonsEphPairs;
 import edu.caltech.ipac.astro.net.Resolver;
 import edu.caltech.ipac.firefly.data.ServerParams;
 import edu.caltech.ipac.firefly.server.ServCommand;
 import edu.caltech.ipac.firefly.server.SrvParam;
 import edu.caltech.ipac.visualize.plot.ResolvedWorldPt;
+import org.json.simple.JSONObject;
 
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -34,6 +37,30 @@ public class ResolveServerCommands {
                 return wp.toString();
             } catch (Exception e) {
                 throw new Exception("Could not resolve " +  name + " using " + resolver.toString());
+            }
+        }
+
+        public boolean getCanCreateJson() { return false; }
+    }
+
+    public static class ResolveNaifidName extends ServCommand {
+
+
+        public String doCommand(SrvParam sp) throws Exception {
+
+            try {
+                HorizonsEphPairs.HorizonsResults[] horizons_results = HorizonsEphPairs.lowlevelGetEphInfo(sp.getRequired(ServerParams.OBJ_NAME));
+
+                Map<String, Integer> values = new HashMap<>();
+                for (HorizonsEphPairs.HorizonsResults element : horizons_results) {
+                    values.put(element.getName(), Integer.parseInt(element.getNaifID()));
+                }
+
+                JSONObject jsonObj = new JSONObject(values);
+                return jsonObj.toJSONString();
+
+            }catch (Exception e){
+                throw new Exception("Could not resolve "+ ServerParams.OBJ_NAME);
             }
         }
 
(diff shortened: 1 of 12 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
```diff
@@ -37,7 +41,40 @@ public String doCommand(SrvParam sp) throws Exception {
             }
         }
 
-        public boolean getCanCreateJson() { return false; }
+        public boolean getCanCreateJson() { return true; }
```
loitly: This is why TargetPanel stopped working.

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -0,0 +1,200 @@
...
+            const rval = resolveNaifidObj(val);
+            if (!rval.p) return undefined;
+            return rval.p.then((response)=>{
+                if(response.valid) {
+                    let suggestionsList = response.data || [];
+
+                    if(this.iAmMounted && this.state.suggestions !== suggestionsList) {
+                        this.setState({suggestions: suggestionsList});
+                    }else if(this.state.suggestions === suggestionsList){
+                        this.setState({suggestions:''});
```
loitly: If you're checking for `this.iAmMounted` before `setState` on top, I am pretty sure you need to  check here too.

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -0,0 +1,200 @@
...
+        if (get(t,'worldPt')) retVal= t.worldPt.toString();
+    }
+    return retVal;
+}
+
+function getNaifidValue(onCallBack) {
+    return (val, str) => {
+        if (! str) return;
+
+        const naifId = (str.match(/Naifid:.*/))[0].split('Naifid:').pop();
```
loitly: This is error-prone.  When match fail, `[0]` will cause TypeError.
Should replace with this:
`const [,naifId] = str.match(/Naifid:(.*)/) || [];`

src/firefly/js/ui/NaifidPanelWorker.js
```diff
@@ -0,0 +1,142 @@
...
+    if (!objName) {
+        return {
+            showHelp: true,
+            valid : true,
+            feedback: ''
+        };
+    }
+
+    let {p}= makeResolverPromise(objName);
+    p= p.then( (results) =>
```
loitly: This is misleading... why re-assign `p` to a `then` declaration?
It should be written,  `p.then( (results) =>`  without `p = `.

robyww: OK, the problem is that you want to return the then promise not p. So what @deannaji is doing is correct, but confusing. It should just be `return p.then...`

But if you do then you will needs to modify `getSuggestions` so it is expect a `promise` from `resolveNaifidObj` and not an `object`.
 
NaifidPanel.jsx, `getSuggestions`
```
            if (!rval) return undefined;
            return rval.then((response)=>{
```

I think you copied some code from target panel where it was return not only a promise in the object but also a function to abort the promise. Since you are only returning a promise there is no need to wrap it in an object.

src/firefly/java/edu/caltech/ipac/firefly/server/rpc/ResolveServerCommands.java
deannaji: Done, thanks!

robyww:
> The dropdown menu should be above the form and height should be larger to show 5 items at least (currently i see only 2)

This is a problem with the disclosure panel. It is getting cut off by the wrapping component.  This is fixable but not a big problem because almost all panels are bigger and therefore it is not cut off. It should be a bug in a separate ticket.

robyww:
> Dropdown label should be in one line: 'Name: .. , Naifid: …', again, this should be controlled by a property css

Tested this on the most recent push, looks good

src/firefly/js/ui/NaifidPanel.jsx
deannaji: Done, thanks!

src/firefly/js/ui/NaifidPanel.jsx
deannaji: Done!
Thanks :)

src/firefly/js/ui/NaifidPanelWorker.js
deannaji: Done, thanks!!

deannaji:
I did update K8s build, and done very minor tweaks to make it look like what we see here:
(https://sha.ipac.caltech.edu/applications/Spitzer/SHA/#id=MOSQuery&RequestClass=ServerRequest&shortDesc=Solar%20System%20Object/Orbit%20Precovery%20Search&isBookmarkAble=true&isDrillDownRoot=true)

Thanks for help!
:)

src/firefly/js/ui/NaifidPanelWorker.js
```diff
@@ -93,7 +93,7 @@ function resolveObject(objName) {
     }
 
     let {p}= makeResolverPromise(objName);
-    p= p.then( (results) =>
+    p.then( (results) =>
```
loitly: My mistake.  Trey corrected me.  You had it right before.
It was just awkward to see it written that way.  Please revert this change.

deannaji: Completed, Thanks!!

deannaji:
@ejoliet , can I merge and fix other issues in a different ticket. So I can add NAIFID to SOFIA :)

deannaji:
Copy that, thanks!

PR #755, opened 2019-03-01
DM-18102: fixed 2 regression bugs

robyww:
DM-18102: fixed 2 regression bugs

   - firefly not guessing ra/dec in tables like before
        - columns like coord_ra, coord_dec not recognized
        - regular expression change
   - fix regression undeclared variable bug

### DIFF
--- a/src/firefly/js/util/VOAnalyzer.js
+++ b/src/firefly/js/util/VOAnalyzer.js
@@ -416,8 +416,8 @@ class TableRecognizer {
             let lonCol;
             let latCol;
             if (useReg) {
-                const reLon= new RegExp(`^[A-z]*?[-_]?(${lon})[1-9]*$`);
-                const reLat= new RegExp(`^[A-z]*?[-_]?(${lat})[1-9]*$`);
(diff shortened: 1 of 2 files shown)


src/firefly/js/util/VOAnalyzer.js
```diff
@@ -416,8 +416,8 @@ class TableRecognizer {
             let lonCol;
             let latCol;
             if (useReg) {
-                const reLon= new RegExp(`^[A-z]?[-_]?(${lon})[1-9]*$`);
-                const reLat= new RegExp(`^[A-z]?[-_]?(${lat})[1-9]*$`);
+                const reLon= new RegExp(`^[A-z]*?[-_]?(${lon})[1-9]*$`);
+                const reLat= new RegExp(`^[A-z]*?[-_]?(${lat})[1-9]*$`);
```
loitly: You probably mean to replace `?`(0 or 1) with `*`(0 or more).  Not both.

robyww: ok, your right. thanks

PR #760, opened 2019-03-05
IRSA-2553: fixed Add Image Dialog Cancel Button Broken in Gator

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-2553

Fixed `Cancel` button not working.
At the moment, you can test with this url: https://irsadev.ipac.caltech.edu:9201/

Also, noticed that it's a multi-select.  Shouldn't it be single select?


loitly:
> For now, i think you should make it single as it was before.

Done.. merging PR.

PR #762, opened 2019-03-07
DM-18121 TAP search: column field selection, column constraints as a view

tgoldina:
https://jira.lsstcorp.org/browse/DM-18121
Test build: https://irsawebdev9.ipac.caltech.edu/dm-18121_colsel/firefly/firefly-dev.html?__action=layout.showDropDown&visible=true&view=TestTAPSearch

- Positional and temporal columns fields now behave similarly to column selection fields in charts
- Made the search for suggestions case-insensitive. (The field value is case-sensitive though, and must match column name exactly to pass validation.)
- Column constraints and selections are now independent of what is visible in column constraints table
- Cleaned column constrains selection, no field is necessary.
- TAP search requests are not cached (per Xiquin's request). New search will produce new asynchronous request.
- Added error reporting, but did not fix filter validation partially working - will move to a relevant ticket. (Filter validation is done in FilterInfo.conditionValidatorNoAutoCorrect)
- Removed unimplemented features from the UI.
- CADC TAP service should now work. (Switched to https.)

### DIFF
--- a/src/firefly/js/charts/ui/ColSelectView.jsx
+++ b/src/firefly/js/charts/ui/ColSelectView.jsx
@@ -57,9 +57,9 @@ export function showColSelectPopup(colValStats,onColSelected,popupTitle,buttonTe
     for (let i = 0; i < colValStats.length; i++) {
             data[i] = [
                         colValStats[i].name,
-                        colValStats[i].unit,
+                        colValStats[i].units,
                         colValStats[i].type,
-                        colValStats[i].descr
+                        colValStats[i].desc
             ];
     }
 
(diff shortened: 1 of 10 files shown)


cwang2016:
It works well. Just some minor questions, 
- when CADC TAP service is selected, all columns are selected, is that expected?
- for the filter  made on columns table, a list which contains all the possible selection is shown when the little triangle is clicked.  In some case like unit, some column may not have unit, is there any way (or if it is needed to have some way) to filter the column table to show those no-unit columns? 
- the title on column table is a little cut off, like 'datatype'.
 
I will integrate the change made on spatial search panel in this PR into the development I am making on target and spatial panel updates, https://jira.lsstcorp.org/browse/DM-18122, including the replacement of 'Position Columns' field  by 'Longitude Column' and 'Latitude Column' fields and the addition of column search suggestion and column name list pop-up to column related fields on the spatial search panel.

tgoldina:
Cindy, thank you for the observations. 

Table headers have been probably slightly cut off on the bottom for years, and you are the first one who have noticed. It's noticeable when there is 'y' or 'p' in the title. @loitly is looking into it.

We preselect all columns with principal set to 1. (It's a visual indication that a column is principal.) 
In CADC caom table, all columns are principal, so they are all preselected. We do detect all column selection, and select `*` in this case. 

Filter enum values do not include empty strings. Perhaps, on purpose. An alternative way to find these values is by a filter `like ''`.

PR #765, opened 2019-03-08
DM-13630: Fixed bug: flip fail when there is a fail image plot in the group

robyww:
Fixed bug: flip fail when there is a fail image plot in the group

_to test:_

https://irsawebdev9.ipac.caltech.edu/dm-13630-dist-tool/firefly/

Make image plots with m34 on SEIP. Four images should come up and one should fail- then do a flip. It if works and there are not errors in the console the bug is fixed.


deannaji:
Looks good, I was able to flip the 3 images on Y with no console error.
:)

PR #769, opened 2019-03-14
DM-16667 Use TAP provided by DAX team when available

tgoldina:
https://jira.lsstcorp.org/browse/DM-16667
A parallel PR is created for SUIT: https://github.com/lsst/suit/pull/13
Please, check it for steps to test.

- Added "LSST TAP" and "Upload" buttons.
- Made the list of TAP services a property.
- Added UCD to the popup with column selections for Spatial and Temporal column fields.
- Removed "schema_name" and "table_name" from columns constraints table, reordered columns into more logical sequence.
- Use ISO time constraint format if time column is not numeric. Even though ISO format is guaranteed to be converted to the native format only for timestamp fields (xtype="timestamp"), it is still more useful than MJD, which would break the query for char time fields.

Bug fixes:
- Schema and table descriptions in select boxes a little bit cut on the bottom. (Noticeable when looking at 'y' or 'p' characters.)
- Center columns not found when there there are unequal number of related fields for ra and dec columns, like there is variance for ra, but not for dec.
- Columns not loaded into schema tree in ADQL view.

### DIFF
--- a/src/firefly/js/ui/DateTimePickerField.jsx
+++ b/src/firefly/js/ui/DateTimePickerField.jsx
@@ -251,7 +251,7 @@ export function validateMJD(tStr, nullAllowed=true) {
     const retval = {valid: true, value: tStr};
 
     const min = Number.MIN_VALUE;
-    const max = Number.MAX_VALUE;
+    const max = 2973483.999988426;  // for ISO 9999-12-31T23:59:59Z
 
     retval.valid = (!tStr && nullAllowed) ? true : Validate.floatRange(min, max, 6, 'Time range in MJD', tStr).valid;
     if (!retval.valid) {
--- a/src/firefly/js/ui/DateTimePickerField.jsx
+++ b/src/firefly/js/ui/DateTimePickerField.jsx
@@ -208,6 +208,10 @@ function handleOnChange(momentVal, params, fireValueChange) {
     fireValueChange({value, valid, message});
 }
 
+function convertMomentToMJD(m) {
+    return m.isValid() ? `${(m.valueOf()/msecPerDay + DiffUnixMJD)}`: '';
+}
+
 export const DateTimePickerField = fieldGroupConnector(DateTimePicker, getProps, propTypes);
 
 const DiffUnixMJD = 40587.0;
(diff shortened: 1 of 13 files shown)


cwang2016:
just copy my comment from the parallel PR, lhttps://github.com/lsst/suit/pull/13

review completes. looks good!
Here are 2 concerns regarding the generation of ADQL on both ISO and MJD columns,

should we also consider to other possible time types? such as JD. (in the future).
the validation of MJD & ISO may need to be calculated on the consistent min/max bounds. Currently, if ISO entry is invalid, MJD will be set as an empty string and shown as valid, and vice versa. I concern that may result in inconsistent ADQL statement for ISO or MJD columns.
I will discuss with you on this next week.

tgoldina:
Cindy, thank you for your help. I've also modified validation to use consistent date range. Hopefully, this clears conversion inconsistency and either input format can be used as a source truth.
I don't see any problem adding Julian Date input in the future. (Choosing between MJD and JD in ADQL could be done based on the user selection.)

src/firefly/js/ui/DateTimePickerField.jsx
```diff
@@ -208,13 +208,19 @@ function handleOnChange(momentVal, params, fireValueChange) {
...
+
 export const DateTimePickerField = fieldGroupConnector(DateTimePicker, getProps, propTypes);
 
 const DiffUnixMJD = 40587.0;
 const DiffUnixJD = 2440587.5;
 const DiffMJDJD = 2400000.5;
 const msecPerDay = 86400000.0;
 
+const MIN_VAL_MJD = 1; // for ISO 1858-11-18T00:00:00Z
+const MAX_VAL_MJD = 2973483;  // for ISO 9999-12-31T00:00:00Z
```
cwang2016: It is good to have consistent validation range for both ISO & MJD. However, what if the user enters '0' for MJD, or '1858-11-17T00:00:00Z' or ' 9999-12-31T00:01:00Z' for ISO, those number will be validated as invalid. Actually those numbers are not that meaningful for temporal search, however theoretically how to explain that those numbers are 'invalid MJD' or 'invalid ISO'. We may inform the user the valid range in the documentation or the tooltip in UI.

tgoldina: The valid date-time range is random, we can always adjust it to something better. We give user the supported range in the error message next to the field ("date is out of range ... to ..."). I think this is good enough for now.

PR #771, opened 2019-03-18
Convert field group connector 

robyww:
Convert field group connector HOC to use Hooks

### DIFF
--- a/src/firefly/js/ui/FieldGroupEnable.jsx
+++ b/src/firefly/js/ui/FieldGroupEnable.jsx
@@ -1,0 +1,206 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {PureComponent, memo, useContext, useState, useEffect, useRef} from 'react';
+import PropTypes from 'prop-types';
+import {get,omit} from 'lodash';
+import {dispatchMountComponent,dispatchValueChange} from '../fieldGroup/FieldGroupCntlr.js';
+import FieldGroupUtils, {getFieldGroupState} from '../fieldGroup/FieldGroupUtils.js';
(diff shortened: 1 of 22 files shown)


src/firefly/js/ui/FieldGroupEnable.jsx
```diff
@@ -0,0 +1,116 @@
...
+        return {...initS, value: confirmInitialValue(initS.value,props,initS)};
+    }
+    const [fieldState, setFieldState] = useState(getInitialState());
+
+
+    useEffect(() => {
+        const {prevFieldKey, prevGroupKey}= infoRef.current;
+        if (prevFieldKey!==fieldKey || prevGroupKey !== groupKey) {  // called the first time or when fieldKey or groupKey change
+            let value= fieldState.value;
+            if (prevFieldKey && prevGroupKey) { // if field and group key changed, whole thing must reinit
```
loitly: I may not understand this right.  But, if the Component is keyed by `groupKey + fieldKey`, then this will happen without any logic.  It looks like you're trying to reuse the component when a new one could be created instead.

loitly:
k8s build: https://irsawebdev9.ipac.caltech.edu/fieldgroup/firefly/

src/firefly/js/ui/FieldGroupEnable.jsx
robyww: the problem is  that the properties can change on a component with out is going away.  This is a react thing.  We deal with in in the HOC now, but it is a log less readable.

robyww: The problem is that the properties can change on a component without it going away. This is a react thing. We deal with this now with the HOC now, but it is a log less readable.

PR #776, opened 2019-03-26
Firefly-50: update immutability-helper and replace react-addon-update

robyww:
Replace the deprecated library react-addons-update with immutability-helper, also upgrade immutability-helper to the most recent version (3.0).

Note- We have been using immutability-helper for 2 years but did not replace it in WebUtil.js because it had a bug in certain use cases. This bug appears to have been fixed a while back so I think it is safe to replace. This will fix the bug we were seeing with Atmos when they added Prototype.js.

https://irsawebdev9.ipac.caltech.edu/firefly-50-immutability-helper/firefly/

### DIFF
--- a/buildScript/webpack.config.js
+++ b/buildScript/webpack.config.js
@@ -183,6 +183,11 @@ export default function makeWebpackConfig(config) {
         );
     }
 
+    if (!ENV_DEV_MODE) { // Adding this so we see it in the log file of our builds
+        console.log('Building client with Global Props:');
+        console.log(globals.__PROPS__);
+    }
+
     const webpack_config = {
         name    : config.name,
         // mode    : 'none',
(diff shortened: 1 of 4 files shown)


buildScript/webpack.config.js
```diff
@@ -183,6 +183,11 @@ export default function makeWebpackConfig(config) {
         );
     }
 
+    if (!ENV_DEV_MODE) { // Adding this so we see it in the log file of our builds
+        console.log('Building client with Global Props:');
+        console.log(globals.__PROPS__);
+    }
+
```
loitly: In the need future, we should start adding a `-verbose` flag.  The additional details is useful, but makes it harder to identify warnings and errors from build.

robyww: Makes sense.

PR #777, opened 2019-03-26
IRSA-1093: vertical line appears due to an external outer <center> tag.

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-1093

Problem:  
See the attached image in the ticket.

Solution:
Add style to rootStyle to remove `center` tag affect.

Test:  https://irsadev.ipac.caltech.edu:9201/

Note: There's a side effect to this solution.  Although, this remove the unwanted vertical line, it also remove the `center` affect as well.  
In general, we do not want external style to affect our firefly component.  This is the reason why rootStyle were introduced.  
But, since IRTF is already in production.  I just wanted to point out the visual differences after this PR.

There's a ticket created already to provide more styling to a table.

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -29,9 +29,10 @@
     font-family: tahoma,arial,helvetica,sans-serif;
     font-size: 11px;
     line-height: 1;
+    text-align: initial;
 }
 
-.rootStyle, div.rootStyle>img, div.rootStyle > html {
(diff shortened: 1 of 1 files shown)


loitly:
`Just so i understand correctly, currently Firefly can't have both world fixed at the same time, having a centered alignment without a 'ghost' line?
`
Yep, having a `center` outside of Firefly will center everything, not just the table content.  So, I right thing to do is to pass style into the Table for table's specific styling.

PR #778, opened 2019-03-28
Firefly-51: Make FieldGroupConnector use function components with hooks

robyww:
Firefly-51: Make FieldGroupConnector use function components with hooks

  - Created a new field group connector hook
  - converted all `FieldGroup` related components to be functions with hooks
  - updated and cleaned up `NaifidPanel`, it needed a little work
  - updated many propTypes to include shape of initialState
  - removed value from some components props, replaced with initialState:{value}
  - a attempt to better define what should be in initial state
  - hook give a warning is value is passed as prop

_Code to look at_:
 
  - Look at `FieldGroupConnect.jsx`, look at the hook `useFieldGroupConnector`. Don't bother looking at the differences, just look at the new file.
  - All other files are pretty straight forward. You might want to look at a couple of files or focus on the components that you wrote. 
 - Before all code used the HOC `fieldGroupConnector`, now they use the hook `useFieldGroupConnector`. Notice how much easier it is follow how to setup the hook compared to the HOC.
 - @deannaji look at NaifidPanel.jsx.  I had to make some changes, we should discuss it soon.

_to test:_
https://irsawebdev9.ipac.caltech.edu/firefly-51-field-group-with-hooks/firefly/
https://irsawebdev9.ipac.caltech.edu/with-hooks/irsaviewer/

- Check console while testing, look for errors 
- Mostly everything should work as before
 - Critical Stuff:
    - Gator Catalog Search
    - Tap Search
    - Image Search Panel
    - workspace connection
    - naif resolving

### DIFF
--- a/src/firefly/js/ui/FieldGroupConnector.jsx
+++ b/src/firefly/js/ui/FieldGroupConnector.jsx
@@ -1,194 +1,152 @@
-import React, {PureComponent} from 'react';
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {useContext, useState, useEffect, useRef} from 'react';
 import PropTypes from 'prop-types';
-import {get,omit} from 'lodash';
+import {get,omit, isUndefined} from 'lodash';
 import {dispatchMountComponent,dispatchValueChange} from '../fieldGroup/FieldGroupCntlr.js';
 import FieldGroupUtils, {getFieldGroupState} from '../fieldGroup/FieldGroupUtils.js';
 import {flux} from '../Firefly.js';
 import {GroupKeyCtx} from './FieldGroup';
+import {isDefined} from '../util/WebUtil';
 
 const defaultConfirmInitValue= (v) => v;
+const defValidatorFunc= () => ({valid:true,message:''});
+
+
+
+const STORE_OMIT_LIST= ['fieldKey', 'groupKey', 'initialState', 'fireReducer',
+    'confirmInitialValue', 'forceReinit', 'mounted'];
+
+function buildViewProps(fieldState,props) {
+    const {message= '', valid= true, visible= true, value= '', displayValue= '',
+        tooltip= '', validator= defValidatorFunc, ...rest}= fieldState;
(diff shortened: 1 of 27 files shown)


src/firefly/js/ui/FieldGroupConnector.jsx
```diff
@@ -1,194 +1,175 @@
...
+ */
+export const fgMinPropTypes= {
+    fieldKey: PropTypes.string.isRequired,
+    groupKey: PropTypes.string,  // normally passed in context
+};
 
+/**
+ *
+ * Hook to connect a field to the FieldGroup Store. Pass the props object, make sure it includes the required props
+ * to connect to the store (fieldKey is the only requirement, see below). The hooks returns qn object with the new
```
tgoldina: qn object - typo?

src/firefly/js/ui/FieldGroupConnector.jsx
```diff
@@ -1,194 +1,175 @@
-import React, {PureComponent} from 'react';
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {useContext, useState, useEffect, useRef} from 'react';
 import PropTypes from 'prop-types';
-import {get,omit} from 'lodash';
+import {get,omit, isUndefined} from 'lodash';
```
tgoldina: isUndefined are unused.

src/firefly/js/ui/FieldGroupConnector.jsx
```diff
@@ -1,194 +1,175 @@
-import React, {PureComponent} from 'react';
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React, {useContext, useState, useEffect, useRef} from 'react';
```
tgoldina: Looks like no React components are created in this file any more. Should we keep it jsx?

robyww: Yes, I think it can be removed.  I think it is required if you use JSX but in the case it does not render anything.

PR #779, opened 2019-03-29
DM-18591 Bugs found in TAP demo

tgoldina:
https://jira.lsstcorp.org/browse/DM-18591
Test deployment: https://lsst-lsp-int.ncsa.illinois.edu/portal/suit/

The following bugs are fixed in this branch or in bug fix commits:
- Table info button does not show up.
- Legacy PDAC does not show up.
- Image metadata do not show up.
- Failed authorization on getting column metadata or images.

Broken chart options and suggest box not accepting input when switching to a different chart type were fixed in https://jira.ipac.caltech.edu/browse/FIREFLY-46 and should be working in the test deployment.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTMetaSearch.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTMetaSearch.java
@@ -1,9 +1,9 @@
 package edu.caltech.ipac.firefly.server.query.lsst;
 
 import edu.caltech.ipac.firefly.server.network.HttpServiceInput;
+import edu.caltech.ipac.firefly.server.network.HttpServices;
 import edu.caltech.ipac.table.io.IpacTableWriter;
 import edu.caltech.ipac.firefly.data.CatalogRequest;
-import edu.caltech.ipac.firefly.data.FileInfo;
 import edu.caltech.ipac.firefly.data.ServerRequest;
 import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.table.TableMeta;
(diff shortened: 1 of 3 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/lsst/LSSTMetaSearch.java
```diff
@@ -58,10 +54,10 @@ private DataGroup  getDataFromURL(TableServerRequest request) throws Exception {
         }
 
         //use fully specified path, which includes logical database, schema, and table name
-        //ex. URL http://lsst-qserv-dax01:5000/meta/v1/db/W13_sdss_v2/sdss_stripe82_01/tables/RunDeepForcedSource
+        //ex. URL http://lsst-qserv-dax01:5000/meta/v1/db/W13_sdss_v2/sdss_stripe82_01/tables/RunDeepForcedSource/
         String url = LSSTQuery.getMetaservURL() + URLEncoder.encode(parts[0], "UTF-8") + "/" +
                 URLEncoder.encode(parts[1], "UTF-8") + "/tables/"  +
-                URLEncoder.encode(parts[2], "UTF-8");
+                URLEncoder.encode(parts[2], "UTF-8") + "/";
```
loitly: Hmm.. haven't thought about needing to encode url path.  This rely on the resource provider doing the same.
Are you sure it's needed?

tgoldina: You are right. We should not use URLEncoder for path segments, only for query string. The set of reserved characters is different for path segments and query parameter values is different. I vaguely remember path segments like `"//lsst-qserv-master01:4040"` at some point (a workaround for missing schema name). URLEncoder was good enough for this.

PR #780, opened 2019-03-29
FIREFLY-24: Reenable projection tests in javascript

lznakano:
make new json file for projection test
   Enable the javascript unit test for projection
   Update the unit test to use assert instead of Chia moudles.
To test, please check out this branch in Firefly and the same named branch in firefly_test_data.
To run the unit test, please run this under hydro/cm/firefly:

 gradle :firefly:jsTest


lznakano:
If you clicked files change, it told you that that file was deleted.   I don't have it in the branch.

PR #781, opened 2019-04-03
IRSA-2626: dynamically computing the size of page number box

deannaji:
Making the size of "current page box" dynamic.

Jira: https://jira.ipac.caltech.edu/browse/IRSA-2626
K8s: https://irsawebdev9.ipac.caltech.edu/irsa-2626/applications/sofia/

Testing: 
- Run SOFIA and do an "All-Sky" search, with all levels (1,2,3,4) added. Then click search.
- Select 'Forcast' instrument.
- Navigate back and forth from the first to the last page.

### DIFF
--- a/src/firefly/js/ui/PagingBar.jsx
+++ b/src/firefly/js/ui/PagingBar.jsx
@@ -47,7 +47,7 @@ export class PagingBar extends Component {
                         style={{textAlign: 'right'}}
                         validator = {intValidator(1,totalPages, 'Page Number')}
                         tooltip = 'Jump to this page'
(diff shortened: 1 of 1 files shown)


ejoliet:
Just a comment here for testing it, the problem was shown in Chrome and Safari browsers. Firefox was ok somehow. I'll tested later.

deannaji:
Yes, but the fix will work on all browsers. Initially, FireFox was working ok as you mentioned, now the box is a slightly bigger (in FireFox) than the other tow browsers.

PR #783, opened 2019-04-03
DM-18690: implement maximum record limit in TAP query screens

tgoldina:
https://jira.lsstcorp.org/browse/DM-18690
Test build: https://irsawebdev9.ipac.caltech.edu/dm-18690_maxrec/firefly/firefly-dev.html?__action=layout.showDropDown&visible=true&view=TAPSearch

"Row Limit"  field is displayed at the bottom of the TAP Search, along with "Search" and "Cancel" buttons. It is propagated to the server and is translated into MAXREC parameter of the search request. We allow to disable it by clearing.

tap.maxrec.hardlimit property (default 10000000) is a config property. If MAXREC is set, it can not exceed this value.
defaultMaxrec is an application property (default 50000 for Firefly). It is the default value for MAXREC parameter.

Other changes:
- Filters are now shown by default in the TAP result tables
- Removed 'tables.showInfoButton' app property and made it a table option instead.
- Fixed Service URL select box style: the text was truncated a bit on the bottom and selected item was shown in white, which was hard to see on light green.

In a followup ticket, we might want to capture and propagate OVERLOAD indicator. It is passed via "results" resource INFO element, and can be displayed as a part of table info. We are currently not capturing INFO elements outside the Table element, but in most cases there is just one resource element with one table in it. In this case it makes sense for the table to store its parent's INFO.

### DIFF
--- a/config/app.config
+++ b/config/app.config
@@ -64,6 +64,9 @@ lsst.dax.metaservURL = "/api/meta/v1/db/"
 lsst.dax.imgservURL = "/api/image/v1/"
 lsst.dax.imgserv.repository = "DC_W13_Stripe82"
 
+// maximum number of table rows Firefly can support for TAP queries
+__$tap.maxrec.hardlimit = 10000000
+
 environments {
     local {
         BuildType = "Development"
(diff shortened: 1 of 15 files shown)


tgoldina:
I don't know how to make the text inside select box selectable. Setting user-select='text' does not help. I noticed that if you start selecting outside the select box, the text is selectable. But not when you start inside. It must be interfering with other handlers.

PR #784, opened 2019-04-03
Firefly-44: Make data product display work with ObsCore

robyww:
Firefly-44: Make data product display work with ObsCore

  - Recognize ObsCore tables
  - Show products if it is a FITS image and less than 1 GIG
  - Put in simple support for data link: only access_url so far
  - Data link tries show the #this semantics image if available
  - Give option to show grid of images when filtered down to image and/or cube

_To test:_
https://irsawebdev9.ipac.caltech.edu/firefly-44-obs-core-meta/firefly/firefly-dev.html

Load a ObsCore table: 
 - TAP Searches => **Tap Service**: CDAC or MAST => **Schema**: ivoa => **Table**: ivoa.obscore
 - Spacial search around small area e.g. 200 arcsec

What happens:
- When you click on a row with the data produce image or a cube it will attempt to load the image
- For CDAC it will do a Data Link indirection and if possible load the `#this` image.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/LockingVisNetwork.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/LockingVisNetwork.java
@@ -156,9 +156,9 @@ public class LockingVisNetwork {
             if (_key != null) {
                 String offStr = "";
                 if (ev.getMax() > 0) {
-                    offStr = " of " + FileUtil.getSizeAsString(ev.getMax());
+                    offStr = " of " + FileUtil.getSizeAsString(ev.getMax(),true);
                 }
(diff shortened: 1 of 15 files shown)


cwang2016:
Works well.
In addition to the issue as Gregory mentioned regarding the inconsistent highlight between the 8 images and the table, I found another highlight issue for the 8 image case - the highlight of the image (supposed to be the first image) and the table is not consistent right after filtering the table on 'dataproduct_type'.  This doesn't happen every time. Here is the case I found, 
TAP service: CADC, filter dataproduct_type on 'image', the result is that the 8th image and the first row of the table are highlighted.

robyww:
I think I have fixed some of the issues.  I think there is a bug in table highlight that I will have to get @loitly  to look at.  I am going to merge and work with @loitly  and the remaining issue when he get back.

PR #786, opened 2019-04-05
IRSA-1673:HiPS WCS grid disappears if "Match Image to HiPS" and EQ J2000-->Galactic

lznakano:
Fixed HiPs and image match issue

Test URL: https://irsawebdev9.ipac.caltech.edu/irsa-1637/firefly/

### DIFF
--- a/src/firefly/js/visualize/task/PlotHipsTask.js
+++ b/src/firefly/js/visualize/task/PlotHipsTask.js
@@ -35,6 +35,7 @@ import {resolveHiPSIvoURL} from '../HiPSListUtil.js';
 import {addNewMocLayer, makeMocTableId, isMOCFitsFromUploadAnalsysis, MOCInfo, UNIQCOL} from '../HiPSMocUtil.js';
 import HiPSMOC from '../../drawingLayers/HiPSMOC.js';
 import {doUpload} from '../../ui/FileUpload.jsx';
+import CoordinateSys from '../CoordSys';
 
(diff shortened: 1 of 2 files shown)


lznakano:
This ticket also fixed the issue reported in IRSA-1672.  I am sorry that I got a typo in the build name.  it should be irsa-1673, I typed to irsa-1637.

src/firefly/js/visualize/task/PlotHipsTask.js
```diff
@@ -496,13 +496,13 @@ export function matchHiPSToImage(pv, hipsPVidAry) {
         Object.entries(attributes).forEach( (entry) => dispatchAttributeChange(id, false, entry[0], entry[1]));
         dispatchAttachLayerToPlot(ImageOutline.TYPE_ID, id);
         dispatchChangeCenterOfProjection({plotId:id, centerProjPt:wpCenter});
-
+        dispatchChangeHiPS( {plotId:id,  coordSys:plot.imageCoordSys});
```
robyww: There is a slight problem here.  HiPS only supports `galactic` and `j2000`.  it is possible that that `plot.imageCoordSys` might be something like `b1950`.  So this line should be wrapped in a `if` to check if it is one of those two coordinate system. If not is should just keep the existing coordinate system.

lznakano: I added the if block as you suggested.  Thanks!

PR #787, opened 2019-04-10
Firefly-61: Headers are not passed for every cube plane, show correct cube headers

robyww:
Firefly-61: Headers are not passed for every cube plane, show correct cube headers

  - pass the correct cube headers
  - optimize size of returns
  - fixed crop file mouse readout bug
  - better feedback
  - large cube significantly faster
  - cube result json much smaller
  - standard result json a little smaller

_To Test:_

https://irsawebdev9.ipac.caltech.edu/firefly-61-cube-loading/firefly/

go to `firefly_test_data/multi-ext-fits` and read some file.

 - `sofia-10_ext-10130_images.fits` is the best test case since this file highlighted the bug.
 - read others such as `F0202_FI_IFS_0300595_RED_WXY-10-hdu-69-plane-cube.fits`
 - read other fits files.

Performance should be good.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotBuilder.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotBuilder.java
@@ -3,8 +3,8 @@
  */
 package edu.caltech.ipac.firefly.server.visualize;
 
-import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.data.FileInfo;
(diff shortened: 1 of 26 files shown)


cwang2016:
works very well and much faster.
Just be nitpicking, when we have cube header window open, could the window content looks no change (no flashing) when the plane number is updated as clicking the left and right arrow but not the HDU?

robyww:
> Just be nitpicking, when we have cube header window open, could the window content looks no change (no flashing) when the plane number is updated as clicking the left and right arrow but not the HDU?

Good suggestion. I will do it.

PR #791, opened 2019-04-16
FIREFLY-22 Fix the reader for FITS table or VOTable with fields of 'byte' or 'short' type

cwang2016:
This PR contains developments for  https://jira.ipac.caltech.edu/browse/FIREFLY-22 
- fix fits table reader and votable reader to handle the fields of type 'byte' or 'short' 
- fix the missing label on upload panel when URL is selected.

test 
start Firefly
upload the FITS sample from URL
http://web.ipac.caltech.edu/staff/ejoliet/gaia/query-sels/result.fits

k8s: https://irsawebdev9.ipac.caltech.edu/firefly-22/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
+++ b/src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
@@ -6,6 +6,7 @@ package edu.caltech.ipac.table;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 
 /**
  * Date: 6/15/18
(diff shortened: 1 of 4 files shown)


src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
```diff
@@ -20,17 +21,35 @@
...
     }
 
     default void checkType(Object val) {
-        if (val != null && !val.getClass().isAssignableFrom(getDataClass())) {
+        int vIdx = val == null ? -1 : numberTypeList.indexOf(val.getClass());
+        int dIdx = numberTypeList.indexOf(getDataClass());
+
+        // number type value or non number type value
+        if ((vIdx >= 0 && dIdx >= 0 && vIdx > dIdx) ||
+            (vIdx < 0 && val != null && !val.getClass().isAssignableFrom(getDataClass()))) {
```
loitly: We've talked about this, and I was unsure of the solution.  After looking at it now, I think the solution is to check at the point where the casting happens.  checkType() is not working nor is it needed.

src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
```diff
@@ -222,7 +241,7 @@ public Object get(int idx) {
         public void set(int idx, Object val) {
             checkType(val);
             ensureCapacity(idx);
-            data[idx] = val == null ? Integer.MIN_VALUE : (int) val;
+            data[idx] = val == null ? Integer.MIN_VALUE : getIntValue(val);
```
loitly: You can cast Byte or Short into int.  Will just have to re-arrange the code slightly so that we only do ensureCapacity when `val` can be casted.. i.e.
```
    int ival = val == null ? Integer.MIN_VALUE : (int) val;
    ensureCapacity(idx);
    data[idx] = ival;
```
Sorry if you have to remove your code.

src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
```diff
@@ -20,17 +21,35 @@
...
+    default int getIntValue(Object val) {
+        int toIdx = numberTypeList.indexOf(Integer.class);
+        int fromIdx = numberTypeList.indexOf(val.getClass());
+
+        if (toIdx == fromIdx) {
+            return ((Integer) val).intValue();
+        } else {
+            return (toIdx == 0) ? ((Byte) val).intValue() : ((Short) val).intValue();
+        }
+    }
```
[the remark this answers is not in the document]
cwang2016: I cast the return value as an 'int' since an Object return (even in fact it is a instanceof Byte or Short) can not be cast by 'int'.

PR #794, opened 2019-04-19
FIREFLY-67: Fixes TabPanel

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-67

Tests:  
https://irsawebdev9.ipac.caltech.edu/FIREFLY-67/firefly/
https://irsawebdev9.ipac.caltech.edu/FIREFLY-67/applications/sofia/

Additional changes:  https://github.com/IPAC-SW/irsa-ife/pull/63

TabPanel was not working in some cases.  It was also very complex in the way it handles its multiple use cases.
Changes:
- Refactor TabPanel
  - separate the different tab types using composition.
  -  convert to functional components with hooks
- Fixes Example Js Dialog 'third' tab prop validation error.
- Create replacement for SimplePanel using hooks.

Also:
I think `useFieldGroupConnector` relies on the view not getting unmounted.  When I use StatelessTabs for its view, onChange was not called.  Not a big deal, switching to Tabs(internal state) works fine.

### DIFF
--- a/src/firefly/js/ui/SimpleComponent.jsx
+++ b/src/firefly/js/ui/SimpleComponent.jsx
@@ -60,7 +60,7 @@ export function useStoreConnector(...stateGetters) {
                     setter(getter());
                 });
         });
-    }), [];     // only run once
+    }, []);     // only run once
 
     return rval;
 }
--- a/src/firefly/js/ui/SimpleComponent.jsx
+++ b/src/firefly/js/ui/SimpleComponent.jsx
@@ -54,14 +54,11 @@ export function useStoreConnector(...stateGetters) {
         return [getter, setter];
     });
 
-    const isMounted = true;
     useEffect(() => {
         return flux.addListener(() => {
-            if (isMounted) {
                 setters.forEach(([getter, setter]) => {
                     setter(getter());
                 });
-            }
         });
     }), [];     // only run once
 
--- a/src/firefly/js/ui/SimpleComponent.jsx
+++ b/src/firefly/js/ui/SimpleComponent.jsx
@@ -1,4 +1,4 @@
-import React, {PureComponent} from 'react';
+import React, {PureComponent, useEffect, useState} from 'react';
 import shallowequal from 'shallowequal';
 
 import {flux} from '../Firefly.js';
@@ -33,4 +33,37 @@ export class SimpleComponent extends PureComponent {
             this.setState(this.getNextState(this.props));
         }
     }
+}
+
+
+
+/**
+ * A replacement for SimpleComponent.
+ * This function make use of useState and useEffect to
+ * trigger a re-render of functional components when the value in the store changes.
+ *
+ * @param stateGetters  one or more functions returning a state
+ * @returns {Object[]}  an array of state's value in the order of the given stateGetters
+ */
+export function useStoreConnector(...stateGetters) {
+
+    const rval = [];
+    const setters = stateGetters.map((getter) => {
+        const [val, setter] = useState(getter());
+        rval.push(val);
+        return [getter, setter];
+    });
+
+    const isMounted = true;
+    useEffect(() => {
+        return flux.addListener(() => {
+            if (isMounted) {
+                setters.forEach(([getter, setter]) => {
+                    setter(getter());
+                });
+            }
+        });
+    }), [];     // only run once
+
+    return rval;
 }
--- a/src/firefly/js/ui/panel/TabPanel.jsx
+++ b/src/firefly/js/ui/panel/TabPanel.jsx
@@ -167,6 +167,7 @@ TabsView.defaultProps= {
 };
 
 
+/*----------------------------------------------------------------------------------------------*/
 /**
  * TabPanel with internal state
  * State will be lost after unload.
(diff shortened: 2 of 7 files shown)


src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -156,71 +159,74 @@ Tabs.propTypes= {
...
+        onTabSelect && onTabSelect(index,id,name);
+    }, []);
 
-        let tabClassName = 'TabPanel__Tab' ;
-        if (selected) {
-            tabClassName += ' TabPanel__Tab--selected';
+    useEffect( ()=> {
+        if (selectedIdx >= children.length) {
+            // selectedIdx is greater than the number of tabs.. update store's state
+            selectedIdx = children.length-1;
```
robyww: this is a constant being reassigned.

src/firefly/js/ui/SimpleComponent.jsx
```diff
@@ -34,3 +34,36 @@ export class SimpleComponent extends PureComponent {
...
+export function useStoreConnector(...stateGetters) {
+
+    const rval = [];
+    const setters = stateGetters.map((getter) => {
+        const [val, setter] = useState(getter());
+        rval.push(val);
+        return [getter, setter];
+    });
+
+    const isMounted = true;
```
robyww: I think you could remove isMount.  It is never false and I don't think it is need for functional components anyway.

src/firefly/js/ui/SimpleComponent.jsx
```diff
@@ -34,3 +34,36 @@ export class SimpleComponent extends PureComponent {
...
+    useEffect(() => {
+        return flux.addListener(() => {
+            if (isMounted) {
+                setters.forEach(([getter, setter]) => {
+                    setter(getter());
```
robyww: don't you want to check for equality here like you use to?

src/firefly/js/ui/panel/TabPanel.jsx
loitly: thanks..  eslint didn't catch it.  did the error show up for you?

src/firefly/js/ui/SimpleComponent.jsx
loitly: I was pretty sure I was getting the setState on unmounted error before when I attempted to do this last time.  Googled and the recommendation was to do this.  But, I will try again.  We can easily add back if the error shows up again.

src/firefly/js/ui/SimpleComponent.jsx
loitly: It's not needed since React will do an Object.is() on the value and only trigger render when it return false.

src/firefly/js/ui/panel/TabPanel.jsx
robyww: it show up through intellij not ESlint.  We probably should have some eslint rule.

src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -156,71 +159,74 @@ Tabs.propTypes= {
...
+/**
+ * TabPanel with ComponentCntlr supported state
+ * Selected state is stored as <componentKey>.selectedIdx
+ */
+export const StatefulTabs = React.memo( (props) => {
+    const {children=[], defaultSelected=0, onTabSelect, componentKey} = props;
+
+    let [selectedIdx = defaultSelected] = useStoreConnector( () => {
+        return get(getComponentState(componentKey), 'selectedIdx');
+    });
```
tgoldina: I have a hard  time understanding what's happening here. We want selectedIdx state to be reset on every store update? Why not explicit useState here and set it in onSelect?

src/firefly/js/ui/SimpleComponent.jsx
```diff
@@ -34,3 +34,33 @@ export class SimpleComponent extends PureComponent {
...
+        return [getter, setter];
+    });
+
+    useEffect(() => {
+        return flux.addListener(() => {
+                setters.forEach(([getter, setter]) => {
+                    setter(getter());
+                });
+        });
+    }), [];     // only run once
```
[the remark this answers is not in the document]
loitly: you're right.  I'll fix it and redo my tests.

src/firefly/js/ui/panel/TabPanel.jsx
loitly: Oh, it's because this will be a common use case.  That's why I created `useStoreConnector`.  This is a convenience way to listen/react to store changes.  This one function allow you to use one or many states connected to the `reduc store` and help cleanup the listener once unmount.

loitly: I think the case you're talking about is `Tabs`, line `ui/panel/TabPanel.jsx:175`.  It uses internal state.. with useState.  Does this answer your question?

tgoldina: Not quite. I though using store connector might be an overkill in this case, because you can set selectedIdx state in onSelect below, right after you do dispatchComponentStateChange. Because this is the only place selectedIdx should be changed. Or I am missing something?

PR #797, opened 2019-04-25
FIREFLY-71: MultiProductViewer component

robyww:
This component allows us to take in a table that represents any type of data product and show the product. A sample implementation uses obs core but it is made for any type of data source.

  - refactor to ImageMetaDataWatcher.
  - Created new DataProductsFactory
  - improved some image return feedback and messaging
  - Now supports menu of data products
  - support multiple product types
  - support both tables, png, images, and download

Much of this work is really positioning firefly for products. There is still some issues that are not yet done and need future tickets:

- chart support, this is easy but I have not chart test data.
- the table handling the 'more...' dropdown. 
- the obscore converter could be worked on significantly more to get better at recognizing data but I did enough to demonstrate the potential.

Next steps:

- implement a data products converter for Sophia and 
- implement a data products converter Spitizer (if we do the convertions)
- Improve and generalize the data product converter for Datalink.

How it works in a nutshell:
`DataProductsFactory` recognizes a table type and produces a converter for that table. The converter will produce an object that knows how to load a data type using an activate function that it creates. It can also produce a menu of other data that the row represents. The `MultiProductViewer` will change to the data type of call an `activate` function that will do the data load. It will also show the menu of other options.

Key files to look at.

- `MultiProductsViewer.jsx`
- `ImageMetaDataWatcher.js`
- `DataProductsFactory.js`

_To test:_

https://irsawebdev9.ipac.caltech.edu/firefly-71-data-products/firefly/firefly-dev.html

Do an obscore search with the left hand side should show various data products.  Also with CADC you will see a `More...` menu that will allow you to switch your data products.


robyww:
> When selecting Show Table in More, More.. button disappeared

I mentioned that as future work.  The dropdown is pass as a parameter to the component. Table container does not yet support it.

robyww:
> The More... button may need more decoration. It does not look like an actionable icon/button.

I need some ideas for that.  Maybe we can talk on Monday.

ejoliet:
Awesome!! I played around and it's very useful!
That gives a lot of flexibility and gives much more insight of the data content than we used to have. I guess this is the power to leverage ObsCore/DataLink, right? What happen with other tables non-ObsCore?

I saw a minor issue that could become an improvement: when one of the file i selected had a table in it and the default preview was an image (or a cube, doesn't matter), the table selected is displayed but the "More..." button disappeared. It reappears if i close the table. I was expecting to keep selecting and changing to another HDU. 

I guess we could discuss several way to go from there:
- always have the button "More..." (or different label) in the top toolbar?
- "More..." could be either: 
 - replace the viewer with the selected item OR
 - add the item to the viewer

What do you think?

Question? Can i set from the API  the HDU to be displayed by default?  

So far is impressively nice and useful feature!

robyww:
> More...

@ejoliet There are several things we can do with the UI with the `More...`. Though show different types of data in the same view would be messy and probably not the best approach. I know the work with the `More...` is incomplete

With this ticket I purposely did not try to do too much UI because the ticket or more the infrastructure.
I just did the `More...` to demonstrate capability.  A common idea that @gpdf has had is to put the dropdown in the table rows. 

I feel we should do some UI design and consider what is the best to do.  However, we are going to have to wait until we have the priority to work on it.  As it is now it is well positioned for future work with obscore, sofia, and spitzer.

> Question? Can i set from the API the HDU to be displayed by default?
 
The `DataProductsFactory` recognizes a table and creates a `DataProductsConvertType`. We can add a `DataProductsConvertType` looks at the table and does anything we want. i.e. _For x type of sofia table always plot the HDU 3 image._

loitly:
Tested the new changes.  It looks good.  No more crashing when `dataproduct_type` is blank.
However, I noticed a `double-toolbar` appearing when you expand then collapse an image.  Same thing happens when you use the `More...` button to load up the first entry(#this).

These are not a big deal, since the PR is more about the capability.  Feel free to merge when ready.

PR #798, opened 2019-04-29
Firefly-68:Firefly can't display SOFIA FITS cube (4-d)

lznakano:
Please using the testing data attached in the jira ticket: Firefly-68 to test this pull request.

URL: https://irsawebdev9.ipac.caltech.edu/firefly-68/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/plotdata/FitsReadUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/plotdata/FitsReadUtil.java
@@ -243,21 +243,44 @@ public class FitsReadUtil {
         header.resetOriginalSize();
     }
 
-    
 
-    private static BasicHDU[] splitFitsCube(BasicHDU inHdu) throws FitsException {
-
+    private static BasicHDU[] splitFits3DCube(BasicHDU inHdu, float[][][] data32) throws FitsException {
         ImageHDU hdu = (inHdu instanceof ImageHDU) ? (ImageHDU) inHdu : ((CompressedImageHDU) inHdu).asImageHDU();  // if we have to uncompress a cube it could take a long time
-        float[][][] data32 = (float[][][]) ArrayFuncs.convertArray(hdu.getData().getData(), Float.TYPE, true);
-
         BasicHDU[] hduList = new BasicHDU[hdu.getHeader().getIntValue("NAXIS3", 0)];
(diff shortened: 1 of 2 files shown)


lznakano:
When I tried the build, I got the same issue.  I can not even load 3-d cube.  But it worked before.  I am investigating the cause now.

lznakano:
I found the cause but don't understand it.  I just deleted a not-used line in a loop.  I believe that there are some memory issue.  The SPOT_PL is defined in the header for the Cube data, but the client side was not able to get it correctly.    I pushed my old version back to let @ejoliet test it.

ejoliet:
Yes! I've tested and now it works. I've added @robyww here because he might need to have a heads-up for his tickets about cube FITS and headers.
One thing didn't work and might be great to identify why is that when i upload the sample file `Cycle4_GR_OT_04_0038_CKramer_CII.lmv.fits` , some image features didn't work: cropping, footprint, compass rose but grid and marker overlay works. Any idea? Probably need to file another ticket for this although it might just be that the file doesn't have proper WCS which we won't support for now. Could you investigate it?

lznakano:
After investigating the issue, now I understood the cause.  In splitFitsCube method (FitsReadUtil),  the line
" int bitpix = header.getIntValue("BITPIX", -1);" is not used. So I deleted it.  After deleting it, my pull request stopped working.  The reason is that when this line is there, the pointer in the Header pointed to BITPIX, thus, a new header card is added before this pointer.  After the line is removed, the header point is pointing to the end of the header.  Thus, the new header card is added after the "END" header key.  Thus, the SPOT_PL is undefined in the UI.   I also found some inconsistence.  The default planeNumber in FitsRead is set to 0.  But the first plane in a cube is also 0.  I saw some place the fitsRead.getPlaneNumber() method is used and some place the header.getInitValue("SPOT_PL") is used.

lznakano:
I can confirm that both adding a footprint and cropping are failing.  The reason is failing is that the drawData is null in drawLayer.  More investigation is needed to pin point the issue.  I changed the naxis=4 to only  naxis4=1.  When naxis>4 and naxis4>1, the exceptions are thrown.  Since the cropping and footprints are not working.  I don't think that we should merge this pull request.

lznakano:
I added my analysis in the ticket.  The footprint issue is caused by the SFL projection definition.  The cropping for cube is not implemented. The current implementation only works for multi extension FITs file. The cube images are also multi images without multi extension.

PR #800, opened 2019-05-02
FIREFLY-31: make any popup panel text selectable and copyable

ejoliet:
This small change is to allow user to select and copy text from a popup panel and dialog which was requested in those other tickets:
IRSA-1613 (marker coordinates not copyable) and FIREFLY-65 (hips properties info dialog).

Please test that any dialog or popup text can be selected and copied and confirm.


ejoliet:
Ok, thanks. I'll make a ticket to address the concern in the future.

PR #805, opened 2019-05-16
IRSA-2421: retain table states between updates including mount/unmount

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-2421
Test: https://irsawebdev9.ipac.caltech.edu/IRSA-2421/firefly/

copied from ticket:

> This applies both to client-side and server-backed tables
> In Firefly tables, sorting, filtering, or removing a column resets all of the following:
> horizontal scroll
> vertical scroll
> column widths
> None of these extra things should happen.
> 

Also:
- Convert BasicTableView into a React function component with hooks.
  - Remove the use of `UNSAFE_componentWillReceiveProps`

Things to test for:
Make sure table works as before beside scroll position and column widths.
Scroll position and column widths should remain the same between table updates.  This includes sort, filter, show/hide columns, mount, and unmount.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
@@ -223,7 +223,7 @@ public class TableServerRequest extends ServerRequest implements Serializable, C
      * by '='.  If the keyword contains a '/' char, then the left side is
      * the keyword, and the right side is its description.
      * @return the serialize version of the class
-     * @deprecated
+     * @deprecated  use JSON format instead.  @see edu.caltech.ipac.table.JsonTableUtil#toJsonTableRequest
      */
     public String toString() {
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
@@ -30,6 +30,7 @@ public class TableServerRequest extends ServerRequest implements Serializable, C
     public static final String INCL_COLUMNS = "inclCols";
     public static final String FIXED_LENGTH = "fixedLength";
     public static final String META_INFO = "META_INFO";
+    public static final String META_OPTIONS = "META_OPTIONS";
     public static final List<String> SYS_PARAMS = Arrays.asList(REQUEST_CLASS,INCL_COLUMNS,SORT_INFO,FILTERS,PAGE_SIZE,START_IDX,FIXED_LENGTH,META_INFO,TBL_ID);
     public static final String TBL_INDEX = "tbl_index";     // the table to show if it's a multi-table file.
     public static final String SELECT_INFO = "selectInfo";  // a property of META_INFO.  deserialize into SelectionInfo.java
@@ -39,6 +40,7 @@ public class TableServerRequest extends ServerRequest implements Serializable, C
     private ArrayList<String> filters;
     private SelectionInfo selectInfo;
     private Map<String, String> metaInfo;
+    private transient Map<String, String> metaOptions;               // options to apply for this request.  it will not persist nor returned to the client
 
     public TableServerRequest() {
     }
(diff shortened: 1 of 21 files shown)


loitly:
> Test case: m31, 500 arcsec, click on the leftmost chart point when the tri-view comes up.

This is how it was before.  You can use the nightly build to confirm.  https://irsawebdev9.ipac.caltech.edu/nightly/firefly/

`scrollToRow` is a feature provided by fixed-data-table-2.  I don't have much control over that.

tgoldina:
Indeed. I don't quite understand why in some cases the last visible row highlight is out of view qhile in the others it's in the view.

Anyway after this ticket highlight might not be visible at all. For example, if you scroll and then sort, first row will be highlighted and scroll position preserved. This means the highlighted row will be out of view.  I am not sure if this is ideal behavior. 

For filter and sort, we probably want to preserve horizontal scroll position, but may be not vertical. For visible column changes we do want to preserve vertical scroll position, but we might not care about horizontal.

loitly:
I agree the behavior is not quite right after sort and filter.  But, I did it exactly as described in the ticket.  I don't mind changing the behavior of highlighted row after sort and filter.  But, I think there's a separate ticket for that.  I'll ask this same question in the ticket and see what the response is.

gpdf:
@vandesai1 and I attempted to respond on the original ticket to the above question, as well as to other complications that occurred to us.

I'll talk to @loitly on Monday to see what's reasonable to implement on this ticket vs. later.

ejoliet:
@gpdf @vandesai1 what was the verdict no Monday? Can @loitly merges this one and fix the highlight out of sync problem in another ticket you raise? Thanks.

loitly:
The ticket is updated after my discussion with @gpdf.  Please look at it again to see what the changes are.  https://jira.ipac.caltech.edu/browse/IRSA-2421
A new commit is added to reflect the changes.  You can test it here: https://irsawebdev9.ipac.caltech.edu/IRSA-2421/firefly/

gpdf:
I reviewed the behavior (not the code) with @loitly just now.  We observed one bug: it appears that sorting and filtering currently reset the horizontal scroll to the leftmost position.  Otherwise, thumbs up.  Please let me know when the next build is available; we only looked at the client-side table behavior so far.

loitly:
@gpdf, the bug you found is fixed and deployed.
https://irsawebdev9.ipac.caltech.edu/IRSA-2421/firefly/

loitly:
Fixed a bug identified by @tgoldina and also implemented a workaround for fixed-data-table's scrollToRow feature because it sometimes hides the highlightedRow behind the scrollbar, making it nearly invisible.
test deployment updated.

loitly:
@tgoldina, Sorry, I made another commit to fix the filtering bug you found.  I also fixed a couple of other things.  Please review.

src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
```diff
@@ -222,6 +223,7 @@ public static TableServerRequest parse(String str) {
      * by '='.  If the keyword contains a '/' char, then the left side is
      * the keyword, and the right side is its description.
      * @return the serialize version of the class
+     * @deprecated
      */
```
tgoldina: When something is deprecated, we might want to add a note what to use instead.

src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
```diff
@@ -30,6 +30,7 @@
     public static final String INCL_COLUMNS = "inclCols";
     public static final String FIXED_LENGTH = "fixedLength";
     public static final String META_INFO = "META_INFO";
+    public static final String META_OPTIONS = "META_OPTIONS";
```
tgoldina: Just curious, why you decided to introduce META_OPTIONS rather than using META_INFO? META_INFO is something that is not supposed to be returned to the client, while META_OPTIONS are not returned? So the new highlighted Idx will be returned in META_INFO?

loitly: META_INFO is returned to the client as part of the table's request and is used whenever you need to modify the table.  I introduce META_OPTIONS to pass along information that is needed once and should not be used in subsequent requests.  So yes, META_INFO is returned to the client while META_OPTIONS is not.

PR #812, opened 2019-05-23
FIREFLY-83_moving_object_resolver_improvements

deannaji:
Sorting the suggestion list of the moving objects (NAIFID) in descending order. See ticket:
https://jira.ipac.caltech.edu/browse/FIREFLY-83 

K8s:https://irsawebdev9.ipac.caltech.edu/firefly-83/applications/sofia/

To Test:
- Select 'Moving Object'
- Type 2, the suggestions list sorting values from highest to lowest.

### DIFF
--- a/src/firefly/js/ui/NaifidPanel.jsx
+++ b/src/firefly/js/ui/NaifidPanel.jsx
@@ -42,9 +42,10 @@ class NaifidPanelView extends PureComponent {
         if (!rval.p) return [];
         //if value has been searched previously, no need to call the api again.
         if(searchHistory.length > 0){
-            const cachedSuggList = Object.entries(searchHistory).find(([k,v])=>(v.searchVal === val));
-            if(cachedSuggList){
-                return sortBy(Object.entries(cachedSuggList.pop().searchRes).map( ([k,v]) => ({name:v.naifName, naifid:v.naifId})), 'naifid').reverse();
+            const cachedSuggList = Object.values(searchHistory).find((v)=>(v.searchVal === val));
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -44,15 +44,15 @@ class NaifidPanelView extends PureComponent {
         if(searchHistory.length > 0){
             const cachedSuggList = Object.entries(searchHistory).find(([k,v])=>(v.searchVal === val));
             if(cachedSuggList){
-                return Object.entries(cachedSuggList.pop().searchRes).map( ([k,v]) => ({name:v.naifName, naifid:v.naifId}));
+                return sortBy(Object.entries(cachedSuggList.pop().searchRes).map( ([k,v]) => ({name:v.naifName, naifid:v.naifId})), 'naifid').reverse();
```
loitly: nitpicking:  this should be written like this...
- The `if` ensure you don't get TypeError on the next line.
- Using .values instead of .entries because you don't use the key.
- Separating the array creation and the sorting makes is easier to read and debug.
```
            if (get(cachedSuggList, '0.searchRes.length') > 0) {
                const namedNaifIdAry =  Object.values(cachedSuggList.pop().searchRes).map( (v) => ({name:v.naifName, naifid:v.naifId}));
                return sortBy(namedNaifIdAry, 'naifid').reverse();
            }

```

deannaji: Fixed, thanks!!

PR #819, opened 2019-06-05
IRSA 2895: TPV projection fixed in JS/Java

ejoliet:
This PR fixes the offset seen in the overlay projection when FITS image is using TPV type (from ZTF dataset for example).
I had to force push this branch because of a bad merge beginning this week in dev. 
But the fix was made by @stargaser.
I've made the same changes in Java. (Still need to do it in gwt, will make another branch for ZTF GUI itself).

I've also updated the json sample header/data and java side with the updated tests.

NOTE:
To run `gradle test` (js or java), please use together with `firefly_test-data/IRSA-2895-debug-test-tpv` branch.

Test the UI: use the IRSAViewer deployed version here:
https://irsawebdev9.ipac.caltech.edu/irsa-2895/irsaviewer/

And upload the image with `ztf-` prefix in the repos/branch (`firefly_test_data/edu/caltech/ipac/visualize/plot/projection`) mentioned above or any of your favorite ZTF image:

Then do a search on Gaia catalog that cover the image (polygon type for example), then check that the overlay catalog boxes are positioned on top of the stars/object apparent in the image. (a way to see more contrast is to stretch it with sigma -2,10 range for example)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/projection/TpvProjection.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/projection/TpvProjection.java
@@ -4,6 +4,8 @@ import edu.caltech.ipac.visualize.plot.ProjectionException;
 import edu.caltech.ipac.visualize.plot.ProjectionPt;
 import edu.caltech.ipac.visualize.plot.Pt;
 
+import static edu.caltech.ipac.visualize.plot.projection.Projection.DtoR;
+
 
 /**
  * See ref https://fits.gsfc.nasa.gov/registry/tpvwcs.html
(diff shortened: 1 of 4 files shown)


slgroom:
The overlay alignment on my quick checks look very good, but I am now seeing an east-west flip in the image as indicated by the compass overlay. Still looking.

ejoliet:
Yes, @stargaser noticed it too.
But remember, GWT is still not correct. The overlay in GWT version (Java) is only correct if the image is north-up, otherwise, the offset is apparent.

ejoliet:
Thanks @stargaser ! I've rebuild and redeploy the fix in same url for reviewer to test. (I've also copied the fix over to Java/GWT and is still failing though...)

PR #820, opened 2019-06-06
FIREFLY-136: Support right justification of numeric columns in Firefly table

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-136

Changes:
- right-align columns with numeric types.  This is based on the column's type and not its values.
- for links:
   - when there's only one link, align it based on its link `text`.
   - When there's more than one, left align.

To test:
https://irsawebdev9.ipac.caltech.edu/FIREFLY-136/
Make sure all tables are behaving this way.  For links, look at test #6 of this page.
https://irsawebdev9.ipac.caltech.edu/FIREFLY-136/firefly/test/tests-main.html#test-6

### DIFF
--- a/src/firefly/html/test/tests-main.html
+++ b/src/firefly/html/test/tests-main.html
@@ -182,9 +182,9 @@
     <div id="expected" >
         <div>Using META-INFO API to show cell value as links
             <ul class='expected-list'>
-                <li><b>band</b> should be shown as links</li>
-                <li>first one links to irsa</li>
-                <li>second links to ivoa with constant text</li>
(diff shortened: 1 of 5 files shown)


loitly:
> Minor comment about the test 6: i did noticed that the in_row_id value is always 1, i was expecting different values.

This is "input" row ID.  This table must be the results of a single row multi-position search.   The values of that column's data is all 1s.  It is expected behavior.

xiuqin:
Tested by uploading a file and the results look good. Thanks!

PR #823, opened 2019-06-11
FIREFLY-49: Table filter shows 'null' in the column category but fails to filter

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-49
There's a test table file attached to the ticket.  It has a variety of null situations.
You may upload it to test:  https://irsawebdev9.ipac.caltech.edu/FIREFLY-49/firefly/

- substitute NULL with %NULL so it can be used in an IN condition
- show NULL as <NULL> and empty string as <EMPTY_STR>
- added same behavior to client table
- added IS and IS NOT support for client table
- added multiple tests
- added clear link to column's enum values drop-down
- fix AsyncTapQuery not failing on UNKNOWN returned state

BTW, 
A good client side table to test is Images -> View HiPS Images
Also, a good server-side table to test.  
https://irsawebdev9.ipac.caltech.edu/FIREFLY-49/firefly/firefly-dev.html
  TAP Searches -> ADQL -> CADC -> ivoa.ObsCore


loitly:
https://github.com/Caltech-IPAC/firefly/pull/823#pullrequestreview-248498303
@cgelino , thanks for the testing.  I've fixed the problems reported and the test link above is updated with the changes. 

> The table is returned with the properly filtered rows, but now the row with the filter text boxes is missing. The funnel icon is set to clear the filters. You have to clear the filters and click the funnel again to bring back the filter text boxes.

The hidden filtered row is fixed.  However, you don't have to clear filters to show/hide the filter header.  The funnel with a crossed out circle is for clearing.  It only appears when there are filters applied to the table. The funnel icon with or without a numbered badge is to show/hide the filtered row.  It will always be there next to the `TEXT` icon.

loitly:
@olgapevunova Thank you for testing.  In regard to _null_,  NullString value of this table is _null_.  Where you see _null_, it means NULL_VALUE.  When you filter on `is null` or `is NULL`, it will returns rows where **objname** contains NULL_VALUE.
The reason why you don't see NULL is because NULL in this case is a string "NULL".

As for `is` operator, it can only be used on null condition.  If you want to filter on rows where **objname** is 'm41', you would say:  `= 'm41'`  for exact match or `like '%m41%'` for partial match.

Does that answer all of your questions?

@cgelino if my latest commit satisfy your change request, please approve.  Thanks.

cgelino:
The bugs I had reported previously have been fixed.

PR #824, opened 2019-06-11
IRSA-2894: ZTF images flipped east-west in time series tool

stargaser:
ZTF images are mirror-reversed with respect to the sky, with North down and East to the left. Account for this case by computing the cross product of local North and East vectors, and taking the sign to know when to flip the compass.

Link to ticket in JIRA: https://jira.ipac.caltech.edu/browse/IRSA-2894

### DIFF
--- a/src/firefly/js/drawingLayers/NorthUpCompass.js
+++ b/src/firefly/js/drawingLayers/NorthUpCompass.js
@@ -164,7 +164,14 @@ function makeCompass(plotId, action){
     var zf= cc.zoomFactor || 1;
     var wpt2= makeWorldPt(wpStart.getLon(), wpStart.getLat() + (Math.abs(cdelt1)/zf)*(px), CoordinateSys.EQ_J2000);
     var spt2= cc.getScreenCoords(wpt2);
-    var sptE2 = getEastFromNorthOnScreen(cc, sptStart, spt2, Math.PI/2);
+    var wpt3= makeWorldPt(wpStart.getLon() + (Math.abs(cdelt1)/(Math.cos(wpStart.getLat()*Math.PI/180.)*zf))*(px),
+                           wpStart.getLat(), CoordinateSys.EQ_J2000);
+    var spt3= cc.getScreenCoords(wpt3);
+    // don't use spt3 because of funny effects near the celestial poles
+    // the sign of the cross product of compass vectors tells us if the image is mirror-reversed from the sky
+    var cross_product= (spt3.x - sptStart.x)*(spt2.y - sptStart.y) -
+                       (spt3.y - sptStart.y)*(spt2.x - sptStart.x);
+    var sptE2= getEastFromNorthOnScreen(cc, sptStart, spt2, Math.sign(cross_product));
 
     if (sptStart===null || spt2===null || sptE2===null) {
         return null;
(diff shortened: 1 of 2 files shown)


ejoliet:
I will build IRSAViewer so we can test it with different images from different projects.

ejoliet:
Version to test is deployed here: https://irsawebdev9.ipac.caltech.edu/irsa-2894/irsaviewer/

stargaser:
@slgroom noticed that in the magnifier view, the compass has the wrong orientation in ZTF images. I'll try to apply the same fix to `src/firefly/js/visualize/ui/ThumbnailView.jsx`.

cgelino:
A few other things I should have noted before.
1) The compass rose doesn't always show on the image even if the button in the menu bar indicates that option is active.  I haven't been able to figure out how to make it happen reliably.  It just seems random.
2) An image's orientation is reverted back to its original state (North down for ZTF) after turning off WCS matching.  This happens even if you put North up prior to the WCS matching.  I don't think the image orientations should change when you deactivate WCS matching.  I guess I think of a WCS match as a one time action.  Whether or not they stay matched after that depends on the user.
3) Sort of related to 2).  Bring up three images of the same field (M5 is a good one): 2 from ZTF and 1 from WISE.  Click on the WISE image to activate and then WCS match the ZTF images to it.  The ZTF images are rotated as expected.  Turn off the frame "locking" so that each image can move independently.  Move one of the ZTF images.  Now activate the other ZTF image (the one you did not move).  Turn off WCS matching and notice the activated ZTF image is rotated back to the original orientation and the ZTF image that you moved still has North up.

ejoliet:
Thanks.

> * The compass rose doesn't always show on the image even if the button in the menu bar indicates that option is active.  I haven't been able to figure out how to make it happen reliably.  It just seems random.
I've also noticed that and the fact that the compass is positioned randomly too, when it should stay in upper corner or any place consistently in the image frame. 
Unfortunately, this happens in OPS so it wasn't introduced in the PR.
I'll make a ticket (minor priority though, compare to others things).

> * An image's orientation is reverted back to its original state (North down for ZTF) after turning off WCS matching.  This happens even if you put North up prior to the WCS matching.  I don't think the image orientations should change when you deactivate WCS matching.  I guess I think of a WCS match as a one time action.  Whether or not they stay matched after that depends on the user.
> * Sort of related to 2).  Bring up three images of the same field (M5 is a good one): 2 from ZTF and 1 from WISE.  Click on the WISE image to activate and then WCS match the ZTF images to it.  The ZTF images are rotated as expected.  Turn off the frame "locking" so that each image can move independently.  Move one of the ZTF images.  Now activate the other ZTF image (the one you did not move).  Turn off WCS matching and notice the activated ZTF image is rotated back to the original orientation and the ZTF image that you moved still has North up.

Those interactions/issues related to WCS match and lock are probably going to change because of FIREFLY-75 updates - see #825 .
Will be good to test though.

slgroom:
- Seems to me that the compass was positioned on some part of the *image*,
not the *display*. so after zoom etc, it may go out of view. Would
be better to just park it in some corner of the *display* and
leave it there, not pan/zoom with the image.

- when wcs-match gets turned off, images should stay as they are, they should not
instantly change rotation/position/zoom when match is switched off. I know there
is some confusion about meaning of match vs lock, but when we turn off match
please don't change the layout.

ejoliet:
>- Seems to me that the compass was positioned on some part of the *image*,
not the *display*. so after zoom etc, it may go out of view. Would
be better to just park it in some corner of the *display* and
leave it there, not pan/zoom with the image.

I've filed ticket FIREFLY-144.

robyww:
@slgroom 
> - when wcs-match gets turned off, images should stay as they are, they should not
instantly change rotation/position/zoom when match is switched off. 

See ticket https://jira.ipac.caltech.edu/browse/FIREFLY-75 now a PR #825. You are welcome to test it. https://irsawebdev9.ipac.caltech.edu/firefly-75-wcsmatch/firefly/

PR #828, opened 2019-06-20
FIREFLY-59: Filtering fail when value contains semicolon

loitly:
ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-59
test:    https://irsawebdev9.ipac.caltech.edu/FIREFLY-59/firefly/firefly-dev.html

As reported, filtering fail when the value contains semicolon.  Because we were using semicolon to separate conditions, this messes up the underlying query statement.

Instead of just fixing the problem, I went ahead and allow the conditions delimiter to be either `AND` or `OR`.  Although this added a useful requested feature, it also may require updates to our online help.  Please coordinate with the appropriate people after merging.

Sample conditions:
`> 0 and < 50 or is NULL`
`< 0 or > 100`
`in ('m31', 'm41') or is NULL`

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/TableServerRequest.java
@@ -42,6 +42,8 @@ public class TableServerRequest extends ServerRequest implements Serializable, C
     private Map<String, String> metaInfo;
     private transient Map<String, String> metaOptions;               // options to apply for this request.  it will not persist nor returned to the client
 
+    private static String FILTER_SEP = "_AND_";         // need to match what's defined in FilterInfo.js
+
     public TableServerRequest() {
     }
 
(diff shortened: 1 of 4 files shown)


vandesai1:
Just tried it. Looks good.

xiuqin:
Definitely a good improvement using 'AND' 'OR' for conditions.  This was something we wanted to add for some time. Great job!

loitly:
> 2\. Putting a string in double quotes instead of single quotes has some weird consequences.  Sometimes the string is ignored (is "blah" -> is NULL) and sometimes it tries to be parsed as is (= "blah" -> ='"blah"').

- `is` and `is not` can only be used with NULL.  That's why auto-correct replace everything after that with NULL.  This is something I am not sure of.  I can either auto-correct, or let is error out when it's not NULL.
- String literal must be enclosed in single-quotes.  When it's not, auto-correct will add it.

PR #829, opened 2019-06-24
FIREFLY-53: Saving a table with large number of columns results in HTTP 400 (Bad Request)

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-53

As described in the ticket, when the URL length exceed its limit, table save will get a `Bad Request` error.  In this PR, the logic is to switch to a HTTP `post` method to avoid the limit constraint.  I've also refactored Image's PNG save to make use of common code.
Please include that in your testing as well.

Test: https://irsawebdev9.ipac.caltech.edu/FIREFLY-53_bad_request/firefly/

Things affected:  tables and images save/download.

### DIFF
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -4,13 +4,15 @@
 
 import shallowequal from 'shallowequal';
 import {get, set, has, omit, isObject, union, isFunction, isEqual,  isNil,
-        last, isPlainObject, forEach, fromPairs, merge, mergeWith, isArray} from 'lodash';
+        last, isPlainObject, forEach, fromPairs, mergeWith, isArray, truncate} from 'lodash';
 
 import {getRootURL} from './BrowserUtil.js';
 import {getWsConnId, getWsChannel} from '../core/AppDataCntlr.js';
 import {getDownloadProgress, DownloadProgress} from '../rpc/SearchServicesJson.js';
+import {showInfoPopup} from '../ui/PopupUtil.jsx';
 
 import update from 'immutability-helper';
+import {ServerParams} from '../data/ServerParams';
 
 
 const MEG          = 1048576;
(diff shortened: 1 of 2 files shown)


src/firefly/js/util/WebUtil.js
```diff
@@ -413,32 +415,41 @@ export function downloadWithProgress(url, numTries=1000) {
...
+export function download(url, filename) {
+    const {protocol, host, path, hash, searchObject={}} = parseUrl(url);
+    const cmd = searchObject[ServerParams.COMMAND];
+    url = `${protocol}//${host}${path}?${ServerParams.COMMAND}=${cmd}` + (hash ? '#' + hash : '');      // add cmd into the url as a workaround for server-side code not supporting it
+
+    const params = Object.fromEntries(
+                        Object.entries(searchObject)
+                              .map(([k, v]) => [k, (isPlainObject(v) ? JSON.stringify(v) : v)]) );          // convert object back into JSON if needed.
+
+    fetchUrl(url, {method: 'post', params})
```
robyww: So you are do a form type `post` without, building DOM form.

src/firefly/js/util/WebUtil.js
```diff
@@ -413,32 +415,41 @@ export function downloadWithProgress(url, numTries=1000) {
...
+        .then((resp) => {
+            filename = filename || resolveFileName(resp);
+            return resp.blob(); })
+        .then( (blob) => {
+            downloadBlob(blob, filename); })
```
robyww: I am a little concerned about this step.  Maybe I am wrong but it appears that we don't stream a file out anymore but hold it in memory.  This might not work with a large FITS or table file. Have you tested that?

loitly: It's not clear to me based on the documentation of Response.blob() that it will returns a fully populated content from the response or it'll return a blob backed by the response's input stream. 
I've tried a 1GB fits file and a 1 million rows table (800MB).  I don't see any increase in memory usage on the browser.  It looks like it's doing the later.
I did notice a BIG performance problem with TableSave when it's a big table.  The UI may need to change to account for a slow prep-stage as well.  I will create a new ticket for this.
@robyww is this acceptable?

robyww: That's fine as long as you tested it.  I guess it could be backed my the stream.

loitly: ticket created for UI changes when file is large.  https://jira.ipac.caltech.edu/browse/IRSA-2975

PR #831, opened 2019-06-28
Firefly-91: Add centering drop down menu

robyww:
Firefly-91: Add centering drop down menu

 - Center on target
 - Center on image
 - Add drop down
 - Modified for brief target panel / cleaned up target panel
 - Added recent cascade
 - CompleteButton will now optionally respond to enter
 - 10 Recent targets saved as a preference
 - added start up option support
 - added watcher to center image on table row

_to test:_

https://irsawebdev9.ipac.caltech.edu/firefly-91-center-dropdown/firefly/

- Load a HiPS (or image but HiPs makes for a better test)
  - Goto the Center dropdown center-dropdown
  - Try Center on target (will be enable if you entered a target)
  - click on the position input and try positions, watch HiPS move
  - look to see if it saves your recent positions
- Load two catalogs from two different positions
  - click on various rows of the table and see the hips move to the position


cwang2016:
The centering dropdown menu looks good and works well. 

I may not understand all functions of the dropdown menu well, here are a few questions, 
- if doing image search (not hips) and catalog search and  'pan by table row'  is checked, how is the image affected when the table  highlight (or select) changs? I see the change on hips, but not very sure about the change on fits image. 
-'center on target' and 'center on image' on the menu looks active for fits image, and grey out for hips image. what are the usage of those two items as shown on the menu? 
- when we do hips search, if no target is given, the hips is centered at 0,0 in default, should the dropdown menu shows '0, 0' at 'Center On' when the centering button is clicked after such  hips search?

robyww:
>I may not understand all functions of the dropdown menu well, here are a few questions,

> - if doing image search (not hips) and catalog search and 'pan by table row' is checked, how is the image affected when the table highlight (or select) changs? I see the change on hips, but not very sure about the change on fits image.

The image will only move if the new target in outside of 90% of the screen.

> - 'center on target' and 'center on image' on the menu looks active for fits image, and grey out for hips image. what are the usage of those two items as shown on the menu?

 'center on target' is not grayed out if you entered a target, you probably did not.

'center on image' is always grayed out for image since it does not makes sense for HiPS.

> - when we do hips search, if no target is given, the hips is centered at 0,0 in default, should the dropdown menu shows '0, 0' at 'Center On' when the centering button is clicked after such hips search?

It could, we can ask the scientist.

vandesai1:
- when I hover over the the words "Center on Target" or "Center on Image" or "Recent Positions", they get highlighted in white.

- when I put in a position that is not on a FITS image, and then recenter the FITS image, the old position is still in the input box, which could lead people to think that position is now centered. Probably just get rid of it at that point?

- if I open a HiPS image with the target "Arp220", and then open the "center on" menu, I see "0,0" in the position, making me think that's where the image is centered. Would be nice if somehow the dialog could tell you where you are currently centered, or go blank. Related to comment above.

Otherwise, this looks really good!

gpdf:
@cwang2016 wrote:
> * when we do hips search, if no target is given, the hips is centered at 0,0 in default, should the dropdown menu shows '0, 0' at 'Center On' when the centering button is clicked after such  hips search?

HiPS maps *may* come with an initial position and FOV that is used as the default.  It would be nice to offer the user the ability to go back to that point through this new menu.  (I can't recall, does the master "reset" button do that when used on a HiPS image?)

gpdf:
@robyww wrote:
> added start up option support

What is this?

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
```diff
@@ -0,0 +1,166 @@
...
+            <ToolbarButton text='Pan by table row' tip='Center selected images position of the highlighted row'
+                           enabled={coordTables}
+                           horizontal={false} key={'pan-table'}
+                           hasCheckBox={true} checkBoxOn={vr.autoScrollToHighlightedTableRow}
+                           onClick={() => dispatchChangeTableAutoScroll(!vr.autoScrollToHighlightedTableRow)}/>
+
+            <DropDownVerticalSeparator useLine={true}/>
+
+
+            <ToolbarButton text='Center on Target' tip='Center on search target'
```
gpdf: Trying to come to this with a blank slate, I found it confusing to think about what the user will understand as a/the "target" or even "search target".  It is *very* easy to imagine a user thinking of what they typed in the "Center On..." field as a "target".

What is really is, I think, is the name or coordinates most recently used in an image or catalog search, is that right?  Let's see if we can think of a way to make that clearer.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
```diff
@@ -0,0 +1,166 @@
...
+                <div style={{display:'flex', flexDirection:'column', alignItems: 'flex-end'}}>
+                    <CompleteButton style={{paddingLeft:10}} text= 'Go' onSuccess={moveToTarget} fireOnEnter={true}
+                                    dialogId='ExampleDialog' />
+                    {/*                //todo- button to add marker*/}
+                    {/*<CompleteButton style={{paddingLeft:10, marginTop:4}} text= 'Marker' onSuccess={createMarkerAndMoveToTarget}*/}
+                    {/*                dialogId='ExampleDialog' />*/}
+                </div>
+            </FieldGroup>
+            {recentAry.length>0 && <DropDownVerticalSeparator useLine={true}/>}
+            <DropDownSubMenu text={'Recent Positions'} visible={recentAry.length>0}>
```
gpdf: I would like the tooltips for these to show their resolved coordinates, if what was entered was a name.  I am OK if that is done in a separate ticket at a later date.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
```diff
@@ -0,0 +1,166 @@
...
+    //         const drawLayerId = title;
+    //         dispatchCreateMarkerLayer(drawLayerId, title, plotId, true, wp, true);
+    //     }
+    // };
+    //
+
+    const dropDown= (
+        <SingleColumnMenu>
+
+            <ToolbarButton text='Pan by table row' tip='Center selected images position of the highlighted row'
```
gpdf: It looks to me like the original specification that there be a "fiducial region" of the displayed image within which no panning would occur was not, in the end, followed.

I still think this is important to avoid excessive panning and even possibly confusing behavior.

It is also pretty important, now that I'm playing with this, for no panning to occur in response to *clicking on the displayed mark on the image* for a table row.  This produces an unsettling effect, plays poorly with "lock by click", and makes the UX very sensitive to whether the user hits the symbol exactly or just off by a little.

robyww: On FITS or zoomed in HiPS not panning occurs if the object object in with in 90% of the displayed area. Zoom out hips do pan.  I think this makes a little more sense.  

However, I did not test clicking on the image or HiPS.  I agree that in that case is should _never_ jump. I am fixing that.

robyww:
> . (I can't recall, does the master "reset" button do that when used on a HiPS image?)

The reset button moves HiPS to the original position and size.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
```diff
@@ -0,0 +1,166 @@
...
+            <ToolbarButton text='Center on Image' tip='Center on the image'
+                           enabled={Boolean(plot && isImage(plot))}
+                           horizontal={false} key={'center-image'}
+                           onClick={() => dispatchRecenter({plotId, centerOnImage:true})} />
+
+            <DropDownVerticalSeparator useLine={true}/>
+            <FieldGroup style={{display:'flex', fontSize:'10pt', padding:'5px 0 0 0'}}
+                        groupKey='TARGET_DROPDOWN' validatorFunc={null} keepState={true}>
+                <TargetPanel groupKey={'target-move-group'} labelWidth={80}
+                             label={'Center On:'}
```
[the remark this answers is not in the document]
robyww: If I can do this in a couple of hours I will, but this probably needs to be a separate ticket.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
robyww: That makes sense and should not take long. I will add it.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
```diff
@@ -0,0 +1,166 @@
...
+
+            <DropDownVerticalSeparator useLine={true}/>
+
+
+            <ToolbarButton text='Center on Target' tip='Center on search target'
+                           enabled={hasSearchTarget}
+                           horizontal={false} key={'center-search-target'}
+                           onClick={() => dispatchRecenter({plotId})} />
+
+            <ToolbarButton text='Center on Image' tip='Center on the image'
```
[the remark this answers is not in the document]
robyww: Just let me know what the tip should be and I will set it to that.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
robyww: Just let me know what the text should be and I will set it to that.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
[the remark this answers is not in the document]
robyww: OK- I can get the basic functionality working. I will use our existing symbols though. We can enhance it later.

robyww:
> - when I hover over the the words "Center on Target" or "Center on Image" or "Recent Positions", they get highlighted in white.

I don't think I understand this one.

> - when I put in a position that is not on a FITS image, and then recenter the FITS image, the old position is still in the input box, which could lead people to think that position is now centered. Probably just get rid of it at that point?

OK- I think it makes sense to clear it.  I will do that.

> - if I open a HiPS image with the target "Arp220", and then open the "center on" menu, I see "0,0" in the position, making me think that's where the image is centered. Would be nice if somehow the dialog could tell you where you are currently centered, or go blank. Related to comment above.

Yes, I think going blank makes most sense.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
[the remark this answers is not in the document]
gpdf: Just as in the first sentence above.  The rest of the comment was a rationale.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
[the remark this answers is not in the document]
gpdf: (Just for the record, @robyww went even further and created an actual Firefly "marker" from the entered position.)

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
gpdf: I do not have an actionable suggestion for now.  We can address this in a future ticket if the CCB-D has a good idea.

src/firefly/js/visualize/ui/ImageCenterDropDown.jsx
gpdf: CCB-D agrees, after experience with the tool, that we should abandon the idea of the "fiducial region".  The behavior of "pan to table row" should be to move the image *only when necessary to make the selected row's symbol visible*.

PR #832, opened 2019-07-10
firefly-84-TimeSeries tool: Add workspace option to download dialog for packaging images

lznakano:
See ticket here: https://jira.ipac.caltech.edu/browse/FIREFLY-159
and here: https://jira.ipac.caltech.edu/browse/FIREFLY-84

 1. Add LcDownloadPanel.jsx
 2. Instead of having three DownloadPanels, only one LcDownloadPanel is use for all missions.
 3. Instead of duplicating onNewRawTable, rawTableRequest and onFieldUpdate for each mission, these three functions are put in LcUntils.js.

I tried to use existing DownloadDialog but I have to make quite some modifications.  Instead, I added a 
LcDownloadPanel.  I modified the BackgroundCntrl to save the files in the workspace.  Please let me know if there is any better approach.

URL: https://irsawebdev9.ipac.caltech.edu/firefly-Lijun-firefly-84-firefly-159/irsaviewer/ts.html

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
@@ -165,7 +165,18 @@ public class AnyFileDownload extends BaseHttpServlet {
     }
 
     private static void trackProgress(HttpServletRequest req, BackgroundEnv.DownloadProgress progress) {
-        StringKey statusKey= new StringKey(req.getQueryString());
+        String queryString =req.getQueryString();
+
+        if (queryString==null){
+            try {
+                throw new Exception("Status can not be tracked because the queryString is empty");
+            } catch (Exception e) {
+                Logger.error("HttpServletRequest has a null query string, the status can not be tracked");
+
+
+            }
+        }
+        StringKey statusKey= new StringKey(queryString);
         getCache().put(statusKey, progress);
     }
 
--- a/src/firefly/js/core/background/BackgroundCntlr.js
+++ b/src/firefly/js/core/background/BackgroundCntlr.js
@@ -176,7 +176,7 @@ function bgPackage(action) {
                     bgStatus = bgStatusTransform(bgStatus);
                     const url = get(bgStatus, ['ITEMS', 0, 'url']);
                     if (url && isSuccess(get(bgStatus, 'STATE'))) {
-                        download(url);
+                        download(url, bgStatus.fileName);
                     } else {
                         dispatchJobAdd(bgStatus);
                     }
--- a/src/firefly/js/core/background/BgMaskPanel.jsx
+++ b/src/firefly/js/core/background/BgMaskPanel.jsx
@@ -28,13 +28,15 @@ export class BgMaskPanel extends SimpleComponent {
             bgStatus && dispatchJobAdd(bgStatus);
         };
 
+
         if (inProgress) {
             return (
                 <div style={maskStyle}>
                     <div className='loading-mask'/>
                     {bgStatus &&
                     <div style={{display: 'flex', alignItems: 'center'}}>
-                        <button type='button' style={maskButton} className='button std' onClick={sendToBg}>Send to background</button>
+                      <button type='button' style={maskButton} className='button std' onClick={sendToBg}>Please click to send to background</button>;
+
                     </div>
                     }
                 </div>
--- a/src/firefly/js/templates/lightcurve/LcResult.jsx
+++ b/src/firefly/js/templates/lightcurve/LcResult.jsx
@@ -24,9 +24,7 @@ import {LcImageViewerContainer} from './LcImageViewerContainer.jsx';
 import {SplitContent} from '../../ui/panel/DockLayoutPanel.jsx';
 import {LC, getViewerGroupKey, updateLayoutDisplay} from './LcManager.js';
 import FieldGroupUtils from '../../fieldGroup/FieldGroupUtils.js';
-import {ValidationField} from '../../ui/ValidationField.jsx';
 import {LcImageToolbar} from './LcImageToolbar.jsx';
-import {DownloadOptionPanel, DownloadButton} from '../../ui/DownloadDialog.jsx';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
 import CompleteButton from '../../ui/CompleteButton.jsx';
 import {HelpIcon} from '../../ui/HelpIcon.jsx';
@@ -34,13 +32,14 @@ import {getTblById, doFetchTable, isTblDataAvail} from '../../tables/TableUtil.j
 import {MAX_ROW} from '../../tables/TableRequestUtil.js';
 import {dispatchMultiValueChange, dispatchRestoreDefaults}  from '../../fieldGroup/FieldGroupCntlr.js';
 import {logError} from '../../util/WebUtil.js';
-import {getConverter, getMissionName,  DL_DATA_TAG} from './LcConverterFactory.js';
+import {getConverter, getMissionName} from './LcConverterFactory.js';
 import {convertAngle} from '../../visualize/VisUtil.js';
-
 const resultItems = ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'searchDesc', 'images',
     LC.MISSION_DATA, LC.GENERAL_DATA, 'periodState'];
+import {LcDownloadOptionPanel } from './LcDownloadPanel.jsx';
+import {BgMaskPanel} from '../../core/background/BgMaskPanel.jsx';
 
-
+export const LcXYPlotCompKey='XYPLOTAREA';
 export class LcResult extends PureComponent {
 
     constructor(props) {
@@ -130,44 +129,13 @@ const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables,
 
     const converterId = get(settingBox, ['props', 'missionEntries', LC.META_MISSION]);
     const convertData = getConverter(converterId);
-    const cutoutSize = get(convertData, 'noImageCutout') ? undefined : get(settingBox, 'props.generalEntries.cutoutSize', '5');
     const mission = getMissionName(converterId) || 'Mission';
     const showImages = isEmpty(imagePlot);
 
     // convert the default Cutout size in arcmin to deg for WebPlotRequest, expected to be string in download panel
+    const cutoutSize = get(convertData, 'noImageCutout') ? undefined : get(settingBox, 'props.generalEntries.cutoutSize', '5');
     const cutoutSizeInDeg = (convertAngle('arcmin','deg', cutoutSize)).toString();
-    const currentTime =  (new Date()).toLocaleString('en-US', { hour12: false });
-    const style = {width: 223};
-    const defaultOptPanel = (m, c) => {
-        return (
-            <DownloadButton>
-                <DownloadOptionPanel
-                    groupKey = {mission}
-                    dataTag = {DL_DATA_TAG}
-                    cutoutSize={c}
-                    title={'Image Download Options'}
-                    style = {{width: 400}}
-                    dlParams={{
-                        MaxBundleSize: 200 * 1024 * 1024,    // set it to 200mb to make it easier to test multi-parts download.  each wise image is ~64mb
-                        FilePrefix: `${m}_Files`,
-                        BaseFileName: `${m}_Files`,
-                        DataSource: `${m} images`,
-                        FileGroupProcessor: 'LightCurveFileGroupsProcessor'
-                    }}>
-                    <ValidationField
-                        style={style}
-                        initialState={{
-                            value: `${m}_Files: ${currentTime}`,
-                            label: `${m}:`
-                        }}
-                        fieldKey='Title'
-                        labelWidth={110}/>
-                </DownloadOptionPanel>
-            </DownloadButton>
-        );
-    };
 
-    const downloaderOptPanel = convertData.downloadOptions || defaultOptPanel;
 
     let tsView = (err) => {
 
@@ -183,7 +151,10 @@ const StandardView = ({visToolbar, title, searchDesc, imagePlot, xyPlot, tables,
                                 </div>
                             </div>
                         </SplitContent>
-                        <SplitContent>{xyPlot}</SplitContent>
+                        <SplitContent>
+                            <BgMaskPanel componentKey={LcXYPlotCompKey}/>
+                            {xyPlot}
+                        </SplitContent>
                     </SplitPane>
                     <SplitContent>{imagePlot}</SplitContent>
                 </SplitPane>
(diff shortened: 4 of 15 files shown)


loitly:
A few things off the top.. 
1.  Please include URL to ticket.
2.  Ticket is horrible.  Nothing in description.  I went and read all of the comments and discovered that it's was all dated in 2018.
When this is the case, will you update the ticket with basic information on what this ticket  is requesting for and/or steps to test it.
I know the title say a lot.  But, I was looking for more and found nothing.

loitly:
Without requirements, I can only assume how it should work.  Based on that, here are my comments.

1.  `File Location:  Local vs Workspace`
This only works if it does not go into Background Monitor.  This happens when it's a bigger job and takes longer than a few seconds.
2. When `Workspace` is selected and it successfully saved into workspace, there's no indication that it happened.  I thought it failed silently.
3.  While preparing, it used to offer a `send to background` option.  I don't see that anymore.

Also, a few items I noticed that is not related to this PR changes, but need to address:
1. The `Upload` dialog is clipped on the bottom.  It was not the case in OPS.
2. On the `Download Dialog`, when you click `Save`, the dialog hides immediately.  No indication it's working, and no option to `Send to background`.
3. `Background Monitor`, click on `Download Now` and the spiny never stopped.  It is working in OPS.

loitly:
I did not review the code.  I went and look at FinderChart and noticed that it's behaving the same.
I don't think this behavior is acceptable.  There is also a lot of things that make downloading really awkward to use.  The same can be said about FinderChart as well.  I recommend putting more effort into making it work properly.

lznakano:
> Without requirements, I can only assume how it should work. Based on that, here are my comments.
> 
> 1. `File Location:  Local vs Workspace`
>    This only works if it does not go into Background Monitor.  This happens when it's a bigger job and takes longer than a few seconds.
> 2. When `Workspace` is selected and it successfully saved into workspace, there's no indication that it happened.  I thought it failed silently.
> 3. While preparing, it used to offer a `send to background` option.  I don't see that anymore.
> 
> Also, a few items I noticed that is not related to this PR changes, but need to address:
> 
> 1. The `Upload` dialog is clipped on the bottom.  It was not the case in OPS.
> 2. On the `Download Dialog`, when you click `Save`, the dialog hides immediately.  No indication it's working, and no option to `Send to background`.
> 3. `Background Monitor`, click on `Download Now` and the spiny never stopped.  It is working in OPS.

lznakano:
> Without requirements, I can only assume how it should work. Based on that, here are my comments.
> 
> 1. `File Location:  Local vs Workspace`
>    This only works if it does not go into Background Monitor.  This happens when it's a bigger job and takes longer than a few seconds.
> 2. When `Workspace` is selected and it successfully saved into workspace, there's no indication that it happened.  I thought it failed silently.
> 3. While preparing, it used to offer a `send to background` option.  I don't see that anymore.
> 
> Also, a few items I noticed that is not related to this PR changes, but need to address:
> 
> 1. The `Upload` dialog is clipped on the bottom.  It was not the case in OPS.
> 2. On the `Download Dialog`, when you click `Save`, the dialog hides immediately.  No indication it's working, and no option to `Send to background`.
> 3. `Background Monitor`, click on `Download Now` and the spiny never stopped.  It is working in OPS.

lznakano:
Thanks you @loitly very much for your review and great input and suggestions. 
> > Without requirements, I can only assume how it should work. Based on that, here are my comments.
> > 
> > 1. `File Location:  Local vs Workspace`
> >    This only works if it does not go into Background Monitor.  This happens when it's a bigger job and takes longer than a few seconds.
 I was kind confused what is the proper behavior here.  I thought that to save to local, it can go through background monitor, to save to workspace, it does not go through background monitor.  
> > 2. When `Workspace` is selected and it successfully saved into workspace, there's no indication that it happened.  I thought it failed silently. 
 It is true.  What kind indicator would you like to see?  Is it OK to use a message box?
> > 3. While preparing, it used to offer a `send to background` option.  I don't see that anymore.
> >  
  I will take a look about this.  
> > Also, a few items I noticed that is not related to this PR changes, but need to address:
> > 
> > 1. The `Upload` dialog is clipped on the bottom.  It was not the case in OPS.
> > 2. On the `Download Dialog`, when you click `Save`, the dialog hides immediately.  No indication it's working, and no option to `Send to background`.
> > 3. `Background Monitor`, click on `Download Now` and the spiny never stopped.  It is working in OPS.
   OK, I will spend more effort to address the issues both in Finderchart and TimeSeries.

loitly:
@lznakano  The PR is reopened, but I did not get a notice to redo my review.
The test build is still the old one.  Behavior has not changed.  
@ejoliet , you don't care about the issues I raised in regard to backgrounding?

ejoliet:
Wait, i messed up, the comment was not for this PR , sorry! I just start reading. The comment 

> Looks perfect!

was for #836  ;-)

lznakano:
I think that I accidentally closed the pull request.  I want to keep it open until I am done with the issues Loi raised.  Most of issues were in the dev and were not bugs introduced in this PR.   I issued two tickets for those issues presented in the current dev.  I am working on one of the ticket, FIREFLY-159, in this PR.  I will let you know when it is ready to review again.   Thanks!

loitly:
@lznakano , you may want to edit your previous commit.  I am sure you did not mean to include that comment.  It's over 150 lines of repeated error messages.

lznakano:
This pull request included the bug fixes for Firefly-159.  Please test it against both tickets. 
When the file is to save to the local file system, the "Send to background" option is enabled.  When the file is to save to the workspace, the "Send to background" option is disabled.  The local job is sent to background monitor automatically if it is a big job.  A message box is displayed after the file is saved to the workspace successfully.  I think that we need to discuss what is the expected behavior to save files.  Currently, the version in ops does not have "Send to background" button, the job is automatically sent to the background.  Please review it again!  Thanks!

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
```diff
@@ -165,7 +165,16 @@ private static void logActivity(File f) {
...
-        StringKey statusKey= new StringKey(req.getQueryString());
+        String queryString =req.getQueryString();
+
+        if (queryString==null){
+            try {
+                throw new Exception("Status can not be tracked because the queryString is empty");
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
+        }
```
loitly: I don't understand this block.  You throw an exception, then catch it to print its stacktrace.
Also, if you plan to log the error, use a logger, not standard out.

src/firefly/js/core/background/BackgroundCntlr.js
```diff
@@ -166,17 +173,27 @@ function bgSetEmail(action) {
         dispatch(action);
     };
 }
-
 function bgPackage(action) {
     return (dispatch) => {
-        const {dlRequest, searchRequest, selectionInfo} = action.payload;
+        const {dlRequest, searchRequest, selectionInfo, isWs, params} = action.payload;
```
loitly: Why add `isWs` and `params` separately when `params` is only used when isWs is true?  Why not use an object and just call it `wspaceInfo` or something.

src/firefly/js/core/background/BgMaskPanel.jsx
```diff
@@ -28,13 +28,20 @@ export class BgMaskPanel extends SimpleComponent {
             bgStatus && dispatchJobAdd(bgStatus);
         };
 
+        const button = bgStatus && bgStatus.isLocal && bgStatus.isLocal?
```
loitly: `bgStatus.isLocal` is repeated twice

src/firefly/js/core/background/BgMaskPanel.jsx
```diff
@@ -28,13 +28,20 @@ export class BgMaskPanel extends SimpleComponent {
...
+
         if (inProgress) {
             return (
                 <div style={maskStyle}>
                     <div className='loading-mask'/>
                     {bgStatus &&
                     <div style={{display: 'flex', alignItems: 'center'}}>
-                        <button type='button' style={maskButton} className='button std' onClick={sendToBg}>Send to background</button>
+ {/*                       <button type='button' style={maskButton} className='button std' onClick={sendToBg}>Please click to send to background</button>
+*/}                  {button}
```
loitly: So now `Send to background` is removed for everything?  Is it still available for catalog search?

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -1,377 +1,353 @@
...
-                }
-            ).catch(
-                (reason) => {
-                    logError(`Failed to get full raw table: ${reason}`, reason);
-                }
-            );
-       }
-    }
-}
+/*
```
loitly: What happen here?  Why github thinks you've changes to the whole file?  Are you sure it's not some sort of git related issues?  Double check to make sure you don't override previous changes.

src/firefly/js/templates/lightcurve/LcUtil.jsx
```diff
@@ -3,14 +3,18 @@ import PropTypes from 'prop-types';
 import {get, has, set, isEmpty} from 'lodash';
```
loitly: Same here.  It looks like you took other people changes as your own.

src/firefly/js/templates/lightcurve/ptf/PTFMissionOptions.js
```diff
@@ -1,15 +1,10 @@
 import React, {PureComponent} from 'react';
```
loitly: same

src/firefly/js/templates/lightcurve/LcUtil.jsx
lznakano: No idea what happened.  In my local branch, it is fine.  It showed the last modification is July 19.  I also compared with dev.

lznakano:
@loitly @ejoliet 
Thank Loi very much for fixing the line separator issue.  I compared a few files and found only LcResult is bad.  I copied the one Loi corrected to my branch.   I also fixed two issues Loi noticed
and a bug I introduced.  Can you please test it when you get time?   Thanks!
Test URL: https://irsawebdev9.ipac.caltech.edu/firefly-84-Lijun/irsaviewer/ts.html

ejoliet:
I can't test it via the link but i will locally. 
(I see this : 

> Bad Request
Your browser sent a request that this server could not understand.
Reason: You're speaking plain HTTP to an SSL-enabled server port.
Instead use the HTTPS scheme to access this URL, please.

lznakano:
I rebased with dev and rebuilt the 
URL: https://irsawebdev9.ipac.caltech.edu/firefly-Lijun-firefly-84-firefly-159/irsaviewer/ts.html.  

Please test it using this URL.  Please check the ticket Firefly-84 and Firefly-159 for implementation information.

lrebull:
https://github.com/Caltech-IPAC/firefly/pull/832
https://jira.ipac.caltech.edu/browse/FIREFLY-84
https://jira.ipac.caltech.edu/browse/FIREFLY-159
https://irsawebdev9.ipac.caltech.edu/firefly-Lijun-firefly-84-firefly-159/irsaviewer/ts.html

Firefox 68.0.2 (Mac OS 10.11.6)

Downloading:

one table to disk via diskette icon - fine
one image to disk via diskette icon - fine
all images to disk via “prepare download” - worked.
*** tick box for “Also send me email with URLs to download” does nothing. (as in, I can’t enter an email address, not just that it doesn’t send email.)
Gave me an option to send it to the background monitor, which worked. 
Asked for massive download. Got a chance to put in an email in the background monitor itself. It sent me email. Link in the email works. 

Can’t make it automatically get dropped into background monitor (mentioned in ticket).  

I note that readme in zip files must be timestamped UT because right now (~1pm) they are timestamped ~8pm tonight. This messed with my head when looking for the most recently modified files, where I expected a zip file to be top of the list of most recently modified files. 

I can’t log in to this server’s workspace, so I cannot test any of the saving files to the workspace.
(Instead of a login screen, I get “default backend - 404”)

Loading a WISE light curve. From the input data tab, I can select all data, download. This process takes a while.
From the phase folded data tab, I can select all data, download. This process appears to complete instantaneously, regardless of what download options I select (cutouts, full images), which makes me think it’s not working. Indeed, the zip files contain solely a “readme” that says, “Successfully packaged 0 files: 0 B.”


Ticket FIREFLY-84 says:
*When the "Prepare Download" button is pressed after making selection in the table, a new Dialog is displayed,  except having other fields, it should have two radio buttons, one is local and the other is workspace.  When the workspace button is selected, the workspace should not be opened.  - 	WORKS FOR DISK BUT CAN’T TEST WORKSPACE 
*When the "Save" is pressed in the Download dialog, the "Send to background" button should be displayed.  — CAN’T GET IT TO DO THAT.
*The big job should be sent to background monitor regardless the "Send to background" is pressed or not. — CAN’T GET IT TO DO THAT.
*If the job is not send to the background, after the search and fetch finishes, if the saving location is "local", the local file system should be opened, if the saving location is workspace, the workspace dialog window should be opened.  - I DON’T KNOW WHAT ‘LOCAL FILE SYSTEM SHOULD BE OPENED’ MEANS. IT SAVES THE FILE AND ASKS ME IF I WANT TO JUST SAVE IT OR OPEN IT WITH AN APPLICATION (WHICH IS WHAT IT ALWAYS DOES). CAN’T TEST WORKSPACE
*If the job is sent to the background, when the "Download Now" button is clicked in the background Monitor, the same procedure as above should be performed, and the check box should be placed after finishing downloading the file. - WORKS

Ticket FIREFLY-159 says:
*On the "Prepare {{Download" Dialog}}, when you click {{Save}}, if it is a big job, it was sent to the monitor directly without showing  the  option to {{Send to background}}. — I HAVE THE OPTION TO SEND IT TO THE BACKGROUND
*{{In the Background Monitor}}, click on {{Download Now}} and the spiny never stopped. - CAN’T REPRODUCE, MUST BE FIXED
*The Download Dialog can not be hidden after the "prepare download" is pressed.  - I DON’T KNOW ABOUT ‘HIDDEN’ BUT I CAN KILL THE WINDOW IN ANY OF A  NUMBER OF WAYS DURING THE DOWNLOAD PROCESS.
*When the file in background monitor is downloaded, there is no check placed to indicate it is already downloaded. - CHECK MARK APPEARS ONCE DOWNLOADED BUT SEEMS MAYBE TO VANISH IN AN IRREPRODUCIBLE WAY. IF I COULD REPRODUCE IT, I WOULD, AND I KNOW YOU CAN’T FIX IT IF I CAN’T REPRODUCE IT. SO UNTIL I CAN REPRODUCE IT, I AM JUST NOTING THAT IT SEEMS TO APPEAR AND THEN DISAPPEAR LATER.



The behavior after finding a period is completely broken. I know that this wasn’t the focus of the ticket, but that makes me even more confused as to why it’s so thoroughly broken.  Send the light curve to the periodogram, get the power spectrum, select/accept, and when it returns ostensibly with a phase-folded light curve, it’s added bunch of outlier points that were not in the original light curve and cannot be filtered out. Made a movie documenting this. Light curve attached too.

lrebull:
light curve attached but it doesn't like the movie. comments and movie and light curve all sent in email and attached to the two Jira tickets.

ejoliet:
@lrebull @lznakano  i don't think is worth chasing other 'weird' or problem you might found because this version is based on an old branch, right?
If what you found in here is also broken in irsadev, then i would raise a ticket against dev.

lznakano:
@ejoliet 
 I rebased with dev this morning.  If something is broken here, it could be broken in the dev.

lrebull:
@ejoliet @lznakano it is indeed broken on dev. made a ticket IRSA-3112

lznakano:
@lrebull @ejoliet 
I confirmed that the issue Luisa found existed in the dev.  I reproduced it using irsadev.ipac.caltech.edu.   FYI

PR #833, opened 2019-07-15
FIREFLY-74_upgrade_babel_and_eslint

deannaji:
Upgrade Babel and eslint:
 - babel to v7.5.0
 - eslint to v5.1

Jira: https://jira.ipac.caltech.edu/browse/FIREFLY-74
K8s: https://irsawebdev9.ipac.caltech.edu/firefly-74/firefly/

IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/75


package.json
```diff
@@ -47,18 +53,22 @@
...
-    "react-datetime": "^2.16.0"
+    "validator": "^9.4"
   },
   "devDependencies": {
-    "babel-jest": "^23.6.0",
+    "@babel/core": "^7.5.0",
+    "@babel/polyfill": "^7.0.0",
+    "@babel/register": "^7.4.4",
+    "babel-core": "7.0.0-bridge.0",
+    "babel-jest": "^23.4.2",
```
loitly: Does it has to be declared at 2 places?

package.json
```diff
@@ -8,15 +8,20 @@
...
   },
   "dependencies": {
-    "babel-core": "^6.26.0",
-    "babel-loader": "^7.1.2",
+    "@babel/core": "^7.5.0",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
+    "@babel/plugin-proposal-json-strings": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-syntax-import-meta": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
```
loitly: The above plugins are needed now?  Do you know what is requiring it?

deannaji: These plugins are used instead of stage-3 plugin. As far as I understood from:
https://github.com/babel/babel-upgrade
(it is mentioned at the end of that link).

package.json
deannaji: Yes, I think those are extra ones. I will go ahead and remove them.
Thanks!

deannaji: Fixed!

package.json
robyww: I don't think we are using all these. I think we where using stage-3 there before for object expanding (ie.`{...x, {y:1}}`) when it is not part of the standard.

Let me look at each of these.

deannaji: Cool, I will take them out.

robyww: Take them out and test, make sure nothing breaks.

deannaji: Fixed, thanks!

__jest__/jest.transform.js
```diff
@@ -1,6 +1,9 @@
 // Custom Jest transform implementation that wraps babel-jest and injects our
 // babel presets, so we don't have to use .babelrc.
 
+/*eslint no-undef: "error"*/
+/*eslint-env node*/
+
 module.exports = require('babel-jest').createTransformer({
-  presets: ["env", "react", "stage-3"],
+  presets: ['@babel/preset-env', '@babel/preset-react', 'stage-3'],
```
robyww: Do you need to fixed the stage-3 thing here as well?

deannaji: Fixed!

deannaji:
Some dependencies (babel/core, babel/preset-env, and babel-core bridge) has to be on devDependencies in order for jest to run.

__jest__/jest.transform.js
```diff
@@ -1,6 +1,9 @@
 // Custom Jest transform implementation that wraps babel-jest and injects our
 // babel presets, so we don't have to use .babelrc.
 
+/*eslint no-undef: "error"*/
+/*eslint-env node*/
+
 module.exports = require('babel-jest').createTransformer({
-  presets: ["env", "react", "stage-3"],
+  presets: ['@babel/preset-env', '@babel/preset-react']
```
robyww: this is not exactly what I meant. in the other webpack we now doing:
```js
plugins: [
                    '@babel/plugin-transform-runtime',
                    '@babel/plugin-proposal-class-properties'
                ]
```
so I think it needs to be:
```js
presets: ['@babel/preset-env', 
                '@babel/preset-react', 
                plugins: [
                    '@babel/plugin-transform-runtime',
                    '@babel/plugin-proposal-class-properties'
                ]]
```
but you should test it.

deannaji: Fixed, thanks!

deannaji:
K8s updated.

PR #835, opened 2019-07-16
Firefly-88: Improve coverage, catalog & Image layer handling

robyww:
Firefly-88: Improve coverage/Catalog/Image layer handling 
   - icons for center and copy to clipboard
   - Coverage layer working
   - image layer will container search target and will hide image
   - new type of drawing symbol - POINT_MARKER
   - center working
   - copy to clipboard working
   - parsing CIRCE from ADQL query
   - cleaned up ToolbarButton
   - added new test to test api setting of HiPS
   - fixed api sagas that were not turned on in API Mode
   - added ACTIVE CSS style to show button press
   - coordinate sys will parse as ICRS as j2000 for now

https://jira.ipac.caltech.edu/browse/FIREFLY-88

_***To Test:***_

https://irsawebdev9.ipac.caltech.edu/firefly-88-coverage-map/firefly/

- Search a catalog, noticed improved drawing layer controls
- Gor catalog control notice how to control the color/size/symbol of the catalog and the image center separately.
- Search a second catalog, one again look a drawing layer controls
- notice base image drawing layer control
- notice new marking for the target look similar to the new icon: 20x20-center-small

- After reload do an image search, notice drawing layers controls for image with center control enabled.
- Try new test page: https://irsawebdev9.ipac.caltech.edu/firefly-88-coverage-map/firefly/test/test-misc.html

**_Fixes:_** 

- Regression color issue in dropdown menus is fixed.
- Changing the HiPS image for coverage is not longer reverted back when a new table is loaded.


ejoliet:
Suggestion to be discussed: when doing a catalog search, center searched might come up with different color - it's already a different symbol which is good.

robyww:
> Looks like a lot of changes. I'm just starting to look at different ticket related that triggered this improvements. I've tried to look at TimeSeries (IRSA-789). When i go to drawer layer and uncheck the target, the image disappear.

I think you are unchecking the image. I provided a separate way undo the target but I did not use a check box to maintain the one-checkbox-per-layer look

robyww:
> Suggestion to be discussed: when doing a catalog search, center searched might come up with different color - it's already a different symbol which is good.

We could do a different color. Might be better.

ejoliet:
> I think you are unchecking the image. I provided a separate way undo the target but I did not use a check box to maintain the one-checkbox-per-layer look

no, i'm unchecking the checkbox in the layer dialog, which is acting ok anywhere else in IRSAViewer or SOFIA, but in TS, is removing the image (???).

ejoliet:
@deannaji i think SOFIA is not using the updated image toolbar - could you check. I've built SOFIA locally with this PR and i don;t see the changes, also in https://irsadev.ipac.caltech.edu/applications/sofia/ with the updated center target menu too is not appearing. @robyww is there an option to be passed or is just a newer component class to use? 
Thanks!

ejoliet:
@robyww you are right, the checkbox in the drawer layer is not making the image disappear on purpose. Was that a design request? i don't really follow the reasoning behind but if that the request, then the feature works ;-)

robyww:
I went though so much writeup it is hard to keep track whether that was written or a oral request in the last year. The point of being able to hide the image is that if you have a lot of overlay and want to focus on that instead of the image (or just want a consistent background) hiding the image allows for that.

deannaji:
Tested SOFIA, looks good from my end.

xiuqin:
Allowing the background image disappear was a request from LSST at one point. I think both Gregory and Vandana liked the idea too.

Xiuqin

On Jul 17, 2019, at 1:46 PM, Trey Roby <notifications@github.com<mailto:notifications@github.com>> wrote:


I went though so much writeup it is hard to keep track whether that was written or a oral request in the last year. The point of being able to hide the image is that if you have a lot of overlay and want to focus on that instead of the image (or just want a consistent background) hiding the image allows for that.

tgoldina:
When I do catalog search (m31, 100''), the center of catalog appears as a yellow target symbol on the 'WISE Atlas W1' coverage image.

I then replace coverage image with 'WISE W4'. The center of catalog symbol is is still in the same place.

Now I see more controls under Image, the first entry in drawing layers. If I Show/Hide the yellow target symbol moves. Now it probably means the center of the image.

Is it the  expected behavior? Should the coverage image have this Show/Hide control?

If yes, is it possible to show the center of the catalog on the coverage image at this point?

robyww:
> When I do catalog search (m31, 100''), the center of catalog appears as a yellow target symbol on the 'WISE Atlas W1' coverage image. .........

I did not try that. I think you found a new issue. They way it is working is that when the coverage watch plots an image it specifically has a setting not to show the center from the image control.  It handles that. However, the test you did replaces the image and then the image control shows the center.  

I will try to figure out what to do.  This whole ticket involves juggling who shows the search center.  In some cases there is the image search target and in other it is the catalog search target.  When there are multiple catalog search targets they really don't care about the image one.

robyww:
@tgoldina I think you found at least 2 bugs. I will work on fixing them.

vandesai1:
Gregory Dubois-Felsmann and I reviewed this ticket on https://irsawebdev9.ipac.caltech.edu/firefly-88-coverage-map/firefly. 

Some very nice functionality there! A few comments:

BUG. If I do a FITS image search followed by a catalog search that results in a HiPS coverage map, the FITS image layers dialog gets "contaminated" with information about the HiPS image.

BUG/Improvement. If I do a search in galactic coordinates, the copy-to-clipboard only copies the numeric part of the coordinates, so that if you paste it into an IRSA Viewer search box, your search will not be done in galactic coordinates. To fix this, make sure that the clipboard is always producing decimal ra and dec in ICRS or J2000.

Labels: In the layers dialog, the mouseover for the new centering icon has two instances of "on" and the dashes kind of look like minus signs next to the coordinates. Also, the translation of coordinates is interesting information, but frustrating because it cannot be cut and pasted. Change the mouseover to simply state "Center on this position."

UI: In the image layers dialog, the entered object name should not be bolded, and "by NED" should be changed to "(NED)"

UI: In the image layers dialog, the Show/Hide buttons took us awhile to figure out. Instead, put a checkbox next to the search position, so the associated marker can be toggled off and on.

UI: In the coverage layers dialog, there is no way to turn off the catalog search position marker. Can we have a checkbox to control that?

robyww:
> BUG. If I do a FITS image search followed by a catalog search that results in a HiPS coverage map, the FITS image layers dialog gets "contaminated" with information about the HiPS image.

This bug has been there for a while.  I think we should fix that in another ticket.

> BUG/Improvement. If I do a search in galactic coordinates, the copy-to-clipboard only copies the numeric part of the coordinates, so that if you paste it into an IRSA Viewer search box, your search will not be done in galactic coordinates. To fix this, make sure that the clipboard is always producing decimal ra and dec in ICRS or J2000.

This is easy to do.  However, I can imagine cases where the user would expect what he copied to be what is shown and convert.  We could include the coordinate system as well in what is copied.

> Labels: In the layers dialog, the mouseover for the new centering icon has two instances of "on" and the dashes kind of look like minus signs next to the coordinates. Also, the translation of coordinates is interesting information, but frustrating because it cannot be cut and pasted. Change the mouseover to simply state "Center on this position."

OK

> UI: In the image layers dialog, the entered object name should not be bolded, and "by NED" should be changed to "(NED)"

OK

> UI: In the image layers dialog, the Show/Hide buttons took us awhile to figure out. Instead, put a checkbox next to the search position, so the associated marker can be toggled off and on.

I purposely used a different UI element because I did not what it is be considered at the same level as the UI checkbox which will turn the _whole_ layer off. I will play with this but I has some concerns.

> UI: In the coverage layers dialog, there is no way to turn off the catalog search position marker. Can we have a checkbox to control that?

I can make a on/off here but I have the same concerns the same concerns as above.  Also- the layer checkbox will turn everything off in the layer. If I had a checkbox here it will not be independent of the layer checkbox.  It will simple hide target when the layer is on. If the layer is off then the target will be hidden even it a checkbox is on. If we want complete independence then they should probably be managed as separate layers.

ejoliet:
@robyww might be fixed or not in this PR but when i look at SOFIA coverage and click "restore to default", the viewer goes in FITS viewer mode, HiPS option disappear... 
https://irsawebdev9.ipac.caltech.edu/sofia-ff-148/applications/sofia/

robyww:
i see it. It is really unrelated to this work but I will go ahead and fix it in this PR.

PR #837, opened 2019-07-17
Ff94 webdav bug

tgoldina:
Follow-up for FIREFLY-94: fixes for bugs discovered while testing LSST workspace.
https://jira.lsstcorp.org/browse/DM-13112

See SUIT pull request at https://github.com/lsst/suit/pull/15 for the info how to test.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/WebDAVWorkspaceManager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/WebDAVWorkspaceManager.java
@@ -55,7 +55,7 @@ public class WebDAVWorkspaceManager extends WebDAVWorkspaceManagerBase {
             return propVal;
         }
         if (propFound.equals(PROPS.ROOT_URL)) {
-            propVal = WS_HOST_URL;
+            propVal = getWsHostUrl();
         } else if (propFound.equals(PROPS.ROOT_DIR)) {
             propVal = getWsHome();
         } else if (propFound.equals(PROPS.AUTH)) {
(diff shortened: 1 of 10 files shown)


tgoldina:
In the two latest commits:

- fixed Bad Request, when saving a table to workspace. This was a missed case from firefly-53.

- added missed authorization credentials to a connection in IPACTableFromSource. This was causing 401 Unauthorized, when querying TAP schema from an LSST TAP server, which now requires authorization.

I've also added more error checking to IPACTableFromSource. Since URLDownload treats only the statuses between 300 and 400 as failures, any other bad status can go unnoticed. In my case, we happily parsed the returned html into an IPAC table with one column "<html>". After my change, IPACTableFromSource will treat HTTP status less than 200 or more or equal than 300 as a failure. This is consistent with HTTPServices. 
@loitly , @robyww, please let me know if you see a problem with this.

loitly:
> HTTP status less than 200 or more or equal than 300 as a failure. This is consistent with HTTPServices

Technically, 3xx are not errors.  It may just require additional steps.  In HttpServices case, follow-redirect is turned on.  So, already some of the 3xx should be handled already.  In other cases, we may decide to say it's an error.  But in general, 3xx are not errors.

tgoldina:
I don't know why URLDownload throws FailedRequestException on `responseCode>=300 && responseCode<400`, I did not want to change this behavior, in case something is relying on it. No changes there since 2017. So I just wanted to make sure we catch all other error statuses when getting table from URL.

loitly:
@tgoldina what you did is correct.   I was only commenting on the 3xx status and how HttpServices are treating it.

PR #839, opened 2019-07-19
FIREFLY-162: fixed download feature not working in firefox and safari

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-162
Test link: https://irsawebdev9.ipac.caltech.edu/FIREFLY-162_save_firefox_safari/firefly/

As reported in ticket, downloading is no longer working in Firefox and Safari.
This is due to a recent chance in the code where the `<A>` tag was not added to the HTML document before triggering a click event.  Not testing it in Firefox and Safari was an oversight.
You should definitely test the fix in all browsers.  :)

- also, minor tweaked to File Upload layout

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -286,7 +286,6 @@ div.rootStyle>img, div.rootStyle > html {
 .keyword-value  {
     color: maroon;
     margin-right: 10px;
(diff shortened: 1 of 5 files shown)


lrebull:
Yes, I confirm that it works for me on Firefox now.

PR #840, opened 2019-07-19
FIREFLY-166: fixed condition when Link Substitution fails, table fail to load.

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-166

When VOTable Link contains bad column reference, table should not fail to load.
This fix applies to both `href` and `value` substitution.  When the substitution fail to resolve, it will use what is given.

To test:
https://irsawebdev9.ipac.caltech.edu/FIREFLY-166_link_substitution_error_handling/firefly/

- upload the attached file in the ticket.  In the 4th column, the 5th links should say 
`info-${non-existing-column}-NGC6543`, because `non-existing-column` is not a valid column in this table.

### DIFF
--- a/src/firefly/js/util/VOAnalyzer.js
+++ b/src/firefly/js/util/VOAnalyzer.js
@@ -1075,7 +1075,7 @@ export function resolveHRefVal(tableModel, href='', rowIdx, defval='') {
         vars.forEach((v) => {
             const [,cname] = v.match(/\${([\w -.]+)}/) || [];
             const col = getColumnByID(tableModel, cname) || getColumn(tableModel, cname);
-            const rval = getCellValue(tableModel, rowIdx, col.name);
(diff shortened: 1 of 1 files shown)


ejoliet:
(ok, but in my local safari, firefox and chrome, i've tried to upload the table but it keep spinning...)

loitly:
> (ok, but in my local safari, firefox and chrome, i've tried to upload the table but it keep spinning...) - aprat from that, i also realized that the result table in the viewer shows extra column `ID` that looks suspiciously the continuation of the table_name... but is not part of the original VOtable.

@ejoliet  this sound like an issue not related to this PR.  Can I go ahead and merge this PR and you create a new ticket for the issues you found?

ejoliet:
Ok, np, i'll create a new ticket. Thanks.

PR #846, opened 2019-08-09
Firefly-179: Fixed right-east align and align w/rotation issues

robyww:
Firefly-179: Fixed right-east align and align w/rotation issues
 
  - right-east image align now work
   - fixed issue with aligning to rotated images
   - fixed align left-east with right-east image
   - clean up of rotation dialog: (fixes some issues in FIREFLY-146)
        - fixed multiple issues with rotation dialog and improved it.
        - now does immediate results
        - operates on active image.
        - has a north up button
        - show active images's rotation
        - Added slider
        - rotation dialog hides when all image removed
 
https://jira.ipac.caltech.edu/browse/FIREFLY-179

_To Test:_

 https://irsawebdev9.ipac.caltech.edu/firefly-179-wcs-match-bug/firefly/

  - load two images at same target (i.e. 2mass, wise), rotate one then match the other to it. Check that they scroll around together
  - load two east-right images (firefly_test_data/east-right-images)
  - you should be able to match them
  - load two east-right images, rotate one, then match
  - load a each right image from `firefly_test_data/east-right-images` then load a wise at 
`3h32m02.85s, -27d26m20.7s`, 792 arcsec.  try matching.
   - bring up the updated rotation dialog and play with it after loading a set of images.

### DIFF
--- a/src/firefly/js/data/ServerRequest.js
+++ b/src/firefly/js/data/ServerRequest.js
@@ -6,6 +6,7 @@ import {has} from 'lodash';
 import validator from 'validator';
 import {parseWorldPt} from '../visualize/Point.js';
 import {replaceAll, isDefined} from '../util/WebUtil.js';
+import {toBoolean} from '../util/WebUtil';
 
 
 const REQUEST_CLASS= 'RequestClass';
@@ -222,7 +223,7 @@ export class ServerRequest {
 //  convenience data converting routines
 //====================================================================
     getBooleanParam(key, def=false) {
-        return has(this.params,key) ? Boolean(this.params[key]) : def;
+        return has(this.params,key) ? toBoolean(this.params[key]) : def;
     }
 
     getIntParam(key, def=0) {
--- a/src/firefly/js/ui/FitsRotationDialog.jsx
+++ b/src/firefly/js/ui/FitsRotationDialog.jsx
@@ -1,262 +1,114 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import React, {PureComponent} from 'react';
-import {dispatchShowDialog} from '../core/ComponentCntlr.js';
-import {Operation} from '../visualize/PlotState.js';
+import React, {useEffect} from 'react';
+import {dispatchShowDialog, dispatchHideDialog} from '../core/ComponentCntlr.js';
 import Validate from '../util/Validate.js';
-import {ValidationField} from './ValidationField.jsx';
-import {RadioGroupInputField} from './RadioGroupInputField.jsx';
 import CompleteButton from './CompleteButton.jsx';
-import {FieldGroup} from './FieldGroup.jsx';
 import DialogRootContainer from './DialogRootContainer.jsx';
 import {PopupPanel} from './PopupPanel.jsx';
-import FieldGroupUtils from '../fieldGroup/FieldGroupUtils.js';
-import {primePlot} from '../visualize/PlotViewUtil.js';
+import {getActivePlotView, primePlot} from '../visualize/PlotViewUtil.js';
 import {visRoot, dispatchRotate} from '../visualize/ImagePlotCntlr.js';
+import {SimpleLayerOnOffButton} from '../visualize/ui/SimpleLayerOnOffButton';
 import {RotateType} from '../visualize/PlotState.js';
+import {StateInputField} from './StatedInputfield';
+import {useStoreConnector} from './SimpleComponent';
+import {isEastLeftOfNorth} from '../visualize/VisUtil';
+import {RangeSliderView} from './RangeSliderView';
 
 import HelpIcon from './HelpIcon.jsx';
+import ROTATE_NORTH_OFF from 'html/images/icons-2014/RotateToNorth.png';
+import ROTATE_NORTH_ON from 'html/images/icons-2014/RotateToNorth-ON.png';
 
-function getDialogBuilder() {
-    var popup = null;
-    return () => {
-        if (!popup) {
-            const popup = (
-                <PopupPanel title={'Rotate Image'}>
-                    <FitsRotationDialog groupKey={'FITS_ROTATION_FORM'}/>
-                </PopupPanel>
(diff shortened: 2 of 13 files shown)


cwang2016:
Woud you please check if the following issue could be reproduced? 
- search image on target: n1, get image WISE w1 & 2MASS J. 
- rotate one image by an angle, ex. 25 deg, and match the other one on to it. 
- load image from east-right-images/twinkles_v230.....fits
- match the rotate angle of 2MASS (i.e. 25deg)
- zoom out "twinkles_v230..." to util it is seen rotated.
- load image from east-right-image/twinkles_v235.....fits (the second one)
- match the rotate angle of 'twinkles_v230_....' (note: activate twinkles_v230, not others)
- zoom out twinkes_v235 in order to view the image in full with the rotate angle,
   => got no response.
  If you do zoom in then zoom out, then you see the image in full. 



-

cwang2016:
Another finding, but not sure  if it is as expected, 
- search one image on target: n1, WISE w1. 
- load an image from east-right-image/twinles_v230.....fits
- open up rotate dialog to rotate twinkles_v230 to like 53 deg, close the rotate dialog
- match ALLWISE WI to be the same as twinkles_v230
- point to ALLWISE WI, open up rotate dialog, it shows the angle 409.95 with warning "angle must be a number within 0 and 359"

robyww:
> Would you please check if the following issue could be reproduced?

 - I am having trouble following this one.  I can't reproduce.

> Another finding, but not sure if it is as expected,

 - fixed.

- I also found a rotation bug or two that if fixed. I am going to do a new push.

cwang2016:
please check the following, I am not sure if I understand the result as I thought
- search one image on target n1, WISE w1
- load an image form  east-right-image/twinkles_v230.....fits
- rotate loaded image, twinkles_v230 by 30 deg
- match WISE, w1 to the rotate angle of twinkles_v230
- rotate 'twinkles_v230' back to 0 deg
- match WISE, w1 to twnkles_v230 (I expect WISE W1 is also rotated back to 0 deg)
- open up the rotate angle for WISE WI, it shows angle '102.95', 
=> not sure if this is the right number?

ejoliet:
Although i didn't expect working on this, i would suggest if we are going to update UI, we could also put back the "apply all" checkbox we had before and disappear after migration? I'll check other tickets related tot eh dialog popup.

ejoliet:
@robyww i can see 'mask layer found' enable in IRSAViewer, this shouldn't be in the layer dialog, could you please remove it or add a property and mark it as false? I guess it's derived from mask overlay in Firefly... [see snapshot](http://web.ipac.caltech.edu/staff/ejoliet/Firefly-179.png)

robyww:
The mask layer found has always been in firefly it an uploaded file has a mask.  You have not seen this because no mask are recognized in IRSA images.

robyww:
> Although i didn't expect working on this, i would suggest if we are going to update UI, we could also put back the "apply all" checkbox we had before and disappear after migration? I'll check other tickets related tot eh dialog popup.

I don’t know why we would have an apply all. Rotation together is based on matching. The apply all is really a relic of when we had to call the server to rotate.

ejoliet:
I see. Great. Now i understand why is see this option now ;-)

ejoliet:
> I don’t know why we would have an apply all. Rotation together is based on matching. The apply all is really a relic of when we had to call the server to rotate.

ok. I was just brining the feature back on the table because i remember @vandesai1 mentioning it a couple of times. But it maight be obsolete now that we have changed the lock/match feature completely.

vandesai1:
Found what seems to be a problem:

Load AllWISE W1 and 2MASS All-Sky K images of Arp220. Rotate AllWISE W1 image. Align and Lock by WCS. Pan the images around. All is well.

Keeping previously loaded images, load SDSS r-band image at the same location. The W1 and K band images move on the page, but their rotation stays the same. However, the SDSS image is not WCS matched to the others. And the lock icon says that they should be locked by WCS. If I pan the SDSS r band image around, the other images pan around, too, but not in an expected way. Make sure all the compass roses are on. If I then click on the AllWISE image and change the lock state to Align by WCS, the compass rose for the SDSS image shifts to the other side of the image, but the image itself and the orientation of the compass rose does not shift. THIS SEEMS TO BE A BUG.

vandesai1:
Another problem:

I loaded an AllWISE channel 1 image of Arp220. When I click the “North Up” button in the rotation dialog, I get an equivalent angle of 359.69. If I click it again, I get an angle of 0. If I click it a third time, I get 359. And it cycles between these three numbers.

vandesai1:
Small decisions that need to be made:

1. The title of the dialog is “Rotate Image In Degrees”. The “degrees” info is important and it is likely to get lost in a title because people have learned to ignore titles. It would be better if the title simply said “Rotate Image”, and the input box was labeled “Angle in degrees East of North”. But is this in fact the sense of the rotation?

2. The angle is only allowed to be 0 <= angle <= 360. Is that appropriate?

robyww:
> Small decisions that need to be made:

> The title of the dialog is “Rotate Image In Degrees”. 

That makes sense, I will change it.

> The angle is only allowed to be 0 <= angle <= 360. Is that appropriate?

I would prefer to keep it this way. It has never been an issue before and internally we already convert to between 0 and 360.  There it would be more book keeping. I just don't think it is necessary.

robyww:
> Found what seems to be a problem:

Your right this is a problem. I did not test my new changes and load a new image.  I will fix.

> I loaded an AllWISE channel 1 image of Arp220. When I click the “North Up” button in the rotation dialog, I get an equivalent angle of 359.69. If I click it again, I get an angle of 0. If I click it a third time, I get 359. And it cycles between these three numbers.

Just saw this as well. I will fix.

robyww:
> It would be better if the title simply said “Rotate Image”, and the input box was labeled “Angle in degrees East of North”. But is this in fact the sense of the rotation?

**Yes.** The rotation does rotate towards the east. For east/left image that is towards to left, for east/right it goes in the other direction.

PR #847, opened 2019-08-13
Firefly 259 filter bug

tgoldina:
https://jira.ipac.caltech.edu/browse/FIREFLY-259
Test build: https://irsawebdev9.ipac.caltech.edu/firefly-259_filterBug/firefly/

- Fixed a bug which was indirectly introduced by firefly-88: a list of drawing drawing objects for a catalog overlay now can have extra objects (like search center), which do not correspond to catalog table rows. These objects should be excluded when filter by ROW_IDX is constructed. 
To test, select on image a subset of points that includes search center and use filter button.

- Also fixed non-returning spinning wheel in charts when there are no data. To test, select an empty area on image and filter or filter from the table with such constraints that no data are left. An error message "No data available" should show up for the related charts.


cwang2016:
review completes. works well. Nice to have the chart work for no data case.

PR #848, opened 2019-08-13
FIREFLY-185: 3-image composition doesn't work

tgoldina:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-185
Test deployment: https://irsawebdev9.ipac.caltech.edu/firefly-185-3color/firefly/

Fixed two bugs related to 3-color: 
1) exception described in the ticket (when less than 3 bands are used) 
2) no image (server-side) failure when red band is not used

Two test, create a 3-color image with 1 or 2 bands, with and without red.


ejoliet:
I've upload a FITS image with blue and red, leaving green out but the layout stuck in "processing images"...

tgoldina:
@ejoliet which images did you use for upload?

ejoliet:
For records, seems that the problem i was seeing is that the 3-color image doesn't work with GZIP images. File analyzer used behind the scene is the culprit and missing this. I'll raise ticket. Other images are correctly rendered after this fix.

PR #849, opened 2019-08-13
FIREFLY-113: Filtering images by BAND doesn't filter properly the sub selection

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-113
test: https://irsawebdev9.ipac.caltech.edu/FIREFLY-113_image_filtering_band/firefly/

The behavior should be as follow:
- Filtering should only affect the view of the panel.  So when images are selected, it will remained selected even though it is no longer visible by filtering.
- The collapsible level checkbox are based on the currently filtered view.  It is checked when all of its filtered checkboxes are checked.  Action(check/uncheck) performed on this checkbox will only affect the filtered image checkboxes.
- Selected images are shown in brown under Selection: at the top of the panel.

### DIFF
--- a/src/firefly/js/ui/ImageSelect.jsx
+++ b/src/firefly/js/ui/ImageSelect.jsx
@@ -5,7 +5,7 @@
 import React, {PureComponent, useEffect} from 'react';
 
 import PropTypes from 'prop-types';
-import {uniqBy, get, countBy, xor, sortBy, isNil} from 'lodash';
+import {uniqBy, get, countBy, remove, sortBy, isNil} from 'lodash';
 
 
 import {CheckboxGroupInputField} from './CheckboxGroupInputField.jsx';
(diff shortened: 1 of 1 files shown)


tgoldina:
Looks good, I see just one unexpected behavior:
1. With "COSMOS" and "mid-IR" filters, I select all COSMOS bands by clicking COSMOS checkbox - 5 bands selected, correct.
2. I add "radio" filter, I see 2 more bands added to the display. I'd like now to select all COSMOS bands in mid-IR and radio range by clicking on COSMOS checkbox again. Radio bands get selected, but mid-IR unselected. Is it what should happen? 

I realized that the checkbox above bands is more complex than select/unselect all, and may be this was a part of the original design, I was just not aware of this feature.

loitly:
> I realized that the checkbox above bands is more complex than select/unselect all, and may be this was a part of the original design, I was just not aware of this feature.

No, this is a bug.  I'll fix it.

PR #855, opened 2019-08-19
IRSA-2998:WCS HiPS/FITS alignment and coord overlay

lznakano:
Fixed the  missing image grid issue after performing wcs align

Test URL: https://irsawebdev9.ipac.caltech.edu/irsa-2998-grid-lijun/irsaviewer/

### DIFF
--- a/src/firefly/js/visualize/reducer/DrawLayerReducer.js
+++ b/src/firefly/js/visualize/reducer/DrawLayerReducer.js
@@ -72,7 +72,7 @@ function handleOtherAction(drawLayer,action,factory) {
     else if (drawData!==newDl.drawData) {
         return Object.assign({},newDl,{drawData});
     }
-    else {
+    else {/**/
         return drawLayer;
     }
 }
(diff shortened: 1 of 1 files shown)


ejoliet:
I've added @cgelino you might want to help testing it.

ejoliet:
('i've tested very quickly and looks fixed to me.)

ejoliet:
Thanks @cgelino, it sounds familiar, i think they are known by @robyww , if not i will consider to make ticket for future.

robyww:
> but dragging the HiPS image does not move the FITS image. 

Working on this in https://jira.ipac.caltech.edu/browse/FIREFLY-173

> #2 

I think it is a bug.  Since we refactored locking and made "lock color and overlay" a separate function I fear we have created some bug and need more testing with overlay locking.

robyww:
@lznakano While this have very little code change it affects about half the actions related to drawing. You are moving the decision of the layers to operate on, from the action to the existing layer. I really don't know if this a good or bad idea. It might be fine, however have you gone though _all_ the action that call `updateFromLayer`?

lznakano:
@robyww 
Please check my analysis in the ticket IRSA-2998.  The plotId is sent correct.  When the HiPs aligns with the image, the HiPs plotId is sent because the HiPs plot needs to be updated.  The issue is that the plotId array in each layer contains all the plotIds.  This leads the failure.

robyww:
This fix is much better.

lznakano:
@robyww I spent quit some time to trace the plotIdAry.  I think that I understand the design now. I slightly modified the changes so that no all draw lawyers are updated.  But all grid layers have to be updated due to the fact that all of them are counted as one layer with one drawLayerId.

lznakano:
@robyww You are too fast.  You checked before I finished writing.  Thanks!

PR #859, opened 2019-08-22
FIREFLY-340: Filter Editor can not be invoked from a chart

tgoldina:
https://jira.ipac.caltech.edu/browse/FIREFLY-340
Test deployment: https://irsawebdev9.ipac.caltech.edu/firefly-340_chartFilterEditor/firefly/

Filter Editor can be now invoked from the chart, the behavior is similar to the its behavior when it's invoked from a table. 

Bug fixes 
- Users should be able to update or remove expression filters set from the chart or ROW_IDX filters set from the image using Filter Editor's "Filters" box, which shows all filters.
- The number of filters on the badge on table toolbar should reflect the number of individual filters.

### DIFF
--- a/src/firefly/js/tables/FilterInfo.js
+++ b/src/firefly/js/tables/FilterInfo.js
@@ -26,11 +26,13 @@ export const FILTER_TTIPS =
 * when IN is used, enclose the values in parentheses
 * you may combine conditions with either 'and' or 'or'
 * grouping by parentheses is not supported at the moment
-* '${FILTER_SEP}' serves as groups separator at the moment 
 Examples:  
-  "ra" > 5 and "color" != 'blue' or "band" IN (1,2,3)
-  "ra" < 5 or "ra" > 6 _AND_ "dec" < -1 or "dec" > 1
+  "ra" > 5 and "color" != 'blue' or "band" IN (1,2,3) 
 `;
+
+// ('${FILTER_SEP}' serves as groups separator for now when grouping is necessary)
+// "ra" < 5 or "ra" > 6 _AND_ "dec" < -1 or "dec" > 1
+
 
 export const NULL_TOKEN = '%NULL';          // need to match DbAdapter.NULL_TOKEN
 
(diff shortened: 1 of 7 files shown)


src/firefly/js/tables/FilterInfo.js
```diff
@@ -22,14 +25,30 @@ export const FILTER_TTIPS =
 * string value must be enclosed in single quotes
 * when IN is used, enclose the values in parentheses
 * you may combine conditions with either 'and' or 'or'
-* grouping by parentheses is not supported at the moment 
-Examples:  "ra" > 12345 and "color" != 'blue' or "band" IN (1,2,3)
+* grouping by parentheses is not supported at the moment
+* '${FILTER_SEP}' serves as groups separator at the moment 
+Examples:  
+  "ra" > 5 and "color" != 'blue' or "band" IN (1,2,3)
+  "ra" < 5 or "ra" > 6 _AND_ "dec" < -1 or "dec" > 1
```
loitly: I see why this is hard.  But, I really prefer we don't expose _AND_.  It's intended for `internal` use as commented above.

src/firefly/js/tables/FilterInfo.js
```diff
@@ -22,14 +25,30 @@ export const FILTER_TTIPS =
...
-const FILTER_SEP = '_AND_';                 // internal use need to match what's defined in TableServerRequest.FILTER_SEP
+
+/**
+ * returns the number of individual filters
+ * @param filterInfoStr
+ * @returns {number}
+ */
+export function getNumFilters(filterInfoStr) {
+    if (filterInfoStr) {
+        return Math.round((filterInfoStr.split(new RegExp(`( and | or |${FILTER_SEP})`, 'i')).length+1)/2);
```
loitly: Took me awhile to figure out why you needed the round and `/2`.  Not sure you meant to do this, but if you remove the capturing `()`, you can just return the length of the result.

`filterInfoStr.split(new RegExp(` and | or |${FILTER_SEP}`, 'i')).length`

src/firefly/js/tables/FilterInfo.js
tgoldina: I've removed mention of the _AND_ from the tooltip. I agree, that the fact that Filters currently show our internal representation of filters is not ideal, but currently it is the only way we have to clear or change non-table filters selectively. 
I think clearing this logic should be in a separate ticket. It's not a quick bug fix.

src/firefly/js/tables/FilterInfo.js
tgoldina: Thank you.

PR #867, opened 2019-09-04
IRSA-2296: added code to support IBE2 inconsistency for 2mass LH

ejoliet:
Ticket [IRSA-2296](https://jira.ipac.caltech.edu/browse/IRSA-2296) is about adding a dataset to the image search.

I've added the dataset 2MASS LH to the image search choice and needed to change the Atlas IBE source code to take into account an inconsistency of the backend data file access. Usually ATLAS api gives the file path with the same table name of the query but in this case the image folder is not respecting the same name of the table.

The table to query is `2mass_lh` but the folder to get the image and build the url to retrieve is only `lh`.

In order to test, search images of 2mass LH around this target coordinates:  159.738690 54.932062 using the IRSAViewer built and deployed here:

https://irsawebdev9.ipac.caltech.edu/irsa-2296/irsaviewer/


lrebull:
confirm that it works!

PR #868, opened 2019-09-04
FIREFLY-80: switch to Java 11 openJDK and Java 1.8 openJDK

loitly:
see https://github.com/IPAC-SW/irsa-ife-env/pull/2


src/firefly/java/edu/caltech/ipac/util/KeyVal.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     *  @return <code>String</code> representation of this <code>KeyVal</code>
+     */
+    @Override
+    public String toString() {
+        return key + "=" + value;
+    }
+}
+/*
+ * THIS SOFTWARE AND ANY RELATED MATERIALS WERE CREATED BY THE CALIFORNIA
```
robyww: The long license is not longer necessary, use our now style of the line at that top.

docker/base/Dockerfile
```diff
@@ -1,5 +1,5 @@
 #FROM tomcat:7.0-jre8
-FROM tomcat:9.0.11-jre10
+FROM tomcat:9.0-jdk11-openjdk
```
tgoldina: I have a general question.
"OpenJDK will have a feature release every 6 months which is only supported until the next feature release". Does it mean this tag is updated every 6 months?

PR #869, opened 2019-09-04
FIREFLY-80: switch to Java 11 openJDK and Java 1.8 openJDK

loitly:
see https://github.com/IPAC-SW/irsa-ife-env/pull/2


buildScript/depends.gincl
```diff
@@ -24,6 +24,7 @@ ext {
                  ':jaxrpc',
                  ':saaj',
                  ':activation',
+                 ':jta',
```
robyww: What is this parameter?

PR #870, opened 2019-09-05
FIREFLY-360: revert changes made previously to onSort behavior

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-360
Test: https://irsawebdev9.ipac.caltech.edu/FIREFLY-360_revert_highlight_onsort/firefly/

As described in the ticket, highlighted row should be set to 0 after sorting.  This behavior should be for all tables, including `client` tables.

Some known client tables:
Images -> View HIPS Images -> Select Data Set
Then click the `i` icon on the image toolbar to bring up another client table.

### DIFF
--- a/src/firefly/js/tables/TablesCntlr.js
+++ b/src/firefly/js/tables/TablesCntlr.js
@@ -471,7 +471,9 @@ function tableSort(action) {
             const [nreq, tableStub, tableModel] = setupTableOps(tbl_id, request);
             if (!tableStub) return;
 
+            // rollback changes to keep current highlighted row.  instead set highlighted to 0.
             // TblUtil.setHlRowByRowIdx(nreq, tableModel);
+            nreq.startIdx = 0;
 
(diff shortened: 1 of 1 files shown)


src/firefly/js/tables/TablesCntlr.js
```diff
@@ -471,7 +471,7 @@ function tableSort(action) {
             const [nreq, tableStub, tableModel] = setupTableOps(tbl_id, request);
             if (!tableStub) return;
 
-            TblUtil.setHlRowByRowIdx(nreq, tableModel);
+            // TblUtil.setHlRowByRowIdx(nreq, tableModel);
```
robyww: might be worth saying why it is commented out, since it is rollback behavior

loitly:
@robyww You're right.  The highlighted row should be set to the first row of the returned table and not the first row in that page.  That is fixed.  I also added to my comment to indicate rollback intention.  Test URL is updated if you want to test again.  Thanks.

PR #871, opened 2019-09-05
IRSA-3112:Time Series tool problems on irsadev

lznakano:
Ticket [IRSA-3112](https://jira.ipac.caltech.edu/browse/IRSA-3112?focusedCommentId=230285&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-230285).
Fixed the wrong phase value issue

Test URL: https://irsawebdev9.ipac.caltech.edu/irsa-Lijun-irsa-3112-lc/irsaviewer/ts.html
Please use the attached testing data in irsa-3112 to test it.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcPhaseTable.js
+++ b/src/firefly/js/templates/lightcurve/LcPhaseTable.js
@@ -99,7 +99,9 @@ export function getPhase(time, timeZero, period,  dec=DEC_PHASE) {
     var q = (time - timeZero)/period;
     var p = q >= 0  ? (q - Math.floor(q)) : (q + Math.floor(-q) + 1.0);
 
+
     return p.toFixed(dec);
+
 }
 
 /**
@@ -227,7 +229,7 @@ function repeatDataCycle(phaseTable) {
     slice(tableData.data, 0).forEach((d) => {
         var newRow = slice(d);
 
-        newRow[fIdx] = `${parseFloat(d[fIdx]) + 1}`;
+        newRow[fIdx] = `${ (parseFloat(d[fIdx]) + 1).toFixed(DEC_PHASE)}`;
         tableData.data.push(newRow);
     });
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/lc/IrsaLightCurveHandler.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/lc/IrsaLightCurveHandler.java
@@ -129,7 +129,7 @@ public class IrsaLightCurveHandler implements LightCurveHandler {
     }
     /**
      * This method will get the "origin_table"'s parameter value from request.  The value can be a json string passing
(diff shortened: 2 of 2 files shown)


lrebull:
Confirm that this seems to be working now! thank you!

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -23,6 +23,18 @@ const DEC_PHASE = 3;       // decimal digit
...
 export function uploadPhaseTable(tbl, flux) {
     const {tbl_id, title} = tbl;
+    const data = tbl.tableData.data;
+
+    let badIndex=[];
+    let k=0;
+    for (let i=0; i<data.length; i++){
+        for(let j=0; j<data[i].length; j++){
+            if (data[i][j].startsWith('999999') ){
+                badIndex.push(k);
```
ejoliet: I understand that by filtering out those numbers we fix the problem but why in the first place those number occured? Is it a side effect of the phase folding formula?

lznakano: Sorry I forgot to remove it.

ejoliet: ok, great!

src/firefly/js/templates/lightcurve/LcPhaseTable.js
```diff
@@ -227,7 +229,7 @@ function repeatDataCycle(phaseTable) {
     slice(tableData.data, 0).forEach((d) => {
         var newRow = slice(d);
 
-        newRow[fIdx] = `${parseFloat(d[fIdx]) + 1}`;
+        newRow[fIdx] = `${ (parseFloat(d[fIdx]) + 1).toFixed(DEC_PHASE)}`;
         tableData.data.push(newRow);
```
ejoliet: Now we have to understand why this is not failing in OPS (???) to make sure we are not forgetting to fix something else behind that might trigger somewhere else.

lrebull:
for the record, i just tried in ops, and ops is fine.

src/firefly/js/templates/lightcurve/LcPhaseTable.js
lznakano: That was exactly my question which I asked in the ticket IRSA-3112.   It was very strange.
I played in the console:
0.257+1
1.2570000000000001
0.128+1
1.1280000000000001

I searched online and learned that there are precision problems in javascript.  Does OPS run the same version if javascript/react?  Maybe Loi @loitly has more insight about this issue.

loitly: I can't explain why it's different in OPS and DEV.  Unless they have different data/numbers and those data happens to fall into the category where there's no `bad` round off at the end.

PR #872, opened 2019-09-05
Firefly 342 improve jenkins to manage multiple tags

deannaji:
This ticket does:
1- Modifying gradle "dockerPublish" task to take multiple tags for a K8s build.
2- Adding new "dockerDelete" task to gradle tasks. This task deletes docker image form docker hub.

To test:
1- Use Jenkins's [snadbox](http://irsawebdev5:8100/job/sandbox/) job to build K8s build, and specify multiple tags for that build: 
2- Use Jenkins's [sandbox2](http://irsawebdev5:8100/job/sandbox2/) job to delete the K8s build. It should delete the image tag from docker hub too.

irsa-ife-env PR: https://github.com/IPAC-SW/irsa-ife-env/pull/4

### DIFF
--- a/buildScript/tasks.gincl
+++ b/buildScript/tasks.gincl
@@ -589,7 +589,7 @@ task dockerDelete (dependsOn: loadConfig){
     group = DOCKER_GROUP
 
     ext.docker_repo = "ipac/firefly"
-    ext.docker_registry = ''
+    ext.docker_registry = "https://cloud.docker.com/v2/repositories" //Docker registry defaults to Docker Hub. In case this needs to change, then either change this value here, or using 'project.appConfigProps'
 
     doLast {
         try {
@@ -610,7 +610,7 @@ task dockerDelete (dependsOn: loadConfig){
         docker_registry = docker_registry == '' || docker_registry.endsWith('/') ? docker_registry : docker_registry + '/'
 
         //DELETE Restful call:
-        def str = "https://cloud.docker.com/v2/repositories/${docker_repo}/tags/${docker_tag}/"
+        def str = "${docker_registry}${docker_repo}/tags/${docker_tag}/"
         def encoding = "${docker_user}:${docker_passwd}".bytes.encodeBase64().toString()
 
         def conn = new URL(str).openConnection();
--- a/buildScript/tasks.gincl
+++ b/buildScript/tasks.gincl
@@ -584,7 +584,8 @@ task dockerPublish (dependsOn: dockerImage) {
 
 
 task dockerDelete (dependsOn: loadConfig){
-    description = 'Delete a docker image from docker hub.'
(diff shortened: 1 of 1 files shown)


deannaji:
Oh, I will correct these missing's. However, the branch name is specified in the PR and should be used as FF branch (instead of dev). Sorry for not making it clear :)
I went ahead and specified the PR branch name (in the sandbox job) to be the default instead of dev. So I think you can build with multiple tags now.
Will continue working on the other notes.
Thanks!!

deannaji:
Oops, I take it back! I understood what you are saying about the open jdk.
I'm unable to test the multiple tags until merging :)

buildScript/tasks.gincl
```diff
@@ -555,14 +555,73 @@ task dockerPublish (dependsOn: dockerImage) {
...
+
+        def conn = new URL(str).openConnection();
+        conn.setRequestMethod("DELETE")
+        conn.setDoOutput(true)
+        conn.setRequestProperty("Content-Type", "application/x-www-form-urlencoded")
+        conn.setRequestProperty("Authorization", "Basic " + encoding)
+        conn.connect()
+
+        if(conn.getResponseCode() >= 400) {
+            throw new GradleException("Fail to delete docker image")
```
ejoliet: Could you add the URL String in the exception message? Thanks.

deannaji:
Loi, for sandbox2 the deployment you mentioned (firefly-loi--firefly-360-revert-highlight-onsort). It has a different name for the docker image (on docker hub), the image has underscore's instead of the dashes. That's why its not being deleted. Couldn't understand why thou!

loitly:
> Loi, for sandbox2 the deployment you mentioned (firefly-loi--firefly-360-revert-highlight-onsort). It has a different name for the docker image (on docker hub), the image has underscore's instead of the dashes. 

Yes, that's why I said:  `** deleting a built where tags contains '_'(underscore) does not work.`
At line tools/k8s_deploy.sh:22, we convert a  name/label into this limited character set.  This was needed because kubernetes's name/label has a more limited character set. 
I just revisited their documentation and it looks like they've expanded their charset.  https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

Kubernetes chartset is now pretty much the same as our label/tag limit except for the 128 max length.  My suggestion is to decrease our 128 max length to 64, and remove the conversion logic mentioned above.  This should fix the mismatch problem.

deannaji:
Got it, will do that :)

deannaji:
Testing using Jenkins is available now. PR description is updated!

buildScript/tasks.gincl
deannaji: Done, thanks!

PR #873, opened 2019-09-05
Firefly 175 image selection

robyww:
Firefly 175 image selection

- Show a table of image when ListOptions is pushed.
- refactor WebPlotRequest and Plot Attributes.

https://jira.ipac.caltech.edu/browse/FIREFLY-175

_To Test_: 
https://irsawebdev9.ipac.caltech.edu/firefly-175-image-selection/firefly/
- Search a bunch of images
- bring up the dialog using  ListOptions

_For code reviewers:_

- Most of the changes code was a refactor and cleanup. Focus on `ExpandedOptionsPopup.jsx` and `ImageSelectPanelV2.jsx`
- The refactor was for `WebPlotRequest.js`


ejoliet:
@robyww This is very useful feature and so works as intended. I've also search images and send it to the viewer and the table get's updated. Very cool.
Couple of minor issues: the last right hand side column is not expanded to fit the URLS width and is difficult toexpand it by sliding to the right the column header border to expand, is there a way to expand it so you can horizontally scroll without having to first expand the column width?
In my opinion, I would suggest to make the table resizable like the FITS info dialog window. 

I have a more important remark and doubt though: didn't we discuss to apply the filter of the table to the image grid, such that when i filter or sort in the table, the grid gets updated? In my test, it doesn't and that's why i'm asking if it's a bug or not implemented or even discussed before? Thanks!

robyww:
> I have a more important remark and doubt though: didn't we discuss to apply the filter of the table to the image grid, such that when i filter or sort in the table, the grid gets updated? In my test, it doesn't and that's why i'm asking if it's a bug or not implemented or even discussed before? Thanks

My understanding this the the filtering is just for helping the use choose images.  So you can filter, decide what to check then filter again and check/uncheck some more.  When OK is pushed the checked images are displayed.  That is a way it works not and I think the ticket specifically requested this behavior.

ejoliet:
> My understanding is that the filtering is just for helping the use choose images. So you can filter, decide what to check then filter again and check/uncheck some more. 

Mhm, ok, i thought it was a way to control the grid, but i might be confused. 

>When OK is pushed the checked images are displayed. That is a way it works not and I think the ticket specifically requested this behavior.

You mean an 'ok' button? I didn't see a button to apply the selection (testing in Safari)

ejoliet:
OK, in Safari browser, i don't see the bottom of the table ;-)... now i'm testing Chrome and i can see what you are talking about. Thanks, i'll continue testing.

ejoliet:
Couple of remark quickly now that i'm testing in Chrome and playing with the new feature:

UI feedback 1: I didn't expect to hide the failed images would removee the button from the table. What happen if you want to see which one were failing after the fact?  

UI feedback 2:I think if we are adding button to 'hide failed' images (or any other label), we could have a reset button to get back all the (non failed or also the failed ones?) images.

robyww:
> UI feedback 1: I didn't expect to hide the failed images would removee the button from the table. What happen if you want to see which one were failing after the fact?

Hide failed does not remove the buttons at the bottom.  Only delete failed.  

> UI feedback 2:I think if we are adding button to 'hide failed' images (or any other label), we could have a reset button to get back all the (non failed or also the failed ones?) images.

The check box at the top will reselect everyone, then OK.  The reason I did not put a third button is because it appears what the user really wants to do is get rid of the fail images.  If this is not the case I can add more buttons.  I considered not even having a "Hide Failed" button since what the user really wants to do is delete all the failed. 

We probably need more feedback. These adjustments are easy.

robyww:
> I've noticed that in Safari browser, the bottom of the table with the controls/buttons are not appearing.

Yes. @gpdf also found it and showed it to me.  He also found another bug. I am doing a pushed with the fixes.

robyww:
Just pushed fixes with the following:

  - fixed bug with safari cutting of the ok button row.
   - fixed bug with going to single image mode, then losing the toolbar
   - added enums to the status and wavelength row- request by gpdf
   - Made the dialog wider and the URL field wider.

_rebuilt:_ https://irsawebdev9.ipac.caltech.edu/firefly-175-image-selection/firefly/

src/firefly/js/core/AppDataCntlr.js
```diff
@@ -211,7 +211,7 @@ export function getConnectionCount(channel) {
  * @param wp center WorldPt
  * @param corners array of 4 WorldPts that represent the corners of a image
  */
-export const dispatchActiveTarget= function(wp,corners) {
+export const dispatchActiveTarget= function(wp,corners=undefined) {
```
loitly: is this for documentation purposes?  it can only be undefined when it's undefined.

src/firefly/js/ui/SimpleComponent.jsx
```diff
@@ -45,7 +46,9 @@ export class SimpleComponent extends PureComponent {
  * By default, this will call Object.is() to ensure the state has changed before
  * calling setState.  To override this behavior, use useStoreConnector.bind({comparator: your_compare_function}).
  *
- * @param stateGetters  one or more functions returning a state
+ * @param stateGetters  one or more functions returning a state, when called the oldState is passed as a parameter.
+ * This allows a stateGetter to optionally act as a comparator.  If the stateGetter does not care about the changes
+ * then it can return the oldState and there will be no state update.
```
loitly: This is a good addition.  I like it.

src/firefly/js/core/AppDataCntlr.js
robyww: yes, I stared doing this more because is kind of self documents.

vandesai1:
Comments:

1. In the TLI, "project" --> "Project Type" to match the terminology on the search screen.
2. In the TLI, "type" --> "Band" to match the terminology on the search screen.
3. Documentation is strangely colorful. Will talk to Luisa.
4. "Hide Failed" and "Delete Failed" are a bit confusing in terms of push buttons. I thought it was telling me that something failed.
5. In the TLI, the status for the failed images is too long. 
6. The "Hide Failed" button should only hide images, not unhide images.

vandesai1:
7. The dialog should not be named "Choose which"

vandesai1:
9. The wavelength column needs to be able to be sorted physically.

robyww:
_Action Items for Trey:_
In the TLI, "project" --> "Project Type" to match the terminology on the search screen.

- [ ] 2.  In the TLI, "type" --> "Band" to match the terminology on the search screen.
- [ ] 5. In the TLI, the status for the failed images is too long.
             Will change to Just `Failed`
- [ ] 8. The wavelength column needs to be able to be sorted physically.


_Comments:_
> 4. "Hide Failed" and "Delete Failed" are a bit confusing in terms of push buttons. I thought it was telling me that something failed.

When you see these buttons something has failed. 

> 6.  The "Hide Failed" button should only hide images, not unhide images.

That is the way it works now, it only will hide.

robyww:
Action Items for @gpdf or @vandesai1 

- [ ] 1. In the TLI, "project" --> "Project Type" to match the terminology on the search screen. **I need a one work Name for this column**

- [ ] 7. The dialog should not be named "Choose which" **I need a title for the dialog**

- [ ] I need you to determine if we should remove the `hide images` button

- [ ] I need to you to give the the text for the `remove images` button

vandesai1:
Continuing to test based on confluence plan: https://confluence.ipac.caltech.edu/pages/viewpage.action?spaceKey=FIREFLY&title=Table-of-loaded-images+proposal

9. If I do a search on arp220, cutout size = 500 arcsec, AllWISE 4 bands, the 4 images display and all look the same size. If I then open the TLI and uncheck W4, the relative sizes of the remaining images look different.
 
10. When an image pane is "x"-d away, the corresponding row should be unchecked in the TLI. Right now it is removed. This might be a little unclear in the confluence page, but see the red text in the priority 1 section.

11. An "info" ("i"-in-a-circle) icon should be present in each non-failed row, which brings up the same image-info dialog box that the image-toolbar's info button does. Instead, we have a long link. Is it possible to replace it with the same icon as in the search page?

vandesai1:
> Action Items for @gpdf or @vandesai1
> 
> * [ ]  1. In the TLI, "project" --> "Project Type" to match the terminology on the search screen. **I need a one work Name for this column**

The only one-word possibility is "Type".

> * [ ]  7. The dialog should not be named "Choose which" **I need a title for the dialog**

I would suggest "Loaded Images".

> * [ ]  I need you to determine if we should remove the `hide images` button

No, we should not remove the "hide images" button. I thought the proposal was to remove the "delete images" button. Yes, I think we should remove the "delete images" button.

> * [ ]  I need to you to give the the text for the `remove images` button

"Hide Failed Images" or "Hide Failures".

dlmcelroy:
> 
> "Hide Failed Images" or "Hide Failures".

I was going to suggest "Hide Failures" - it makes it clear what the subject of the Hide action is.

In general I hesitate to make a lot of detailed critiques as I'm not as familiar with the ins and outs of Firefly BUT as a general critique, I think every effort should be made to have verbiage on the new UI component match other existing verbiage.  Vandana has that pretty well covered in her critiques

At least in the Demo dated 9/17/19, when you hit 'Hide Failed', it re-checked unchecked images.  User selections in terms of checked/unchecked should not be modified by the Hide button.

robyww:
I made a new commit with the following (mostly request) changes:

- The displayed images shown are now controlled by filtering
- Sorting of the table is not reflected in the image layout
- Bottom Buttons (there are none)
   - OK button is removed.
   - Other "delete" buttons are removed
   - Table actions are immediate, no OK button, dialog closed with the x
- Table toolbar is enabled with only options and filter buttons

- Wavelength column is always in microns and numerically sortable.

- Title and Labels
  - Changed dialog title to "Loaded Images"
  - Change "Project" column title to "Type"
  - Changed "Type" column title to "Band"
  - Help column cells: now says "Help" as a link, if no help url it is blank


- None visible changes:
   - TablePanel.jsx now has a `showToggleTextView` property
   - Links code change to show a blank when no URL is generated, instead of the given text
        - `TableRenderer.js`, line 291, @loitly might want to look at this.

ejoliet:
Thanks for updating and make the changes.
I tested it and looks awesome. Toolbar and reset feature is very useful. I've laso noticed the URL is now a link 'help' which i think is better.

Minor feedbacks/comments: 
- Should we keep an 'Close' button to close the dialog? Most of the dialog have a ok or close button, see foe example FITS info dialog.
- I think we need to a help anchor at the bottom right-hand corner to link to a documentation page for explanation of the table content and controls.

robyww:
Task for Trey: 

- [ ] Add 'Close' button
- [ ] Add  help anchor

gpdf:
There seems to be a bug or surprising interaction.  When I have a bunch of 500x500 cutouts loaded, and I filter away enough of them that the grid of image viewers changes its dimensions (e.g., from a 4x4 grid to  3x4 grid), whichever one was the selected image (in the image UI, not in the TLI) before the filter action ends up zoomed in more than the others.  It looks like all the others retain their original size in on-screen, now not filling out their now-larger viewer subframes, while the selected one is zoomed to fit its viewer.

image

gpdf:
@ejoliet made a comment which clarified for me what's going on above: it depends on whether the WCS/etc.-lock is in force or not.  That _sort of_ makes sense but the results are counter-intuitive:

- When the lock is OFF, the highlighted (in the image viewer grid, not the TLI) image DOES grow but the others do not.
- When the lock is ON, upon filtering and changing the grid dimensions, NO images change their on-screen size.  That is, when removing images makes the image viewer subframes larger on-screen, the actual images do NOT grow to fill the newly available space.

If this arose from a consistent underlying behavior, I would expect that when the lock was ON, ALL images would grow.

By the way, when removing filters, all the images shrink back down as expected to their previous states.

I think the CCB-D may want to think about what the ideal behavior should be.

robyww:
> I think the CCB-D may want to think about what the ideal behavior should be.

It does need work and is a bug, but I think it is preexisting. I would like to approach this in a separate ticket. Feedback on how it should work would be helpful.

robyww:
> > I think the CCB-D may want to think about what the ideal behavior should be.
> 
> It does need work and is a bug, but I think it is preexisting. I would like to approach this in a separate ticket. Feedback on how it should work would be helpful.

I did do a little digging on this issue. The problem is that when we changed the default matching it created this behavior. I am going to go ahead and put in a fix to make it work similar to before. It is simply this: When WCS match lock is on it will make all the images lock to the active one as expected. The active will fit or fill and the others will size appropriately. In fact it works like that now.  When match lock is off it will make all the images fit or fill when the grid changes.

src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -288,6 +288,7 @@ export const LinkCell = React.memo((props) => {
                         const target = action || '_blank';
                         const rvalue = resolveHRefVal(tableModel, value, absRowIdx);
                         const rhref = resolveHRefVal(tableModel, href, absRowIdx, val);
+                        if (!rhref) return '';
```
loitly: Yes, perfectly reasonable.

PR #876, opened 2019-09-11
Firefly-159:The issues in "prepare download" that are seen in Time Series, Sofia etc.

lznakano:
See ticket here: https://jira.ipac.caltech.edu/browse/FIREFLY-159

  1. Fixed bugs in BackgroundCntlr.js, BackgroundMonitor.jsx, BackgroundUtil.js, so that the file name can be changed
  2. Add common method "doDownload"  to WebUtil.js
  3. Modify the DownloadOptionPanel so that it can handle background job

There are 5 bugs seen in Finderderchart, Time Series and Sofia.  The bugs come from both Firefly and IRSA.   This branch fixed the issues in Firefly.  How to test it, we have to run the applications that use those components.  Please use the test link below:
URL: https://irsawebdev9.ipac.caltech.edu/firefly-159-irsa-3113/applications/sofia/

Please refer the ticket Firefly-159 and IRSA-3113 for the bug lists.

### DIFF
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -13,6 +13,9 @@ import {showInfoPopup} from '../ui/PopupUtil.jsx';
 
 import update from 'immutability-helper';
 import {ServerParams} from '../data/ServerParams';
+import {validateFileName} from '../ui/DownloadOptionsDialog';
+import {getWorkspacePath, WS_SERVER_PARAM} from '../visualize/WorkspaceCntlr';
+import {doDownloadWorkspace} from '../ui/WorkspaceViewer';
 
 
 const MEG          = 1048576;
@@ -376,11 +379,13 @@ export function copyToClipboard(str) {
 }
 
 /**
- * @param {string} url  the url to download.  It should be based on AnyFileDownload
- * @param {number} [numTries=1000]  number of time to check for progress until giving up
- * @returns {Promise}  resolve is called on DONE and reject when FAIL.
+ *
+ * @param url
+ * @param fileName
+ * @param numTries
+ * @returns {Promise<any>}
  */
-export function downloadWithProgress(url, numTries=1000) {
+export function downloadWithProgress(url, fileName, numTries=1000) {
     return new Promise(function(resolve, reject) {
         const {search} = parseUrl(url);
         let cnt = 0;
@@ -405,7 +410,7 @@ export function downloadWithProgress(url, numTries=1000) {
             }, interval);
         };
         doIt();
-        download(url);
+        download(url, fileName);
     });
 }
 
@@ -434,7 +439,8 @@ function resolveFileName(resp) {
 export function download(url, filename) {
     const {protocol, host, path, hash, searchObject={}} = parseUrl(url);
     const cmd = searchObject[ServerParams.COMMAND];
-    url = `${protocol}//${host}${path}?${ServerParams.COMMAND}=${cmd}` + (hash ? '#' + hash : '');      // add cmd into the url as a workaround for server-side code not supporting it
+    url = cmd? `${protocol}//${host}${path}?${ServerParams.COMMAND}=${cmd}` + (hash ? '#' + hash : ''):      // add cmd into the url as a workaround for server-side code not supporting it
+        url;
 
     const params = Object.fromEntries(
                         Object.entries(searchObject)
@@ -849,3 +855,31 @@ export function hashCode(str) {
 export function isNumeric(n) {
     return !isNaN(parseFloat(n)) && isFinite(n);
 }
+
+
+export function doDownload (bgStatus, isWorkSpace, wsSelect, fileName) {
+    const url1 = get(bgStatus, ['ITEMS', 0, 'url']);
+    const urlInfo =parseUrl(url1);
+    const searchObj = urlInfo.searchObject;
+    const file = searchObj.file;
+    const url = isWorkSpace ? `${getRootURL()}sticky/CmdSrv`
+        : url1;
+
+    if (isWorkSpace) {
+        if (!validateFileName(wsSelect, fileName)) return false;
+    }
+
+    if (isWorkSpace) {
+        const zipFileName = fileName + '.zip';
+        const params = {[WS_SERVER_PARAM.currentrelpath.key]:getWorkspacePath(wsSelect, zipFileName),
+            [WS_SERVER_PARAM.newpath.key]: zipFileName,
+            file,
+            [ServerParams.COMMAND]: ServerParams.WS_PUT_IMAGE_FILE,
+            [WS_SERVER_PARAM.should_overwrite.key]: true};
+
+        url && doDownloadWorkspace(url, {params});
+        //showInfoPopup('File:' + fileName + '.zip is saved to workspace!', 'Workspace Info');
+    } else {
+        url && download(url, fileName);
+    }
+};
--- a/src/firefly/js/core/background/BackgroundCntlr.js
+++ b/src/firefly/js/core/background/BackgroundCntlr.js
@@ -174,11 +174,12 @@ function bgSetEmail(action) {
 function bgPackage(action) {
     return (dispatch) => {
         const {dlRequest, searchRequest, selectionInfo, bgKey} = action.payload;
-
+        const fileName = dlRequest.fileName|| dlRequest.Title.split(':')[0];
         const onComplete = (bgStatus) => {
             const url = get(bgStatus, ['ITEMS', 0, 'url']);
+
             if (url && isSuccess(get(bgStatus, 'STATE'))) {
-                download(url);
+                download(url, fileName);
             } else {
                 showInfoPopup(getErrMsg(bgStatus));
             }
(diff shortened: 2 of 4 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
```diff
@@ -165,7 +165,18 @@ private static void logActivity(File f) {
...
+        String queryString =req.getQueryString();
+
+        if (queryString==null){
+            try {
+                throw new Exception("Status can not be tracked because the queryString is empty");
+            } catch (Exception e) {
+                Logger.error("HttpServletRequest has a null query string, the status can not be tracked");
+
+
+            }
```
loitly: Why throw an exception just to catch it?  Do you want to log it?  Do you want execution to stop?  Do you want the caller to handle the exception?  What do you want it to do?

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -244,24 +261,24 @@ function hasProprietaryData(tableModel={}) {
...
- * @param {boolean} [show=true] show or hide this dialog
+ *
+ * @param panel
+ * @param title
+ * @param show
  */
-function showDownloadDialog(panel, show=true) {
-    const ttl = panel.props.title || DOWNLOAD_DIALOG_ID;
+export function showDownloadDialog(panel, title, show=true) {
+
```
loitly: Why did you remove logic here outside?  It's being called by several other places.  Were they all changed?

src/firefly/js/util/WebUtil.js
```diff
@@ -848,4 +854,32 @@ export function hashCode(str) {
 
 export function isNumeric(n) {
     return !isNaN(parseFloat(n)) && isFinite(n);
-}
\ No newline at end of file
+}
+
+
+export function doDownload (bgStatus, isWorkSpace, wsSelect, fileName) {
```
loitly: What is the difference between `download` and `doDownload`?  Why is it exported and not used in this PR, but is used in another PR, that's not linked to this one.
If you merge the other PR and not this one, build will fail.  This change should have been with the other PR.  Or, link them.

ejoliet:
> I will reject this PR because I don't think the behavior is acceptable. I also did not see how finderchart download work. They should be the same.

> When downloading, the working mask is over the right panel. It will stay there until it's finish.
Or, you can download something small(one image), then it will clear the previous working mask.
The previous request is still going. You have no indication of it though.
Another problem with this is the location of the mask. It's in the right panel in SOFIA. Where will it be for FinderChart? How about the next application we decide to build? It'll behave differently for each one?

> For some reason, login is not working anymore. I can't test saving into workspace.

I think you are suggesting same behavior in any Firefly based app? I agree that in term of UI visual, we should be consistent when and where the 'send' to background monitor button should appear. 
For that reason, I would recommend that it appear on top of the frame filling all the window(s) or at least overlap the app completely, so we can use the same behavior in different context/app. SOFIA for example should have the 'send' to monitor in the middle and not covering the detail tab on the right.

In this case (or any other such as FinderChart), should we change in Firefly the common way or is it an app by app basis change? 
Again, better to have a meeting about this feature, it's difficult to see/explain the details here in my opinion.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
lznakano: This is the part of the bug I found.  The query status is NULL, thus, the download progress run forever.  I fixed the status.  In order to guard it,  I added the exception to log the error.

src/firefly/js/ui/DownloadDialog.jsx
lznakano: Yes, I changed it in all places.  The DownloadButton was a class before.  Then this worked.  I don't know when and who changed to a stateless function.  After that change, the "this" keyword no longer worked.  Thus, the Dialog was not able to be closed as it reported in IRSA-3113.  If you know a better way to handle this, please feel free to push your changes.  THANKS!

src/firefly/js/util/WebUtil.js
lznakano: The doDownload is the function I moved from Finderchart to here.   It will be used for workspace download.  In this pull request, it was only used by finderchart.   The download is an existing function.  I just keep it there for now.   Once Firefly-84 is done,  I will make the final clean up.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
robyww: I agree with @loitly this code is confusing.  The code is working just like before. It is just logging a message.  Why not just do this?
```
if (queryString==null){
         Logger.error("HttpServletRequest has a null query string, the status can not be tracked");
}
```

lznakano:
@loitly 
I don't understand why you reject this pull request.  This pull request is not the same as Firefly-84.  @ejoliet issued another bug report IRSA-3113.  I decided to include the Firefly issue from IRSA-3113 to Firefly-159 since it is still in progress.  Please see my analysis in both tickets.  I tended to write my findings and implementations in the tickets because I can come back to see them if I need to.  The codes changes early this year in dev introduced a few bugs and four of them you noted when you review my Firefly-84 last month.  Although you noticed in my pull request, the bugs are from the dev.  I spent quite some effort to locate the bugs.  What I want you to review is to check if my fix is appropriate.  This pull request has nothing to do with workspace.  In this pull request, all three applications, Sofia, LightCurve and Finderchart behavior the same (for local download ONLY).   I have one pending branch for LightCurve.  After I am done, I will build the LightCurve.   As where to show the backgroundable mask, we can decide it later.  I just copied what you did for Finderchart.   LightCurve and Sofia did not use the background mask and did not implement the backgroundable option.  That is why I am working in three tickets, Firefly-159 that fixed bugs in firefly, and IRSA-3113 that enabled backgroundable option for Sofia and IRSA-3112 that enabled backgroundable option for LightCurve.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
lznakano: If it works before, I would change it.  I am OK to remove the exception, however, if the null string comes again, it may need to debug it again.

lznakano:
> > I will reject this PR because I don't think the behavior is acceptable. I also did not see how finderchart download work. They should be the same.
> 
> > When downloading, the working mask is over the right panel. It will stay there until it's finish.
> > Or, you can download something small(one image), then it will clear the previous working mask.
> > The previous request is still going. You have no indication of it though.
> > Another problem with this is the location of the mask. It's in the right panel in SOFIA. Where will it be for FinderChart? How about the next application we decide to build? It'll behave differently for each one?
> 
> > For some reason, login is not working anymore. I can't test saving into workspace.
> 
> I think you are suggesting same behavior in any Firefly based app? I agree that in term of UI visual, we should be consistent when and where the 'send' to background monitor button should appear.
> For that reason, I would recommend that it appear on top of the frame filling all the window(s) or at least overlap the app completely, so we can use the same behavior in different context/app. SOFIA for example should have the 'send' to monitor in the middle and not covering the detail tab on the right.
> 
> In this case (or any other such as FinderChart), should we change in Firefly the common way or is it an app by app basis change?
> Again, better to have a meeting about this feature, it's difficult to see/explain the details here in my opinion.

This pull request is about the bugs.  The Finderchart is the only application has the "send to background" option implemented.  I did the same for sofia and lightCurve to make them behave the same.  If that is not expected behavior, it is a design issue.

lznakano:
Here is the URL for finderchart:
URL: https://irsawebdev9.ipac.caltech.edu/firefly-159-Lijun/applications/finderchart/

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileDownload.java
lznakano: I reverted this file the same as the one in dev.  The "UNKNOWN" status is assigned when the cache key (added in this function) does not match the key received from the client in the BackgroundEnv.

ejoliet:
I was looking at all the other ticket related to this one and i think we need to discuss the UI design part but all the fixes and refactoring Lijun did are indeed useful. See IRSA-3113 and IRSA-3121. Those are fixed. Let's discuss the placeholder of the monitor button but i feel that we might not be able to either find a common place or we might need to put it always in the middle of the applications window. What do you think?

lznakano:
> I was looking at all the other ticket related to this one and i think we need to discuss the UI design part but all the fixes and refactoring Lijun did are indeed useful. See IRSA-3113 and IRSA-3121. Those are fixed. Let's discuss the placeholder of the monitor button but i feel that we might not be able to either find a common place or we might need to put it always in the middle of the applications window. What do you think?

I think where to put the button should be a new ticket because it changes the current behavior.  It was not a bug introduced  in this pull request.

lznakano:
Rebased with dev, redid the bug fixes and tested.

PR #877, opened 2019-09-17
IRSA-3121-showing the option to Send to background in TimeSeries

lznakano:
1. Remove the redundancy of the  download panel (ZtfDownloadOptionPanel, PtfDownloadOptionPanel), and all download among missions share the same download panel
  2. Modify the downloaderOptPanel in LcResult,  to make it backgorundable and share with all missions
  3. Add BgMaskPanel in LcResult

@ejoliet 
When I worked on Firefly-159 and IRSA-3113, I found that the multi-downloand panel is very problem pron.  I fixed the bug in DefaultPanel, I found only WISE worked right and ZTF, PTF all did not work.  After I looked into these two, I found there is no need to maintain three similar download panels.   If you compare the ptf/ztf panel with the default panel, you find the only difference is the mission parameters.  Thus, I removed ZTF/PTF download panel and modified the default download panel to share among the missions.  When I fix bug, I only need to fix it in one download panel. 

This pull request is depending on Firefly-159.  FYI

Test URL is
 https://irsawebdev9.ipac.caltech.edu/irsa-3121-Lijun/irsaviewer/ts.html

### DIFF
--- a/src/firefly/js/core/background/BackgroundCntlr.js
+++ b/src/firefly/js/core/background/BackgroundCntlr.js
@@ -173,10 +173,12 @@ function bgPackage(action) {
         SearchServices.packageRequest(dlRequest, searchRequest, selectionInfo)
             .then((bgStatus) => {
                 if (bgStatus) {
+                    const fileName = dlRequest.fileName|| dlRequest.Title.split(':')[0];
+                    bgStatus.fileName = fileName;
                     bgStatus = bgStatusTransform(bgStatus);
                     const url = get(bgStatus, ['ITEMS', 0, 'url']);
                     if (url && isSuccess(get(bgStatus, 'STATE'))) {
-                        download(url);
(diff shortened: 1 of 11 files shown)


ejoliet:
You might want wait for the #881 PR to be merged and rebase because some refactoring code here is probably still needed.

loitly:
Please don't merge.  I believe my changes addresses all of these issues.  What is it in this PR that is not fixed in mine?  Most of the changes here conflict with mine.

ejoliet:
Loi is correct, and i know what we said Lijun about having the mask on the table but now it will be consistent to inherit the Firefly one which has all the visual included so nothing to be added. I think you still want to keep the refactoring of the duplicated function.

ejoliet:
@lznakano i think most of the issues were taking care by the update in #881 .

PR #880, opened 2019-09-20
FIREFLY-381: Prepare download dialog updates

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-381

Add progress indicator and option to background prepare download jobs.  This should be implemented such that the behavior is consistent among all applications using Firefly.
See `Intended behavior: ` section of the ticket for more details.

To test, use Time Series:  https://irsawebdev9.ipac.caltech.edu/FIREFLY-381_download_dialog_mask/firefly/ts.html
- Upload a file
- Select all from the `Input Data` result table
- Click `Prepare Download` located top right corner.

PS.  
https://irsawebdev9.ipac.caltech.edu/FIREFLY-381_download_dialog_mask/applications/finderchart/
I built finderchart to see if the above changes would propagate down to finderchart, and it did not.  The reason is because finderchart is using a copy of the above dialog and not using one made through widget composition. (FinderchartDownloadDialog.jsx)  This is an example of the multiple versions issue I mentioned in the meeting.

### DIFF
--- a/src/firefly/js/core/background/BackgroundCntlr.js
+++ b/src/firefly/js/core/background/BackgroundCntlr.js
@@ -2,7 +2,6 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {take} from 'redux-saga/effects';
 import {set, get, has, pick, isNil} from 'lodash';
 
 import {flux} from '../../Firefly.js';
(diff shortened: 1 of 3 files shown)


lznakano:
I saw your changes.  You added the MaskPanel in DownloadDialogPanel, Great! The existing mask property previously did not have the "Send to background" ( I tried once).   It worked nicely now!

You can test Sofia.  The changes should be propagated to it.   

Question:
When the job is sent to background, should the dialog be closed?  Is it your intension to keep it open always?

loitly:
> When the job is sent to background, should the dialog be closed? Is it your intension to keep it open always?

It's not absolutely clear that closing the dialog after download completed or when the user clicked on send to background is a good idea.  Plus, the added behavior is not trivial, requiring additional states to track.  
Because of this, I've decided to leave it as is until the feature is requested by CCB.

ejoliet:
I think for now let's not close the dialog and make more work in that direction. 
I think we can discuss it later in CCB and among us but at least for now the purpose is met, which is to have a unique component and visual for any other app to use it.

PR #881, opened 2019-09-24
FIREFLY-384: update to download dialog

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-384

additional changes here: https://github.com/IPAC-SW/irsa-ife/pull/90

- create new workspace selection panel as an example of how to write composable widgets
- add Title to dialog
- add SaveAs plus workspace support
- add show* props to allow for more flexibility when laying out fields.
- refactor LightCurve download dialogs.

To test:
https://irsawebdev9.ipac.caltech.edu/FIREFLY-384_update_download_dialog/irsaviewer/timeseries
https://irsawebdev9.ipac.caltech.edu/FIREFLY-384_update_download_dialog/applications/sofia/

PS:  Workspace selection UI and related selection data is added to the dialog and also to the download request.  But, the logic to `save to workspace` is not yet implemented.
If you wish for me to add that logic to the PR as well, let me know.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcResult.jsx
+++ b/src/firefly/js/templates/lightcurve/LcResult.jsx
@@ -24,7 +24,6 @@ import {LcImageViewerContainer} from './LcImageViewerContainer.jsx';
 import {SplitContent} from '../../ui/panel/DockLayoutPanel.jsx';
 import {LC, getViewerGroupKey, updateLayoutDisplay} from './LcManager.js';
 import FieldGroupUtils from '../../fieldGroup/FieldGroupUtils.js';
-import {ValidationField} from '../../ui/ValidationField.jsx';
 import {LcImageToolbar} from './LcImageToolbar.jsx';
 import {DownloadOptionPanel, DownloadButton} from '../../ui/DownloadDialog.jsx';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
@@ -37,8 +36,10 @@ import {logError} from '../../util/WebUtil.js';
 import {getConverter, getMissionName,  DL_DATA_TAG} from './LcConverterFactory.js';
 import {convertAngle} from '../../visualize/VisUtil.js';
 
-const resultItems = ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'searchDesc', 'images',
-    LC.MISSION_DATA, LC.GENERAL_DATA, 'periodState'];
+const resultItems = () => {
+    return ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'searchDesc', 'images',
+            LC.MISSION_DATA, LC.GENERAL_DATA, 'periodState'];
+};
 
 
 export class LcResult extends PureComponent {
@@ -46,7 +47,7 @@ export class LcResult extends PureComponent {
     constructor(props) {
         super(props);
 
-        this.state = Object.assign({}, pick(getLayouInfo(), resultItems));
+        this.state = Object.assign({}, pick(getLayouInfo(), resultItems()));
     }
 
     componentDidMount() {
@@ -61,7 +62,7 @@ export class LcResult extends PureComponent {
 
     storeUpdate() {
         if (this.iAmMounted) {
-            const nextState = pick(getLayouInfo(), resultItems);
+            const nextState = pick(getLayouInfo(), resultItems());
             this.setState(nextState);
         }
     }
@@ -115,6 +116,28 @@ export class LcResult extends PureComponent {
     }
 }
 
+export function defaultDownloadPanel(mission='', cutoutSize, addtlParams={}) {
+    mission = mission.replace(/[\/ ]/g, '_');       // clean up mission description to be used for save as value.
+    return (
+        <DownloadButton>
+            <DownloadOptionPanel
+                groupKey = {mission}
+                dataTag = {DL_DATA_TAG}
+                cutoutSize={cutoutSize}
+                title={'Image Download Options'}
+                dlParams={{
+                    MaxBundleSize: 200 * 1024 * 1024,    // set it to 200mb to make it easier to test multi-parts download.  each wise image is ~64mb
+                    TitlePrefix: mission,
+                    BaseFileName: `${mission}_Files`,
+                    DataSource: `${mission} images`,
+                    FileGroupProcessor: 'LightCurveFileGroupsProcessor',
+                    ...addtlParams
+                }}/>
+        </DownloadButton>
+    );
+}
+
+
 const ExpandedView = ({expanded, imagePlot, xyPlot, tables}) => {
     const view = expanded === LO_VIEW.tables ? tables
         : expanded === LO_VIEW.xyPlots ? xyPlot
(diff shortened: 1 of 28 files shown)


src/firefly/js/ui/WorkspaceSelectPane.jsx
```diff
@@ -0,0 +1,128 @@
...
+                    : isEmpty(wsList) ? <div style={{color:'maroon', fontStyle:'italic', padding:10}}> {'Workspace access error: ' + getWorkspaceErrorMsg()} </div>
+                    : <WorkspaceSave fieldKey={wsSelectKey} files={wsList} value={wsSelect} />;
+
+    return (
+        <div style={{ border:'1px solid #eaeaea', padding:3, position:'relative', minHeight:60, minWidth:550}}>
+            {content}
+        </div>
+    );
+}
+
```
ejoliet: Very minor: You might want to remove those empty lines below ;-)

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -153,8 +160,17 @@ export function DownloadOptionPanel (props) {
...
     const maskPanel = <BgMaskPanel componentKey={bgKey} style={maskStyle}/>;
 
+    const saveAsProps = {
+        fieldKey: 'BaseFileName',
+        initialState: {
+            value: get(dlParams, 'BaseFileName')
+        }
+    };
+
+    const dlTitle = get(dlParams, 'TitlePrefix', 'Download') + '-' + dlTitleIdx;
```
ejoliet: Although the index is increased, in my test, the title stays the same on each submit.

src/firefly/js/templates/lightcurve/LcResult.jsx
```diff
@@ -115,6 +116,28 @@ export class LcResult extends PureComponent {
...
+    return (
+        <DownloadButton>
+            <DownloadOptionPanel
+                groupKey = {mission}
+                dataTag = {DL_DATA_TAG}
+                cutoutSize={cutoutSize}
+                title={'Image Download Options'}
+                dlParams={{
+                    MaxBundleSize: 200 * 1024 * 1024,    // set it to 200mb to make it easier to test multi-parts download.  each wise image is ~64mb
+                    TitlePrefix: mission,
```
lznakano: Title needs to have date and time.  That was a ticket a year ago to request that kind format.  But I can add it after you merge to the dev.

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -109,8 +112,11 @@ const noticeCss = {
     marginBottom: 3,
 };
 
+let dlTitleIdx = 0;
+
 export function DownloadOptionPanel (props) {
     const {groupKey, cutoutSize, help_id, children, style, title, tbl_id, dlParams, dataTag} = props;
+    const { showZipStructure=true, showEmailNotify=true, showFileLocation=true, showTitle=true } = props;
```
lznakano: Does showFileLocation=true not cause issue in LSSTCatalogSelectViewPanel?  LSSTCatalogSelectViewPanel is the only NONE IRSA application which uses DownloadOptionPanel.   LSSTCatalogSelectViewPanel does not have workspace.  I forgot how to build LSST so I am not able to test it.

src/firefly/js/templates/lightcurve/LcResult.jsx
loitly: Check with @ejoliet before putting it back.  If I understand it correctly, it was requested because there was only one field used to hold both Title and Save As.  I've now separate them.  It should be clearer now without having to introduce timestamps into the label.

src/firefly/js/ui/DownloadDialog.jsx
loitly: It should not be an issue since the panel will only appear when `useWs` is also true. 
But, this remind me I need to show Save As even if workspace is not used.  I will add this into the PR.

ejoliet: Yes, the `label` is fine (even though CCB could argue to change it to something else in the future) but the value is the one you can pass outside and this one will contain time stamp. So i think we are good. @lznakano should confirm once you merge.

src/firefly/js/templates/lightcurve/LcResult.jsx
ejoliet: Yes, the label is fine and should be fixed (even though CCB could argue to change it to something else in the future) but the value is the one you can pass outside and this one will contain time stamp. So i think we are good. @lznakano should confirm once you merge.

loitly:
@ejoliet @lznakano 
 - fix Title not updating
   Title only update when dialog is re-opened.  This is so that the value is not changing while it's opened.
 - add null check to Title
 - show Save As even if workspace is not used

Also, verify that `email` still works.  There were tickets reporting that it's not working.  We should check on them.

Changes are pushed and test instances rebuilt.  Let me know if there's more you would like me to do.

lznakano:
Look great!
Here is a few issues I encountered:
1.  If I save just one file, the Save As fileName is not used.  If I save a package, it works fine. 
2. If I enabled email option, and I receive email only when the job is sent to the background.  I think this is expected behavior, right?  
3. Maybe we can discuss with scientists later if the time stamp is still needed. 
4. The bugs I fixed in IRSA-159 are still there.   After this pull request is merged, I can rebase with dev and the bugs should all be fixed.  There are no conflict at all. 

For saving to workspace, my logic may not be the same as what Loi's one.  The implementation is very trivial (I mean mine).  Loi can feel free to add it.

ejoliet:
>Also, verify that email still works. There were tickets reporting that it's not working. We should check on them.

I've tested the email in this PR but it didn't work, although it worked in irsadev SOFIA at some point (recently).

ejoliet:
@loitly go ahead and add the workspace implementation. I think it make more sense to have it in the PR now based on the may other changes already @lznakano had to do that might be in conflict now. Thanks.

loitly:
> 1. If I save just one file, the Save As fileName is not used.  If I save a package, it works fine.
I can't duplicate.

> 2. If I enabled email option, and I receive email only when the job is sent to the background.  I think this is expected behavior, right?
Yes, that's working as designed.

lznakano:
Screen Shot 2019-09-25 at 2 08 52 PM

When you select ztf, here it what looked at.

loitly:
@lznakano @ejoliet Lots of changes just updated.  Please test thoroughly, using multiple browsers when possible.

- update "Send to background" style  *** (UI changes)
- add multiparts download support for interactive download(not background)
- implements backend support for 'Save to Workspace' option
- validate workspace path before downloading
- add progress message to maskpane

https://irsawebdev9.ipac.caltech.edu/FIREFLY-384_update_download_dialog/irsaviewer/timeseries
https://irsawebdev9.ipac.caltech.edu/FIREFLY-384_update_download_dialog/applications/sofia/

Some cases to pay attention to:
- Interactive download that produces multiple files.  I use wise with original images.
- Save to workspace with multiple files.. watch for file naming convention
- Workspace UI updated when it suppose to.  User should always see the latest.
- Try to find fail cases and validate whether or not it's handled correctly.

loitly:
> > If I save just one file, the Save As fileName is not used. If I save a package, it works fine.

> I can't duplicate.

I am able to duplicate this now.  It has nothing to do with ZTF.  When you select only one row, where only one file is downloaded, it does not go through the packaging logic and therefore the original file name is present to the user instead.
@ejoliet , you or CCB should spend a few minutes on this and decide what the appropriate behavior should be.   Regardless, a new ticket should be created for this.

lznakano:
> > > If I save just one file, the Save As fileName is not used. If I save a package, it works fine.
> 
> > I can't duplicate.
> 
> I am able to duplicate this now. It has nothing to do with ZTF. When you select only one row, where only one file is downloaded, it does not go through the packaging logic and therefore the original file name is present to the user instead.
> @ejoliet , you or CCB should spend a few minutes on this and decide what the appropriate behavior should be. Regardless, a new ticket should be created for this.

@loitly That is not what I encountered.  When I uploaded a wise table, and entered a "Save As" name myWise, it was saved as "myWise.zip" file.  When I uploaded a ztf table, and entered a "Save As" name, myZTF, it saved with the original table name (I guess it was) without using the name I entered.  In both cases, I selected one row ONLY.  Maybe it just happened to me.  I will clean my environment and cache to try it again.

ejoliet:
Unfortunately, dev is unstable with respect to SOFIA data, the data is being ingested with new authentication wiring.
Although the SOFIA dev database looks broken, i've tested it with AOR and UI looks good.
I think we should merge and take from there.

PR #883, opened 2019-09-26
FIREFLY-266 Update plotly.js

tgoldina:
https://jira.ipac.caltech.edu/browse/FIREFLY-266
Test deployment: https://irsawebdev9.ipac.caltech.edu/firefly-266_plotly/firefly/

Updated plotly.js to v1.49.4
- The main change is that title is now an object with text and font attributes. String titles have been deprecated. API users might need to convert, if they are setting title explicitly. 
- 'scattergl' behavior did not change, the bug with error bars disappearing on relayout is still there.

Bug fixes: 
- fixed basic options update for non-firefly charts (the update was ignored before)
- fixed creating chart with color map (color map was not taking effect when a chart was created from UI)
- fixed reverse axis behavior (was broken)


tgoldina:
@ejoliet, @loitly I've added handleApiChanges to ChartsCntlr.js CHART_ADD action creator. It should  care of title format conversion. Reverted back Wei's test SED charts to make sure it works: https://irsawebdev9.ipac.caltech.edu/firefly-266_plotly/firefly/demo/ffapi-sed.html

tgoldina:
The test deployment has been updated. Please, review.

PR #884, opened 2019-10-03
DM-20248: client side format

loitly:
https://jira.lsstcorp.org/browse/DM-20248
https://github.com/Caltech-IPAC/firefly/pull/884

- remove format guessing logic
- remove default format
- send original values to the client
- format on the client as needed
- add sprintf library as external source
- add testing for sprintf and tables
- add `showClientTable` to public API

Tests:
https://irsawebdev9.ipac.caltech.edu/DM-20248_client_format/firefly/
https://irsawebdev9.ipac.caltech.edu/DM-20248_client_format/firefly/test/tests-table.html
src/firefly/js/externalSource/__tests__/sprintf-test.js

Known issues:
IPAC tables:  Because we don't apply `guess` logic anymore, values where there were trailing zeroes will no longer show up as such.  1.2000 => 1.2


tgoldina:
Apart from truncated trailing zeros, I noticed that scientific notation is not used consistently. Don't we want to display all values in a column in scientific notation or without it?

One of the best test cases for it is `firefly/demo/ffapi-sed.html`

Notice, that in SED: ARP-220 table 
- "Frequency" (double) does not show in scientific notation as it used to (now I see 1450000000000000000 instead of 1.45e+18).
- "Photometry Measurement" (double) has mixed format: sometimes it's displayed in scientific notation and sometimes not. 

Regarding guessing, maybe we can preserve this for the tables of known origin, when we know that the number precision and notation are consistent through all the values in the column? (Like in tables coming from Gator.) When getting a table in these cases, maybe we can set a flag that format guessing is OK?

loitly:
> http://numeraljs.com/

It looked at it.  It's supporting a format similar to Java's DecimalFormat.  What we need is a `printf` style of format.  Let me know if that's not the case.

loitly:
@tgoldina , thanks for the feedbacks.
The IPAC table is likely a bug, and it will be fixed.  How it will look is the next issue you brought up; default format.

Originally, I plan to address the default format to another ticket since it's an issue CCB have interest in.  But, if you want to address in this PR as an initial implementation then have CCB comment on, that's good for me as well.

Some information for consideration:
There's discrepancies between Java and JavaScript default numeric format.
For integer and long, Java will print is as is. But because there's no integer/long data type in JavaScript, all values are treated as doubles.

For float/double:
Java uses decimal when value is 10<sup>-3</sup> <= v < 10<sup>7</sub>
JS uses decimal when value is 10<sup>-6</sup> <= v < 10<sup>21</sub>
Outside that range, it uses exponential format.  In both cases, it will carry full precision.

I have yet to figure out how to retain the full precision in Java format.  Any suggestion?

ejoliet:
Added @wmiipac as reviewer here for her to check Gator, Herschel, Atlas and Radar UI.

loitly:
UPDATE:  
- Fixed bad default format when writing IPAC table
- Made client and server default format the same
  using Java's convention: use decimal format where m = abs(value); 10<sup>-3</sup> <= m < 10<sup>7</sup>
  Otherwise, use exponential.  Retain original precision.

src/firefly/java/edu/caltech/ipac/table/TableMeta.java
```diff
@@ -24,28 +24,29 @@
...
-    public static final String MIN_VALUE_TAG = "col.@.MinValue";
-    public static final String MAX_VALUE_TAG = "col.@.MaxValue";
-    public static final String VALUE_TAG = "col.@.Value";
-    public static final String LINKS_TAG = "col.@.Links";
+    public static final String UTYPE_TAG = "col.@.utype";
+    public static final String REF_TAG = "col.@.ref";
+    public static final String MIN_VALUE_TAG = "col.@.minValue";
+    public static final String MAX_VALUE_TAG = "col.@.maxValue";
+    public static final String VALUE_TAG = "col.@.value";
+    public static final String LINKS_TAG = "col.@.links";
```
tgoldina: Just curious, why is it changed to lowercase? (I see that later in IpacTableUtil.consumeMeta you have to worry about backward compatibility.)

src/firefly/js/externalSource/__tests__/sprintf-test.js
```diff
@@ -0,0 +1,98 @@
...
+        expect(sprintf('%d', 2))    .toBe('2');
+        expect(sprintf('%i', 2))    .toBe('2');
+        expect(sprintf('%d', '2'))  .toBe('2');
+        expect(sprintf('%i', '2'))  .toBe('2');
+
+        expect(sprintf('%j', {foo: 'bar'}))	.toBe('{"foo":"bar"}');
+        expect(sprintf('%j', ['foo', 'bar']))	.toBe('["foo","bar"]');
+
+        expect(sprintf('%e', 2))    .toBe('2.000000e+0');
+        expect(sprintf('%.6e', 2))    .toBe('2.000000e+0');
```
tgoldina: Does it support `%E` and `%G`?

src/firefly/js/externalSource/__tests__/sprintf-test.js
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    test('J format:', () => {
+        // Java toString() format:   https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#toString(float)
+        expect(sprintf('%J', 0))            .toBe('0.0');
+        expect(sprintf('%J', 0.123456))     .toBe('0.123456');
+        expect(sprintf('%J', 0.000123456))  .toBe('1.23456e-4');
+        expect(sprintf('%J', 1234567))      .toBe('1234567');
+        expect(sprintf('%J', 12345678))     .toBe('1.2345678e+7');
+    });
```
tgoldina: So this one is non-standard format for our convenience? Our default format?

src/firefly/java/edu/caltech/ipac/table/TableMeta.java
loitly: Original server-side code was based on Spot's code which uses the uppercase version.
When we migrated from GWT to React, I switched over to the conventional camelCase style so that it'll look better in JSDocs.  This was probably a mistake because soon after we had to support `META_INFO` sent from client to server.  It was confusing having them different.

The changes here basically favor the new camelCase style, while deprecating the old.  I've updated the docs to reflect this change.

src/firefly/js/externalSource/__tests__/sprintf-test.js
loitly: As far as sprintf is concern, no, we only support `%e` and `%g`.  This is probably because JavaScript's toExponential() prints it in `%e` format.  On the server-side, we support both. 
 On the client, I convert it to lowercase before formatting.

But now, since I've made quite a bit of changes to sprintf, It would be trivial to support uppercase E and G.  Do you think I should add it?

src/firefly/js/externalSource/__tests__/sprintf-test.js
loitly: This is a new format I added.  It mimic the behavior of Java's toString() on numeric values.  And yes, this is the default format on the client side because we want to match toString() on the server.

src/firefly/js/externalSource/__tests__/sprintf-test.js
tgoldina: It might make comparing with the original table easier (when you implement the original table format of IPAC tables from Gator).

src/firefly/js/externalSource/__tests__/sprintf-test.js
ejoliet: this is really good idea.

ejoliet:
Seems that between your changes (ipac table fix) and @tgoldina changes in a recent commit  
14fdd43 https://github.com/Caltech-IPAC/firefly/pull/884/commits/14fdd43093962cdb36dc250bc16d021ada461ce4 the SED table looks better and the tooltip in the plot has now the right notation/format! I guess we will have to consider another cut of dev and build IRSAViewer release candidate from it.

ejoliet:
Also, i just noticed that Gator like table (WISE search 100" around 'm31') shows 'wsigpro4' 'nan' where it should read 'null'... could you confirm? I think IRSA wants to see 'null' for 'null' values, isn't it?

gpdf:
> I just noticed that Gator like table (WISE search 100" around 'm31') shows 'w4sigmpro' 'nan' where it should read 'null'... could you confirm? I think IRSA wants to see 'null' for 'null' values, isn't it?

We do need to distinguish between NaN and _null_ - they both arise in LSST tables, for sure, and are distinguished from each other in VOTable's `TABLEDATA` AND `BINARY2` formats.

I suspect the `w4sigmpro` values @ejoliet is seeing are really _null_, because that's an important flag value in the AllWISE data model (usually indicating that the corresponding `w4mpro` measurement is an upper limit).

gpdf:
By the way, is there code in this PR for displaying IEEE floating-point infinities as `+Inf` / `-Inf` ?

loitly:
> Also, i just noticed that Gator like table (WISE search 100" around 'm31') shows 'w4sigmpro' 'nan' where it should read 'null'... could you confirm? I think IRSA wants to see 'null' for 'null' values, isn't it?

@ejoliet I cannot duplicate.  Both my PR built and irsadev shows it as `null`. 

@gpdf We do not distinguish between NaN, +Inf, -Inf, and null.  They all will be represented as `null`.  This is because JSON does not support NaN and infinity. (https://www.ietf.org/rfc/rfc4627.txt)
```
   Numeric values that cannot be represented as sequences of digits
   (such as Infinity and NaN) are not permitted.
```

loitly:
@ejoliet @tgoldina @gpdf  I am adding the final changes requested to this PR.  I noticed our FITS reader is setting `NaN` as the nullString for float and double.  This is inconsistent with IPAC table and VoTable.  Would you like me to remove it?

gpdf:
On your last question: if you are saying that some current application is displaying "NaN" for values which are actually (null) in the data back end, to my mind that's really misleading.

gpdf:
On the larger question: I am _really_ uncomfortable with accepting a system that can't distinguish (null), NaN, and infinities in the era of data science that we are in.  I won't go so far as to resist this PR, as I can see that this kind of work will have to be spread out over multiple edits.  But this is important.

Perhaps if JSON has these limitations we can transmit some kind of flag array to handle out-of-band values?

I'd like to understand more systematically:
* Do we have a way to distinguish (null) for integer columns?
* What will we do for integer values that are not representable in double precision?  Many database IDs, which we will have to get exactly right (e.g., for DataLink purposes) are now 64-bit.
* How do we distinguish (null) and "" for string columns?
* How do we transport and display floating-point special values, and distinguish them from (null) values?

loitly:
> I tested uploading a VO table with width and precision metadata, they were displayed accordingly.
> But I found that Catalogs=>load catalog did not work. I used Upload to do the test

Do you have the sample file?  I tried with a few of mine, and they were working fine.

ejoliet:
@loitly , yes, you are correct, i can't see the 'nan' anymore, might be a glitch or refresh from old cache i have been testing... nvm. All good!

xiuqin:
Attached the file to FIREFLY-149.

tgoldina:
Interesting. I get identical results for Xiuqin's table (wiseDemoTable.xml) weather I load it via Catalogs->Load Catalog or Upload menu.

xiuqin:
It is working for me now!  I am glad it is not an introduced bug. 
But it is really weird because I tried at least 5 times before I reported it this morning.

ejoliet:
>  I am adding the final changes requested to this PR. I noticed our FITS reader is setting NaN as the nullString for float and double. This is inconsistent with IPAC table and VoTable. Would you like me to remove it?

In case o NaN present in FITS, what is the displayed value in the 'flux' readout? because if the FITS has NaN , i think scientist want to see 'NaN' in the readout. See https://jira.ipac.caltech.edu/browse/FIREFLY-238 comment from @vandesai1 .

wmiipac:
I did local build of firefly on this PR branch and reconfigured IRSA proxy to point my local firefly
I tested on the dev port 9073 with IRSA apps Gator, Data Discovery and IRTF. The tables are loading and functioning as expected, did not notice any obvious abnormality in above IRSA apps.

PR #887, opened 2019-10-14
IRSA-3158

lznakano:
https://jira.ipac.caltech.edu/browse/IRSA-3158
https://jira.ipac.caltech.edu/browse/IRSA-3166

Additional changes here: https://github.com/IPAC-SW/irsa-ife/pulls

This pull request is about the changes in DownloadDilaog.jsx that is used to handle finderchart download.  One property was added to DownloadButton and two properties were added to DownloadOptionPanel to handle Finderchart Download.

Please review the properties and let me know if you have any better suggestions.  I spent more time on the properties because I tried to avoid introducing new properties.  But after trying alternate ways, I think that  adding the properties has the least impact and changes to the existing codes in front and back ends. 

But I could be wrong.  Please check and let me know your suggestions.  Thanks!

Test  URL: https://irsawebdev9.ipac.caltech.edu/irsa-3158-3166-Lijun/applications/finderchart/

### DIFF
--- a/src/firefly/js/ui/DownloadDialog.jsx
+++ b/src/firefly/js/ui/DownloadDialog.jsx
@@ -4,7 +4,7 @@
 
 import React, {useCallback, useState} from 'react';
 import PropTypes from 'prop-types';
-import {set, cloneDeep, get} from 'lodash';
+import {set, cloneDeep, get, omit} from 'lodash';
 import DialogRootContainer from './DialogRootContainer.jsx';
 import {PopupPanel} from './PopupPanel.jsx';
 import {dispatchShowDialog, dispatchHideDialog} from '../core/ComponentCntlr.js';
@@ -76,7 +76,7 @@ export function DownloadButton(props) {
     const selectInfoCls = SelectInfo.newInstance(selectInfo);
 
     const onClick = useCallback(() => {
-        if (selectInfoCls.getSelectedCount()) {
+        if (selectInfoCls.getSelectedCount() || props.checkSelectedRow===false) {
             if(hasOnlyProprietaryData(getTblById(tbl_id))){
                 showInfoPopup('You do not have permission to download the selected data set(s).', 'Private Data Selected');
             }else if(!hasOnlyProprietaryData()){
@@ -103,9 +103,13 @@ export function DownloadButton(props) {
 DownloadButton.propTypes = {
     tbl_id      : PropTypes.string,
     tbl_grp     : PropTypes.string,
+    checkSelectedRow:PropTypes.bool
 };
 
 
+DownloadButton.defaultProps = {
+    checkSelectedRow:true
+};
 const noticeCss = {
     backgroundColor: 'beige',
     color: 'brown',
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -119,16 +123,24 @@ let dlTitleIdx = 0;
...
+
+        let {request,  selectInfo} = getTblById(tbl_id);
+        const {FileGroupProcessor, SearchProcessor} = dlParams;
+        if (SearchProcessor){
+            request.id = SearchProcessor;
+        }
+        if (imageSource){
+            request.selectImage = imageSource;
+        }
+
```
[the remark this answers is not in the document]
ejoliet: yes, i think this has to be passed from the parent, no need to have distinction here in Firefly code. If i understood Loi correctly, you need to add a function as a prop to the dialog and pass it from outiside, correct?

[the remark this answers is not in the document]
lznakano: @Loi 
Can you please check the function prop added in DownloadDialog?  The function is passed from FinderchartDownloadDialog.  Every component who is using DownloadDialog can pass its own function if it is needed.  Thank you for your good advice.  Using the function property, it is easy to fix the bug you noted in multi-search.   

@ejoliet The three color issue is from the old dev.  Since I did not update my pull request, there was an exception in PlotImageTask when the 3-color is requested.   I will test it after I rebase the dev.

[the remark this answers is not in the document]
lznakano: @loitly @ejoliet 
Please only check the function properties if you wish.   I fixed the issue for current target.  
I rebased with dev and am testing/debugging color image and multi-position search for all target.  The color image issue is not related to what I am doing.   It is uploading image failure.  Once I have these two issues solved, I will build the test URL.  FYI

lznakano:
I rebased firefly with dev, the color image is loaded after the rebase.  The bugs in download multi-position are fixed.   
Please test:
 1. Check the function property introduced to DownloadOptionPanel and let me know if it is OK
 2. I did not introduce the function property for DownloadButton,  I also don't like the property introduced.  I checked the props and could not find anything else that can be used for Finderchart single position search which has no table involved.   Any idea or suggestions are welcome.
3. All downloads seem working fine except the issue with switching from workspace back to local.  As I noted,  it is a bug in the dev version of workspace.  
Thanks!

ejoliet:
@loitly could you look at it again? i agree with the refactor proposed from Lijun and the cleanup/.checks she made gives confident that the dialog is useful and correctly used by external Firefly-based UI.

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -103,9 +103,13 @@ export function DownloadButton(props) {
 DownloadButton.propTypes = {
     tbl_id      : PropTypes.string,
     tbl_grp     : PropTypes.string,
+    need_selectInfo:PropTypes.bool
```
loitly: Minor:  Can you change the property to `checkSelectedRow`?

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -217,7 +240,12 @@ DownloadOptionPanel.propTypes = {
 
 DownloadOptionPanel.defaultProps= {
     groupKey: 'DownloadDialog',
-};
+    updateRequest:null,
+    updateDownloadRequest:null,
+    dlParams:   PropTypes.shape({
+        SearchProcessor: undefined
+       })
```
loitly: `updateRequest` should be `updateSearchRequest`
`dlParams` should not contains `shape`.  This is where you define default values.
Actually, they whole block of update is not needed.  It's setting default values that are either wrong or is already the default behavior.  I suggest removing the whole block.

lznakano: Thanks for spotting the left over stuff.  I changed the way to implement it and left those unused properties there.

src/firefly/js/ui/DownloadDialog.jsx
lznakano: Yes, I will change it.

PR #888, opened 2019-10-15
IRSA-3179: implement proprietary data access for SOFIA

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-3179
Test: https://irsawebdev9.ipac.caltech.edu/IRSA-3179_sofia_proprietary_not_working/applications/sofia/

Additional changes:  https://github.com/IPAC-SW/irsa-ife/pull/93

SOFIA: proprietary queries/downloads not working.

When accessing proprietary data aware services, user's credential has to be explicitly requested when the connection is created.  This has not been done to any of IRSA's React based applications.  This PR added a few things to enable this feature and also ensure that SOFIA's server-side code implements it.

Also, refactor HttpServices to simplify its API.

To test:
Query SOFIA proprietary data.  (I use `m17` with Level 2,3,4)
This should return some FORECAST with pink(no access) rows.

Sign in as someone with access to SOFIA data.
Repeat the same query.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/spring/JdbcFactory.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/spring/JdbcFactory.java
@@ -121,7 +121,7 @@ public class JdbcFactory {
     }
 
     private static DataSource getDirectDataSource(DbInstance dbInstance) {
-        logger.briefDebug("Getting a new database connection for " + dbInstance.dbUrl + " using DriverManager");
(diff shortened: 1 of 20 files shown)


ejoliet:
@loitly Thanks for chasing and updating the http framework in Firefly. 
I've added @wmiipac as reviewer - Wei, could you build and check that latest (react based) ZTF and TimeSeries with PTF private data for example are behaving also ok with this change? Thanks!

PR #898, opened 2019-10-22
Firefly-144,400,388: bug fixes

robyww:
Firefly-144,400,388: bug fixes:
  - Firefly-144: fixed: compass is not redrawn is certain cases.
  - Firefly-400: Compress images show as table in the upload tool
  - Firefly-388: fixed image should not select when in decimation mode
  - Fixed: image jumping on selection
  - also, NorthUpCompass.js code cleanup

_to test:_

https://irsawebdev9.ipac.caltech.edu/firefly-144-400-388-bug-fixes/firefly/

  - load an image and turn on compass: flip, rotate, zoom compass should always redraw
  - upload and image with a compress FITS images. It will show and load correctly.
  - load catalog m16, 1800 arcsec. This will put the image catalog in decimation mode. It should not show selection.
  - In the same case, filter down (with chart) then start clicking on table to show selection on image. It should not jump around.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/FitsHDUUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/util/FitsHDUUtil.java
@@ -8,16 +8,17 @@ import edu.caltech.ipac.table.DataGroup;
 import edu.caltech.ipac.table.DataObject;
 import edu.caltech.ipac.table.DataType;
 import edu.caltech.ipac.table.TableUtil;
-import nom.tam.fits.ImageHDU;
-import nom.tam.image.compression.hdu.CompressedImageHDU;
-import org.json.simple.JSONObject;
+import nom.tam.fits.BasicHDU;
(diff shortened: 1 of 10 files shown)


robyww:
Oh, this is not a change but a fix. The selection is incorrect in the decimation selection case.  It has  never been correct. When you to select for the plot in the large data decimation case we give a pop up saying that is is not allowed. In the case, selection is happing from the charts. Because of decimation we can’t compute it in the image. So the best solution is just not to show it instead of showing the wrong thing. The ticket shows this as well. 

This is an edge case that we were not watching for.

ejoliet:
Oh, Ok, understood.
And the default icon on the image toolbar showing filter on crossed?

robyww:
> And the default icon on the image toolbar showing filter on crossed?

I don't understand this comment but @tgoldina (above) did find some problems with the wrong unselect button showing when there is nothing to unselect. I am fixing that as well.

PR #902, opened 2019-10-28
Firefly-396: fixed: filtering  coverage not working with decimated data

robyww:
Fixed: filtering coverage now works with decimated ddata.

https://jira.ipac.caltech.edu/browse/FIREFLY-396

_to test:_ 

https://irsawebdev9.ipac.caltech.edu/firefly-396-filter-coverage/firefly/test/test-misc.html

The following test simulates the gator problem:

1. Go the the second test on the page
1. Using the charts select a large area
1. the image coverage should filter
1. unfilter from the chart and the image should unfilter


ejoliet:
Did you notice that the branch you are pulling is firefly-396-...?

PR #903, opened 2019-10-29
Correct Markdown syntax

gpdf:


gpdf:
What's the Firefly standard for merges?  Is one of the three Github "green button" options acceptable, or is it like LSST where it has to be done by hand on the command line because none of the three is acceptable?

robyww:
Green button is executable. We usually change the beginning of the comment to begin with `Firefly-xxx:` but in this case just make it `Documentation: `

PR #904, opened 2019-10-30
FIREFLY-56 Change ruler tool behavior for image with or without WCS and HiPS

cwang2016:
The developement fixes the ruler tool behavior based on the following JIRA tickets, 
https://jira.ipac.caltech.edu/browse/FIREFLY-56
https://jira.ipac.caltech.edu/browse/FIREFLY-1

Here is the  summry of the fixing, 
- increase the line thickness and font size and style of the ruler tool
- display PA along angular distance without requiring the user to click in the layer dialog
   --  PA angle and angular disatance are displayed in parallell, above and along the anglular distance vector 
   -- PA angle is measured East from North, in unit 'degree'
- Add an arrow head at the end of angllar distance vector 
- Unit update including that on layer dialog, 
   -- add unit 'pixel' to image with WCS case 
   -- image without WCS has only unit 'pixel' option
   -- HiPS has no 'pixel' unit
-  lines and the relevant text from offset calculation
   -- add handles (squares) at both ends of the lines. 
   -- text is rotated per the direction of each line

test: 
https://irsawebdev9.ipac.caltech.edu/FIREFLY-56_RulerToolChange/firefly/
- start image search on regular image, image without WCS, or HiPS
   (samples for image without WCS are located at firefly_test_data/no-projection-images)
- turn on the distance tool icon and locate the the tool on the image. 
- start the layer dialog, click 'offset calculation' and select the unit to check the display of the tool. 
- for non HiPS case, rotate the image and view the distance tool is rotated accordingly.

### DIFF
--- a/src/firefly/js/drawingLayers/DistanceToolUI.jsx
+++ b/src/firefly/js/drawingLayers/DistanceToolUI.jsx
@@ -13,6 +13,10 @@ import CsysConverter from '../visualize/CsysConverter';
 
 
 export const getUIComponent = (drawLayer,pv) => <DistanceToolUI drawLayer={drawLayer} pv={pv}/>;
+const worldUnit = [ {label: 'degrees', value: 'deg'},
+    {label: 'arcminutes', value: 'arcmin'},
+    {label: 'arcseconds', value: 'arcsec'}];
+const pixelUnit = [{label: 'pixels', value: 'pixel'}];
 
 function DistanceToolUI({drawLayer,pv}) {
 
@@ -26,12 +30,7 @@ function DistanceToolUI({drawLayer,pv}) {
     const plot = pv.plots[pv.primeIdx];
     const cc = CsysConverter.make(plot);
     const isWorld = hasWCSProjection(cc);
-    const worldUnit = [ {label: 'degrees', value: 'deg'},
-                        {label: 'arcminutes', value: 'arcmin'},
-                        {label: 'arcseconds', value: 'arcsec'}];
-    const pixelUnit = [{label: 'pixels', value: 'pixel'}];
     const unitStyle = getUnitStyle(cc, isWorld);
-
     const pref = getUnitPreference(unitStyle);
     const options = (unitStyle === UNIT_ALL) ? worldUnit.concat(pixelUnit)
                                              : ((unitStyle === UNIT_NO_PIXEL)? worldUnit : pixelUnit);
--- a/src/firefly/js/drawingLayers/DistanceToolUI.jsx
+++ b/src/firefly/js/drawingLayers/DistanceToolUI.jsx
@@ -7,7 +7,9 @@ import PropTypes from 'prop-types';
 import {RadioGroupInputFieldView} from '../ui/RadioGroupInputFieldView.jsx';
 import {dispatchModifyCustomField, dispatchForceDrawLayerUpdate} from '../visualize/DrawLayerCntlr.js';
 import * as AppDataCntlr from '../core/AppDataCntlr.js';
-import {DIST_READOUT,ARC_SEC,ARC_MIN,DEG} from './DistanceTool.js';
+import {DIST_READOUT, getUnitPreference, getUnitStyle, UNIT_NO_PIXEL, UNIT_ALL} from './DistanceTool.js';
+import {hasWCSProjection} from '../visualize/PlotViewUtil';
+import CsysConverter from '../visualize/CsysConverter';
 
 
 export const getUIComponent = (drawLayer,pv) => <DistanceToolUI drawLayer={drawLayer} pv={pv}/>;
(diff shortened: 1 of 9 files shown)


src/firefly/js/visualize/draw/ShapeDataObj.js
```diff
@@ -605,16 +605,40 @@ function drawLine(drawObj, ctx,  plot, drawParams, onlyAddToPath) {
...
+        rOptions.rotAngle = rAngle;
+
+        if (style !== Style.ENDHANDLED) {
+            rOptions.translation = {x: devPt0.x, y: devPt0.y};
+            DrawUtil.fillRec(ctx, color, -3, -3, 7, 7, rOptions);
+        }
+        if (style !== Style.STARTHANDLED) {
+            rOptions.translation = {x: devPt1.x, y: devPt1.y};
+            DrawUtil.fillRec(ctx, color, -3, -3, 7, 7, rOptions);
+        }
```
robyww: I am having trouble understanding this logic. It looks like if the is `HANDLED` it does stuff twice.  I think is would be better to say:
```js
if (style === Style.HANDLED) {
   // do this
}
else if (style === Style.ENDHANDLED) {
   // do this
}
else  if (style === Style.STARTHANDLED) {
   // do this
}
```

src/firefly/js/visualize/draw/ShapeDataObj.js
```diff
@@ -706,39 +731,46 @@ export function drawText(drawObj, ctx, plot, inPt, drawParams) {
         }
 
         angle = pvAngle + lineAngle;
+    } else {
+        const pv = getPlotViewById(visRoot(), plot.plotId);
+        angle = pv.flipY? 180 - pv.rotation:pv.rotation;
```
robyww: I don't think the flip logic is right. I flipped an image and used the distance tool  and the text was not correct.

src/firefly/js/drawingLayers/DistanceToolUI.jsx
```diff
@@ -20,17 +22,19 @@ function DistanceToolUI({drawLayer,pv}) {
...
-
-    var pref= AppDataCntlr.getPreference(DIST_READOUT) || DEG;
+    const checked= drawLayer.offsetCal;
+    const plot = pv.plots[pv.primeIdx];
+    const cc = CsysConverter.make(plot);
+    const isWorld = hasWCSProjection(cc);
+    const worldUnit = [ {label: 'degrees', value: 'deg'},
+                        {label: 'arcminutes', value: 'arcmin'},
+                        {label: 'arcseconds', value: 'arcsec'}];
+    const pixelUnit = [{label: 'pixels', value: 'pixel'}];
```
robyww: I think you can make  this stuff constants outside of the function.

xiuqin:
Rotation has the same issue as flip. Otherwise it looks good.

ejoliet:
After discussion with @cwang2016 and @vandesai1 / @gpdf / Joe , here are the issues to be fixed:
- The user can no longer control the position of the origin of the arrow, in IRSA OPS site he/she can.
- When you load more than 1 image, draw an arrow, then change units to Pixels, the overlays on all of the images report the same number of pixels, even when the images have different pixel scales.  Decision made as least effort solution: When multiple images (and/or hips), pixels option should be for now because it's tricky to fix in reasonable amount of effort/time

Thanks.

robyww:
From @ejoliet, @vandesai1 @gpdf 

I spoke to Vandana and Gregory, here are the reported issues found from testing the PR (FIREFLY-56) with the new ruler, they would need to be fixed before merging if possible:

1. When you load >1 image, draw an arrow, then change units to Pixels, the overlays on all of the images report the same number of pixels, even when the images have different pixel scales. This is scientifically misleading, and must be changed.

2. When the units are changed to pixels, the arrow and angle disappears on HiPS images. They should stay. This is a bug and should be fixed.

3. When you zoom out on an image, the overlay sometimes jumps to another position. If I change units, it appears to refresh and jump back. This is a bug and should be fixed.

4.  The user can no longer control the position of the origin of the arrow. On ops, you can. This is a regression, and it would be good to fix.

Let me know if the fixes are doable in a reasonable time.

cwang2016:
Regarding the above comments from @robyww  & @ejoliet 
- issues related to when the image >= 2:  issue 1 & 2 from @robyww 
   the distance tool layer was initially implemented as one single data applied to all attached plots.  
   -- If there is the time, It is doable to make modification on the layer plus the dialog by setting the layer to 'hasPerPlotData'. If that, more detail needs to be designed, such as how to synchronize the distance tool data and units among regular image, image without wcs and hips since they all have various unit selections.
-- the solution listed by @ejoliet (removing unit pixel from the dialog) could at least resolve the issue when image >= 2 cases (mix of regular image & Hips). However, there is still the issue if  a image with no wcs is mixed with either regular image or hips since image with no wcs has only 'pixel' as the option for the unit. 
- issue 3 from @robyww: I need to find a test case to reproduce it. Could a case be provided in order to reproduce the issue? 
- issue 4 from @robyww: it will be fixed as the comment from @ejoliet

vandesai1:
To reproduce the third item:

1. search on M81, AllWISE ch1

2. Draw an arrow from the big galaxy nucleus to the centermost south-east object.

3. Zoom in once. The arrow jumps away from these objects.

cwang2016:
Per review comments, the following issues are updated, 
- The user can no longer control the position of the origin of the arrow. On ops, you can. This is a regression, and it would be good to fix.
-  When you zoom out on an image, the overlay sometimes jumps to another position. If I change units, it appears to refresh and jump back. This is a bug and should be fixed.
- Bug that needs to be fixed: When I flip and image the text is not placed correctly.
- A little code cleanup and docs.

Other issues regarding multiple images and distance tool display under various unit selection, here is  how the distance tool works when mulitiple images are shown and one of them is highlighted as the active image. 
- if 'pixel' unit is applied to the active image, 
   -- distance tool layer are shown on image with wcs or no wcs (except HiPS). 
   -- the location of the ruler on all images has the same image pixel coordinate values  as those of active image. 
   -- distance is measured in image pixel unit. 
   -- no position angle is shown. 
- if world unit ('degree', 'arcmin', 'arcsec') is applied to the active image, 
   -- distance tool layer are shown on image with wcs or HiPS (except image with no wcs) 
   -- the location of the ruler on all images has the same world coordinate values as those of active image. 
   -- distance is measured in world domain unit. 
   -- Position angle is shown in 'degree' unit.

cwang2016:
the test build is 
https://irsawebdev9.ipac.caltech.edu/FIREFLY-56_RulerToolChange/firefly/

cwang2016:
made the following fixes per review comments, 
- remove unit 'pixel' option for fits with wcs. 
- enable to drag the base of the arrow as well as the head. 
- number with three significant figures are displayed in the overlay.

PR #916, opened 2019-11-15
FIREFLY-112: Client table filter does not fail when strings are used …

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-112

When filtering numeric columns, client table shows `No data..` when non-numeric values are used in the criteria.  In server-backed tables, an error is displayed instead.  Make the client tables behavior consistent with server-backed tables.

Also, 
- added support for 'and' and 'or' to client table.
This was missing from previous ticket in which `and` and `or` was added to the filtering syntax for server-back tables.

Tests: https://irsawebdev9.ipac.caltech.edu/FIREFLY-112_client_table_filter_noresults/firefly/

Try entering `abc` into a numeric column.  Do this on a client table(View HiPS Images) as well as a server-backed table(catalogs).  The behavior is the same now.  Do the same thing on irsadev to see what it did before.
( although not consistent, I thought it was more user friendly on irsadev )

Client table now support `and` and `or`.  Try:
`> 0 and < 1000`  or `< 0 or in (8,9,10)`

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/TableMeta.java
+++ b/src/firefly/java/edu/caltech/ipac/table/TableMeta.java
@@ -50,7 +50,7 @@ public class TableMeta implements Serializable {
 
 
     public static final String RESULTSET_ID = "resultSetID";            // this meta if exists contains the ID of the resultset returned.
-    public static final String RESULTSET_REQ = "resultSetRequest";      // this meta if exists contains the Request used to create this of the resultset.
+    public static final String RESULTSET_REQ = "resultSetRequest";      // this meta if exists contains the Request used to create this resultset.
 
     public static final String IS_FULLY_LOADED = "isFullyLoaded";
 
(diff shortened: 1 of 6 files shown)


ejoliet:
BTW, when there is no data match, there is no 'back' button either, i don't remember what we said about it but the only way to go back in a client table that doens't show the funnel to reset, you will need to to open the settings and then click on reset, which might be a little overhead. We should think about putting back the 'back button' i think.

ejoliet:
Another comment maybe that should be addressed is the error message in a client table, when i do 'abc' filtering on a numerical column, i get:
"data exception: invalid character value for cast"
which is not user friendly, we should put another message such as :
"Using non numerical characters on a numerical column is not possible" or something, CCB-D can give a better sentence.

loitly:
> We should think about putting back the 'back button' i think.

The button was never there before under normal usage.  
Applying a filter that resulted in zero rows returned is normal usage.  
Removing or changing the filters will update the results.  
This is not a hard task.  Simply remove the filter you've just entered or change it to something else, like, if `band > 5` returns no data, change it to `band > 2`.  Or, remove the filter.


The `back` button only appears when it fails to load a table due to exceptions caused by a variety of reasons including SQL syntax errors, like `"ra" = 'abc'` where `ra` is a numeric column.
Because client tables are not backed by a database, we did not try to captured what is deemed as syntax error on server-based tables.  Hence, the difference in behavior.

The reason for this PR is to make the tables consistent.  That's why the cryptic error message.  That is the exact message you'll get if you do the same on a server-backed table.

Maybe we need to re-evaluate what we really want to do here, both client and server-backed.

PR #922, opened 2019-12-13
FIREFLY-150: Handle data array type

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-150

General changes:
- `data type` is shown by default.  (most tables)
- `Text View` UI is updated to look more like an SQL client and not IPAC table.
- Data overflow in table cell will be appended with `...`(more indicator).
- By default, column width is set to the data’s length or 100 whichever is less.
- significant update to VoTableWriter - @cwang2016 please review and test to make sure it support all previous use-cases.

Data array:
- Display
    - Full value of the array is displayed in the table cell as JSON string
    - Column width is default to the length of the value or 30 whichever is less.
    - Overflow values are hidden and replaced with `...`
    - In `Text View`, the same JSON string value is shown in full 

- Save
    - VOTable: data array is saved as defined by their supported format: TABLEDATA, BINARY2, FITS
    - IPAC Table:  (array type is not supported)
        - Add data array support to Firefly
        - Data is saved like VOTable's TABLEDATA with additional meta in the header to describe type and shape of the array.
        - This feature is only available in Firefly.  Otherwise, it’ll be recognize as ’char’ or string.
    - for CSV and TSV: (data array is not supported)
        - It will be saved as JSON string and read back as a string

Testing:  https://irsawebdev9.ipac.caltech.edu/FIREFLY-150_table_array_data/firefly/
- Upload the file `tbl.fits` attached to the ticket.
- The first column(`flags`) of the first table has a type of `boolean[57]`.
You should see a partial JSON string representing an array of 57 booleans.  Resize the column to see more/less of the data.
- Switch to `Text View` to see the new UI.  The boolean array is shown in full.
- Click save -> select IPAC Table -> a.tbl
- Open the saved file (`a.tbl`).  It should look exactly like the one you see in `tbl.fits`
- Now, do the same for the other VOTable formats.  They should work the same way.
- Now, if you save it as CSV or TSV, it'll be saved as a string and read back also as string

Multi-dimensional array:
- Save first table in `tbl.fits` as `VOTable - TABLEDATA` (`a.vot`)
- Edit `a.vot` and change `arraysize="*"` to `arraysize="3x*"`
- Load this file.  You should see the array is nicely folded into 3-boolean parts
- You can do the same with IPAC Table
- You can try 3 or 4 dimensions.
- You can also test other types, like `char`
   - try changing `flgs` into a multi-dimensional char array.  i.e. `arraysize="2x2x*"` and type `datatype="char"`
   - you should see an array of 2-char pairs.
   - StarLink right-trim the strings.  I think that makes sense.
- Now, see if you can move from one format to another, i.e open VOTable and save as IPAC

@gpdf if you have time, please take a look.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
@@ -189,7 +189,7 @@ abstract public class BaseDbAdapter implements DbAdapter {
     }
 
     public String toDbDataType(DataType dataType) {
-        if (!StringUtils.isEmpty(dataType.getArraySize())) return "other";
+        if (dataType.isArrayType()) return "other";
 
         Class type = dataType.getDataType();
         if (String.class.isAssignableFrom(type)) {
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
@@ -58,7 +58,7 @@ abstract public class BaseDbAdapter implements DbAdapter {
             ", minValue varchar(64000)" +
             ", links    other" +
             ", dataOptions varchar(64000)" +
-            ", arraySize varchar(64000)" +
+            ", arraySize varchar(255)" +
             ")";
 
     private static final String META_INSERT_SQL = "insert into %s_meta values (?,?,?)";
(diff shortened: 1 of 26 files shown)


cwang2016:
please check the following test sample, 
FileUpload-samples/CSV/dum1.csv

vandesai1:
I loaded tbl.fits. The tri-view shoes a context image, a table, and a chart. 

What works:
- Table display looks great, in interactive table and text model.
- Saving the table in every mode works great. Was able to successfully re-upload all of the formats.

What doesn't work:
- Filtering the array column doesn't work, but that is ok for now.
- something seems to be wrong with the context image overlays. Pay by table row is on, but if I highlight a different row and then highlight the original row again, I don't see the same overlays. Not sure this is related to this ticket, though.

loitly:
- disable sort and filter when it's a data array
- handle csv numeric parsing errors by making the value null.  (old behavior)
- add data array info into the File Upload dialog
  - show type the same way it's shown on table and text view
  - for FITS table, you'll have to look at the FITS header.. `TFORMn`
I've made another commit to address the issues reported.  I've also updated the k8s built if you want to test it again.
@ejoliet let me know when you want this PR to be merged.

The image overlay issue @vandesai1 reported is currently in `dev`.  I left it out of this PR.

ejoliet:
@loitly looks good to be merged.
@vandesai1 i will raise ticket for the overlay problem which is outside the scope of this PR.

PR #928, opened 2020-01-28
FIREFLY-471: Table Options UI Improvement

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-471

- add ability to set precision and null_string
- add HMS/DMS format support into precision
- add advanced filter (not available on client table)
- display table meta
- display additional column meta

To test:  https://irsawebdev9.ipac.caltech.edu/FIREFLY-471_table_ui_improvements/firefly/
On any table, click the options icon (gears) to bring up the options dialog.
Most of the changes are visible here.  However, make sure to test to ensure that the changes are consistent throughout, like text view and save.

* Not sure what to do about Chart's filter dialog.  Depending on the amount of changes, it will either be addd to this PR or a new ticket will be created.


tgoldina:
We might want to discuss this change as group, especially Advanced Filter UI.

I did not realize at first that "Advanced Filter" is completely separate from the column filters applied in Column Options tab. It is an ADDITIONAL, free-form filter. It would be more clear if the existing column filters were shown in a non-editable box somewhere close to AND/OR filter switch. Without understanding that the filter is additional, this switch does not even make sense.

I do like that it's now possible to access this additional column metadata. Since now there are many more options, it might be worth adding a short description and (may be) use human-friendly titles. Also consider making a column invisible if there no values in each row. 

At this point, I did not test the functional behavior well enough. Will do.

src/firefly/js/tables/ui/TablePanelOptions.jsx
```diff
@@ -2,103 +2,346 @@
...
+    const nColumns = cloneDeep(get(getTableUiById(tbl_ui_id), 'columns', []));
+    const selCol = nColumns.find((col) => col.name === selColName);
+    return {selColName, nColumns, selCol};
+}
+
+function makePrecisionRenderer(tbl_ui_id, ctm_tbl_id, onChange) {
+    const tooltips = 'A string Tn where T is either F, E, G, HMS, or DMS\n' +
+        ' When T is F or E, n is the number of significant figures after the decimal point\n' +
+        ' When T is G, n is the number of significant digits\n' +
+        ' When T is HMS or DMS, n is ignored';
```
robyww: maybe an example here. `F4`, `E5`

loitly:
- make the format and null_string columns non editable for now.
- Try to find another icon to describe the column name in the advanced filter to show visually that this is a link to copy and paste into the text area
- 'Meta' tab title should be 'Table Meta'
- fix 'Contraints' label mispelled
- add -like to first SQL in the help description
- move the 'reset' and 'close' buttons at the bottom left and same look&feel as other popup dialog

This latest commit address the above issues plus these:
- add indicator to Advanced Filter tab when there's one used.
- fix word wrap in the SQL examples
- hide format and null_string columns when there's no data.

@ejoliet In regard to `check why the panels can't show the scrollbar when scrolling`
It works as expected in my environment with all 3 browsers; chrome, safari, and firefox.
Please check `System Preference -> General -> Show scroll bars` section to ensure you got the right settings.  Or, try this on other people browsers to see if this is only an issue with your laptop.

PR #929, opened 2020-01-28
Firefly- 451 removed `numeral` package and revamp `package.json`

robyww:
Clean up of package.json

  - Allows any version of node v8 or later
  - Most packages updated
  - Packages now locked at their revisions instead of minor releases
  - Babel updated, to most recent, 7.8
  - ESlint updated to most recent, and now better supports ES2020
  - Jest update some, can go to most recent with node 8
  - JSDoc updated to most recent
  - Webpack updated a few revisions, (minor)
  - Other package updated when there is compatibility
  - Cleaned out several unused packages
  - Added `package-json-doc.md` to help better document package.json
  - transformation-matrix-js moved from npm package to externalSource
     - it is not longer supported as a npm package.
  - Fixes Firefly-476
  - changed a couple of files of use `?.` to make sure it works
     - optional chaining


src/firefly/js/data/form/PositionFieldDef.js
```diff
@@ -190,8 +190,8 @@ var makePositionFieldDef= function(properties) {
 
 
 
-        var lonStr = numeral(wp.getLon()).format('#.0[00000]');
-        var latStr = numeral(wp.getLat()).format('#.0[00000]');
+        var lonStr = sprintf('%6.5f',wp.getLon());
+        var latStr = sprintf('%6.5f',wp.getLat());
```
loitly: I noticed you're  using `width.precision`.  `width` is the minimum number of characters to output.  This is usually use for padding.  In your case, it makes no difference since the results will always be more than 6 characters, meaning `%.5f` would have suffice.

PR #932, opened 2020-02-11
firefly-487-Add validation to WAVE-TAB WCS implementation

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-487 

When the WAVE-TAB was implemented,  the validation process was ignore.  When the FITs has multiple HDUs, it can have more binary tables.  Thus, the validation is necessary to guarantee that the right table is used.  This pull request is to add WAVE-TAB WCS validation according to the paper " Representations of spectral coordinates in FITS by E. W. Greisen et al."   All changes are in sever side.  
What it does it to check all the binary tables and find the right one whose extver number, extlevel number and extname are the same as thePV3_i  (i=1, 2) and PS3_i  respectively.  Also the extname in the binary table should start with "WCS-". 

To test it, please use previously known working FITs files (in the ticket). 
URL: https://irsawebdev9.ipac.caltech.edu/firefly-487-waveTab-validation/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
@@ -18,10 +18,8 @@ import nom.tam.fits.BasicHDU;
 import nom.tam.fits.Header;
 
 import java.io.File;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * @author Trey Roby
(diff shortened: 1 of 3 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
```diff
@@ -51,37 +49,109 @@ public static int findWaveTabHDU(BasicHDU[] HDUs) {
...
         if (HDUs.length<2) return -1;
+        if (!isValidWCSForTab(HDUs)) return -1;
+
         BasicHDU hdu= HDUs[hduIdx];
         Header header= hdu.getHeader();
         String xtensionPrim= header.getStringValue("XTENSION");
         if (xtensionPrim!=null && xtensionPrim.equals("BINTABLE")) return -1;
         String ctype= header.getStringValue("CTYPE3");
         if (!"WAVE-TAB".equals(ctype)) return -1;
         String waveHDUName= header.getStringValue("PS3_0");
```
robyww: I think you should do this section before calling isValidWCSForTab

lznakano: Good suggestion!  Thanks, I will change it.

lznakano: @robyww  I changed the order of the validation as you suggested.  I did not pay attention about the efficiency issue about the existing codes. Thanks for pointing it out.

PR #933, opened 2020-02-11
Firefly-489: caching of HiPS properties and tiles should only last until a reload happens

robyww:
### Firefly-489: caching of HiPS properties and tiles should only last until a reload happens

   - After a reload all HiPS tile and property caching should be cleared
   - new function Firefly.js: getFireflySessionId()
   - Firefly-490: allow reload caching to happen in local mode
   - Firefly-361: docker clean will cleanup hips cache over 90 days

### Testing
https://irsawebdev9.ipac.caltech.edu/firefly-489-hips-cache/firefly/

You will need to bring up developer tools and watch the network tab. Watch for the message `(disk cache)` or `(memory cache)` when fits tiles are loading.

_First try_
1. Chose a target and HiPS selection, you will use this same one again and again.
1. make sure developer tools, network tab is up
1. load the HiPS and zoom down to around 5 degree FOV, and scroll around that area
1. you will see a lot of activity on the network tab, watch the size
1. as you scroll around you will see more `(disk cache)` or `(memory cache)` under size

_Second try_
1. delete that HiPS (don't reload) and do the same thing again, zoom down to around 5 degrees
1. this time you will see and lot more `(disk cache)` or `(memory cache)` under size

_Now Reload, Third try_
1. load the same HiPS and same target and zoom down to around 5 degress.
1. This time you should not see many `(disk cache)` or `(memory cache)`

### DIFF
--- a/buildScript/tasks.gincl
+++ b/buildScript/tasks.gincl
@@ -119,6 +119,7 @@ task createVersionTag  {
         props.setProperty('BuildMajor', major)
         props.setProperty('BuildMinor', minor)
         props.setProperty('BuildRev', rev)
+        props.setProperty('ClientBuildEnv', project.hasProperty("env") ? project.env : "local")
         props.setProperty('BuildType', type)
         props.setProperty('BuildNumber', buildNum)
         props.setProperty('BuildDate', build_date)
--- a/docker/base/cleanup.sh
+++ b/docker/base/cleanup.sh
@@ -28,7 +28,7 @@ function doCleanup() {
         fi
 
         ${find_cmd} "${workarea}/upload" -type f -mtime +7 -exec /bin/rm '{}' \+ -print >> "${log_file}" 2>&1
-        #${find_cmd} "${workarea}/HiPS" -type f -mtime +90 -exec /bin/rm '{}' \+ -print >> "${log_file}" 2>&1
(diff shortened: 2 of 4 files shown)


robyww:
> However, at the meeting, we said we would make this extended caching an option for developer

Why does it need to be an option since a developer can easily do a clear cache and reload? I considered it but It does not really seem necessary to me.

loitly:
> Why does it need to be an option since a developer can easily do a clear cache and reload? I considered it but It does not really seem necessary to me.

While it's not hard to `rm -r .../workarea` (which does not work atm)` then reload browser, it is not a `mode` you can stay in.  You simply cannot run the application the way it should be running in 'local' env.  But, if you say no one will ever want it, then I am fine with the decision.

robyww:
I will be easily enough to put in later if it is wanted. Since it would be only 2 lines in `task.gincl` and modifying the one line in `Firefly.js`.

_Something like:_
```js
export const getFireflySessionId= once(() =>
    (getProp('version.ClientBuildEnv')==='local' && !toBoolean(getProp('ClearCachingOnReload'),false))? 'LOCAL_SESSION' : `FF-Session-${Date.now()}`);
```
If, when you implement Firefly-488, you think it is really necessary go ahead and put it in. I would just want it to be the default.

In general, when developing, I want very aggressive caching at least 95% of the time.

PR #937, opened 2020-02-25
Firefly-484: HiPS/Image WCS match rotate updates

robyww:
Firefly-484: HiPS/Image WCS match rotate updates
  - The actual fix to this bug was to make the image rotate to the FITS orientation
  - Image rotating to the HiPS orientation on WCS match is actually a new feature that was requested in the past

When WCS matching:
   - When FITS rotates the HiPS will change coordsys it if it EQJ2000 north or Galactic north
   - When HiPS changes coordinateSys FITS will rotate to it.
   - FITS will rotate when locking to a HIPS

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-484

### Testing

https://irsawebdev9.ipac.caltech.edu/firefly-484-hips-rotate/firefly/

1. Attempt repeat the ticket, the problem should be fix.
1. Bring in a FITS and HiPS, wcs match lock
    - switch coordinate system on the HiPS, the fits will rotate
    - rotate the hips as you cross J2000 north or galactic north the HiPS will change coordinate systems.
    - rotate a image north, it the hips is galactic it will change to J2000

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ There are several branches the this repository.  Here are the ones that you shou
  -  [Tomcat 7+] (http://tomcat.apache.org/download-70.cgi)
     Apache Tomcat is an open source software implementation of the Java Servlet and JavaServer Pages technologies.
 
- -  [Node v6.9+] (https://nodejs.org/) >= 12
+ -  [Node v12+] (https://nodejs.org/) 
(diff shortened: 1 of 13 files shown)


lrebull:
I agree, issue from FIREFLY-484 seems to be fixed.

PR #938, opened 2020-02-25
FIREFLY-488: enforce new cache policy

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-488

Implement new cache policy for table searches.
Table results are only saved for the duration of the page.  Upon reload, all table searches should refetch data and not use cached version.

Also in this PR, fix Gator master table processor (CatMasterTableQuery) so that it does not do any caching of its own.  It should follow the same caching policy mentioned above.  In doing so, the code is rewritten to parse the IPAC table directly from IO stream avoiding having to write out the IPAC table into `perm-files` first.  This is one step closer to removing `perm_files` directory.  This should also fix the exception encountered when `perm_files` directory is deleted during runtime.

Test:  https://irsawebdev9.ipac.caltech.edu/FIREFLY-488_table_cache_update/firefly/
I test this using catalog search with larger radius.  It's obvious when a new fetch happens versus a cached load.

Also, added the ability to disabled  extended caching feature added for developer in a previous PR.  This allow a developer to experience the behavior of production environment locally.

To do disable extended caching:

`echo 'ExtendedCache = "false"' >> ~/.gradle/build.config `

### DIFF
--- a/buildScript/tasks.gincl
+++ b/buildScript/tasks.gincl
@@ -119,13 +119,14 @@ task createVersionTag  {
         props.setProperty('BuildMajor', major)
         props.setProperty('BuildMinor', minor)
         props.setProperty('BuildRev', rev)
-        props.setProperty('ClientBuildEnv', project.hasProperty("env") ? project.env : "local")
         props.setProperty('BuildType', type)
         props.setProperty('BuildNumber', buildNum)
         props.setProperty('BuildDate', build_date)
         props.setProperty('BuildTime', build_time)
         props.setProperty('BuildTag', tag)
         props.setProperty('BuildCommit', getCommitHash())
+
+        props.setProperty('ExtendedCache', project.env == "local" ? (project.hasProperty("ExtendedCache") ? project.ExtendedCache : "true") : "false")
 
         if (fireflyPath != rootDir.getPath()) {
             props.setProperty('BuildCommitFirefly', getCommitHash(fireflyPath))
@@ -823,7 +824,7 @@ task buildAndPublish( dependsOn: war ) {
 }
 
 
-task devMode (dependsOn: loadConfig) {
+task devMode (dependsOn: createVersionTag) {
     description= 'JavaScript Dev Mode'
     group = DEV_GROUP
     outputs.upToDateWhen { false }
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
@@ -3,15 +3,14 @@
  */
 package edu.caltech.ipac.firefly.server.catquery;
 
+import edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil;
+import edu.caltech.ipac.firefly.server.network.HttpServiceInput;
+import edu.caltech.ipac.firefly.server.network.HttpServices;
+import edu.caltech.ipac.firefly.server.query.EmbeddedDbProcessor;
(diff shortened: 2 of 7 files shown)


buildScript/tasks.gincl
```diff
@@ -119,14 +119,15 @@ task createVersionTag  {
         props.setProperty('BuildMajor', major)
         props.setProperty('BuildMinor', minor)
         props.setProperty('BuildRev', rev)
-        props.setProperty('ClientBuildEnv', project.hasProperty("env") ? project.env : "local")
         props.setProperty('BuildType', type)
         props.setProperty('BuildNumber', buildNum)
```
robyww: I think it would still be nice to keep access to the build environment.

src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
```diff
@@ -35,110 +35,48 @@
...
+    public DataGroup fetchDataGroup(TableServerRequest req) throws DataAccessException {
 
-            } else {
-                // concat dg to dgExtra
-                Iterator<DataObject> iter = dg.iterator();
-                while (iter.hasNext()) {
-                    dgExtra.add(iter.next());
+        AtomicReference<DataGroup> dgRef = new AtomicReference<>();
+        AtomicReference<DataAccessException> exRef = new AtomicReference<>();
+        HttpServices.getData(new HttpServiceInput(MASTER_LOC), method -> {
```
robyww: I think you should keep using URLDownload.  The way out stuff has more consistent logging. 
Eventually we might want to replace the insides of URLDownload with HttpServices.

buildScript/tasks.gincl
loitly: Build env is currently global:
https://github.com/Caltech-IPAC/firefly/blob/585c2a2704de025e134d1dffc9f454cb6a473d23/buildScript/webpack.config.js#L65

Let me know if you would like me to move it into our global `__PROPS__` instead.

src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
loitly: What I've done is creating a DataGroup from the response's OutputStream directly without writing to disk.
I need to be able to handle the response's OutputStream directly.
I don't think it's possible to do this with URLDownload as it is written.

buildScript/tasks.gincl
robyww: I did not realize that you were putting in in `NODE_ENV`. I thought `NODE_ENV` was suppose to have only two values `production` or `development`. 
I guess webpack and node only do code such as 
```js
NODE_ENV !== "production"
```
If it works that is fine.

Yes, I think it should be in __PROPS__ since that is how the client gets at it from the inside.

src/firefly/java/edu/caltech/ipac/firefly/server/catquery/CatMasterTableQuery.java
robyww: ok

robyww: I still have a general logging concern.  I have recently been investigating problems and find URL download easy to track while using HttpService more challenging.

We should think about it in the future.

loitly:
@robyww I removed the global NODE_ENV and added BUILD_ENV into global PROPS.  BUILD_ENV reflects our build env; local, dev, test, ops, etc.
@ejoliet It has been a few days now.  If you don't have any comments, I'll merge without your review.

PR #940, opened 2020-02-27
Firefly-486: Add file name in tab to catalog upload

robyww:
Show the file names on tabs when loaded from the catalog upload.   

 - Catalog tab show file name
 - fixed a small issue with coverage

_for review:_ also look at `UserCatalogQuery.java:prepareTableMeta`

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-486

_test_: https://irsawebdev9.ipac.caltech.edu/firelfy-486-file-in-tab/firefly/

- upload two file using the catalog upload (not the more general file upload)
- There must be two files, to test the tabs. (you can use the file in the ticket.)

### DIFF
--- a/src/firefly/js/tables/ui/TablesContainer.jsx
+++ b/src/firefly/js/tables/ui/TablesContainer.jsx
@@ -17,6 +17,11 @@ import {hashCode} from '../../util/WebUtil.js';
 import {LO_VIEW, LO_MODE, dispatchSetLayoutMode, getExpandedMode} from '../../core/LayoutCntlr.js';
 import {CloseButton} from '../../ui/CloseButton.jsx';
 
+const updateTitles= (tables) => Object.entries(tables)
+    .reduce( (obj,[key,{title,tbl_id, ...tableInfo}]) => {
+        obj[key]= {...tableInfo, tbl_id, title:(TblUtil.getTblById(tbl_id)?.tableMeta?.title ?? title)};
+        return obj;
+    },{});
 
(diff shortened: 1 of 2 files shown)


lrebull:
Yes, this seems to work now!

src/firefly/js/tables/ui/TablesContainer.jsx
```diff
@@ -17,6 +17,11 @@ import {hashCode} from '../../util/WebUtil.js';
 import {LO_VIEW, LO_MODE, dispatchSetLayoutMode, getExpandedMode} from '../../core/LayoutCntlr.js';
 import {CloseButton} from '../../ui/CloseButton.jsx';
 
+const updateTitles= (tables) => Object.entries(tables)
+    .reduce( (obj,[key,{title,tbl_id, ...tableInfo}]) => {
+        obj[key]= {...tableInfo, tbl_id, title:(TblUtil.getTblById(tbl_id)?.tableMeta?.title ?? title)};
+        return obj;
+    },{});
```
loitly: I see that it's working.  However, the logic should be placed here:
https://github.com/Caltech-IPAC/firefly/blob/585c2a2704de025e134d1dffc9f454cb6a473d23/src/firefly/js/tables/reducer/TableUiReducer.js#L93

This is where we take the tableModel and convert it into UI states to be rendered.  This way, it'll work even if we use TablePanel directly without TablesContainer.

I am not 100% sure.  So, if it does not work as described above, I would go ahead and merge this as is.

PR #944, opened 2020-03-03
Firefly-501: fixes region saving

robyww:
fixes firefly-501.

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-501

_to test:_
- https://irsawebdev9.ipac.caltech.edu/firefly-501-region-save/firefly/
- save a region file


lrebull:
Yes, confirmed that this fixes it.

PR #945, opened 2020-03-03
Firefly-499: fixed actions being lost in WebsocketConnector

robyww:
Firefly-499: fixed actions being load in WebsocketConnector
   - created test to send many table to Firefly
   - fixed a problem with unnecessary HiPS message

Fixes: Firelfy-509

https://jira.ipac.caltech.edu/browse/FIREFLY-499

_To test:_ 
1. use: https://irsawebdev9.ipac.caltech.edu/firefly-499-api-losing-tables/firefly/test/tests-viewer.html
1. go to test 5 and click
1. a firefly should load with all ten tables.
1. delete the firefly tab and repeat several times to make sure it always gets 10 tables.


ejoliet:
(BTW, adding the test is great idea!)

robyww:
I think is was out of scope of the ticket to put a "Firefly Ready..." panel.  However, we have had the in slate for 2 or 3 years. I  could add this to the Tri-view as well.

ejoliet:
IT's out of scope, agree. 

Whatever we decide to add we should discuss but i think we should show visually that 'something' is happening instead of the default empty image search form. 
Probably not "Firefly ready..." but something like "Working..." or "Loading..."  
I'm sure CCB can contribute to agree on the wording. 
That's a detail we can iterate on another ticket. 
I'll raise a ticket and we can discuss it later (maybe to be included in Summer). 
Thanks!

loitly:
Tested and it looks good.  Not sure why it waited for all tables to return before rendering.
Also, agree that we should add some indicators while waiting.  This should go into a new ticket.  We should talk about how it should be done.

PR #949, opened 2020-03-16
Firefly-518: Added better method method of managing overlay position

robyww:
Added "OverlayPostion" table meta data that will place search target on coverage or catalog.

_to test:_

https://irsawebdev9.ipac.caltech.edu/firefly-518-overlay-pos/firefly/test/test-coverage.html

- go down to test #7 
- click on next search
- each time the overlay position should update


cwang2016:
the resolution works well. Thanks for the update.

PR #950, opened 2020-03-16
firefly-514: revamped how we use starlink fits table reading so that we can close the FITS file.

robyww:
firefly-514: revamped how we use starlink fits table reading so that we can close the FITS file.

_to test:_
-  use the upload to read a table fits file. it should close when it is done.
- I am not making a jenkins build
- the only way to test is to build it on a local machine and stop in the debugger on line 436 and observe the file being closed.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
@@ -22,6 +22,7 @@ import nom.tam.fits.UndefinedHDU;
 import nom.tam.image.compression.hdu.CompressedImageHDU;
 import nom.tam.util.ArrayFuncs;
 import uk.ac.starlink.table.*;
+import uk.ac.starlink.fits.FitsStarTable;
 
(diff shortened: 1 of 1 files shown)


src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -389,24 +399,45 @@ public static StarTable getStarTable(String fits_filename, int table_idx) {
...
         }
+        finally {
+            try {
+                if (fits!=null && fits.getStream()!=null) fits.getStream().close();
+            } catch (IOException e) {
+                // do nothing
+            }
+        }
 
         FitsFactory.useThreadLocalSettings(false);
```
robyww: looking over the code this morning. I think i will move 
```java
 FitsFactory.useThreadLocalSettings(false);
```
to the finally block

cwang2016:
Just tested locally by setting stop in the debugger. Works well! Thanks for the fixing.

PR #965, opened 2020-04-22
Firefly-527: Web API Infrastructure

robyww:
### Firefly-527: Web API Infrastructure

  - Infrastructure done
  - Help page done
  - API is created for each application, not there by default
  - 4 searches,
      - image
      - HiPs
      - table
      - tap panel
  - Auto execute of tap panel
  - Updated showDropDown so it now fully supports initArgs
  - Now uses `api=<command>`

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-527
_ife PR:_ https://github.com/IPAC-SW/irsa-ife/pull/112

### Testing
_firefly:_ https://fireflydev.ipac.caltech.edu/firefly-527-url-api/firefly/
_irsaviewer:_ https://irsakudevlb1.ipac.caltech.edu/firefly-527-url-api/irsaviewer/
_sofia:_ https://irsakudevlb1.ipac.caltech.edu/firefly-527-url-api/applications/sofia/

adding   `?api`  will get you the help page

_firefly:_ https://fireflydev.ipac.caltech.edu/firefly-527-url-api/firefly/?api
_irsaviewer:_ https://irsakudevlb1.ipac.caltech.edu/firefly-527-url-api/irsaviewer/?api
_sofia:_ https://irsakudevlb1.ipac.caltech.edu/firefly-527-url-api/applications/sofia/?api

- try the URLs in the document
- copy a URL and add parameters to test errors
- copy a url change the command to test errors
- try other parameters from the document
- notice IRSAViewer does not have tap
- includes a starting implementation for the sofia web API


robyww:
> - once the URL is expanded, the same URL can't be copied and pasted again, the expanded URL looks like different than the API

This is the api vs state URL thing.  We said during the design session that this is what we were going to do.  However, we should talk about it more

> - We might never talked about it but does the API would allow POST? Like uploading a table from curl/wget POST? 

I don't think this would work. We could talk about where you are thinking about here.

> - I didn't notice a catalog search in the URL examples listed

I only was planning to add 3 commands (I did 4) and 1 example on to another app (maybe sofia, but not done yet). Adding more commands should be other tickets.

> - The HiPS example seems to land first on the empty form and couple of seconds depending on the latency i guess, the hips gets loaded. This is similar behavior seen when using IRSA Viewer from Atlas, the user see an empty form first while the image or catalog gets loaded in the back. I think there was a ticket to change the default page to be "loading" visual" or something so this probably is out of the scope of the PR but wanted to record it here so we don't forget.

I think this a problem with the way our HiPS loading works, since it is the same of the JS api, we need to pursuit it in another ticket.  I don't think it is so much related to the Web API

loitly:
Looking good.  Most of the functionalities are implemented.  Didn't think you were going to do so much in this PR.
Here are a few comments I have:

**Simple and Consistent**
This is tedious to implement, but I think it's important from an API perspective.
For example, when `url` is used for source URL, it should be used everywhere.  It should not be different for table vs image vs hips.
- parameters
  - Don't expose all parameters, just the important ones.
  - The same parameter should not means something else in another command. i.e. `service`
  - Unit should be same in all parameter.  i.e. `SizeInDeg` and `radiusInArcSec`  
  - Simple name, i.e. `radius`, `fov`, `band`, `pos`, etc.

- UI options
  - Should not be mixed with query parameters.  If you want to include them, place them under an `options:` sections
  - Not all options need to be exposed, but when it is, it should be consistence across the commands.  Like `title` to display should be done the same in image and table.


**Minor**
- example pointing to /staff/roby should be replaced with something else.. i.e. `https://irsa.ipac.caltech.edu/ibe/data/twomass/allsky/allsky/000102n/s014/image/hi0140009.fits.gz`
- syntax error detection is nice, but the error message should be formatted differently..
maybe having a section above the help... with a message and instructions to read the help below.

robyww:
@loitly I think you comments are helpful but out of scope for this PR. 
 

- Rationalizing all the parameters is a big task (that I am not sure we have the man power for) and should be done after we get the commands that we want. 
- My implementation are really just a starting points for the commands.
- I am not sure what I think about having an options parameter.  If so then that should be managed by the  infrastructure, but we still should decide that in another round.
- I think you are right about the help page but `WebApiHelpInfoPage.jsx` will undergo many iterations. I have already iterated on it 3 or 4 times. It is probably best to call it done for a while.

PR #969, opened 2020-05-01
Firefly-547: clean up docker

robyww:
Firefly-547: clean up docker
   - simplify `Dockerfile` and `launchFirefly.sh`
   - removed unused options
   - remove setenv.sh, work now done in launchTomcat
   - removed optional logging

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-547

@loitly please look over the following, I took most of your recommendations and improved the logging.

- simplified `Dockerfile`
- simplified and cleaned up `launchFirefly.sh`
- It is difficult to looks at diffs of these files since the changes are pretty extensive
- I took your (_very cool_) sendLogsToConsole function idea and make it into a script.
  - added a little logging before the tail commands
  - tail will show the whole file.
  - by having it as a script `ps` form inside the containers  gives me a better idea of what is going on


docker/base/sendLogsToConsole.sh
```diff
@@ -0,0 +1,19 @@
...
+echo "sendLogsToConsole: started: sleeping ${sleepTime}"
+sleep ${sleepTime}
+ALL_LOGS="${CATALINA_HOME}/logs/*[a-zA-Z].log ${CATALINA_HOME}/logs/*.out"
+IFS=" " read -r -a logArray <<< "$(echo ${ALL_LOGS})"
+namelist=' '
+for element in "${logArray[@]}"
+do
+    namelist="$(basename -- ${element}) ${namelist}"
+done
+echo -e '\n\n>>>>>>>>> sendLogsToConsole: Tailing logs: ' "${namelist}" '\n'
```
loitly, line 16: Because this is specifically for firefly, you could declaratively define the log files you want to output.

robyww: I know, there is actually a couple of places that I do computations. However, for now I though I would just keep it flexible.

PR #973, opened 2020-05-13
FIREFLY-465: adding sofia analyzer and include spectra 1d extraction from FITS image data

ejoliet:
The purpose of this PR is to add an DP analyzer that will be used for SOFIA data product purposes. See ticket here: [FIREFLY-465](https://jira.ipac.caltech.edu/browse/FIREFLY-465)
This PR has a counterpart PR for SOFIA in IFE [here](https://github.com/IPAC-SW/irsa-ife/pull/115).

The analyzer will trigger and add an extracted spectra to the report returned using a custom FITS HUD image extractor (also included) and a spectra model defined for SOFIA using VO UCDs and units as standard as possible.
Extracted and added spectra will depend on a couple of columns: instrument, product type and level.

The interesting code to look for and complete or update is here:
- The analyzer: `edu.caltech.ipac.firefly.server.dpanalyze.SofiaAnalyzer`
- The extractor: `edu.caltech.ipac.firefly.data.sofia.Sofia1DSpectraExtractor`
- The meta info added to the client dispatch in IFE to make use of the Firefly analyzer can be found here: `src/irsa/js/sofia/Sofia.jsx`

The analyzer will extract the data and add it to the report as a VOTable following a drafted spectral model. (See `edu.caltech.ipac.firefly.data.sofia.VOSpectraModel`)
The analyzer should be completed with other analysis, replacement or extraction as we go.

Caveats: 

1. The combinations to identify 1D spectra included in the FITS images are still in progress ([see confluence page](https://confluence.ipac.caltech.edu/pages/viewpage.action?spaceKey=IRSA&title=SOFIA+-+Plotting+flux+versus+wavelength
 )). 
2. Spectral model used is also in progress, ucds, units, etc. That will need more expertise/review.
3. What is absolutely left out the PR is how to deal with multiple plane (case of NAXIS3 or 'cube') , visualizing multiple spectra.

A way to test the 1D spectra is to 
- go to Firefly build here:
https://fireflydev.ipac.caltech.edu/firefly-465-sofia-analyzer/firefly/
- upload the table [here](http://web.ipac.caltech.edu.s3-us-west-2.amazonaws.com/staff/ejoliet/sofia/FORCAST-with-meta-analyzer.tbl), 
- select the second row
- noticed that the 'File Contents' now has 3rd part which correspond to the extracted data from the primary HDU and the chart represent flux and wavelength.
 
Another way to test is to 
- go to the SOFIA app build here:
https://irsakudevlb1.ipac.caltech.edu/firefly-465/applications/sofia/
- do a search around moving object 'Pallas' with level 2,3,4 checked and look for FORCAST, EXES or FLITECAM instruments on different product type and level. Using FORCAST table as an example, filter the table by product type column with `IN ('combspec','mrgspec','rspspec')`, first row highlighted should show a chart wavelength vs flux.
The menu 'File Contents' on those data product and other combination of instrument, product types and levels should show an additional part which represent the spectra extracted

For example, FLITECAM LEVEL 3 and product type 'IN ('calspec')'... etc.

If you got to this point and not satisfied with the testing, look at the if-then-else statement in the code `edu.caltech.ipac.firefly.server.dpanalyze.SofiaAnalyzer#analyzeFits` or see confluence page [here](https://confluence.ipac.caltech.edu/pages/viewpage.action?spaceKey=IRSA&title=SOFIA+-+Plotting+flux+versus+wavelength
 ).

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/DataExtractUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/DataExtractUtil.java
@@ -1,0 +1,47 @@
+package edu.caltech.ipac.firefly.data.sofia;
+
+import edu.caltech.ipac.firefly.data.ServerRequest;
+import edu.caltech.ipac.firefly.server.ServerCommandAccess;
+import edu.caltech.ipac.firefly.server.ServerContext;
+import edu.caltech.ipac.firefly.server.SrvParam;
+import edu.caltech.ipac.firefly.server.cache.UserCache;
+import edu.caltech.ipac.firefly.server.query.EmbeddedDbProcessor;
+import edu.caltech.ipac.firefly.server.util.multipart.UploadFileInfo;
+import edu.caltech.ipac.table.DataGroup;
+import edu.caltech.ipac.util.cache.Cache;
+import edu.caltech.ipac.util.cache.StringKey;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.File;
+
+/**
+ * Should implement child that returns datagroups,
+ * source file from request or overwritten in child or unit/integration tests
+ * Could be called as link service or return a datalink table
+ */
+public abstract class DataExtractUtil {
+
+
+    /**
+     * @param f File to extract data from
+     * @return File the data extracted/transformed
+     * @throws Exception
+     */
+    public abstract DataGroup extract(File inf) throws Exception;
+
+    /**
+     * @param req
+     * @return
+     */
+    public File getSourceFile(ServerRequest req) {
+        String fileKey = req.getParam("file");
+        Cache sessionCache = UserCache.getInstance();
+        File sourceFile = ServerContext.convertToFile(fileKey);
+        if (sourceFile == null || !sourceFile.canRead()) {
+            UploadFileInfo tmp = (UploadFileInfo) (sessionCache.get(new StringKey(fileKey)));
+            sourceFile = tmp.getFile();
+        }
+        return sourceFile;
+    }
+}
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
@@ -3,7 +3,6 @@ package edu.caltech.ipac.firefly.server.dpanalyze;
 import edu.caltech.ipac.firefly.core.FileAnalysisReport;
 import edu.caltech.ipac.firefly.data.sofia.Sofia1DSpectraExtractor;
 import edu.caltech.ipac.firefly.data.sofia.SofiaSpectraModel;
-import edu.caltech.ipac.firefly.data.sofia.VOSpectraModel;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.table.DataGroup;
 import edu.caltech.ipac.table.io.VoTableReader;
@@ -15,7 +14,7 @@ import nom.tam.fits.Header;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.*;
+import java.util.Map;
 
 import static edu.caltech.ipac.table.TableUtil.Format.VO_TABLE_TABLEDATA;
 
@@ -24,12 +23,6 @@ import static edu.caltech.ipac.table.TableUtil.Format.VO_TABLE_TABLEDATA;
  */
 @DataProductAnalyzerImpl(id = "analyze-sofia")
 public class SofiaAnalyzer implements DataProductAnalyzer {
-
-
-    @Override
-    public FileAnalysisReport analyze(FileAnalysisReport inputReport, File inFile, String analyzerId, Map<String, String> params) {
-        return inputReport;
-    }
 
     @Override
     public FileAnalysisReport analyzeFits(FileAnalysisReport inputReport,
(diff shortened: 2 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
...
+    public FileAnalysisReport analyzeFits(FileAnalysisReport inputReport,
+                                          File inFile,
+                                          String analyzerId,
+                                          Map<String, String> params,
+                                          Header[] headerAry) {
+
+
+        String code = params.getOrDefault("product_type", "");
+        String level = params.getOrDefault("processing_level", "");
+        String inst = params.getOrDefault("instrument", "");
```
robyww, line 44: It might be easier to go ahead and convert all to uppper case here. It would make the `if`'s look a little cleaner since you could just use `equals`

```java
String inst = params.getOrDefault("instrument", "").toUpperCase();\
.
.
.
if (inst.equals("FORCAST")) {
```

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
...
+
+            //TODO add error bars?
+            FileAnalysisReport.ChartParams cp = new FileAnalysisReport.ChartParams();
+            String xCol = VOSpectraModel.SPECTRA_FIELDS.WAVELENGTH.getKey();
+            String yCol = VOSpectraModel.SPECTRA_FIELDS.FLUX.getKey();
+            cp.setxAxisColName(xCol);
+            cp.setyAxisColName(yCol);
+            cp.setMode("markers");
+            addPart.setChartParams(cp);
+            retRep.addPart(addPart);
```
robyww, line 146: We might want to change `FileAnalysisReport` so you can add a part to the beginning. You could do it now by getting the parts list, putting it at the beginning, cloning to an empty report, then setting each part in the right order. But that is too much work.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
...
+        DataGroup dataObjects = model.extract(new File(fitsPath));
+        File tempSpectraFile = File.createTempFile("sofia1d-spectra-", ".xml", ServerContext.getTempWorkDir());
+        VoTableWriter.save(tempSpectraFile, dataObjects, VO_TABLE_TABLEDATA);
+        return tempSpectraFile;
+    }
+
+    /**
+     *Add column names and units to a fits image rendered as table
+     */
+    private FileAnalysisReport demoAddColumnNamesAndUnits(FileAnalysisReport inputReport, Header[] headerAry) {
```
robyww, line 170: remember to remove all the demo code.

src/firefly/java/edu/caltech/ipac/firefly/data/sofia/DataExtractUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.File;
+
+/**
+ * Should implement child that returns datagroups,
+ * source file from request or overwritten in child or unit/integration tests
+ * Could be called as link service or return a datalink table
+ */
+public abstract class DataExtractUtil {
```
loitly, line 23: Util classes should be written as public static.  If this is intended to be a DataExtractor base class or interface, then it should be promoted outside of sofia and have well-defined usage.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
+package edu.caltech.ipac.firefly.server.dpanalyze;
```
loitly, line 1: Sorry, I missed this in @robyww PR.  As a reminder, he should create new Java packages directly under firefly and not server.  Everything in firefly is 'server' now.

Also, consider where best to put all future analyzers.  If it's placed here, then it may clutter the core classes and interface, making it harder to distinguish core vs impl.
I would think it belongs with the other sofia classes.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
...
+ Analyzer for SOFIA extracted Spectra from FITS 'image'
+ */
+@DataProductAnalyzerImpl(id = "analyze-sofia")
+public class SofiaAnalyzer implements DataProductAnalyzer {
+
+
+    @Override
+    public FileAnalysisReport analyze(FileAnalysisReport inputReport, File inFile, String analyzerId, Map<String, String> params) {
+        return inputReport;
+    }
```
loitly, line 32: Unnecessary override.  Same as default impl.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -0,0 +1,285 @@
...
+
+    @Override
+    public FileAnalysisReport analyzeFits(FileAnalysisReport inputReport,
+                                          File inFile,
+                                          String analyzerId,
+                                          Map<String, String> params,
+                                          Header[] headerAry) {
+
+
+        String code = params.getOrDefault("product_type", "");
```
loitly, line 42: I didn't even know they added this method to Map.  Nice.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
robyww: I did not know that was out strategy on the server side. It is fine but we should talk about it.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
robyww: I missed this but java now allows interfaces to have default implementations. `DataProductAnalyzer` was the first time I ever wrote an interface like this. It is a nice feature.

ejoliet:
From Vandana and Justin: the EXES spectral model should be based on 'Wavenumber', not wavelength. I will fix that next.

ejoliet:
@robyww i noticed that showing or not the units or filters in the table can be changed here, under visualize/ui/MultiProductViewer.jsx:

`                <TablesContainer tbl_group= {tableGroupViewerId} mode='both' closeable={false} expandedMode={false} tableOptions={{showFilters:true, showUnits:true}} />
`
But it affects any data product. Is there an easy way to control it for a particular row or meta from a table? Or from the analyzer?

robyww:
>  noticed that showing or not the units or filters in the table can be changed here

Let me add that today.  Ok if I just do a push to this branch?

ejoliet:
> Let me add that today. Ok if I just do a push to this branch?
yes, i pushed yesterday some fixes also.
(I wanted to ask Vandana if showing units and filters by default is something we want for any table apart from this PR because it looks like a request everytime we have new features from a table...)

vandesai1:
I think we should be showing units and filters by default for ALL tables,
yes. Units are essential for scientific interpretation. Filtering is such a
great feature that is a little hidden.

Vandana

robyww:
Actually, i just discovered that filter is suppose to be on. So that is a bug I need to fix. @ejoliet I will do a push, hopefully today.

robyww:
@ejoliet I think we got confused. The data products tables have filter and units on by default. It is the search results table that do not. So, I am not going to do a push.

ejoliet:
@robyww ok, that means that this is something i need to change in SOFIA side when i dispatch the tables instrument initially?

robyww:
yes, `TblOptions`, second parameter in the `dispatchTableSearch`
```javascript
{
  showFilters: true,
  showUnits: true
}
```

ejoliet:
Ok, i did change it in `src/irsa/js/sofia/Sofia.jsx` to:

`const tblOptions = {showFilters: true, showUnits: true, removable: false, [LOG_HISTORY]: false};`

The main table only show units by default , not filters. The on efrom the data product though show the right settings, filters and units omn by default.
Maybe Sofia.jsx has something special in or afte rthe dispatch? 

But agree, it should work that way.

robyww:
I think the AOR table may be dispatched separately.

PR #974, opened 2020-05-15
 Firefly-62 Adding examples for  tooltip hovertemplate, with numeric …

wmiipac:
This PR is to address ticket [Firefly-62](https://jira.ipac.caltech.edu/browse/FIREFLY-62). 
When use firefly in NED SED plot  or IRSA Radar (SED/VizieR) plot, the hover tooltip data value showing data with scientific notation. For the NED/IRSA apps, the examples provide snippets to be added in using firefly api.
Test the changes:
https://fireflydev.ipac.caltech.edu/firefly-62/firefly/demo/ffapi-sed.html
you may compare with the dev build to see the difference, dev build:
https://fireflydev.ipac.caltech.edu/firefly/demo/ffapi-sed.html
IRSA app to be updated : Radar (Data Discovery)

### DIFF
--- a/src/firefly/html/demo/ffapi-sed.html
+++ b/src/firefly/html/demo/ffapi-sed.html
@@ -4,7 +4,7 @@
   ~ License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
   -->
 
-<html>
+<html xmlns="http://www.w3.org/1999/html">
 
 <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


wmiipac:
yes, I see that too, when click the highlighted data point it defaults to the unformatted data value, and default label of (x, y). I will change the formatted value to (x,y) as well, so the label/value stay the same

lrebull:
Looks good; I see the same thing about the unformatted tool tip values that Emmanuel found above.

lrebull:
The "case3 Showing proper error message when no photometry data retrieved" has no differences with respect to ops. Is that what I should be seeing?

ejoliet:
@wmiipac  did you managed to fix the bug mentioned: after clicking on a datapoint, the tooltip is changed bakc to full precision?

wmiipac:
> @wmiipac did you managed to fix the bug mentioned: after clicking on a datapoint, the tooltip is changed bakc to full precision?

click value and hover value is different, I have not found an easy way to control the data format for the click point value yet, in our case the point value is calculated from table columns that adds another level of complication. so in short the click point value is not formatted with desired scientific notation.

ejoliet:
Must be a way to stop propagation of the trigger into the formatting. Maybe @robyww or @loitly can take a look later. If no really way to solve it, then ok to merge the example for now and we will have to document the issue for now.

PR #978, opened 2020-05-26
Firefly-562: change some components to be functional and optimize rendering

robyww:
#### Firefly-562: change some components to be functional and optimize rendering

   - refactor multiple react classes into functional components
   - added multiple memo optimizations
   - Added a new hook
   - fixed coverage gridOn option
   - created a comparison function for PlotView that excludes scrolling and is very optimized
   - optimized a `getScreenPixScaleArcSec` which is called many, many times with same parameters

This PR is mostly about conversion to functional components, optimization, and some cleanup.
No behaviors should change. The code that I changed is now about 3/4 the size of the original.

_ticket_: https://jira.ipac.caltech.edu/browse/FIREFLY-562

#### Testing
_url:_ https://fireflydev.ipac.caltech.edu/firefly-562/firefly/

- Everything should work at before
- load some images or fits all


ejoliet:
- When saving a table, the default pre-filled filename is different than irsadev and not really user friendly, `${filename}`.

ejoliet:
-in SOFIA, search 'm17' with 100" radius, go to HAWC+ instrument tab, find a row with product type 'wcs', 'level2', then try to show the table from the file content (under 'data' tab), see how it fails with error message on the table background:
`Unable to read FITS file:/Users/ejoliet/hydra/workarea/sofia/visualize/fits-cache/DataProduct--839994494.fits`
That works in irsadev.

robyww:
> * When saving a table after loading a catalog from file, the default pre-filled filename is no longer the filename, it's `${filename}`.

This bug is is in `rc-2020.2`. It should be fixed but it is not a bug with this PR.

robyww:
> First issue i see is that cropping is not working (it is working in dev) using the selection tool.
> Error in the console: `TypeError: undefined is not an object (evaluating 'r.pt0')`

Yes, i found this, I will fix it.

robyww:
> -in SOFIA, search 'm17' with 100" radius, go to HAWC+ instrument tab, find a row with product type 'wcs', 'level2', then try to show the table from the file content (under 'data' tab), see how it fails with error message on the table background:

I can't duplicate this.  I don't even get a level 2 results for HAWC+

robyww:
Just push a fix to the crop bug, rebuilt on k8s.

PR #983, opened 2020-06-05
firelfy-347:tickbox on top of catalog search column

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-347

This PR fixed the none responsive tickbox on top of catalog search column.  It is in Firefly, so I did not make an ife branch.  I did test it by switching long/short and individual selection, all seem working fine. I did not clean it up since my fix may not be the the best way.  

Here is the test URL:
 https://fireflydev.ipac.caltech.edu/firelfy-347/firefly/


ejoliet:
Switching from standard to long should bring (and populate) the table with more column constraint (this is like in Gator website).
The reset suppose to reset to default the column selected either in standard or long mode.
Should i make another ticket to fix this behavior?

lznakano:
@loitly 
I did not remove the commented codes in case you find out the lost logic sometime later to switch it back it is.

PR #986, opened 2020-06-19
firefly-590:IRSA Viewer: return to defaults doesn't change color table back

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-590  
Fixed restore to default about the color issue.  
I did not change the behavior of the second button.  I don't know what is the expected behavior in the case below:
   1.  color overlay lock is in lock position
   2. change all plots to a different color (all image changed at the same time)
   3.  unlock the color overlay lock, and lock the second lock (image alignment)
   4. click "restore default",  all the image changed to the default color and the image
  
  One issue I noticed, "restore to default" also changed to "North up".  Is this the requested behavior? 

Test URL:
 https://firefly-590-colorissue.irsakudev.ipac.caltech.edu/irsaviewer/

### DIFF
--- a/src/firefly/js/visualize/task/PlotAdminTask.js
+++ b/src/firefly/js/visualize/task/PlotAdminTask.js
@@ -19,7 +19,10 @@ import {clone} from '../../util/WebUtil.js';
 import {detachSelectAreaRelatedLayers} from '../ui/SelectAreaDropDownView.jsx';
 import {getAppOptions} from '../../core/AppDataCntlr';
 import {WcsMatchType} from '../ImagePlotCntlr';
+import {hasOverlayColorLock, findPlotGroup } from '../PlotViewUtil';
 import {onPlotComplete} from '../PlotCompleteMonitor';
+
+
 
 export function autoPlayActionCreator(rawAction) {
     return (dispatcher) => {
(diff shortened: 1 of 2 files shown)


lznakano:
> I see the problem, I broke it when I changed the grouping. So this approach makes sense. However, it is not correct.
> 
> 1. Reset to default mean reset to how it was plotted. For example- we have a ticket to change the default color bar to something other than zero.
> 2. Also - is it possible that a set of plot could have been plotted with different colors.
> 3. You have to filter out 3 color and hips
> 4. The original and active color table are both found in plotState, since plotState hold the original request
> 5. I guess we want to restore if it is color lock or position locked.
> 6. it is all unnecessary in position lock context and may make a race condition since the server is be called with all the plots.
> 7. need to filter out the main plot id since it is going to be call anyway.
> 8. if it is not locked we _don't_ want to restore.
> 
> This is all very complex, more than it should be. Because of all this I wrote a function that I think will work, but I have not tested it.
> 
> ```js
> const changeColorsToDefault= (vr, plotId) =>
>     vr.plotViewAry
>         .map( (pv)=>  primePlot(pv)) // get the primary plot from the plotView
>         .filter( (plot)=>  plotId!== plot.plotId && !isHiPS(plot) && !isThreeColor(plot))  // only image and non-three color plot
>         .filter( (plot)=> plot.plotState.getColorTableId()!==plot.plotState.getWebPlotRequest().getInitialColorTable()) // compare color tables
>         .forEach((plot) =>
>             dispatchColorChange({plotId:plot.plotId,cbarId:plot.plotState.getColorTableId(), actionScope:ActionScope.SINGLE})); // change the color
> ```
> 
> It should be called in the following context in `restoreDefaultsActionCreator`
> 
> ```js
> const plotGroup= findPlotGroup(pv.plotGroupId,vr.plotGroupAry);
> hasOverlayColorLock(pv, plotGroup) && !vr.positionLock && changeColorsToDefault();
>         
> ```
@robyww 
Thank you for your explanation.  From the UI, I did not see the colorId is sent in the webPlotRequest.  So I thought the default is always 0.  I modified the approach based on the fact that the default colorTableId is sent from WebPlotRequest and can be any valid ColorTableId.  At first, I found that some of the images changed the size.  After looking further, I found that the scope is lost in the process.  That caused the group scope is used.  Also, the returning color to default should be irrelevant to the position Lock.  It should only respond to the overlayLock.  If the overlayLock is on, change to default colors for all images, if not, change only the active image.  Please review the changes again and let me know if any codes are not appropriate.  Thanks!

src/firefly/js/visualize/task/PlotAdminTask.js
```diff
@@ -86,23 +88,53 @@ export function changePointSelectionActionCreator(rawAction) {
...
+                         if (plot.plotId===vr.activePlotId && isImage(plot) && plot.plotState.getColorTableId() !== plot.plotState.getWebPlotRequest().getInitialColorTable()){
+                             dispatchColorChange({
+                                 plotId: plot.plotId, cbarId: plot.plotState.getWebPlotRequest().getInitialColorTable(),
+                                 actionScope: ActionScope.SINGLE
+                             });
+
+                         }
+                     }
+                }
+        );
```
robyww, line 126: You don't need to do anything if hasOverlayColorLock is false (i.e. not `else`) because because line 133 will take care of the single case.

lznakano: Yes, you are right on the line 133.

PR #988, opened 2020-06-23
 FIREFLY-580 Fixed bug that causes the phase-folded lc failed to load

wmiipac:
This PR fixes bug report by [FIREFLY-580](https://jira.ipac.caltech.edu/browse/FIREFLY-580).
1. column width now taking null sting into consideration
2. when parsing data to IPAC table taking null string as is

Test url:
https://fireflydev.ipac.caltech.edu/firefly-580/firefly/ts.html
upload the the table in [FIREFLY-580](https://jira.ipac.caltech.edu/browse/FIREFLY-580) and follow the steps to produce phase-folded lc. Check the phase-folded table in both table view and text view,  you should note the column width in the text view is corrected and the "null" string is kept as is in the table view.

### DIFF
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -563,11 +563,6 @@ export function getTypeLabel(col={}) {
     }
 }
 
-export function parsenullValue(cal, val) {
-    "use strict";
-
-}
-
 /**
  * returns an array of all the values for a columns
  * @param {string} tbl_id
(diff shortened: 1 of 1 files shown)


src/firefly/js/tables/TableUtil.js
```diff
@@ -563,6 +563,11 @@ export function getTypeLabel(col={}) {
     }
 }
 
+export function parsenullValue(cal, val) {
+    "use strict";
+
+}
+
```
loitly, line 570: Not used, please removed.

wmiipac: > Looks good, please correct the minor change requests before merging.

good catch. will do, thanks!

PR #992, opened 2020-06-29
FIREFLY-362-LowerLimits fix the text shown in the tooltip of upward pointing arrows for lower limit on Y axis values in scatter plots

cwang2016:
update for 
https://jira.ipac.caltech.edu/browse/FIREFLY-362

test:
https://fireflydev.ipac.caltech.edu/firefly-362-lowerlimits/firefly/test/tests-chart.html

please look at the 3rd chart with both upward and downward pointing arrows. 'Upper Limit' or 'Lower Limit' should be shown at the last line in the tooltip box respectively for downward or upward pointing arrows.


cwang2016:
@robyww I saw it is a list with either null or number. but it will be good to use isNil. Thanks.

PR #993, opened 2020-06-30
FIREFLY-183_scatter_plot_and_histogram_should_not_referesh_when_table…

deannaji:
The histogram should not re render when sorting tables:

Jira: https://jira.ipac.caltech.edu/browse/FIREFLY-183
K8s: https://fireflydev.ipac.caltech.edu/firefly-183/firefly/

Testing: 
- Run a catalog search like m17, with 300 radius. 
- Click on charts button and select 'Histogram'
- Click on the sorting button on any column in the table, the chart shouldn't refresh.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -504,6 +504,10 @@ function tablesourcesEqual(newTS, oldTS) {
 
 
 function updateChartData(chartId, traceNum, tablesource, action={}) {
+    // Histogram doesn't update on a table sort event.
+    if (getChartData(chartId).fireflyData[0].dataType === 'fireflyHistogram'){
+        if (action.payload && action.payload.invokedBy === 'table.sort'){return;}
+    }
     // make sure the chart is not yet removed
     if (isEmpty(getChartData(chartId))) { return; }
     const {tbl_id, resultSetID, mappings} = tablesource;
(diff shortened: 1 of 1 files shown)


xiuqin:
Tested in the specified build and it looks good.
I am going to create a new ticket for scatter plot for future and this ticket can be closed.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -504,6 +504,10 @@ function tablesourcesEqual(newTS, oldTS) {
 
 
 function updateChartData(chartId, traceNum, tablesource, action={}) {
+    // Histogram doesn't update on a table sort event.
+    if (getChartData(chartId).fireflyData[0].dataType === 'fireflyHistogram'){
+        if (action.payload && action.payload.invokedBy === 'table.sort'){return;}
+    }
```
tgoldina, line 510: There should be `traceNum` instead of `0`. 
Also you might want to move it to "else" block and rewrite it a bit to capture other non-scatter trace types, since the order of table rows only matters for scatter traces:
```
        if (action.type === TABLE_LOADED && action.payload.invokedBy === TABLE_SORT) {
            const {data} = getChartData(chartId);
            const traceType = get(data, [traceNum, 'type'], 'scatter');
            if (!traceType.includes('scatter')) return;
        }
```

deannaji: Done, thanks!

PR #995, opened 2020-07-07
Firefly-610: refactor the firefly redux bootsstrap process

robyww:
#### Firefly-610: refactor the firefly redux bootsstrap process 

 - The management of reducers and action creator is now in BootstrapRegistry
  - ReduxFlux import very little
  - Mostly eliminate the uninitialized constant problem that we occasionally encounter.
  - Allow us to create redux test that send dispatches though the store.
  - Allows us to import more code in testing the does not inadvertently bring in react
  - Decrease the mock code in _jest_/initTest.js
  - add babel-plugin-lodash to removed unused lodash functions

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-610

#### Testing
Everything should work like before.
https://fireflydev.ipac.caltech.edu/firefly-610-bootstrap/firefly/

#### Review: File to look at
Code refactor:
- `Firefly.js` - small changes
- `ReduxFlux.js` - mostly rewritten
- `BootstrapRegistry.js`- much of the old ReduxFlux has been moved here

Testing:
 - `initTest.js` - removed many `jest.mock` calls
- `ImagePlotCntlr-test.js` - demo of one test to going though Redux - look at line 39-51

lodash-webpack-plugin
- `webpack.config.js` - very small change

Other files are only just changes from `import firefly` to `import ReduxFlux`

### DIFF
--- a/__jest__/InitTest.js
+++ b/__jest__/InitTest.js
@@ -1,6 +1,6 @@
 import * as jest from 'jest';
 import {bootstrapRedux} from '../src/firefly/js/core/ReduxFlux';
-import {getBootstrapRegistry} from '../src/firefly/js/core/BootstrapRegistry';
+import {getBootstrapRegistry} from '../src/firefly/js/core/BootstrapRegistry.js';
 
 
 jest.mock('firefly/Firefly.js', () => {
@@ -41,8 +41,4 @@ jest.mock('firefly/visualize/DrawLayerCntlr.js', () => {
     };
 });
 
-bootstrapRedux(
-    getBootstrapRegistry(),
-    () => '',
-    () => ''
-);
+bootstrapRedux( getBootstrapRegistry());
--- a/__jest__/InitTest.js
+++ b/__jest__/InitTest.js
@@ -1,32 +1,12 @@
 import * as jest from 'jest';
+import {bootstrapRedux} from '../src/firefly/js/core/ReduxFlux';
+import {getBootstrapRegistry} from '../src/firefly/js/core/BootstrapRegistry';
 
-// jest.mock('firefly/Firefly.js');
-//
-//-- we are going to have to do something like this eventually, I think, right now I can't make it work
+
 jest.mock('firefly/Firefly.js', () => {
     return {
-        flux : {
-            process : () => {},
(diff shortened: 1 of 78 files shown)


src/firefly/js/core/ReduxFlux.js
```diff
@@ -96,238 +12,72 @@ import ImageLineBasedFootprint from '../drawingLayers/ImageLineBasedFootprint.js
...
+        redux.dispatch(ac ? ac(rawAction) : rawAction);
+        postDispatch(rawAction);
+    };
+    return process;
+}
+
+export function bootstrapRedux(bootstrapRegistry, preDispatch= () => undefined, postDispatch= () => undefined) {
+    if (!bootstrapRegistry) {
+        Logger('ReduxFlux').error('bootstrapRegistry parameter is require for firefly startup');
+        return;
```
loitly, line 63: I think allowing Firefly to change the behavior of `process` is a great idea.  I don't really like how `pre/post` is passed from here, all the way to `makeProcess`.
How about, instead of `pre/post` we take a decorator function.  We will apply it if exists.  i.e.

```
export function bootstrapRedux(bootstrapRegistry, processDecor= (process) => process ) {
    const process = processDecor(makeProcess(redux,bootstrapRegistry.getActionCreators)) : 
}  
```

This way, `makeProcess` does only what it suppose without external logic.

robyww: sure, the decorator idea did not occur to me. I will do that.

__jest__/InitTest.js
```diff
@@ -1,33 +1,13 @@
 import * as jest from 'jest';
+import {bootstrapRedux} from '../src/firefly/js/core/ReduxFlux';
+import {getBootstrapRegistry} from '../src/firefly/js/core/BootstrapRegistry';
```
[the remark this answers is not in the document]
robyww: I have been wrestling with this as well. When IntelliJ auto-adds imports it does so without the `.js`. So I have not really been doing this on purpose.  However, just now I found there is an option to tell IntelliJ to include the extension so I am going to enable that setting.

PR #997, opened 2020-07-09
FIREFLY-612:weird lag in chrome/safari layers popup

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-612

The lag reported is caused by the visibility property in two components.  This only happens when there are more images and layers because restoring them to default takes longer time.  In such situation, when the visibility is true, the component is drawn and displayed before the popup dialog is displayed.  In this PR, the visibility property was changed to default, inherit.  When the parent component is visible, these two components are visible as well.

To test:
URL: https://fireflydev.ipac.caltech.edu/firefly-612-layerpopup/firefly/
1. select at lease four sources to do the image search
2. lock the both overlay and position lock
3. change color 
4. change stretch
5. add many layers as many as 11 layers
6. click restore
7. click the layer button, it may not response immediately because the restoring to default is not finished.  

The bug can be reproduced in the same way using irsadev.ipac.caltech.edu

### DIFF
--- a/src/firefly/js/visualize/ui/DrawLayerItemView.jsx
+++ b/src/firefly/js/visualize/ui/DrawLayerItemView.jsx
@@ -55,7 +55,7 @@ export function DrawLayerItemView({maxTitleChars, lastItem, deleteLayer,
                          width:'100%'
                          }} >
                 <div style={{display: 'flex', alignItems: 'center'}}>
-                    <input type='checkbox' style={{visibility: canUserHide?'visible':'hidden'}} checked={visible} onChange={() => changeVisible()} />
+                    <input type='checkbox' style={{visibility: canUserHide?'inherit':'hidden'}} checked={visible} onChange={() => changeVisible()} />
                     {getTitleTag(title,maxTitleChars, autoFormatTitle)}
                 </div>
                 <div style={{padding:'0 4px 0 5px', width: 180, display: 'flex', justifyContent: 'flex-end'}}>
@@ -134,7 +134,7 @@ function makeDelete(canUserDelete,deleteLayer) {
         display:'inline-block',
         whiteSpace: 'nowrap',
         marginLeft: mLeft*2+symbolSize,
-        visibility: canUserDelete ? 'visible' : 'hidden'
+        visibility: canUserDelete ? 'inherit' : 'hidden'
     };
     return (
         <div style={bStyWid}>


lznakano:
> * PointSelection.js

Can you show me how to test if the UI was broken?  I thought the changes only affected the layer popup because the DrawLayerItemView was only used by DrawLayerPanel.   "CanUserDelete" is the property of the DrawLayerItemView.  The ImageRoot and PointSelect still have their own "CanUserDelete" property.  I guess that I am missing the whole picture of it.  I can test to find a better option if the changes affect UI.

robyww:
DrawLayerItemView takes it property from directly from the layer.  
```
   canUserDelete={l.canUserDelete}
```
If you bring up an image you can see the image layer has the delete `x` and you turn on lock by click you can that the layer has the `x`

lznakano:
> lock by click

Sorry, I compared the PR and the irsadev, can not see what was broken.  I know that the layer has "canUserDelete" property passed to the DrawLayerItemView.  As soon as "canUserDelete" is true, the "Delete" div was drawn.  Removing this dependency prevent the "Delete" from drawing earlier.  I thought that the only effect is deferring drawing "Delete".  Let me see if other combination will work the same.

lznakano:
@robyww @ejoliet 
I still don't understand how the UI is affected by changing the Div style  in the DrawLayerItemView.  Please let me know how to reproduce the broken UI issue.  So that I can see what is the cause.

robyww:
This is the easiest to see: bring up an image, show the drawing layers, look at the top layer. On the that layer UI there should not be an `x`.  In you PR there is an `x`. If you click it then it has an undefined behavior.

lznakano:
> This is the easiest to see: bring up an image, show the drawing layers, look at the top layer. On the that layer UI there should not be an `x`. In you PR there is an `x`. If you click it then it has an undefined behavior.

Here is what I did in order to reproduce it:
1. search all wise, m81
2. open the layer popup dialog
3. I saw the x in all images
4. I click the x, the image was deleted.
5. I checked the irsadev, it has the same behavior.  

What did I do wrong?

robyww:
Screen Shot 2020-07-15 at 1 59 38 PM

The delete in the top right should not been there.

lznakano:
> Screen Shot 2020-07-15 at 1 59 38 PM
> 
> The delete in the top right should not been there. If you click on it you get unexpected results

I was looking for the x not the Delete.  Now I see what you mean. Thank you!    Let me see what I can do about it.

PR #999, opened 2020-07-16
TAP UI: Enhance schema browser with qualified name column

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-596
Test:    https://fireflydev.ipac.caltech.edu/firefly-596-tap-ui-updates/firefly/?__action=layout.showDropDown&view=TAPSearch

Changes:
- Insert "SELECT TOP 1000 * FROM <table>" only when `ADQL Query` field is empty, otherwise insert just the table name.
- Add checkbox `Insert fully-qualified column names`, defaults to checked.
  - When checked, insert fully-qualified column names, which means table-name.column-name
  - When unchecked, insert just the column name

### DIFF
--- a/src/firefly/js/ui/tap/AdvancedADQL.jsx
+++ b/src/firefly/js/ui/tap/AdvancedADQL.jsx
@@ -100,7 +100,6 @@ export function AdvancedADQL({adqlKey, defAdqlKey, groupKey, serviceUrl, style={
                     <div className='flex-full'>
                         <div style={{display: 'inline-flex', marginRight: 25, justifyContent: 'space-between', alignItems: 'center'}}>
                             <h3>ADQL Query:</h3>
-                            <label style={{display: 'flex', alignItems: 'center'}}> <input type='checkbox' ref={ffcn} defaultChecked={true} />Insert fully-qualified column names</label>
                             <div style={{display: 'inline-flux'}}>
                                 <button className='button std' title='Reset to the initial query' style={{height: 24, marginRight: 5}} onClick={onReset}>Reset</button>
                                 <button className='button std' title='Clear the query' style={{height: 24}} onClick={onClear}>Clear</button>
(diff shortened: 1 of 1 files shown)


loitly:
> It works and looks good. The ticket is implemented.
> 
> That said, I wonder if you could see if the cursor is sitting right after another column, if so, add a comma as well. Without a fully sql parser I am sure there is all sort of things ghat can go wrong but even a simple implementation might improve the flow. Maybe always insert a comma if there is a non white space in front of the insertion position.

It's hard to guess when a comma is needed.  In the 'select' clause, yes.  But, in the 'where' clause, it's not so straight-forward.  That's why instead of guessing, I place the cursor right after the inserted name.  With left hand on comma key and right hand on mouse, you can pretty quickly/easily create a bunch of name/comma set.

gpdf:
A few comments:

1. Regarding the comma insertion, while it would be awesome to be clever about it, I suspect that users are more likely than not to be using the schema browser to construct WHERE clauses to refine a query started in the simple UI; the simple UI already makes it easy to checkbox a bunch of columns and have the right SELECT generated.
1. I would like to suggest a change to make the point of the new checkbox a little clearer.  Change the subtitle under the ADQL field to "Schema Browser Usage" (from "Hints") and move the checkbox and its associated text below the second line, the "Click on a Column node" one, slightly indented.
1. I see that when the **table** name is already fully qualified in TAP_SCHEMA, e.g., for "ztf.ztf_current_meta_sci", where the schema name is included, it looks like you are deleting the schema prefix when inserting the column name; in this case, for instance, you insert "ztf_current_meta_sci.ra".  I don't **think** this is consistent with the TAP standard.  The table name is whatever TAP_SCHEMA says it is, whether there's a "." in it or not; so a fully-qualified column name has to include the table name; Firefly can't try to guess that it's safe to strip off the schema name.

gpdf:
This is the relevant language from the standard:

> The value of the `table_name` [field in TAP_SCHEMA] should be the string that is recommended for use in querying the table; it may or may not be qualified by schema and catalog name(s) depending on the implementation requirements. The fully qualified table name is defined by the ADQL language and follows the pattern [[catalog.]schema.]table. If the table name is such that the name must be quoted (delimited identifier in ADQL) then the value must include the quotes.

I have not been able to find a TAP service that has a quoted table name against which to test, but it is at least a possibility.

gpdf:
This is what I have in mind:

ADQL-dialog-edit

loitly:
@gpdf Regarding "stripping of schema name".  I was thinking of usability, being that the name get long and hard to read.  But you're right, we should not guess.  Plus, it's wrong in the case where '.' is in the name.
No attempt made to guess when quotes are needed for table or column names.  The user must insert quotes where needed.
Thank you for the useful feedback.  I implemented all of the suggestions.  A new build is created and ready for testing.

PR #1000, opened 2020-07-17
Firefly-505: added support for a loading message

robyww:
#### Firefly-505: added support for a loading message 

 - Loading message is settable from API viewer code
  - `ApiViewer.js` cleaned up
  - ` FireflySlate.jsx` cleaned up, converted to functional component
  - `react-grid-layout` package updated, `GridLayoutPanel.jsx` tweeked to support changes

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-505

#### Testing
https://fireflydev.ipac.caltech.edu/firefly-505-starting-message/firefly/test/tests-viewer.html

- Click on any link and you will see the loading message.
- The message might not be there long, but is more obvious with HiPS and MOC and some tables

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "react-datetime": "~2.16",
     "react-dnd": "~7.0",
     "react-dnd-html5-backend": "~7.7",
-    "react-grid-layout": "~0.17",
+    "react-grid-layout": "~0.18",
     "react-select": "~3.0",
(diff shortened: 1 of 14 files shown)


robyww:
@ejoliet we could do that, we would have to add a little more api to get the string set the way a project wants it. Write now the string is set internally by the components. It would be easy to do though.

PR #1003, opened 2020-07-23
irsa-2716:SOFIA GREAT: Level 4 .fits files can be visualized

lznakano:
https://jira.ipac.caltech.edu/browse/IRSA-2716

Add the analyzer to analysis the SOFIA GREAT not_scale data. And then display the flux vs frequency table and chart.

To Test:

 URL: https://irsa-2716-sofia-v1.irsakudev.ipac.caltech.edu/applications/sofia/
URL: https://irsa-2716-sofia-v2.irsakudev.ipac.caltech.edu/applications/sofia/

  1. click the URL above
  2. select, allSky, Level4
  3. filter the data in file-url field by "not"
  4. click "Data" tab

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/SofiaFitsConverterUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/SofiaFitsConverterUtil.java
@@ -1,11 +1,16 @@
 package edu.caltech.ipac.firefly.data.sofia;
 
+import edu.caltech.ipac.table.DataGroup;
+import edu.caltech.ipac.table.DataObject;
+import edu.caltech.ipac.table.DataType;
 import edu.caltech.ipac.util.download.FailedRequestException;
 import edu.caltech.ipac.util.download.URLDownload;
+import edu.caltech.ipac.visualize.plot.ImageHeader;
 import edu.caltech.ipac.visualize.plot.plotdata.FitsReadUtil;
 import nom.tam.fits.*;
 import java.io.*;
 import java.net.URL;
+import java.util.ArrayList;
 
 
 public class SofiaFitsConverterUtil {
@@ -71,9 +76,10 @@ public class SofiaFitsConverterUtil {
 
     }
 
-    private static  BinaryTableHDU convertWaveTabHDU(BasicHDU hdu) throws FitsException {
 
-        //convert the data to float to do all the calculations
+
+
+    private static  BinaryTableHDU convertWaveTabHDU(BasicHDU hdu) throws FitsException {
 
         double[] coords = FitsReadUtil.getImageHDUDataInDoubleArray(hdu);
         Header header = hdu.getHeader();
@@ -155,6 +161,7 @@ public class SofiaFitsConverterUtil {
         return outFits;
 
     }
+
 
     /**
      * If the FITS is a wave-tab: 1. CTYPEi=WAVE-TAB or Lambda; 2: there is a HDU that has an EXTNAME="WAVELENGTH"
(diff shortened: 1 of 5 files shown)


lznakano:
> Thanks. I've tested the different files and the generated charts are meaningful now.
> 
> In terms of chart, i think you got all the requested features Justin asked but i'm concerned that we don't keep the original FITS image for later download.
> The archive data is the FITS images, not the charts, those are byproduct generated for the UI.
> 
> I think we need to have each HDU image accompanied with table and chart generated. The menu will have indentation for the produced spectra chart and table. See how i did it for the others in the analyzer.
> 
> I think @vandesai1 or Justin missed the point of the orginal data to be exposed and available in the UI as well.

Do you know where the image is now displayed when you do the same procedure as listed in the PR in irsadev.ipac.caltech.edu/applications/sofia?  It only showed 1-d image table and chart.  I can see the image if I download the FITs file and view it in irsaviewer.  I thought that the image is no longer needed.

lznakano:
@ejoliet 
I made two versions so that people can compare the difference and choose one that they like.

ejoliet:
Great you got for those 2 versions! Let me take a look and talk to @vandesai1 so they understand the comparison. 
Thank you@

ejoliet:
@lznakano i think that regardless the decision we will make about this visualization, SOFIA analyzer/product viewer is missing the updates that @robyww did here #979 (requested from Vandana's feedback) to deal with orginal and extracted/generated data where the main HDU is a 'parent' and the extracted data is inserted and 'below' with indentation.
I actually don't see this in irsadev anyway so that's why i don't see it here either.
I think we need to fix this and integrate those changes in SOFIA as well. I will try to fix it (or let you know what is needed) in your PR otherwise will raise a ticket. Thanks.

lznakano:
> @lznakano i think that regardless the decision we will make about this visualization, SOFIA analyzer/product viewer is missing the updates that @robyww did here #979 (requested from Vandana's feedback) to deal with orginal and extracted/generated data where the main HDU is a 'parent' and the extracted data is inserted and 'below' with indentation.
> I actually don't see this in irsadev anyway so that's why i don't see it here either.
> I think we need to fix this and integrate those changes in SOFIA as well. I will try to fix it (or let you know what is needed) in your PR otherwise will raise a ticket. Thanks.

OK.  I will wait for you, then.  Thanks!

ejoliet:
@lznakano i've talked to Vandana. We will keep version 2.

The general idea is that we want to be consistent with the other data product viewer (example, FORCAST or EXES 1D spectra where HUD and extracted data are both shown). 

So, let's keep version 2 of your PR where you show HDU primary as a parent and the extracted data as child indented in the file menu, that is the general idea for the UI in SOFIA.
 
It seems that it works great in your version 2 deployed but somehow it fails in the case of FORCAST or EXES (in irsadev too so is not related to this PR).
 
I have to ask @robyww why, i've tried to follow and use the 'insertAfter' method but it still doesn't show the right UI (parent+child). 
Let me raise ticket for Trey to look at this specific problem. It affects the part where i added extractedSpectra since #973.

In your PR here, keeping version 2, i noticed 2 more issues we need to discuss:
 
1. one is related to the extracted data or table, it needs to use the spectral model to set the units/ucds introduced in #973 which add VO attributes to the table, see in particular the class used for 1D spectra:`edu.caltech.ipac.firefly.data.sofia.SofiaSpectraModel`. 
I know it doesn't currently fully fit the need here wrt the units/column name you have to set but we need to add this flexibility to the model to be more generic and use it for those tables here. Does it make sense to you? 
The use of the model will ensure that the user gets the right VO table with VO UCDs in it. Currently it doesn't. See my PR if that helps: #973
Please take a look and let's discuss it next time if you have questions. 

2. The FITS to visualize are suppose to be 'image' type, so the primary HDU (parent in the 'file contents' menu) should be an image and the UI is suppose to show 3 buttons: 'table', 'chart' 'image'. This is the original data and the general UI Firefly is suppose to show for this type of FITS. (if you upload one of those file with Firefly viewer upload feature, you can see that each HDU is an image which doesn't really help to see any spectra but it happens with other FITS in SOFIA unfortuaneltyel. In order to to be consistent, version 2 is the right UI but it doens't show the 3 buttons. Maybe is related to the issue i mentioned earlier above and we need to talk to Trey about it so maybe wait for fixing or dealing with this issue until we talm to @robyww.

If you have questions, let me know!
Thank you!

lznakano:
> @lznakano i've talked to Vandana. We will keep version 2.
> 
> The general idea is that we want to be consistent with the other data product viewer (example, FORCAST or EXES 1D spectra where HUD and extracted data are both shown).
> 
> So, let's keep version 2 of your PR where you show HDU primary as a parent and the extracted data as child indented in the file menu, that is the general idea for the UI in SOFIA.
> 
> It seems that it works great in your version 2 deployed but somehow it fails in the case of FORCAST or EXES (in irsadev too so is not related to this PR).
> 
> I have to ask @robyww why, i've tried to follow and use the 'insertAfter' method but it still doesn't show the right UI (parent+child).
> Let me raise ticket for Trey to look at this specific problem. It affects the part where i added extractedSpectra since #973.
> 
> In your PR here, keeping version 2, i noticed 2 more issues we need to discuss:
> 
> 1. one is related to the extracted data or table, it needs to use the spectral model to set the units/ucds introduced in #973 which add VO attributes to the table, see in particular the class used for 1D spectra:`edu.caltech.ipac.firefly.data.sofia.SofiaSpectraModel`.
>    I know it doesn't currently fully fit the need here wrt the units/column name you have to set but we need to add this flexibility to the model to be more generic and use it for those tables here. Does it make sense to you?
>    The use of the model will ensure that the user gets the right VO table with VO UCDs in it. Currently it doesn't. See my PR if that helps: #973
>    Please take a look and let's discuss it next time if you have questions.
> 2. The FITS to visualize are suppose to be 'image' type, so the primary HDU (parent in the 'file contents' menu) should be an image and the UI is suppose to show 3 buttons: 'table', 'chart' 'image'. This is the original data and the general UI Firefly is suppose to show for this type of FITS. (if you upload one of those file with Firefly viewer upload feature, you can see that each HDU is an image which doesn't really help to see any spectra but it happens with other FITS in SOFIA unfortuaneltyel. In order to to be consistent, version 2 is the right UI but it doens't show the 3 buttons. Maybe is related to the issue i mentioned earlier above and we need to talk to Trey about it so maybe wait for fixing or dealing with this issue until we talm to @robyww.
> 
> If you have questions, let me know!
> Thank you!

Issue 2 is not relevant to this PR. I reported to you previously.
Issue 1:  What I did is the standard Spectra processing according to the Spectra definition and algorithm that described in a set of a few papers which I had read and researched previously.  The FREQ, VRAD and WAVE are the CTYPEia.   The FREQ in the current situation is the simplest case.  It has no dependency on other axes and the algorithm is linear.  The UNIT should be defined in CUNITia.  All the information is in the header.  I don't know why the model has to be used.  Although the FREQ was implemented for sofia, it is the standard process, just like wavelength.  Maybe you can explain how the SpectralModel is constructed.  The UNIT can only be available from model and not in the FITs header??? You can explain more to me in the meeting.   Thanks!

PR #1004, opened 2020-07-30
irsa-3638:The analyzer for "GREAT" instrument image does not have the image tab

lznakano:
https://jira.ipac.caltech.edu/browse/IRSA-3638

This issue affected all instruments except FIFI-LS because FIFI-LS does not have one-dimension table.  The bug is a small logic issue in the PartAnalyzer.  Please see the attached images in the ticket https://jira.ipac.caltech.edu/browse/IRSA-3638. All the images were screenshots from running the nightly build. 

The issue first was notified in the GREAT instrument which I was adding an analyzer to.   I commented in the ticket that all the instruments have the same issue.  But I did not change the ticket title.

Please test it
URL: https://irsa-3638-analyzer.irsakudev.ipac.caltech.edu/applications/sofia/

If you prefer, you can run the nightly build to compare, thus, you can see the difference between this two build.  The image tag is shown in my PR among all instruments if the data has the analyzer.  The image tab was not shown in the nightly build.

### DIFF
--- a/src/firefly/js/metaConvert/PartAnalyzer.js
+++ b/src/firefly/js/metaConvert/PartAnalyzer.js
@@ -8,7 +8,7 @@ import {RequestType} from '../visualize/RequestType.js';
 import {TitleOptions} from '../visualize/WebPlotRequest';
 import {createChartTableActivate, createChartSingleRowArrayActivate} from './converterUtils';
 import {createSingleImageActivate} from './ImageDataProductsUtil';
-
+import {isEmpty} from 'lodash';
 /**
  *
  * @param part
@@ -24,6 +24,8 @@ export function analyzePart(part, request, table, row, fileFormat, serverCacheFi
 
     const {type,desc, fileLocationIndex}= part;
     const availableTypes= findAvailableTypesForAnalysisPart(part, fileFormat);
+    if (isEmpty(availableTypes)) return {imageResult:false, tableResult:false};
+
     const fileOnServer= (part.convertedFileName) ? part.convertedFileName : serverCacheFileKey;
 
     const imageResult= availableTypes.includes(DPtypes.IMAGE) && type===FileAnalysisType.Image &&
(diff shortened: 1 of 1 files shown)


src/firefly/js/metaConvert/PartAnalyzer.js
```diff
@@ -24,6 +24,8 @@ export function analyzePart(part, request, table, row, fileFormat, serverCacheFi
 
     const {type,desc, fileLocationIndex}= part;
     const availableTypes= findAvailableTypesForAnalysisPart(part, fileFormat);
+    if (availableTypes===undefined || availableTypes.length===0) return {imageResult:false, tableResult:false};
```
robyww, line 27: cleaner to use lodash `isEmpty` here

src/firefly/js/metaConvert/PartAnalyzer.js
```diff
@@ -71,7 +73,8 @@ export function chooseDefaultEntry(menu,parts,fileFormat, dataTypeHint) {
 function findAvailableTypesForAnalysisPart(part, fileFormat) {
     const {type}= part;
     if (type===FileAnalysisType.HeaderOnly || type===FileAnalysisType.Unknown) return [];
-    if (type!==FileAnalysisType.Image || fileFormat!=='FITS' || is1DImage(part)) return [DPtypes.CHART,DPtypes.TABLE];
+
+    if (type!==FileAnalysisType.Image &&  fileFormat!=='FITS' &&  is1DImage(part)) return [DPtypes.CHART,DPtypes.TABLE];
```
robyww, line 77: I see how this logic is very confusing, I was confused myself until I did some testing.  The fix is not correct, it breaks out reading VO tables (and maybe other tables). We should talk about what the right fix is.
Probably just having an if right before.
```
if (type===FileAnalysisType.Table) return [DPtypes.CHART,DPtypes.TABLE];
```
We need to test that though.

lznakano: Can you please let  me know a vo-table test case?  If so, I can figure out a solution.  Thanks!

robyww: Any changes to the core of the data products viewer should be tested with a firefly build using the file:  http://web.ipac.caltech.edu.s3-us-west-2.amazonaws.com/staff/roby/data-products-test/dp-test.tbl

lznakano: I guess if I merge my branch, I can test the vo-table.  Let me do that first.

robyww: I tried my suggested fix and it works for dp-test.tbl. Maybe it fixed the GREAT problem as well.

lznakano: > Any changes to the core of the data products viewer should be tested with a firefly build using the file: http://web.ipac.caltech.edu.s3-us-west-2.amazonaws.com/staff/roby/data-products-test/dp-test.tbl

Good to know.  I will remember that from now on.

PR #1005, opened 2020-07-31
FIREFLY-595: Handle cell with long entries

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-595

Handle cell with long entries
- show '...' icon when hover over cells with value clipped
- on click, show actions as context popup menu
  - Copy to clipboard
  - View as plain text:  show full content as text in a dialog

Also:
- fix minor build related script bugs
- update fixed-data-table-2 to latest version

To test: https://fireflydev.ipac.caltech.edu/firefly-595-long-cell-entries/firefly/

### DIFF
--- a/buildScript/webpack.config.js
+++ b/buildScript/webpack.config.js
@@ -44,7 +44,7 @@ export default function makeWebpackConfig(config) {
         html_dir    : 'html',
         use_loader  : true,
         filename    : '[name]-dev.js',
-        deploy_dir  : (process.env.HYDRA_ROOT || '/hydra') + `/server/tomcat/webapps/${config.baseWarName}`,
+        deploy_dir  : (process.env.tomcat_home || '/hydra/server/tomcat') + `/webapps/${config.baseWarName}`,
         alias       : {
             firefly : path.resolve(config.firefly_dir, 'js'),
             styles : path.resolve(config.firefly_dir, 'html', 'css'),
--- a/buildScript/tasks.gincl
+++ b/buildScript/tasks.gincl
@@ -79,23 +79,29 @@ task loadConfig {
             configRepo.merge(ConfigRepo.parse(project.env, project.appConfig))
         }
 
-        project.ext.appConfigProps = configRepo.flatten();
-//    appConfigProps.each{ k, v -> println "${k} => ${v}" }
+        def props = configRepo.flatten();
+//        props.each{ k, v -> println "^^ ${k} => ${v}" }
 
         if (file("$rootDir/buildScript/build.config").exists()) {
-            def buildProps = ConfigRepo.parse(project.env, file("$rootDir/buildScript/build.config")).flatten()
-            buildProps.each { k, v -> project.ext.set(k,v)}
+            ConfigRepo.parse(project.env, file("$rootDir/buildScript/build.config"))
+                    .flatten().each { k, v -> props.put(k,v)}
         }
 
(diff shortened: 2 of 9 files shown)


buildScript/webpack.config.js
```diff
@@ -44,7 +44,7 @@ export default function makeWebpackConfig(config) {
         html_dir    : 'html',
         use_loader  : true,
         filename    : '[name]-dev.js',
-        deploy_dir  : (process.env.HYDRA_ROOT || '/hydra') + `/server/tomcat/webapps/${config.baseWarName}`,
+        deploy_dir  : (process.env.tomcat_home || '/hydra/server/tomcat') + `/webapps/${config.baseWarName}`,
```
robyww, line 47: This is going to require a change from everyone? right?

loitly: If you have Tomcat installed at a location other then '/hydra/server/tomcat', you are required to set `tomcat_home` to that location.  Here, using 'tomcat_home' to determine the location of Tomcat's webapps is the right thing to do.  
But, you are right.  I am not sure if this will affect anyone or not.  I am guessing not, but I can't be sure.  I should bring this to everyone's attention at the next standup.

robyww: Go I know both @ejoliet and I have updated to the most recent version of OSX, which required use to move tomcat.

gpdf:
I agree with the above comments - I think the "..." icon chosen is really nice, and blends in with the UI well despite being easy to see.

Only one substantive suggestion: please make the pop-up **significantly** wider by default.  Right now it's around 300 pixels wide by default - I think 550 or even up to 700 pixels would be a real improvement.  I've found that in virtually every "natural" case where I've invoked the pop-up the first thing I did was to stretch it horizontally.  If it's a long value like a URL or an ObsCore `s_region` polygon spec, having fewer line breaks is always better.

You can make it a little shallower at the same time so that the amount of screen area covered doesn't grow - this also makes it easier to move the pop-up around without hiding toolbars.  Right now it seems to have a fixed minimum height which is pretty tall!

How about a text area size of 650x125 by default (plus the title bar), with the height able to be reduced to ~75 with the stretch box?

From here on are just observations, not requests for changes:

I like that the text field uses word wrap, rather than breaking at a fixed column.

An observation (for the record): the dialog box is non-modal and re-usable, but it is *not* an "inspector" - meaning that it is re-used if you invoke it on a different table cell, rather than a new one coming up, and stays "floating" over the Firefly UI regardless of what else you do, *but* its contents only update based on explicit request, not on a highlight-state change elsewhere in the UI (contrast this, for instance, with the layers dialog, which updates based on the currently highlighted image).  

Based on short acquaintance, I think this is fine for now.  We might want to re-think this in the future after experience has built up - I'll be interested to see what Louisa thinks.  In any event, I would request notification of the CCB-D if this behavior were ever changed.

I like the logical consequence of the above: that you can invoke the menu without changing the row-highlight of the table.  This is similar to how context menus work in the macOS UI - you can invoke the context menu for a file in the Finder without selecting it.  Again, please don't change this without further review.

Note that the combination of the above two remarks mean that you can't do one potentially nice thing: if you have a table with a super-wide column, set up the pop-up, and then review the contents of all the visible rows by using the up/down arrows to change the highlighted row - with the pop-up changing contents to follow.  I think this is OK, but it might be something we'd want to look at again someday.

Nice job!

loitly:
@gpdf changes made as you requested.  A new build is created and ready for testing.  Thanks.

PR #1008, opened 2020-08-12
FIREFLY-629:FIFI-LS: Image button shouldn't be present in HDU table type

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-629  

 This PR fixed issue that the image tab is present for 1D Image.

To test:
URL: https://lijun.irsakudev.ipac.caltech.edu/applications/sofia/

Target: CasA
Instrument: FIFI-LS

PR #1009, opened 2020-08-12
FIREFLY-631: fixing spectra and other instrument data product viewer related issues

ejoliet:
This is a PR that should fix most of the problem seen lately and related to 1d spectra and SOFIA analyzer. In particular:

+ GREAT L1 1D image default shown as chart
+ GREAT L4 checked
+ FIFI-LS types other than L3 wavelength_resampled to be shown as table (binary table types)
+ FORCAST spectra 1D test/check that table extracted and VO model is aligned with data
+ FILTECAM L2 spec spectra 1D test/check that table extracted and VO model is aligned with data
+ EXES L3 'mrgordspec' spectra 1D test/check that table extracted and VO model is aligned with data

See ticket for more details [here FIREFLY-631](https://jira.ipac.caltech.edu/browse/FIREFLY-631).

Please test the different cases here and confirm:
https://firefly-631.irsakudev.ipac.caltech.edu/applications/sofia/
Thanks!

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/Sofia1DSpectraExtractor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/Sofia1DSpectraExtractor.java
@@ -107,10 +107,13 @@ public class Sofia1DSpectraExtractor extends DataExtractUtil {
         DataType[] dd;
 
         double[][] fdata = (double[][]) ArrayFuncs.convertArray(p.getKernel(), Double.TYPE, true);
+        if(fdata.length < 2) throw new FitsException("Data has less than 2 columns, this is not chartable as 1D spectra. data length is " + fdata.length);
         for (int row = 0; row < fdata[0].length; row++) {
             DataObject aRow = new DataObject(dataGroup);
             dd = dt.toArray(new DataType[dt.size()]);
             for (int dtIdx = 0; dtIdx < fdata.length; dtIdx++) {
+                //if the data has changed without notice, at least we pick up the old model unless there is less data than the model itself but at least 2 column should be present
+                if(dtIdx == dd.length) break;
                 aRow.setDataElement(dd[dtIdx], fdata[dtIdx][row]);
             }
             dataGroup.add(aRow);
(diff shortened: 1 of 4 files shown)


lznakano:
I don't understand why "FIFI-LS types other than L3 wavelength_resampled to be shown as table (binary table types)".  The scientists asked to see the wavelength.  This is the reason the converter was added to convert the FITs to standard WAVE-TAB format.   I still think that all images should be shown as image first.  The table mostly does not make much sense as what Justin put.  The table and chart are just ways to show what the data in the HDU is like.  This is just my two cents.  

FIFI-LS LEVEL4 the analyzer's menu's mark is wrong.   The image is the first HDU, but the mark is at "Image Data:All Images in file".

The mark and the indentation on the rest of instruments look right.

ejoliet:
Example: Level 1 is a binary table:
https://irsadev.ipac.caltech.edu:443/data/SOFIA/FIFI-LS/OC4B/20160304_F285/raw/r9/data/si/Folder1/Data/00112_051250_00051_OrionBar_act3_107_-127_A_sw.fits

LEVEL_2 scan_combined is another binary table:
https://irsadev.ipac.caltech.edu:443/data/SOFIA/FIFI-LS/OC4B/20160304_F285/proc/p4012/data/g18/F0285_FI_IFS_7004087_RED_SCM_100113-100114.fits

Those should be shown without analyzer, the default report should work. No conversion is needed.

ejoliet:
But it appear correctly too in your PR #1008 so all good.

ejoliet:
For FIF-LS LEVEL4, in this case is also correct, we use the default FIREFLY analyzer for cube of images, where the 3rd axis (planes) is a wavelength. default view is show all. There is no problem there or no improvement to be made as far as i understand.

lznakano:
> Example: Level 1 is a binary table:
> https://irsadev.ipac.caltech.edu:443/data/SOFIA/FIFI-LS/OC4B/20160304_F285/raw/r9/data/si/Folder1/Data/00112_051250_00051_OrionBar_act3_107_-127_A_sw.fits
> 
> LEVEL_2 scan_combined is another binary table:
> https://irsadev.ipac.caltech.edu:443/data/SOFIA/FIFI-LS/OC4B/20160304_F285/proc/p4012/data/g18/F0285_FI_IFS_7004087_RED_SCM_100113-100114.fits
> 
> Those should be shown without analyzer, the default report should work. No conversion is needed.

Yes, I did not test FIFI-LS Level1 and 2 because they are not using analyzer.  I tested Level3 and Level4. 
For Level3: it is the wavelength, the image should be shown initially, but I don't see it.  
For Level4, the first HDU is shown but the title above it is marked.   

Those are the two issues I saw.  Otherwise, all are good.

ejoliet:
LEVEL 3 are the one which has wavelength lookup table. Level 4 is something else, it's a cube of images, each image correspond to a wavelength observed - that is ok as UI is showing now.
That's why i added a check in the analyzer to filter only the one you converting to use wave-tab code for level 3. The others are something else.

PR #1010, opened 2020-08-18
firefly-632:More issues found after fixing FIREFLY-629 and 631

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-632

 This PR fixed some of the issues reported in the ticket above.  

NOTE the following two issues:
GRID:  FF can not add grid for the image without projection, ie. the mapType = unrecognized or unspecified.  I spent quit some effort on this issue, but I did not find a better solution.  
GREAT LEVEL-1: Some of the GREAT LEVEL-1 data has 'Table Load" error. The cause is the FF does not support the FITs that has more than one naxisi>1 where i=1,..naxis.  The failing FITs files have both naxis1>1 and naxis2>1.

Test URL:
https://lijun.irsakudev.ipac.caltech.edu/applications/sofia/
https://lijun1.irsakudev.ipac.caltech.edu/applications/sofia/

To test:
1. allSky
2. Select instrument: GREAT, LEVEL_1

In the first URL, the fourth row will give a "table load" error.
In the second URL, the bug was fixed, it just shows the image.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
@@ -41,58 +41,67 @@ public class SofiaAnalyzer implements DataProductAnalyzer {
                                           Map<String, String> params,
                                           Header[] headerAry) {
 
-
         String code = params.getOrDefault("product_type", "");
         String level = params.getOrDefault("processing_level", "").toUpperCase();
         String inst = params.getOrDefault("instrument", "").toUpperCase();
 
-        //
-        // Add spectra extracted data from image
-        //
-        boolean isSpectra = false;
-        if (inst.equals("FORCAST")) {
-            if (level.equals("LEVEL_2") && (code.equals("rspspec") || code.equals("mrgspec"))) { //params.containsKey("replace") && Boolean.parseBoolean(params.get("replace"))) {
-                isSpectra = true;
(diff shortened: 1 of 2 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -78,21 +78,36 @@ public FileAnalysisReport analyzeFits(FileAnalysisReport inputReport,
...
+              return setReportDisplay(inputReport, FileAnalysisReport.ChartTableDefOption.showChart);
             }
         }
 
         return inputReport;
     }
 
+    private FileAnalysisReport setReportDisplay(FileAnalysisReport report,FileAnalysisReport.ChartTableDefOption option ){
+
+        FileAnalysisReport retReport =  report.copy();
```
ejoliet, line 103: I'm not sure that you even need to copy the report to set the parts with the default showing.

lznakano: I saw you used the copy and just kept the same manner.  I don't think that we need to make copy any places at all.  If you are OK, I will remove all the copy and update the input directly.  Again, it is your call.

ejoliet: Ok, please,  if the copy is not needed, we should remove it, and test everything again before merging. Thank you.

lznakano: > Ok, please, if the copy is not needed, we should remove it, and test everything again before merging. Thank you.

I did quite some clean up.   Removed the unnecessary copy's and re-arranged the codes.  I tested and seemed all OK.

PR #1014, opened 2020-08-27
 firefly-279:Update 3-color image labeled incorrectly

lznakano:
https://jira.ipac.caltech.edu/browse/FIREFLY-279

change the 3-color image's label

Test:
URL: https://lijun.irsakudev.ipac.caltech.edu/irsaviewer/

### DIFF
--- a/src/firefly/js/visualize/reducer/PlotTitle.js
+++ b/src/firefly/js/visualize/reducer/PlotTitle.js
@@ -5,7 +5,6 @@
 import {TitleOptions} from '../WebPlotRequest.js';
 import {RequestType} from '../RequestType.js';
 import {PlotAttribute} from '../PlotAttribute';
-
 /**
  *
  * @param {WebPlot} plot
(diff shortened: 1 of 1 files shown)


lznakano:
> I tested with the image search and it worked.
> 
> I tried with uploading my own image but it didn't. It works in irsadev though.
> 
> When trying to upload and search with my own image, I'm seeing also error log in the console if that helps:
> `Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'd.getTitle().split')`
> 
> Could you take a look?
> Thanks!

According to your request, the label is survey 3-color.  From what you described that the uploaded image does not have the survey information.  In this case, how do you want to label the image?  Please put your request in the ticket, I will do accordingly.  
Thanks!

ejoliet:
Thanks. I've added a comment in the ticket because i thinkLuisa didn't know about those other options, and she should confirmed my suggestions on how to treat the label in those cases.

What do you think about those suggestions?

For file upload , i would suggest to see the 'filename' - 3-color.
For example, my file name is ARP220-xyz.fits. I would suggest to see 'ARP220-xyz 3-color'. Or just:
'Upload 3-color' might be easier and more friendly?
 
For URL option, i would say the same: 'file name 3-color' or just 'URL 3-color'?

ejoliet (approved):
Looks good! thank you!

PR #1015, opened 2020-09-03
FIREFLY-574: Predefined cell renderers

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-574

From ticket:
By default the table renders values into the cells based a variety of factors, including data type, format, and other column attributes.

This ticket to be design a way to customize the renderer for more complex cases.  It should includes these features:
- Data can be pulled from other columns within the table
- Allow default style to be overridden as needed.
- Define the renderer declaratively
- Setup renderer via table's META-INFO
- Generic syntax so that it can be applied to future renderers
- Programmatically by creating the renderer and passing it into the table as props
- Create several predefined cell renderers as a proof of concept, include one where it uses three numbers: base, upper bound, and lower bound

Test Url: https://fireflydev.ipac.caltech.edu/firefly-574-cell-renderers/firefly/test/tests-table.html#test-7
Test #7
- First column is a NumberRange renderer where base value is from itself, upper bound is from column `band` and lower is a static 1.  It also override default style to make upper green and lower red.
- Second column is also a NumberRange where base value is from itself but with only one bound specified. 
- Third column is an image renderer with the ability to add before/after description
- Last is a link renderer where href can be composed from data in other columns.

Additional documentation: https://fireflydev.ipac.caltech.edu/firefly-574-cell-renderers/firefly/docs/js/firefly.ui.table.renderers.html


ejoliet (approved):
Great! This is a big deal. I think it'll answer most of the use cases i heard from CCB.

The fact that we are able to combine values and rendered cell more complex will give more difficult behavior that will need to be covered or make sure is not a dead end.

I'm just going in not any particular order about things i've noticed:

- Filter: when i filter +/- band 'like %5%' i was expecting some rows with value 5 but it returns empty. I don't know if it should but just mention it because filter will be a matter of attention.
- Can we extend the render from +/- to almost any expression, meaning with any unicode like Å (angstrom) - in html `&Aring;` or '~' or degree 'º' or inf '∞'?
- is type `i` image coming from VO standards? How it will be converted in VOTable?
- What you see and what you download will be clearly different. The dowloaded data is actually the original data. The cell renderer is a display but is there a way to change it or turn it off from the client?

Something i noticed, i'm not sure if it's part of the test or this PR but i found very nice that the filter for 'band as image' is actually giving me the freedom of full filter but also shoing me the different limited values!

I would say it looks actually very flexible enough for now but we should open the discussion with CCB also to get their feedback. 

Thank you!

robyww (approved):
I like it. After looking at the test it is pretty clean how to set it up. The changes look smaller than I expected so I am optimistic they we can add more new renderers without too much effort.

The only problem that I see is: 
- NextSci will have the original columns hidden and define a new dynamic column for display. This approach choses and existing column and renders it with with other columns.

ejoliet:
> define a new dynamic column for display.
This is the concept of 'synthetic' column and it might relevant also for SEDs/Spectras for IRSA/NED. For example, adding a new column that actually is a result of combining other columns in a formula or expression with different units (and type?)

loitly:
> * NextSci will have the original columns hidden and define a new dynamic column for display. This approach choses and existing column and renders it with with other columns.

As you can see from the examples, we have the ability to decorate a column as we feel fit, i.e. change label, change rendering, enable/disable sorting or filtering, show/hide, etc.

Using a combination of the above features, you should be able to present a table that's very similar to the one they have now.
What is specific to Firefly and cannot easily change is the sorting and filtering.  
- Filter when applied at the column header will always applies to that column's internal data.
- Sorting is the same way.  It sorts based on the column's internal data, not what is shown on the screen.

So, without any additional changes, I would recommend to disabling filter and sort if the rendered value no longer represent the internal data.

PR #1017, opened 2020-09-17
Firefly 656 bitpix 64 support

ejoliet:
While testing Bolocam: Galactic Plane Survey (GPS) dataset, we found that Firefly was displaying NaN flux value without units in the readout.

One of the discussion was about unit missing but actually what happened here is that the FITS file fetched from the search (Bolocam GPS, band 1.1 mm around 'ic1396' object) has BITPIX = 64 and bunit = Jy/Beam (see also file sample attached in https://jira.ipac.caltech.edu/browse/FIREFLY-651).

I've added the case of BITPIX = 64 in the code where the readout is getting the flux out of the FITS image.

To test, either:
- upload the file sample (in FIREFLY-651) and check that the readout is no longer NaN and missing the unit
https://fireflydev.ipac.caltech.edu/firefly-656-bitpix-64-support/firefly/?__action=layout.showDropDown&view=FileUploadDropDownCmd
or
- do a image search of BOLOCAM GPS 1.1mm around 'ic1396' astronomical object
https://fireflydev.ipac.caltech.edu/firefly-656-bitpix-64-support/firefly/?__action=layout.showDropDown&
or
- use the URL API to display the image ;-)
https://fireflydev.ipac.caltech.edu/firefly-656-bitpix-64-support/firefly/?api=image&url=https://irsa.ipac.caltech.edu:443/data/BOLOCAM_GPS/images/v1/IC1396/v1.0.2_ic1396_13pca_map50.fits

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/visualize/plot/PixelValue.java
+++ b/src/firefly/java/edu/caltech/ipac/visualize/plot/PixelValue.java
@@ -243,6 +243,13 @@ public class PixelValue {
 		/*if (Double.isNaN(retval))
 		    throw new PixelValueException("No flux available");*/
 		break;
+		case 64:
+		bytes_per_pixel = 8;
+		file_pointer = data_offset + pixel_offset * bytes_per_pixel;
+		fits_file.seek(file_pointer);
+		file_value = fits_file.readLong();
+		retval = file_value * bscale + bzero;
+		break;
 	    default:
 		retval = Double.NaN;
 		break;
--- a/src/firefly/js/visualize/saga/MouseReadoutWatch.js
+++ b/src/firefly/js/visualize/saga/MouseReadoutWatch.js
@@ -21,6 +21,7 @@ import {
     getWaveLengthUnits, hasPixelLevelWLInfo, hasPlaneOnlyWLInfo,
     isImageCube, wavelengthInfoParsedSuccessfully
 } from '../PlotViewUtil';
+import {getNumberHeader, HdrConst} from '../FitsHeaderUtil.js';
 
 
 const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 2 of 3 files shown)


gpdf (changes requested):
I will not be able to do a live review today.

However, based on what I know about Javascript, and the presence of "Double" in the code snippet shown, I suspect there may be a problem.  If the Firefly client-server interface for returning pixel data values to the client is based on returning a `double` value, this will not work for 64-bit integer data.  There are (many) 64-bit integer values that are not representable in a `double`.  

There is no reason to imagine that any particular image will actually have such values, so the live test is in no way sufficient to explore this.  (We could construct a "torture test" artificial image file for use in future testing.)

The code looks very much like it converts the pixel values to `double` so I'm pretty confident this is a real problem.  Workaround could be either a) returning a formatted string or b) returning some other kind of data-preserving encoding.  The experimental client-side FITS processing would also help with this, of course.

ejoliet:
@gpdf yes, you are right, part of the problem which is not what i'm solving, is that the readout client parse the 'flux' value as float. We might need to introduce at least 2 different parsing one for integer, one for double but i will discuss that further with @robyww .
That's why i wanted to do this in dev and not in rc branch because it could be more refactoring/implementation.

robyww (approved):

robyww:
@gpdf @ejoliet We return the flux value from the server as a json string so I can be shown in the readout. I added formatting so there is not decimal place or scientific notation.

ejoliet:
> @gpdf @ejoliet We return the flux value from the server as a json string so it can be shown in the readout. I added formatting so there is no decimal place or scientific notation.
Looks good! 
i've build a version for testers:
https://firefly-656-readout-update.irsakudev.ipac.caltech.edu/irsaviewer/

lrebull:
It seems ok to me, but I did not bang on this very hard. Gregory, you should take a look.

PR #1020, opened 2020-09-22
Firefly-549: Fixed Crop and context toolbar not going away.

robyww:
Fixes firefly-549

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-549

Best test is to look at @lrebull last comment on the ticket.  Before the context toolbar was staying up after the crop.

_build:_ https://fireflydev.ipac.caltech.edu/firefly-549-toolbar-staying-up/firefly/

### DIFF
--- a/src/firefly/js/visualize/reducer/HandlePlotChange.js
+++ b/src/firefly/js/visualize/reducer/HandlePlotChange.js
@@ -168,7 +168,7 @@ function changePlotAttribute(state,action) {
     const plot= primePlot(pv);
     if (!plot) return state;
     const newAtts= {...changes};
-    if (attKey && attValue) newAtts[attKey]= attValue;
+    if (attKey) newAtts[attKey]= attValue;
 
     if (positionScope) {
         plotViewAry= applyToOnePvOrAll(state.positionLock, plotViewAry, plotId, false,
--- a/src/firefly/js/visualize/task/PlotChangeTask.js
+++ b/src/firefly/js/visualize/task/PlotChangeTask.js
@@ -10,6 +10,7 @@ import {WebPlotResult} from '../WebPlotResult.js';
 import {WebPlot} from '../WebPlot.js';
 import {makeCubeCtxAry, populateFromHeader} from './PlotImageTask';
 import {locateOtherIfMatched} from './WcsMatchTask';
+import {PlotAttribute} from '../PlotAttribute.js';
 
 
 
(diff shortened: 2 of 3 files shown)


lrebull:
Yes, confirm that this is looking better to me now!

ejoliet:
It fixes the alignment problem but @lrebull are you ok that the toolbar is gone after the cropping? 
How do you get back the toolbar if you want keep interacting with it? Maybe we can delay this improvement for later but the solution here changes the previous behavior of the toolbar after cropping, just wanted you to be aware.

ejoliet (approved):
Although the alignment is now fixed, it changes the behavior of the toolbar after cropping. I'm ok with that but we probably need to circulate back this change so we can either agree on keeping it that way or improve it later.

lrebull:
So the appearance/disappearance of this toolbar is ..complicated, and has been changed several times in this release. There are questions about its behavior in the CCB queue. One of the changes is that when you select a region in one image, the toolbar no longer appears in any other image, because it has no meaning over in the other images. That's at least one possible asymmetry that prompted this particular ticket.  After you crop, though, there isn't anything more to be done with the tools. If you want to do more with the tools, you can pick the region selection tool again. Does this make sense? Or should we rethink (again)?

lrebull:
(CCB ticket is linked to this one iin Jira:  FIREFLY-515 )

ejoliet:
It make sense.
I’m ok with re selection but bare in mind you could still get stats out of the selected area without reselecting it - statistics table could be also another round of improvements but this is another topic.
--
Emmanuel Joliet
Office: 170
Email:  ejoliet@caltech.edu
Phone:  (626) 395-1489

On Sep 24, 2020, at 15:16, Luisa Rebull <notifications@github.com> wrote:

﻿

So the appearance/disappearance of this toolbar is ..complicated, and has been changed several times in this release. There are questions about its behavior in the CCB queue. One of the changes is that when you select a region in one image, the toolbar no longer appears in any other image, because it has no meaning over in the other images. That's at least one possible asymmetry that prompted this particular ticket. After you crop, though, there isn't anything more to be done with the tools. If you want to do more with the tools, you can pick the region selection tool again. Does this make sense? Or should we rethink (again)?

robyww:
This is not a change in firefly behavior. It was broken in dev. The toolbar has always be removed after cropping since there is no active selection.  the toolbar is not made to be up if there is not selection. We can consider future changes but this is only a fix.

robyww:
We definitely cannot release with out this ticket being merge, since leaving the toolbar up would create multiple bugs.

PR #1022, opened 2020-09-24
FIREFLY-661: replace DataLine generic column name with title HDU

ejoliet:
[FIREFLY-661](https://jira.ipac.caltech.edu/browse/FIREFLY-661) request is about changing the column name 'DataLine' for 1D image to be matching the title of the HDU or plot title.

Confirm that for GREAT L1, FIFI-Ls L3 and L4 specific product type (details in ticket) you see no more the 'DataLine' column name and the chart axis is also the same as the HDU title.
Test version available here:
https://firefly-661-dataline.irsakudev.ipac.caltech.edu/applications/sofia/
**Note**: I'm not sure what would happened if 1D image doesn't have HDU title.

### DIFF
--- a/src/firefly/js/metaConvert/PartAnalyzer.js
+++ b/src/firefly/js/metaConvert/PartAnalyzer.js
@@ -125,7 +125,7 @@ const C_COL2= ['flux','data','data1','data2'];
  * @param fileFormat
  * @return {{xCol:string,yCol:string,cNames:Array.<String>,cUnits:Array.<String>}|{}}
  */
-function getTableChartColInfo(part, fileFormat) {
+function getTableChartColInfo(title, part, fileFormat) {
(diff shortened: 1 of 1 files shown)


lrebull:
updates in the jira ticket, sorry.

robyww (approved):
Looks good.  My only concern is that the title may be very long. Should we set a max length for the column name? Maybe not, I don't know.

Either way. It is OK to merge.

ejoliet:
>Looks good. My only concern is that the title may be very long. Should we set a max length for the column name? Maybe not, I don't know.

yes, It could be but i would argue that is usually short and usual common names, like flux, wavelength, etc. Of course it can be mission_isntrument_ultimate_super_calibration_flux ;-) or something like that but we can see this for future mission, it is not the case for SOFIA dataset currently so we are good for now.

PR #1023, opened 2020-09-25
IRSA-2986-3690-3675: Email related issues

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-3690
https://jira.ipac.caltech.edu/browse/IRSA-2986
https://jira.ipac.caltech.edu/browse/IRSA-3675

IRSA-2986: Email notification does not respond to click (TimeSeries)
IRSA-3690: SOFIA download email not sent
IRSA-3675: finderchart download email not working
- convert logError in WebUtil.js to use logger
- fix 'Get Download Script' when there are multiple parts not working
- fix TimeSeries help_id prop error showin in console

Test URLs:
https://irsa-2986-email-notification.irsakudev.ipac.caltech.edu/irsaviewer/timeseries
https://irsa-2986-email-notification.irsakudev.ipac.caltech.edu/applications/finderchart/
https://irsa-2986-email-notification.irsakudev.ipac.caltech.edu/applications/sofia/

As described in tickets, 'Enable email notification' box in Download Dialog as well as Background Monitor does not response to click.  Once clicked , it was not sending emails when completed.  When there were multiple parts, 'Get Download Script' was not working.  These should all be fixed in these applications.  Please confirm.

### DIFF
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -6,6 +6,8 @@ import { forEach, fromPairs, get, has, isArray, isBoolean, isEqual, isFunction, 
 import slug from 'slug';
 import shallowequal from 'shallowequal';
 import update from 'immutability-helper';
+import {logger} from './Logger.js';
+
 
 
 const MEG          = 1048576;
@@ -279,7 +281,9 @@ export function encodeParams(params) {
  */
 export const encodeServerUrl= (url, params) => encodeUrl(url, params);
 
-export const logError= (...message) => message && message.forEach( (m) => console.log(has(m,'stack') ? m.stack : m) );
+export const logError= (...message) => {
+    message?.forEach?.( (m) => logger.error(m));
+};
 
 /**
  * Copy the content of the string to the clipboard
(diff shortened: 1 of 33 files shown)


src/firefly/js/util/WebUtil.js
```diff
@@ -6,6 +6,8 @@ import { forEach, fromPairs, get, has, isArray, isBoolean, isEqual, isFunction,
 import slug from 'slug';
 import shallowequal from 'shallowequal';
 import update from 'immutability-helper';
+import {logger} from './Logger.js';
+
```
robyww, line 10: This is just a quick comment on this PR. (I will look at the PR closer on Monday.)

It would be better, because of future workers, to keep the dependencies of `WebUtil.js` as small as possible. Bringing in `Logger.js` then brings in `moment` with makes the worker much bigger. It would be better to move `logError` or deprecate it and make another version in `Logger.js` than to import logger into `WebUtil.js`

loitly: I agree.  `WebUtil.logError` is not needed since that functionality is already available in Logger.js.  I would remove logError from WebUtil and replace it with Logger where it was used.  It would involved touching a few more files, though.

ejoliet (commented):
FinderChart and SOFIA tested and working great!!
I couldn't test **TimeSeries** because the checkbox state doesn't change. (Safari 14.0)
(maybe just me, waiting for other reviewers to confirm)

loitly:
> I couldn't test **TimeSeries** because the checkbox state doesn't change. (Safari 14.0)
You are right.  TimeSeries fail to build and I didn't noticed it.  I rebuilt it and is now working as expected.  Please retest.

ejoliet (approved):
yes, looks good! Thanks!

deannaji (approved):

deannaji:
Thanks!

PR #1024, opened 2020-09-28
irsa-3276:FIFI-LS level 3 file contents menu

lznakano:
https://jira.ipac.caltech.edu/browse/IRSA-3726

 This ticket is about the consistency of file content menu among instruments.  The FIFI-LS LEVEL4 has "All Image In File" in the file content but FIFI-LS LEVEL_3 does not.  The analysis is logged in the ticket above.  
 
This PR removed the dependency on added analyzer with convertedFileName set.  The WCS TAB's table column name is changed to wavelength instead of coords. 

To test:
 URL: https://lijun-irsa-3276.irsakudev.ipac.caltech.edu/applications/sofia/

### DIFF
--- a/src/firefly/js/metaConvert/MultiProductFileAnalyzer.js
+++ b/src/firefly/js/metaConvert/MultiProductFileAnalyzer.js
@@ -324,7 +324,7 @@ function processAnalysisResult({table, row, request, activateParams, serverCache
     const imageParts= partAnalysis.filter( (pa) => pa.imageResult);
     let makeAllImageOption= !disableAllImageOption;
     if (makeAllImageOption) makeAllImageOption= imageParts.length>1 || (imageParts.length===1 && parts.length===1);
-    if (makeAllImageOption) makeAllImageOption= imageParts.every( (ip) => !ip.imageResult.override);
+    if (makeAllImageOption) makeAllImageOption= !(imageParts.every( (ip) => ip.imageResult.override));
     const duoImageTableParts= partAnalysis.filter( (pa) => pa.imageResult && pa.tableResult);
     if (makeAllImageOption && duoImageTableParts.length===1 && imageParts===1) makeAllImageOption= false;
     const useImagesFromPartAnalysis= parts.length>1 || !makeAllImageOption || duoImageTableParts.length;
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/SofiaFitsConverterUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/sofia/SofiaFitsConverterUtil.java
@@ -53,7 +53,7 @@ public class SofiaFitsConverterUtil {
         //PS3_0 is the EXTNAME in the BinaryTableHDU
         header.addLine(new HeaderCard( "PS3_0", waveTableName, "WAVE-TAB binary table name (added by Sofia)"));
         //PS3_1 is the column name for the coordinate array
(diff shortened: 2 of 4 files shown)


ejoliet (approved):
I tested what IRSA has been requested in the ticket and is fixed.

Thank you!

src/firefly/js/metaConvert/MultiProductFileAnalyzer.js
```diff
@@ -324,7 +324,6 @@ function processAnalysisResult({table, row, request, activateParams, serverCache
     const imageParts= partAnalysis.filter( (pa) => pa.imageResult);
     let makeAllImageOption= !disableAllImageOption;
     if (makeAllImageOption) makeAllImageOption= imageParts.length>1 || (imageParts.length===1 && parts.length===1);
-    if (makeAllImageOption) makeAllImageOption= imageParts.every( (ip) => !ip.imageResult.override);
```
robyww, line 327: I agree there is a bug here but I don't think this is the right fix. The line is meant to only enable the "all images" options if there are some parts that have not been replaced with an alternate image. The line above wrongly disables the all images option if _any_ of the parts have been replaced.

The "all images" option should only be disabled if _all_ the parts have been replace with an alternate image.
You will have to test it, but I think the following is a better fix than removing the line.

```js
if (makeAllImageOption) makeAllImageOption= !(imageParts.every( (ip) => ip.imageResult.override));
```

robyww (changes requested):
@lznakano Your right, there is a bug on line 327, but I don't think removing the line is the best fix. I had to study it to remember why it was there but I think I understand now.  See by comment below for a different fix to try.

src/firefly/js/metaConvert/MultiProductFileAnalyzer.js
lznakano: > I agree there is a bug here but I don't think this is the right fix. The line is meant to only enable the "all images" options if there are some parts that have not been replaced with an alternate image. The line above wrongly disables the all images option if _any_ of the parts have been replaced.
> 
> The "all images" option should only be disabled if _all_ the parts have been replace with an alternate image.
> You will have to test it, but I think the following is a better fix than removing the line.
> 
> ```js
> if (makeAllImageOption) makeAllImageOption= !(imageParts.every( (ip) => ip.imageResult.override));
> ```

Yes, I understand that.  That was the question I was trying to ask in the meeting. @ejoliet said that as long as it is a multi-extension FITs, we just show all image option.  In this case, we no longer care if any part is replaced.

robyww: This checks for case that we have not encountered. That is the case when we read a multi-extension FITS file and the replace very image with an alternate.

lznakano: In this case, your proposed change will work better.  Only all the parts are replaced, the makeAllImageOption=false.  I will use this change then.

robyww: thanks.

robyww (approved):

lznakano:
@robyww 
I made minor changes in three files.  To show the "All Image in File" option, it has to meet two conditions: one is that the FITs file has multi-extension, and two is that not all the extension are converted.   Thus, the each analyzer should not set the part's convertedFileName for the part that is changed.  With all changes I did, it worked fine and also preserved the original design.  Please take a look and let me know.

robyww:
Looks good.

PR #1025, opened 2020-09-28
irsa-3721:original HDU view names

lznakano:
https://jira.ipac.caltech.edu/browse/IRSA-3721

This PR is changed the file menu, chart and table titles as requested. 

Test URL
 https://lijun-irsa-3721.irsakudev.ipac.caltech.edu/applications/sofia/

### DIFF
--- a/src/firefly/js/metaConvert/converterUtils.js
+++ b/src/firefly/js/metaConvert/converterUtils.js
@@ -26,7 +26,8 @@ const getSetInSrByRow= (table,sr,rowNum) => (col) => {
 
 
 export function createTableActivate(source, titleStr, activateParams, tbl_index=0) {
-    return createChartTableActivate(false, source, titleStr, activateParams,undefined,tbl_index);
+
+    return createChartTableActivate(false, source, {titleStr, showChartTitle:true}, activateParams,undefined,tbl_index);
 }
 
 const makeCommaSeparated= (strAry) => strAry.reduce( (str,d) => str? `${str},${d}` : d,'');
@@ -50,7 +51,7 @@ const makeCommaSeparated= (strAry) => strAry.reduce( (str,d) => str? `${str},${d
  * Activate a chart and table
  * @param {boolean} chartAndTable - true - both char and table, false - table only
  * @param {String} source
- * @param {String} titleStr - the title of the table or chart
+ * @param {titleInfo} an object that has a titleStr and showchartTile properties
  * @param {ActivateParams} activateParams
  * @param {ChartInfo} chartInfo
  * @param {Number} tbl_index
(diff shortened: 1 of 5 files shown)


ejoliet (changes requested):
I started to test but we probably need to be careful here because i see already some impact on the change based on @lrebull request.
GREAT LEVEL1, when selecting table view, i see for example name such as `${cache-dir}/DataProduct-12...
`
Same for FORCAST, LEVEL_2, 'combspec'. Or FLITECAM.

Pretty much when there is an image and data interpreted and the name of the HDU is not set anymore and the table view is used.

I don't think we want to see this, can we have 'table-1' where 1 is the HDU number? Or something like that? @lrebull could help about this.

That's why i think the used a default generic name in those cases.

Let me know if we need to discuss further with IRSA folks.
Thanks!!!

ejoliet (changes requested):
Thanks!

I tested it and everything looks what Luisa requested apart from GREAT level 1 where i see ‘undefined’ as a column instead of Dataline or Flux or HDU name. And still see ${data} as table name.

src/firefly/js/metaConvert/converterUtils.js
```diff
@@ -183,10 +183,17 @@ function makeChartObj(chartInfo, activateParams, titleStr, chartId, tbl_id ) {
     const {chartParamsAry}= chartInfo;
     const {xAxis, yAxis}= chartInfo;
 
+    /*The default FileAnalysis does not set the chartParam.  Only the added chart has this parameter set.
+     Thus, if the chartParamsAry is set, the "table_" is the name given by the added analyzer.  Only
+     when chartParamsAry is not defined and the title starts with "table_", this is the table in the HDU that has
+     no extension name.
+     */
+    const chartTitle= !chartParamsAry && titleStr.startsWith('table_')? '':titleStr;
```
robyww, line 191: This is a little confusing.  I think it would be simpler if you just put it in the `else` a fiew lines below. Then it could just be
```js
const chartTitle= titleStr.startsWith('table_')? '':titleStr;
```
I have a mild concern that checking for `'table_'` is to too special. It make this function have to have special knowledge about the contents of `titleStr`. This precludes us from using it in this case but that is probably not that big of deal.

Effectively the jsdoc for this param is: "ignore titleStr if it starts with `table_`"

robyww (approved):
I am fine with you merging this PR but I really don't like the solution in `converterUtils.js`. I am glad you documented though.

src/firefly/js/metaConvert/converterUtils.js
lznakano: > This is a little confusing. I think it would be simpler if you just put it in the `else` a fiew lines below. Then it could just be
> 
> ```js
> const chartTitle= titleStr.startsWith('table_')? '':titleStr;
> ```
> 
> I have a mild concern that checking for `'table_'` is to too special. It make this function have to have special knowledge about the contents of `titleStr`. This precludes us from using it in this case but that is probably not that big of deal.
> 
> Effectively the jsdoc for this param is: "ignore titleStr if it starts with `table_`"

I had the same concern as you.  I wanted to minimize the changes.  So checking the string is the least change.  I now introduce an object that has two properties, title and showChartTitle.   The change is still very minimal.

robyww: OK- it is fine to merge.

lznakano: Thank you @robyww !!

ejoliet:
You might want to double check if it’s not best to do the PR against rc-2020.3 to avoid future conflict with changes pushed directly in dev.
What do you think? @lznakano @robyww

PR #1031, opened 2020-10-14
Firefly 667 Firefly related changes needed for irsakubernetes deployment

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-667

- Firefly 'dev' and 'gwt' need to be modified so that it take header ''X-Forwarded-*' header' when available.  This is because when Firefly is behind a reverse proxy, the proxied host information is stored in these headers.
- Add health check endpoint to all Firefly-based applications.  The check will perform a simple operation, like allocating a small amount of memory to ensure the JVM is healthy.  It will be mapped to `/healthz`.
- Fixed error messages coming from the catalog dialog.

I do not have a test build.  Please review code to ensure it does not break existing functionality.  More testing to come when IRSA UI apps are served by irsakudev.

### DIFF
--- a/config/web.xml
+++ b/config/web.xml
@@ -71,6 +71,15 @@
     </servlet-mapping>
 
     <servlet>
+        <servlet-class>edu.caltech.ipac.firefly.server.servlets.HealthCheck</servlet-class>
+        <servlet-name>healthz</servlet-name>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>healthz</servlet-name>
+        <url-pattern>/healthz</url-pattern>
+    </servlet-mapping>
+
+    <servlet>
         <servlet-class>edu.caltech.ipac.firefly.server.servlets.CommandService</servlet-class>
         <servlet-name>FireFly CommandService</servlet-name>
     </servlet>
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
@@ -24,13 +24,19 @@ import javax.servlet.http.HttpServletResponse;
 public class HealthCheck extends BaseHttpServlet {
 
     protected void processRequest(HttpServletRequest req, HttpServletResponse res) throws Exception {
-
-        // memory check
-        int mem = Math.max(Math.min(StringUtils.getInt(req.getParameter("mem"), 5), 1024), 1);
-        byte[] b = new byte[mem * 1024 * 1024];
-        Runtime rt= Runtime.getRuntime();
-        Logger.debug(String.format("HealthCheck(Used/Max/Allocated): %s/%s/%s",
-                            FileUtil.getSizeAsString(rt.totalMemory() - rt.freeMemory()),
-                            FileUtil.getSizeAsString(rt.maxMemory()), FileUtil.getSizeAsString(b.length)));
+        byte[] alloc = null;
+        try {
+            // memory check
+            int mem = Math.max(Math.min(StringUtils.getInt(req.getParameter("mem"), 5), 1024), 1);
+            alloc = new byte[mem * 1024 * 1024];
+        } catch (OutOfMemoryError oome) {
+            Logger.error(oome, "Encountered OutOfMemory during memory check");
+            throw oome;
+        } finally {
+            Runtime rt= Runtime.getRuntime();
+            Logger.debug(String.format("HealthCheck(Used/Max/Allocated): %s/%s/%s",
+                    FileUtil.getSizeAsString(rt.totalMemory() - rt.freeMemory()),
+                    FileUtil.getSizeAsString(rt.maxMemory()), alloc == null ? "null" : FileUtil.getSizeAsString(alloc.length)));
+        }
     }
 }
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
@@ -1,0 +1,36 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+package edu.caltech.ipac.firefly.server.servlets;
+
+import edu.caltech.ipac.firefly.server.util.Logger;
+import edu.caltech.ipac.util.FileUtil;
+import edu.caltech.ipac.util.StringUtils;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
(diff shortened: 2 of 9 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+    protected void processRequest(HttpServletRequest req, HttpServletResponse res) throws Exception {
+
+        // memory check
+        int mem = Math.max(Math.min(StringUtils.getInt(req.getParameter("mem"), 5), 1024), 1);
+        byte[] b = new byte[mem * 1024 * 1024];
+        Runtime rt= Runtime.getRuntime();
+        Logger.debug(String.format("HealthCheck(Used/Max/Allocated): %s/%s/%s",
+                            FileUtil.getSizeAsString(rt.totalMemory() - rt.freeMemory()),
+                            FileUtil.getSizeAsString(rt.maxMemory()), FileUtil.getSizeAsString(b.length)));
+    }
```
robyww, line 35: You might want to catch an `OutOfMemoryError` here. and log that as well.

robyww: I am guessing this is called outside of firefly app context. I would think it should return a string such a `Health Check Passed` or something.

You could also return a `Health Check Failed` for an `OutOfMemoryError` but I doubt it would return in the case. I have enchanted  edge case when we do get an`OutOfMemoryError` right before the server hangs.

config/web.xml
```diff
@@ -70,6 +70,15 @@
         <url-pattern>/admin/status</url-pattern>
     </servlet-mapping>
 
+    <servlet>
+        <servlet-class>edu.caltech.ipac.firefly.server.servlets.HealthCheck</servlet-class>
+        <servlet-name>healthz</servlet-name>
+    </servlet>
+    <servlet-mapping>
+        <servlet-name>healthz</servlet-name>
+        <url-pattern>/healthz</url-pattern>
```
robyww, line 79: why `healthz` and not just `health`?

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
```diff
@@ -43,8 +48,7 @@
 
     protected void processRequest(HttpServletRequest req, HttpServletResponse res) throws Exception {
 
-        boolean showHeaders = Boolean.valueOf(req.getParameter("headers"));
-        String jwt = req.getParameter("jwt");
+        boolean showHeaders = Boolean.parseBoolean(req.getParameter("headers"));
```
robyww, line 51: This appears to be a special view of the server status page. why not just add it to everything else?

robyww (approved):
Looks fine. I have some comments below.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
loitly: This is not meant for interactive use.  It's an HTTP API where a 200 status is pass and everything else is fail.
In the case of OOME or any exception, it will be caught upstream and a 500 will be returned.
This is usually used with a `timeout` threshold.  When it failed to return within a specified time, it is also considered 'unhealthy' or fail.

The `/healthz` check is within an app context.  So, for firefy, it would be `/firefly/healthz`.

config/web.xml
loitly: > It historically comes from Google’s internal practices. They're called "z-pages".

I am following Kubernetes's convention which is from Google.  It is also popular in the kubernetes/docker community.
The 'z' is also to avoid any conflict with services endpoint.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
robyww: If you are logging it is still worth catching the `OOME`

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
loitly: Yes.  I added this a while back.   Displaying headers may take up lots of space that's why I did not add it by default.  Also, I am thinking we can expand on this and have additional parameters as needed in the future, like actions.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
loitly: I can if you like, for logging purposes.  But in most cases, you won't even get an exception.  It will just hang there.

Normally, I don't want to log this.  It's unnecessary noise.  But, `debug` is our lowest log level.  This segue into our need to move most of our logs to 'info' level and set that as the default log level.

ejoliet: the OOME is catch anyway in catalina.out, this is redundant to add it here in my opinion.

ejoliet (approved):
Had a question more about been curious than an issue so feel free to merge. The idea of updating the code to get the header specific values is ok in the code so all good.

src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -406,19 +406,19 @@ export const LinkCell = React.memo(({cellInfo, style, ...rest}) => {
...
             return (
-                <Cell {...CellProps}>
+                <Cell {...{rowIndex, height, width, columnKey}}>
                     {val}
                 </Cell>
             );
         } else {
             return (
-                <Cell {...CellProps}>
+                <Cell {...{rowIndex, height, width, columnKey}}>
```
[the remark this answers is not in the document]
loitly: As you mentioned, CellProps contains more than just these props.  However, `Cell` component is expecting only these and have PropTypes validation on these only.  That's it was giving validation errors.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
loitly: I also think it's redundant, since the code will look like this.  However, I am fine with adding it. 
```
try { 
... 
} catch(oome) {
  log(oome);
  throw oome;
}
```

loitly:
@robyww I added OOME logging to HealthCheck and fixed Description not making into the client `table.resources`.  Please take a look.  If no further request, I will merge this and all related PRs tomorrow.  Thanks for reviewing.

robyww:
Looks good. Fine with me to merge.

PR #1032, opened 2020-10-14
Firefly 667 Firefly related changes needed for irsakubernetes deployment

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-667

Additional changes:  https://github.com/IPAC-SW/irsa-ife/pull/142

- Firefly 'dev' and 'gwt' need to be modified so that it take header ''X-Forwarded-*' header' when available.  This is because when Firefly is behind a reverse proxy, the proxied host information is stored in these headers.
- Add health check endpoint to all Firefly-based applications.  The check will perform a simple operation, like allocating a small amount of memory to ensure the JVM is healthy.  It will be mapped to `/healthz`.

I do not have a test build.  Please review code to ensure it does not break existing functionality.  More testing to come when IRSA UI apps are served by irsakudev.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -970,7 +970,7 @@
       "requires": {
         "anymatch": "^1.3.0",
         "async-each": "^1.0.0",
-        "fsevents": "^1.0.0",
+        "fsevents": "~2.1",
         "glob-parent": "^2.0.0",
         "inherits": "^2.0.1",
         "is-binary-path": "^1.0.0",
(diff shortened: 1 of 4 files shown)


robyww (approved):
Looks good. I noticed that you did not add a `web.xml` entry for the health check.

loitly:
> Looks good. I noticed that you did not add a `web.xml` entry for the health check.

You are right.  It's added in `irsa-ife`.  I forgot to create a PR for that.  Thanks for the reminder.
https://github.com/IPAC-SW/irsa-ife/pull/142

PR #1035, opened 2020-10-20
FIREFLY-674: Add custom Online-Help feature to API

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-674

Allow API user to set custom location of Online-Help.  To test, goto `Test` URL below and scroll down to test #5.
Click on the Help icon(?) at upper right side of the table.  Click on `source: show` to see how it's done.

Test: https://fireflydev.ipac.caltech.edu/firefly-674-help-base-url-api/firefly/test/tests-main.html#test-5


ejoliet (approved):
I tested it and works as described.
The code change required in the HTML code is minimum.
That's pretty much what Firefly-based HTML pages would need to overwrite the default base URL to a specific one. 

Thanks!

PR #1037, opened 2020-10-27
Irsa-3771: coverage bug

robyww:
Irsa-3771: coverage bug

- occures when the tables indicates that is support coverage but coverage cannot be computed
- happening in sofia when some of the corner columns were null
- the fix is general

fixes: https://jira.ipac.caltech.edu/browse/IRSA-3771

To test with sofia: 
- https://irsa-3771-coverage-bug.irsakudev.ipac.caltech.edu/applications/sofia/
- serach m42, 1 degress, EXES only

### DIFF
--- a/src/firefly/js/core/ComponentCntlr.js
+++ b/src/firefly/js/core/ComponentCntlr.js
@@ -63,8 +63,8 @@ export function getDialogOwner(dialogKey) {
     return (d.visible && d.ownerId) ? d.ownerId : null;
 }
 
-export function getComponentState(componentKey) {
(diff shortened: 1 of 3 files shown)


ejoliet:
@robyww i think the fix is a change between OPS. We use to have a symbol overlay in the coverage image indicating the coordinates from the table or the target coordinates at least, even without 4 corners, see use case in ops. 

I think IRSA will want to keep that but we should talk to them if we need to change it.

ejoliet (approved):
Fix looks working well! Thanks!

PR #1038, opened 2020-10-29
Simplify Firefly dockerfile

loitly:
- remove unused variables
- predefine Firefly runtime directories, making it easier for user to mount into
- move statistics directory into logs/statistics
- VISUALIZE_FITS_SEARCH_PATH now defaults to blank.  Setting it will append to the '/external' set by default.

### DIFF
--- a/docker/base/launchTomcat.sh
+++ b/docker/base/launchTomcat.sh
@@ -20,31 +20,38 @@ echo "!!============================================================"
 echo "!!============================================================"
 echo -e "!!============================================================\n\n"
 
-echo "========== Information:  you can set properties using -e on docker run line =====  "
+echo "========== Information:  you can set environment varible using -e on docker run line =====  "
 echo 
-echo "Properties: "
-echo "        Description                  Property                 Value"
-echo "        -----------                  --------                 -----"
-echo "        Min JVM size (*)             MIN_JVM_SIZE            ${MIN_JVM_SIZE}${MM_NOT_USED}"
-echo "        Max JVM size (*)             MAX_JVM_SIZE            ${MAX_JVM_SIZE}${MM_NOT_USED}"
-echo "        Initial JVM size (**)        INIT_RAM_PERCENT        ${INIT_RAM_PERCENT}${PER_NOT_USED}"
-echo "        Max JVM size (**)            MAX_RAM_PERCENT         ${MAX_RAM_PERCENT}${PER_NOT_USED}"
-echo "        CPU cores (0 means guess)    JVM_CORES               ${JVM_CORES}"
-echo "        Admin username               ADMIN_USER              ${ADMIN_USER}"
-echo "        Admin password               ADMIN_PASSWORD          ${ADMIN_PASSWORD}"
-echo "        Run tomcat with debug        DEBUG                   ${DEBUG}"
-echo "        Extra firefly properties     FIREFLY_OPTS            ${FIREFLY_OPTS}"
-echo "        Shared work Area             FIREFLY_SHARED_WORK_DIR ${FIREFLY_SHARED_WORK_DIR}"
-echo "(*)  If MAX_JVM_SIZE is blank, autosizing properties INIT_RAM_PERCENT and MAX_RAM_PERCENT are used instead"
-echo "(**) Autosizing properties INIT_RAM_PERCENT and MAX_RAM_PERCENT are not used if MAX_JVM_SIZE is set"
+echo "Using Environment Variables: \n"
+echo "        Description                  Name                       Value"
+echo "        -----------                  --------                   -----"
+echo "        Admin username               ADMIN_USER                 ${ADMIN_USER}"
+echo "        Admin password               ADMIN_PASSWORD             ${ADMIN_PASSWORD}"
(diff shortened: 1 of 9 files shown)


docker/base/launchTomcat.sh
```diff
@@ -20,31 +20,38 @@ echo "!!============================================================"
...
 echo "Volume Mount Points: "
-echo "        Log directory : ${CATALINA_HOME}/logs : Directory for logs files"
-echo "        Local Images  : /local/data : Root directory for images and tables that firefly can read"
+echo "    /firefly/config           : used to override application properties"
+echo "    /firefly/workarea         : work area for temporary files"
+echo "    /firefly/shared-workarea  : work area for files that are shared between multiple instances of the application"
+echo "    /firefly/logs             : logs directory"
+echo "    /firefly/logs/statistics  : directory for statistics logs"
+echo "    /firefly/alerts           : alerts monitor will watch this directory for application alerts"
+echo "    /external                 : default external data directory visible to Firefly"
```
robyww, line 54: maybe we want to break these into two parts as well, common and advanced.
`logs`, `shared-workarea`, `/external` as common and the others as advanced.

docker/base/launchTomcat.sh
```diff
@@ -67,31 +74,35 @@ if [ "$1" = "--help" ] || [ "$1" = "-help" ] || [ "$1" = "-h" ]; then
...
    JVM_SIZING="-XX:InitialRAMPercentage=${INIT_RAM_PERCENT} -XX:MaxRAMPercentage=${MAX_RAM_PERCENT}"
 else
    JVM_SIZING="-Xms${MIN_JVM_SIZE} -Xmx${MAX_JVM_SIZE}"
 fi
 
+if [ -z ${VISUALIZE_FITS_SEARCH_PATH} ]; then
+   VIS_PATH="/external"
+else
+   VIS_PATH="/external:${VISUALIZE_FITS_SEARCH_PATH}"
+fi
```
robyww, line 87: good, I think this idea

docker/base/launchTomcat.sh
```diff
@@ -67,31 +74,35 @@ if [ "$1" = "--help" ] || [ "$1" = "-help" ] || [ "$1" = "-h" ]; then
...
+  ${JVM_SIZING} \
+  -Dhost.name=${HOSTNAME} \
+  -Dserver.cores=${JVM_CORES} \
+  -Djava.net.preferIPv4Stack=true \
+  -Dwork.directory=/firefly/workarea \
+  -Dshared.work.directory=/firefly/shared-workarea \
+  -Dserver_config_dir=/firefly/config \
+  -Dstats.log.dir=/firefly/logs/statistics \
+  -Dalerts.dir=/firefly/alerts \
+  -Dvisualize.fits.search.path=${VIS_PATH} \
```
robyww, line 100: since you made this change in statistics, does this mean that we will not attempt to make a statistics dir if the use just maps logs?

docker/base/start-examples.txt
```diff
@@ -30,18 +30,18 @@ Map a directory for direct file reading:
...
 Production like:
-   > docker run -p 8055:8080 -p 9050:9050 --memory=30g -e "ADMIN_PASSWORD=myPassword" -e -e DEBUG="FALSE" --name productionServer ipac/firefly:latest
+   > docker run -p 8055:8080 -p --memory=30g -e "ADMIN_PASSWORD=myPassword" -e DEBUG="FALSE" --name productionServer ipac/firefly:latest
    - start firefly with - http://localhost:8055/firefly/
 
 Production like, multiple tomcats, using a shared work directory:
-   > docker run -p 8055:8080 -p 9050:9050 --memory=30g -e "ADMIN_PASSWORD=myPassword" \
-             -e "FIREFLY_SHARED_WORK_DIR=/local/shared/work/area" -v /local/shared/work/area --name productionServer ipac/firefly:latest
+   > docker run -p 8055:8080 -p --memory=30g -e "ADMIN_PASSWORD=myPassword" \
+             -v /local/shared/work/area:/firefly/shared-workarea --name productionServer ipac/firefly:latest
```
robyww, line 46: should we mentions starting a redis server and pointing to it here. I know we did not in the pass, but it is clearer now what to do.

robyww (approved):
This all looks good. I want to talk to you before you merge it though.

docker/base/launchTomcat.sh
loitly: If they mount /firefly/logs, then we would create a statistics directory under logs if one does not exists. Permission is not a problem since we expect the process owner to have write access to this directory.
If they want only the stats logs to persistent outside the container, then they can mount /firefly/statistics instead.

docker/base/start-examples.txt
loitly: > Should we mention starting a redis server and pointing to it here? I know we did not in the pass, but it is clearer now what to do.

I don't think so since redis is only needed when you have multiple instances of Firefly running.  These examples are for a single instance.
With multiple instances, you may need to introduce docker-compose or may be swarm.  Both, are out of our scope.

robyww: ok

PR #1041, opened 2020-11-12
Add syntax highlighting to ADQL editor using Prism Live

brianv0:
This PR updates the ADQL editor to be a bit more like a code editor. Specifically, it uses PrismJS and the associated Prism live to add live syntax highlighting. In order to do this, InputAreaFieldView/Connected has been modified to accept both additional CSS classes, via the `additionalClasses` property, and an id via the `idName` property.

Some additional fixes have been done to ensure the textarea will be transparent instead of white, which works because prism live has to effectively work with the textarea and a some additional HTML elements (div/pre) and sync them up to make this work right, and it relies on the textarea being transparent to make it all work.


gpdf:
It looks like the `prism-sql.js` file was committed with (welcome) changes to accommodate ADQL syntax like its geometry functions.  Would it be possible, for clarity in future maintenance, to have the initial commit be the out-of-the-box version of this file (and any others copied from the Prism code base) with our changes shown as a commit on top of that?

loitly:
Brian, please update yarn.lock file by running `yarn install`, then commit it as part of your PR.

brianv0:
I've separated out the commits, with checking in the original prism-sql.js, and updated yarn.lock file.

brianv0:
The last commit is also included because prism-live uses bliss.js (aka blissfuljs) for a few things. I'm sure some of those things could be rewritten over time, but I added bliss so we don't rely on auto-download from external sources.

robyww (changes requested):
The code looks good.  I just had a couple small code comments. 

I would prefer to have prism live as a npm package but it does not appear to be available. Also, I see that  https://github.com/PrismJS/live does not even have any releases either.  It is still at the experimental level. I don't see this as a big problem since `PrismJS/live` is not in the critical path of what Firefly does.

So in the future we will probably need to watch `PrismJS/live` and figure out how to effectively manage this package. I think the `externalSource` is fine for now.

**Document**
I began an effort to document our packages a while back (since you can't put it in package.json). Will you add a section around `package-json-doc.md` around line line 29?

**Bug**
There is still the bug that when you try to add columns from the selector, it does not show up in prism. I looked into this and the fix is not obvious.  However, after looking at the code it appears to respond to a resize. I think if you send a resize event to window it will update.  I think other firefly components will ignore the event because it does not have all the fields populated.
At `AdvancedADQL.jsx`:`insertAtCursor` add:
```js
    window.setTimeout(() => window.dispatchEvent(new UIEvent('resize')), 5);

```

robyww:
I have an event better idea to do that update.


in the `AdvancedADQL` component at

```
  const pRef = useRef(null);  
``` 
then in you useEffect
```js
pRef.current= new Prism.Live(textArea);
```
then pass it in the onSelect:
```js
  insertAtCursor(textArea, tname, adqlKey, groupKey,pRef.current);
```

you insertAtCursor would not look like:
```js
export function insertAtCursor (input, textToInsert, fieldKey, groupKey, prismLive) {
```

In `insertAtCursor` use `syncSytles()`:
```js
window.setTimeout( () => prismLive.syncStyles(), 5);
```

loitly:
Beside the issue where TextArea is not updating after mouse-clicked events from the schema UI, everything else looks good.

brianv0:
I think I got all the changes right, it seems to be working good with the columns too.

robyww (approved):

PR #1043, opened 2020-12-08
Firefly-688: Blank Hips

robyww:
#### Firefly-688: implement blank HiPS coverage

- blank hips change be changed normal hips
- background color may be set
- added a test
- blank hips is available as an HiPS option

_ticket_: https://jira.ipac.caltech.edu/browse/FIREFLY-688

#### Testing

https://fireflydev.ipac.caltech.edu/firefly-688-blank-hips/firefly/test/test-coverage.html?test=blank+hips

_or_

https://firefly-688-blank-hips.irsakudev.ipac.caltech.edu/irsaviewer/

_Search_:
 - 54.218205903 -37.497959343 
 - 10 arcsec (keep the radius small, the search will take a while, as in minutes)

_or_ 

https://firefly-688-blank-hips.irsakudev.ipac.caltech.edu/irsaviewer/

- choose blank HiPS from the menu


ejoliet (commented):
Looks good. in no particular order, i think we should discuss either after you merge or wait for the conclusion the following (some minor) issues:
- does it require Gator HTMl page to be changed? I see the test has a particular 'CovergaeHips:Blank' node to the table data. Is this needed if the Gator table result has a flag or a header specific attribute that would let the showCoverage function behave by default without passing this value manually. 
- Is overwriting possible? scenario where the simulated data can be use with a particular coverage hips instead of blank?
- label in the layer setting popup: 'Blank hips' is adequate enough for IRSA, need inputhere from CCB.
- Was it black instead of grey background on purpose?
- CCB should decide if the other operations make sense and should keep them enabled? Distance tool for example.
- Label '2MASS...' in the plot title appears although is an empty / black background sky.

The mechanism is there and works great. Maybe need one iteration for the other issues with CCB, what do you think?

robyww:
>does it require Gator HTMl page to be changed? I see the test has a particular 'CovergaeHips:Blank' node to the table data. Is this needed if the Gator table result has a flag or a header specific attribute that would let the showCoverage function behave by default without passing this value manually.

The coverage requires this meta data to be set. It can be in the table meta or in the api call which adds it to the table meta
_Note-_ as of this morning I will also put support for `SIMULATED_TABLE  = 'y'`, as with another other meta this can be added from the api or in the table.

> Is overwriting possible? scenario where the simulated data can be use with a particular coverage hips instead of blank?

When I add support for  `SIMULATED_TABLE  = 'y'` I will all it to be overridden by `CovergaeHips:some url`

> label in the layer setting popup: 'Blank hips' is adequate enough for IRSA, need input here from CCB.

OK

> Was it black instead of grey background on purpose? 

Yes I think it show the coverage colors better. It is a little closer to the way images would look.   If also make it stand out better when you have it zoom all the way out. (180 degrees). The user can set the background and we can change the default if we want. 

> CCB should decide if the other operations make sense and should keep them enabled? Distance tool for example.

I suppose, but it supports everything it did before. Right now I don't see the point of disabling anything.

> Label '2MASS...' in the plot title appears although is an empty / black background sky.

opps... I will fix it.

> Will the mechanism work the same way in IRSAViewer?

If a table with the right metadata comes in, then yes.

robyww:
>> CCB should decide if the other operations make sense and should keep them enabled? Distance tool for example.

> I suppose, but it supports everything it did before. Right now I don't see the point of disabling anything.

OK, i did find one toolbar item that should not be there. I disabled the color dropdown for a blank HiPS display

vandesai1:
So far I have tested here: https://fireflydev.ipac.caltech.edu/firefly-688-blank-hips/firefly/test/test-coverage.html?test=blank+hips

Looks great! A couple of comments:

1. If I "Change HiPS" to a 2MASS HiPS map, I can't "Change HiPS" back to "Blank".

2. If I click on the "Restore to Defaults" button, it replaces the blank HiPS map with a 2MASS HiPS map. I didn't expect that.

I haven't tested in IRSA Viewer, yet, because I was confused, and tried to do a search on the simulated catalog and it takes hours to run. I think I'm understanding that this isn't supposed to work on the actual simulated catalog....

robyww:
> 
> 1. If I "Change HiPS" to a 2MASS HiPS map, I can't "Change HiPS" back to "Blank".

@gpdf asked about this as well. I had not planned to do it since I did not thing there was an actual use case for it. However, since you both has asked independently I think I should just go ahead and do it.

_note:_ with any image or hips you can always hide the base layer.

> 2. If I click on the "Restore to Defaults" button, it replaces the blank HiPS map with a 2MASS HiPS map. I didn't expect that.

Yes, that is a bug. Glad you tested that, I did not think to. I will fix it.

> I haven't tested in IRSA Viewer, yet, because I was confused, and tried to do a search on the simulated catalog and it takes hours to run

You can check with @smonkewitz about this but I was able to do a 10 arcsec search in a couple of minutes.

PR #1045, opened 2020-12-14
Migrate BasicUI to a functional component

brianv0:
This PR migrates BasicUI to be a functional component rather than class-based.

### DIFF
--- a/src/firefly/js/ui/tap/TableSelectViewPanel.jsx
+++ b/src/firefly/js/ui/tap/TableSelectViewPanel.jsx
@@ -1,4 +1,4 @@
-import React, {PureComponent, Fragment} from 'react';
+import React, {PureComponent, Fragment, useState, useEffect, useRef} from 'react';
 import SplitPane from 'react-split-pane';
 
 import {SplitContent} from '../panel/DockLayoutPanel';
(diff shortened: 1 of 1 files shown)


robyww (approved):
The changes look good to me.

 I think it is a good solution with this `useStateRef`.  The problem to solve is that there are up to 3 this different promises firing and when they come back they may or maybe should use the results, so you need a ref, however if they set the results the UI should update, so you need a state.

Converting it to a functional component makes the code much easier to follow.

Good Job.

src/firefly/js/ui/tap/TableSelectViewPanel.jsx
```diff
@@ -271,202 +271,205 @@ function AdqlUI({serviceUrl}) {
...
-    loadSchemas(serviceUrl, schemaName=undefined, tableName=undefined) {
-        this.setState({error: undefined, schemaOptions: undefined, schemaName: undefined,
-            tableOptions: undefined, tableName: undefined, columnsModel: undefined});
+function BasicUI(props) {
+    const tapFluxState = getTapBrowserState();
+    // const isObsTap = props.obsTap !== undefined ? props.obsTap : false;
+    const [error, setError] = useState(undefined);
+    const [serviceUrl, serviceUrlRef, setServiceUrl] = useStateRef(tapFluxState.serviceUrl || props.serviceUrl);
+    const [schemaName, schemaRef, setSchemaName] = useStateRef(tapFluxState.schemaName || props.initArgs.schema);
+    const [tableName, tableRef, setTableName] = useStateRef(tapFluxState.tableName || props.initArgs.table);
```
loitly, line 290: Existing logic relies on local props/states as well as external states from flux/store made this a problem more complicated than it needs to be.  As it is now, the use of useStateRef to solve this problem is a good one.

loitly (approved):
The TAP panel may have started out as a single component but is no longer the case.  When multiple dependent components are relying on shared states, it should be taken from flux/store.  Currently, it's using a mixture of `getTapBrowserState`(flux) and props to pass along states.  A cleaner approach is to have all components reacting to flux's state changes instead of props.  The problem Brian fixed also exists in AdvancedADQL.

That said, the changes Brian made is good and clean.  It addresses the current issue.  I see no reason not to merge.

PR #1050, opened 2021-01-14
ADQL editor fixes when service url changes

brianv0:
Fixes issues with adql text area being out of sync when service url changes as well as z-index related issues when selecting a new service url

### DIFF
--- a/src/firefly/js/ui/tap/TableSelectViewPanel.jsx
+++ b/src/firefly/js/ui/tap/TableSelectViewPanel.jsx
@@ -184,7 +184,7 @@ export class TapSearchPanel extends PureComponent {
 
                         <div className='TapSearch__section'>
                             <div className='TapSearch__section--title'>1. TAP Service <HelpIcon helpId={tapHelpId('tapService')}/> </div>
-                            <div style={{flexGrow: 1, marginRight: 3, maxWidth: 1000}}>
+                            <div style={{flexGrow: 1, marginRight: 3, maxWidth: 1000, zIndex: 9999}}>
                                 <CreatableSelect
                                     options={tapOps}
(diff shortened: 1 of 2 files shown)


src/firefly/js/ui/tap/TableSelectViewPanel.jsx
```diff
@@ -184,7 +184,7 @@ export class TapSearchPanel extends PureComponent {
 
                         <div className='TapSearch__section'>
                             <div className='TapSearch__section--title'>1. TAP Service <HelpIcon helpId={tapHelpId('tapService')}/> </div>
-                            <div style={{flexGrow: 1, marginRight: 3, maxWidth: 1000}}>
+                            <div style={{flexGrow: 1, marginRight: 3, maxWidth: 1000, zIndex: 9999}}>
```
loitly, line 187: Can you describe when this is needed?  I am looking at `dev` now, and it looks like it's working fine without the change.  With this change is when the Schema and Table input fields flow on top of the TAP Service dropdown.

brianv0: image

See image ^. In the case when you are selecting the bottom of GAIA or the top of GAVO, the textarea is actually z-index=1 and it is above those divs (TapSearch__section) and selection won't actually work in that limited area, but it works in every other area of the dropdown.

loitly: ah.  thanks.  i was looking at the wrong section.

loitly (approved):
Looks fine.  Good to go.

robyww (approved):
Looks Good.

PR #1051, opened 2021-01-20
FIREFLY-686: Phase 1, first draft

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-686

Implemented a rough draft of phase 1 described in the above ticket.  I just wanted someone to look at it to provide feedback as I look into FIREFLY-269.  
- added `chartType = 'auto'` to `showChart()`. This allow you to display an auto-select chart based on a given table.  
- added nested table groups support.

Assumptions:
- ucd and unit defined as PARAM in table group override ucd and unit in FIELDRef if exists.
- if only statError is given, then error is symmetric
- when both statErrLow and statErrHigh are given, then error is asymmetric.

Questions:
- what if you have combination of statError, statErrLow, and statErrHigh? i.e..  statError and statErrLow or all three.
- what is binSize, binLow, and binHigh?
- sample data with SpectralAxis errors and how the chart should look like.
- what is upper/lower limit column, and how it that mapped to SpectrumDM.
  - sample data and how the chart should look like  
- I remembered in the meeting, Spectral chart settings dialog has some fields as read-only.  the pdf i got from @ejoliet  did not indicate that.  should i wait for additional requirement before looking into this?

Test: 
API: https://fireflydev.ipac.caltech.edu/firefly-686-spectra-viewer/firefly/test/tests-chart.html?test=Spectrum_DM+charts
For upload, use sample files in the ticket.


ejoliet (approved):
So far it looks pretty solid. I uploaded other tables and looks ok at first glance. I'm sure there is a lot of details but so far so good. I noticed that the UCD groups now can be clicked and more info popped up which give more insight. I'm not sure IRSA folks would like to see so much VO vocabulary but this is part of the details i was expecting to hear feedback from.

I would suggest to give the URL to Vandana/Xiuqin so they can play around too at this stage which i think is already advanced and complete enough. What do you guys think?

Also, i have to make sure now that what we are generating inside SOFIA from the data analyzer when interpreting spectra, it follws now the spectra model used here so the user who save the table from SOFIA or send it to IRSAViewer can display it in this spectra viewer also. I will raise ticket for that.

Regarding the PDF and fixed field in the settings, i think what they expected was to disable the field so people can't change it but then there was a discussion about a button to go to advanced mode which i didn't like. Let's hold on on this settings to be disabled or fixed and see the first feedback.

I think is standing out already as it is!

I mark it as 'approve' just because it has already all the goodies to get to another round of tweaking.

Great job!

robyww (commented):
Testing:

1. If I switch Asym error names to none, apply, and switch back then the title is gone.
2. I don't think there should be an add new trace at the top. Do we?
3. Should the error option be disabled if there is no error?

Comments:

The Table and the chart seem to be two different components, which might be fine for the current work.  However, I think we will eventually need a component that can alternate between showing table(s) that is involved in the spectrum. It might need side-by-side or flip option.

Maybe this needs to be a wrapper around your current work.

PR #1052, opened 2021-01-22
Firefly-689:  Multi search panel - TAP now available

robyww:
### Firefly-689: Make multi search panel - TAP available with other searches

- multi search panel now allow to put tap with other ipac searches
 - configurable
 - all panels now take up available area
 - clean up of highlevel tap panel, functional component

other stuff:

 - obs-core behavior now can be disabled, e.g. sofia
 - make color drop down popout
 - DM-28431: Added version access to logo icon

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-689  

### Testing

 - _IrsaViewer_: https://firefly-689-combine.irsakudev.ipac.caltech.edu/irsaviewer/
 - _Firefly_: : https://fireflydev.ipac.caltech.edu/firefly-689-combine/firefly/
 - Click on cataogs, you should see tabs that include tap.  Try them all


lrebull:
This is a really petty comment, but can we get rid of the 'shadow' under the words in the new tabs? It reads to me like a "comic sans" kind of faux pas.

Spacing seems an issue. (a) IRSA tab: box around upper right section looks weird to me because it is so offset from bottom box. (b) TAP searches tab, section 4 when "single table" is selected in 2, the "try NED then simbad" pulldown is partially buried AND I cannot scroll (and there is a ton of white space to the left of this part of the screen). 

Load catalog file -> upload from URL - is this no longer supposed to be the way you access VO catalogs, because now SCS is separate?  Is "VO" branding important enough that "VO" should appear somewhere? It appears on ops as "VO", and I don't know if that's important.

TAP searches tab: use of "schema"/"table" here - I know what you mean, but in particular we don't call it "schema" anywhere else. It's "project" and "catalog" (even in the IRSA tab of this rebuilt version) which seems more user-friendly. You've lost the links to information about the catalog and columns that are present in the IRSA tab.

oops. it just kicked me out. You must be rebuilding it or something. I can't do more until I can access it, so I'll check back later.

robyww:
> This is a really petty comment, but can we get rid of the 'shadow' under the words in the new tabs? It reads to me like a "comic sans" kind of faux pas.

You might be being a little hard on the physiognomy here. The tabs need to stand out a little more than usual since the represent a header on different types of searches. I think this is a little more subtle than just using a bigger font. However, I could play with the shadow color some and see if there is a happy medium. 


> Spacing seems an issue. (a) IRSA tab: box around upper right section looks weird to me because it is so offset from bottom box. (b) TAP searches tab, section 4 when "single table" is selected in 2, the "try NED then simbad" pulldown is partially buried AND I cannot scroll (and there is a ton of white space to the left of this part of the screen).

 - _a:_ For IRSA tab I am not really sure what to do about that beyond a full revamp of how it is laid out. The upper right box would not look right being resized, or being repositioned as the layout is stretched. The change is an attempt to make better use of the real estate and more consistent with how we have been designing panels recently. I don't think it is perfect but probably good enough.

 - _b:_ The internal layout of the TAP panel is really out of scope for this ticket. I have been aware of the ned/simbad option being cut off.  I think we will get it worked out before the release. @brianv0 is working on that part and still doing some updates in future PRs.

> Load catalog file -> upload from URL - is this no longer supposed to be the way you access VO catalogs, because now SCS is separate? Is "VO" branding important enough that "VO" should appear somewhere? It appears on ops as "VO", and I don't know if that's important.

This is the name that @vandesai1 asked for in the ticket.

> TAP searches tab: use of "schema"/"table" here - I know what you mean, but in particular we don't call it "schema" 
anywhere else. It's "project" and "catalog" (even in the IRSA tab of this rebuilt version) which seems more user-friendly. 

We have discussed the word "schema" extensively in the firefly science CCB. 

Here is my conclusion: I don't particularly like the term but it is the correct TAP term. All the services that support tap are using it.  I might be worse if we choose something more user-friendly since it would be confusing to the user as they switch between services. I think we just have to educate the users as to what it means. 

You can also ask @gpdf, @vandesai1, or @ejoliet for their take on it.

>You've lost the links to information about the catalog and columns that are present in the IRSA tab. Maybe that's ok because they're still in the IRSA tab, but I noticed it.

The details come from the backend IRSA TAP service, Firefly just shows what it gets from the service.  There is another ticket, https://jira.ipac.caltech.edu/browse/IRSA-3879, that might be worth commenting on

lrebull:
Yeah, I’m coming at this totally blind since I didn’t even know this was on the docket. And really just a few minutes in, before I could try anything, I got ‘you are not connected to the server anymore’ followed by ‘504 bad gateway’ and nothing else I did got me back in. So all I had time to do was note cosmetics....

lrebull:
(Just to reiterate, I have NOT been involved in ANY of the discussions of this, so I have no idea what has been decided upon and what hasn't. But our users haven't been involved in those conversations either!)

What does "ADQL edits below will not be reflected in Single Table view" mean? (What is "Single Table view"? I cannot figure it out from playing with it.) My choices from the main TAP panel are "Single table" or "ADQL" - is that what is meant? Is ADQL designed to merge tables and thus be different than "single table"? If so, that should be one of the examples given.

Little of this makes any sense to me:
> Schema browser usage
> Click on a Table node to insert a default SELECT statement of that table into the Query input box.
> Click on a Column node to insert the column’s name at the Query input box’s cursor.
> Insert fully-qualified names
if "schema" is supposed to be the new name for "mission" (e.g., wise, 2mass, etc), then what does "schema browser usage" mean?
If the idea is that clicking on a table name on the left is supposed to insert a select statement, it doesn't for me; it just puts the name of the table wherever my cursor is, breaking the ostensibly working-by-default ADQL into nonsense. This seems to be the behavior described in the subsequent line for columns, not tables.
I still think it very, very odd that you don't want the description of these tables/columns to be accessible anywhere from this tab and only from the irsa tab. Even experts in ADQL will need a mechanism for finding out what all the cryptic names of tables and columns are. I mean, I'm literally one of the world's experts on IRSA holdings and I could not have come up ab initio with "wise.wise_allwise_p3am_cdd" as the AllWISE release, and I had to check that that was really right.
(Also, I find the capitalization in that text to be odd - if you want Query Input Box to be a proper noun and thus capitalized, that's fine, but just capitalizing Query seems weird to me.)

For the examples on the bottom, you should specify in the plain-English text that this is searching 2MASS all-sky because "fp_psc" isn't sufficient to indicate that to most users.

The ADQL that comes up for me by default in the box does not work. I tried it and I get "Table Load Error: UsageFault: CONTAINS on a CIRCLE is not implemented." Whatever script is in there by default needs to work.

The examples given near the bottom do not work either; they blank out the screen and tell me "No coverage available" and "UsageFault: CONTAINS on a CIRCLE is not implemented." I have to restart the tool to regain control. 

Huh. Having reloaded the tool, the ADQL that appears in the boxes now is totally different and I now can't reproduce the failure, so I don't know what I did to break it. ON the other hand, now clicking on any of the table nodes or column names does nothing to the AQDL script box, even after I explicitly put the cursor in the ADQL script box. 

I don't know how to test more than this, which I realize is really all cosmetic stuff, but I can't get more out of it.

stargaser:
While I generally like the TAP functionality, and I do want to see it become more widely available, I have hit a hitch. I want to do queries to find out about the length of a table, in ADQL like `select count(ra) from ztf.ztf_current_meta_sci`. These never return a value in the Firefly TAP UI. I get an answer back in seconds using an interface like pyvo.

Why doesn't this query work?

robyww:
@lrebull @stargaser 
I appreciate your feedback on the TAP UI panel. However, it is all outside the scope of this ticket. We are simply exposing the existing firefly TAP UI, not updating or improving it. 

Over time I hope to make it better and better. However, I think now we need to make a ticket for TAP UI feedback. Maybe we need a meeting.

lrebull:
Yes, please, I don't understand at all how to comment on the TAP UI without commenting on the TAP UI....

robyww:
please see: https://jira.ipac.caltech.edu/browse/FIREFLY-699 and all related tickets.

gpdf:
@lrebull , I'll reply in more detail later, on a Confluence page (to respect @robyww's point about what is and isn't in scope in _this_ ticket).  One key point is that _in this phase of work_ we are mainly adding a _generic_ capability to search any TAP service.  That means we can't just replace the ADQL namespace hierarchy terms ("schema" and "table" - and see below for a further complication) with IRSA-specific ones of our own, or make assumptions about how other publishers use them to organize their datasets.  

One suggestion that has been made is that we could augment the language we can't escape with one small thing: we could say "Table collection (schema)" instead of just "Schema"; this is essentially an accurate characterization of the role "schema" plays in a generic TAP service.

gpdf:
Responses to some of @lrebull's comments here: https://confluence.ipac.caltech.edu/display/FIREFLY/TAP+UI+feedback

PR #1054, opened 2021-02-02
Firefly 693 ts phase precision too low

wmiipac:
Fix the phase fold table column "phase" precision.  
To test on k8 dev build:
https://fireflydev.ipac.caltech.edu/wmi-firefly-693/firefly/ts.html
 upload the attached ts table in [Jira ticket](https://jira.ipac.caltech.edu/browse/FIREFLY-693), follow the phase fold steps.


loitly (approved):
Changes look good.

ejoliet (approved):
Thanks for fixing this! Worth the fix in the upcoming release!

PR #1056, opened 2021-02-03
FIREFLY-707: coordinate transformation defined by PCi_j is not applied

tgoldina:
https://jira.ipac.caltech.edu/browse/FIREFLY-707

Fixed a bug in the the display of images with WCS defined by PCi_j.  Rotation defined by PCi_j had no effect. (PCi_j is a linear transformation matrix without scale. When CDELT values are equal, the PCi_j matrix is a pure rotation. It is a FITS standard alternative to CROTA2 or CDi_j.)

Test at https://fireflydev.ipac.caltech.edu/firefly-707/firefly/
1. Load image from URL http://spherexdev1:8080/simulated/simulator_delivery1/outputs/SPHEREx_simulation_1_field_2_array_2.fits
2. Use E,N button and notice that the image orientation it is now matching DS9 screenshot. 
3. The catalog of the sources from the third extension now overlays exactly with the image.

### DIFF
--- a/src/firefly/js/visualize/projection/ProjectionHeaderParser.js
+++ b/src/firefly/js/visualize/projection/ProjectionHeaderParser.js
@@ -245,10 +245,12 @@ export function parseSpacialHeaderInfo(header, altWcs='', zeroHeader) {
 	    if ( (p.cdelt1) && (p.cdelt2) ) p.cd1_1 = undefined;
 	}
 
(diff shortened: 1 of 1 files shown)


robyww (approved):
The changes look good, I studied the code and it makes sense. Thank you for looking into this. I think this bug must have been introduced when I converted the `ImageHeader.java` to `ProjectionHeaderParser.js`. I tried to simplify the code some and I missed this case.

PR #1058, opened 2021-02-03
Firefly-567: added ecliptic j2000 readout

robyww:
#### Firefly-567: added ecliptic j2000 readout

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-567

#### Testing

- https://fireflydev.ipac.caltech.edu/firefly-567-ecl/firefly/
- click on the readout dialog, there will the a ecliptic j2000 option

### DIFF
--- a/src/firefly/js/drawingLayers/WebGridUI.jsx
+++ b/src/firefly/js/drawingLayers/WebGridUI.jsx
@@ -14,10 +14,10 @@ import {COORDINATE_PREFERENCE} from './WebGrid.js';
 export const getUIComponent = (drawLayer,pv) => < WebGridUI drawLayer={drawLayer} pv={pv}/>;
 
 const coordinateOptionArray = [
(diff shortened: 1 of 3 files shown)


ejoliet (commented):
Does the design ticket mention a specific acronym to use. Is label "ECL J2000" ok? 
Another comment i have is about decimal and HMS. The is only the option for decimal. Equatorial has also HMS. Is that on purpose. Again i would ask CCB-D about this feedback. Maybe for consistency we should have the hms option? What do you think?

ejoliet:
Also, i noticed inconsistency between the labels in the grid layer settings are all different. EQ J2000 is Eq. J2000, and so on.

robyww:
I don't think you use HMS for ecliptic. We never have before.  The ticket gives no detail about the labels. However, I will go ahead and make the grid and options consistent.

PR #1060, opened 2021-02-17
Firefly-685: set the starting the cube plane

robyww:
#### Firefly-685: set the starting the cube plane

Other stuff:

  - upgrade react to 17
  - clean up

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-685

#### Testing

 - https://fireflydev.ipac.caltech.edu/firefly-685-starting-cube-plane/firefly/test/test-image.html
 - You will see two cube each starting on a different plane
 - you will have to click expand to see it in the test
 - plane by be set by a number or a percent: i.e. 5, '5', '20%'


ejoliet (approved):
Looks good. As per the ticket request, i tested that we can now select the middle or any index in a cube of images.

I didn't test beyond that but i noticed a lot of clean up. Would you think useful to test other features or other apps such SOFIA just to make sure that the changes doesn't have any impact? We can also wait for this to be merged and test in dev too.

Thanks!

ejoliet:
Actually, i've noticed that it includes the `upgrade react to 17` so might need more testing - @deannaji / @wmiipac could you please test SOFIA / FinderChart either based on this branch or wait for the merged and test dev in general. Thanks!

PR #1063, opened 2021-03-05
Irsa-3966: Firefly changes to support the prot of the IRSA toolbar an…

robyww:
Irsa-3966: Firefly changes to support the prot of the IRSA toolbar and frontpage to React

  - Two more command in AppServerCommands: GetAlerts & GetUserInfo
 
_Ticket_:  https://jira.ipac.caltech.edu/browse/IRSA-3966

_Testing_: test with front page


loitly:
I added `frontpage` to jenkins.  `frontpage-gwt` is still there for the GWT version.
This is what I found so far testing it in an IRSA like deploy... https://irsa-3966-frontpage.irsakudev.ipac.caltech.edu/frontpage/

- login profile link shows name instead of email.
- toolbar will not load elsewhere beside frontpage, i.e Gator, Atlas, etc.
- all ‘blue’ buttons are absolute link to ‘irsa.ipac.caltech.edu’ instead of relative.
- i think all of the links in the menubar link to ‘irsa’ instead of relatively.
- links in the slider also go to 'irsa' instead of relatively.
- Knowledgebase link is bad.
- ‘Help Desk’ and ‘Contact’ should also be relative link.  (existing bug).

Let's fix these first, then we can go for a second round of testing.

Also, will you help change irsa-ife/.gitignore: `package.json` to `/package.json`

buildScript/webpack.config.js
```diff
@@ -106,6 +109,7 @@ export default function makeWebpackConfig(config) {
         workerFilename= '[name]-[hash].worker.js';
     }
     const output =  {filename, path: out_path};
+    console.log('output:',output);
```
loitly, line 112: Ok to remove  now.

src/firefly/java/edu/caltech/ipac/firefly/server/AlertsMonitor.java
```diff
@@ -30,7 +30,7 @@
      * check for alerts
      * @param forceSend  if true, send alerts regardless
      */
-    public static void checkAlerts(boolean forceSend) {
+    public static Alert checkAlerts(boolean forceSend) {
```
loitly, line 33: Let's add a flag to avoid sending alerts when polling.  As is, alerts are sent but not delivered.  Not sure what happens to them.

loitly:
Everything looks fine now except for a couple of links that I think should be relative instead of absolute:
- User profile
- Footer's `Contact`

Also, you forgot to edit `.gitignore` file to only exclude `package.json` in the root directory, not everywhere. i.e.  `/package.json`

robyww:
> Everything looks fine now except for a couple of links that I think should be relative instead of absolute:

>User profile
Footer's Contact


I can fix `Contact`. The user profile is coming from the server, ultimately from app.config.  Do you have a suggestion for what we should do? Is it that important since I think it is doing the right thing?

loitly:
> I can fix `Contact`. The user profile is coming from the server, ultimately from app.config. Do you have a suggestion for what we should do? Is it that important since I think it is doing the right thing?

If you compare `irsa-ife/config/app.config` to `gwt` branch, you will see that we only use absolute path on local build env.  This allow the proxy to decide which version of the SSO is suitable for its deployment.
Basically, remove `sso.server.url` and `sso.user.profile.url ` from `dev`, `test`, and `ops_int` env block.

robyww:
ok, it that is what you want to do.

robyww:
ahhh, i guess we can do that now since we are `https` across the board

buildScript/webpack.config.js
robyww: done

src/firefly/java/edu/caltech/ipac/firefly/server/AlertsMonitor.java
robyww: done.

PR #1064, opened 2021-03-16
Corrected errors in frequency and flux density unit conversions

gpdf:


loitly (approved):
Looks good.  Thank you for correcting the formula.

PR #1066, opened 2021-03-17
FIREFLY-428: Spectral Viewer Phase 1 RC

loitly:
Test: https://fireflydev.ipac.caltech.edu/firefly-428-chart-auto-split/firefly/

I am listing all of the tickets I've looked at and worked on.  Some, only partially, due to time constraint or requiring additional requirements.

https://jira.ipac.caltech.edu/browse/FIREFLY-691  (phase 1 requirement)
- Different order will be assigned a different color, but with the same symbol.  Both, can be changed from the settings dialog.
- TODO
  - How to identify a spectra vs a SED? `ipac:Spectrum.SED` is this a table utype?
  - The hover text for a point should include order information if it is present.

https://jira.ipac.caltech.edu/browse/FIREFLY-686  (create new spetral viewer component)
- Done
 
https://jira.ipac.caltech.edu/browse/FIREFLY-428  (auto-split spectra with 'order' info)
- Mostly done, except for some advance UI concepts I do not understand

https://jira.ipac.caltech.edu/browse/FIREFLY-269  (only unit conversion)
- Used this ticket for initial unit conversion work.  Other items on the ticket has not been worked on.

https://jira.ipac.caltech.edu/browse/FIREFLY-733  (release multitrace UI to ops)
- Done.  However, due to this change, there are cases when there's only 1 field under 'Trace Options' (Heatmap).

https://jira.ipac.caltech.edu/browse/FIREFLY-745  (modifications to Spectral Viewer Prototype)
- #3: spectral axis upper/lower limit.  Feature is not yet implemented.  This will require more time than what's available in phase 1.  Please move to phase 2 and provide sample data as well as what it should look like.
- TODO
  -  #8 this is not available as part of plotly API.  Need to use annotations which may not work completely in all cases.  Should reconsider the request.  If this is still an issue, please move to phase 2.
  - #11 "Use spectrum preset" is the feature "Escape Spectral/SED-specific Plot Settings" requested in FIREFLY-691.  Please provide new label as well as tooltips.
  - #15 So, instead of 'scatter', you want to see 'spectrum'?  Or, you want both options?  Today is the first time I heard of this request.  I think it should be pushed to phase 2.

The goal of this Phase 1 RC is to tie up loose the ends so it can be released.  Please help thoroughly test this version for bugs.    I will fix as much as I can.  Bigger items will be addressed in phase 2.

Additional changes not specified in the above tickets:
- Updated plotly from v1.49.4 to v1.58.4.  This is needed to add Legend's title to chart.
- Allow active trace switching by clicking on data points on the chart
- Minor UI changes 
  - Remove borders to make dialog less cluttered
  - Add max height to dialog with auto scrollbar


ejoliet:
FIREFLY-686 has one sample including specOrder example. See file name: SPITZER_S0_25343744_0001_3_E7173899_tune.votable

ejoliet (approved):
Great!
I've added the sample file to the FIREFLY-428 ticket with the added reference field to the wavelength axis group so anyone can test it. It works like a charm!

robyww (approved):
Looks good.

Suggestions:
- You might want to try using `async`/`await` in `FireflyGenericData.js`. I find it improves readability.
- You might want to go ahead a convert `ScatterOptions` to be a function component. I am trying to do that when I have to modify any component.  I would like to see the rest of `SimpleComponent` subclasses  get upgraded.

robyww (approved):
#### Code

I is really too much code to review, but what I looked at looks good. I did not notice you changed some `SimpleComponent` classes to functional. That's great!

#### UI

- I look really good to me. stuff seems to work as advertised, though I did not click on everything.
- **One comment:** I don't think our non-editable fields should look like the editable.  `InputFieldView` should probably render a different look when `readonly` is set. The only way a user knows he can't edit them is by clicking and wonder what is wrong. To the user it probably feels like a bug.

vandesai1:
- attempting to change x-axis units does not work
- attempting to change y-axis units does not work
- inconsistent point value tooltip: when order 2 is selected, the tooltip will show wavelength, flux_density with errorbar for either order. When order 3 is selected, it will not show the uncertainty for a point in order 2, but will show the uncertainty for a point in order 3.
- Firefly doesn't recognize SPITZER_S3_3757824_0013_12_E7212632_tune.votable or F0291_EX_SPE_04012012_EXEELONEXEECHL_CMB_0040-0043.votable as having orders, no order testing possible.
- I think we want "Plot Style"-->"Trace Style"
- The "choose trace" menu is really small so my chosen trace name doesn't fit in it even though there is plenty of space.

PR #1068, opened 2021-03-18
FIREFLY-717: Auto escape values when used as part of a URL in a <LINK> tag

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-717

Instead of adding an option or an attribute to a votable LINK tag which will violate its xml schema, it's best to change the way these values are interpreted.

So, instead of taking the value as is, we will determine whether or not encoding is needed based on how it's used.
If it's used as tokens to construct a URL, then we will encode it.  If the value is intended to be a complete URL, then it's taken as is without any encoding

Test case added with two examples: src/firefly/js/util/tests/VOAnalyzer-test.js

Test deployment is here: https://fireflydev.ipac.caltech.edu/firefly-717-escape-link-url/firefly/


cwang2016:
I did the test by using a modified votable sample from NED. In the votable sample, there is a 'link' column with the values as part of the url. The test shows the table and the link well.

cwang2016 (approved):

PR #1071, opened 2021-03-31
IRSA-3625: HERON dataset added, need IBE to be fixed though

ejoliet:
Following the info in the ticket [IRSA-3625](https://jira.ipac.caltech.edu/browse/IRSA-3625), i've added HERON config added to master image file.

UI changes such as labels and filtering can still be tested for now and i've built a version here:
https://irsa-3625.irsakudev.ipac.caltech.edu/irsaviewer/

Sample targets that suppose to give a match is: PGC 10266.
IBE doesn't seems ready because the search returned nothing:
https://irsadev.ipac.caltech.edu/IBE?table=heron.heron_images&POS=40.669629,-0.013281


lrebull:
HERON should be tagged as an extragalactic program. It's not at the moment. All else seems ok.

ejoliet:
Updated based on reviewer feedback about category, should be extragalactic.
Merging.

PR #1072, opened 2021-04-05
FIREFLY-717_full_url_token URL escape option for <LINK> parameters values for single complete token case

cwang2016:
This PR has the solution regarding  the issue addressed in https://jira.ipac.caltech.edu/browse/FIREFLY-717 
In the case like  `<LINK href=“${columm_name}”>` where the assignment to 'href' is a single independent token, then the value from the relevant column is treated as a full URL, no escape is involved while resolving a tag with 'href' value.

Test: 
https://fireflydev.ipac.caltech.edu/firefly-717-full-url-token/firefly/
Please test two xml samples adapted from NED votables  attached in https://jira.ipac.caltech.edu/browse/FIREFLY-717.

### DIFF
--- a/src/firefly/js/util/VOAnalyzer.js
+++ b/src/firefly/js/util/VOAnalyzer.js
@@ -1223,8 +1223,9 @@ export const findTargetName = (columns) => columns.find( (c) => DEFAULT_TNAME_OP
  * @returns {string}    the resolved href after subsitution
  */
 export function applyLinkSub(tableModel, href='', rowIdx, fval='') {
-    if (href) fval = encodeURIComponent(fval);                  // if href is given, then we assume fval is a value token and not a full url.
-    const rhref = applyTokenSub(tableModel, href, rowIdx, '', true);
+    const encode = !!href && !href.match(/^\${[\w -.]+}$/g);      // don't encode if href is blank or consists of exact one token.
+    if (encode) fval = encodeURIComponent(fval);                // if encoding is needed, then fval needs to be encoded.
+    const rhref = applyTokenSub(tableModel, href, rowIdx, '', encode);
     if (rhref === href) {
         return fval ? href + fval : '';       // no substitution given, append defval to the url.  set A.1
     }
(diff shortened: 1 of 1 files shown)


src/firefly/js/util/VOAnalyzer.js
```diff
@@ -1247,6 +1247,7 @@ export function applyTokenSub(tableModel, val='', rowIdx, def, encode=false) {
     const vars = val?.match?.(/\${[\w -.]+}/g);
     let rval = val;
     if (vars) {
+        encode = encode && (vars[0] !== val);
```
loitly, line 1250: `applyTokenSub` is generic and used in many places.  Please move logic into `applyLinkSub` where it's made for resolving links.  I am guessing something like this..

```
const encode = !!href && !href.match(/^\${[\w -.]+}$/g);    // don't encode if href is blank or consist of exactly one token
if (encode) fval = encodeURIComponent(fval);                // if encoding is needed, then fval also need to be encoded.
const rhref = applyTokenSub(tableModel, href, rowIdx, '', encode);

```

loitly (changes requested):
Please move logic into `applyTokenSub` and resubmit.  Thank you.

src/firefly/js/util/VOAnalyzer.js
cwang2016: Yes I agree it is better way to have the change in 'applyLinkSub' and I was doing that originally.  
I moved the change into 'applyTokenSub'  since I thought, 
- It seems only applyLinkSub has passed 'encode: true' to 'applyTokenSub', other calls pass 'encode:false' to `applyLinkSub`.  (It may not be always true), 
- href.match is done twice in both `applyLinkSub` and `applyTokenSub`, but it should be ok. 

I will move the change to be inside `applyLinkSub` as your comment, however, please check if the match like this
`href.match(/^\${[\w -.]+}$/g)`  is good for the case like `${col_name1}${col_name2}`?

cwang2016: Yes,  `/^\${[\w -.]+}$/` is good to test if `href` consists of exact one token! It is for the string starting with `${`, ending with `}`,  and containing at least one of alphanumeric characters, underscore, or characters between space and period in the middle.

loitly (approved):
Looks good.  Ok to merge.

PR #1076, opened 2021-04-12
Firefly-737: add parameter to set sleep time for cleanup

robyww:
Firefly-737: add parameter to set sleep time for cleanup …

  - CLEANUP_INTERVAL sets the max age for most of the temp directories
  - HiPS continues to be 30 days and stage continues to be 7 days
  - revamped the cleanup.sh script

_ticket_: https://jira.ipac.caltech.edu/browse/FIREFLY-737


loitly (changes requested):
Naming it `CLEANUP_INTERVAL` maybe clearer than `CLEANUP_SLEEP_TIME`.
But, that's not the issue.  The issue is when should the cached files be deleted.  Currently, we have 1d, 7d, and 90d.  I suggest categorizing them into meaningful categories, then allow the user to override those values.  In this case, then `cleanup_interval` is not needed, since it should always be shorter or equals to the shortest category.

robyww:
oh, your right. I did not study the script enough and just got focussed on the interval. I will update it.

robyww:
Added a new update.  This required me to revamp `cleanup.sh`.  Now `CLEANUP_INTERVAL` will also set oldest file that is allowed in the temporary dirs. The script will sleep 1 minute longer. So it's sleep time is now tied to the file age. To make all this work and because the linux `find` `-mtime` parameter does not allow allows units (i.e. days, minutes, hours) I have to parse the  CLEANUP_INTERVAL, convert to minute,s and always use `-mmin` parameter.

The parsing actually turned out to be a good thing since it now gives me finer control over the sleep time.

docker/base/cleanup.sh
```diff
@@ -1,66 +1,85 @@
...
+   CLEANUP_INTERVAL=${DEFAULT_CLEANUP_INTERVAL}
+   interval=$(getInterval "${DEFAULT_CLEANUP_INTERVAL}")
+   mult=$(getMultiplier "${DEFAULT_CLEANUP_INTERVAL}")
+fi
 
-        ${find_cmd} "${workarea}/upload" -type f -mtime +7 -exec /bin/rm '{}' \+ -print >> "${log_file}" 2>&1
-        ${find_cmd} "${workarea}/HiPS" -type f -mtime +90 -exec /bin/rm '{}' \+ -print >> "${log_file}" 2>&1
+cleanupMinutes=$((interval*mult))
+sleepTime="$((cleanupMinutes+1))m"
+logfileMaxAge=$((cleanupMinutes*10+5))
```
loitly, line 42: I am not sure why the +1m to sleepTime.  This makes it possible for a file to be older than cleanupMinutes.  Also, logfileMaxAge is not needed.  The size of the log file is insignificant.  It's served for debugging purposes mainly.  It should be set to a fixed time.  But, both are trivial.  Ok if you want to merge as is.

loitly (approved):

docker/base/cleanup.sh
robyww: Add +1 to sleep time allows it get search the entire interval. I think added the +1 make it possible to catch more older files. Otherwise it could miss something that we within a few seconds. It is not a big deal with 24h stuff my since LSST wants to set it to like 1h it might be more important.

Every cleanup run has a new log file. The max again keeps it to around 10 files.  It was hardcoded before.

PR #1078, opened 2021-04-15
For DSS, use TAN if present - FIREFLY-662

tyb987:


ejoliet:
@tyb987 you already merged ? without review or test build URL posted here? 
Would have been nice(er) to get the review of the code/test.

PR #1079, opened 2021-04-17
FIREFLY-691: Create the capability to view spectra and SEDs (Phase 1)

loitly:
This PR covers the 3 tickets below.  I've also included bugs reported in my previous related PR.  Those that I've fixed are checked.  Please test:  https://fireflydev.ipac.caltech.edu/firefly-691-spectra-phase1/firefly/

<br />

https://jira.ipac.caltech.edu/browse/FIREFLY-733
Should be fully implemented, except for items where tooltips were requested.  Please provide the tooltips.
<br />

https://jira.ipac.caltech.edu/browse/FIREFLY-691
> 5.  Handle orders: ... The hover text for a point should include order information if it is present.

Order info is not in hover text, but is in the legend.  Is this acceptable?

> 6. Escape Spectral/SED-specific Plot Settings: Firefly should forget that this is a spectrum or SED...

This will move to phase 2.  The ability to switch back/forth was taken out in phase 1 due to unresolved bugs.
<br /><br />

https://jira.ipac.caltech.edu/browse/FIREFLY-745

> In spectral mode, sort the x-axis by default. Allow the user to sort on whatever column they want, which may make a connected lines plot look goofy, but that's ok. When users change the x-axis units, the table can stay sorted by the original x-axis values. For all scatter plots in general, don't do any sorting for the user.

No default sorting applied but user is allowed to sort on whatever column.  This simplify some of the initial book keeping.  Is this acceptable?

> Upper and lower limits for the spectral axis should map to:
spec:Spectrum.Data.SpectralAxis.Accuracy.BinLow
spec:Spectrum.Data.SpectralAxis.Accuracy.BinHigh
and these limits should be plotted as error bars except without the serifs, e.g. as segments.

I don't think it's currently supported in Firefly(xMax xMin).  I need sample data as well as how it should look visually because I may need to overlay it as an annotation.  This will move to phase 2.

Issues with the UI found from testing:
- [x]  CONFUSING: The point corresponding to the highlighted row is the same color as one of the orders. That's a little confusing.
- [x]  CONFUSING: The legend doesn't have a title to let you know what "2" and "3" refer to. Would title this the same as the column name, in this case "order".
- [ ] CONFUSING: When I click on the blue line in the legend, only the orange points remain. A mouseover would help: "Click to toggle visibility of trace"

This is not available as part of plotly API. Need to use annotations which may not work completely in all cases. Should reconsider the request. If this is still an issue, please move to phase 2.

- [x] BUG: By default, plot y axis linearly, not in log.
- [x] BUG: Why can't I plot the x axis in log? By default, it should be linear, but there should be a checkbox for changing it to log.
- [x] CONFUSING: The plot settings now has a checkbox that says "Use spectrum preset". We need to learn what this is supposed to do before renaming.
- [x] BUG: Why is the trace menu different than the order menu? Aren't the orders traces?
- [x] BUG: Changing the color of the orders doesn't work
- [x] BUG: The plot settings now has a "Symbol" menu, with default = circle. However, if I change to one of the other options, nothing happens.
- [ ] BUG: If I add a new chart on a loaded spectrum, I can't access spectral mode.

Please provide more details and move it to phase 2.

- [x] BUG: The plot settings no longer has the "line" plot style option. We'd like it available for spectra.

<br /><br />

**Reported Bugs**:
- [x] non-editable fields should NOT look like the editable
- [x] attempting to change x-axis units does not work
- [x] attempting to change y-axis units does not work
- [x] inconsistent point value tooltip: when order 2 is selected, the tooltip will show wavelength, flux_density with errorbar for either order. When order 3 is selected, it will not show the uncertainty for a point in order 2, but will show the uncertainty for a point in order 3.
- [x] Firefly doesn't recognize SPITZER_S3_3757824_0013_12_E7212632_tune.votable or F0291_EX_SPE_04012012_EXEELONEXEECHL_CMB_0040-0043.votable as having orders, no order testing possible.

Must include order field as a FIELDref in a GROUP block.. i.e
```
<GROUP ID="Data.SpectralAxis" utype="spec:Spectrum.Data.SpectralAxis">
   <FIELDref ref="wavelength"/>
   <FIELDref ref="order"/>
</GROUP>
```
- [x] I think we want "Plot Style"-->"Trace Style"
- [x] The "choose trace" menu is really small so my chosen trace name doesn't fit in it even though there is plenty of space.
- [x] Plot should auto-scale after unit changes
- [x] utype for Spectral Order should be `ipac:Data.SpectralAxis.Order`

This request is added while still supporting the old `spec:Spectrum.Data.FluxAxis.Order` until asked to remove.
- [ ] XY options:  toggling reverse checkbox does not work reliably when log is also selected.

This is an existing bug.  There were not enough time to fix this in this PR.  Please raise ticket for this bug.


src/firefly/js/ui/FieldGroup.jsx
```diff
@@ -63,9 +65,11 @@ export class FieldGroup extends Component {
         if (shallowequal(this.props, nextProps)) return false;
         const wrapperGroupKey= nextContext;
         const {groupKey, reducerFunc, keepState, actionTypes}= nextProps;
-        if (this.props.groupKey!==groupKey) {// support change the groupKey property on the form without unmounting
+        // added check for 'reducerFunc.ver' to re-register if needed.  Should revisit this for a better solution
+        if (this.props.groupKey!==groupKey || this.props?.reducerFunc?.ver !== reducerFunc?.ver) {// support change the groupKey property on the form without unmounting
```
robyww, line 69: After studying all this, I think it would be better to add a `dispatchUpdateReducer`.
```js
if (this.props.groupKey!==groupKey) 
   dispatchMountFieldGroup(groupKey, false);
   dispatchMountFieldGroup(groupKey, true, keepState, null, reducerFunc, actionTypes, wrapperGroupKey);
}
else if (this.props?.reducerFunc !== reducerFunc) {
   dispatchUpdateReducer(groupKey,reducerFunc)
}
```
The reducer would be something like:
```js
const fg= {...state[groupKey], reducerFunc};
return {....state, [groupKey]: {...fg, fields: fireFieldsReducer(fg, action)}
```
I will do this on Monday and push to this branch, then you can try it out. If it does not work we can revert. It would make the external code simpler.

robyww (approved):
Looks good.

See below for my plan about FieldGroup.

I did notice in the UI you remove the option to not treat as a specturm. Was it just not working?

ejoliet (approved):
Nice work! Testes everything that has been reviewed and requested and looks good.

Apart from the unchecked bugs/feedback not resolved this time which make totally sense and will be in the future, i've found this one from FIREFLY-733:

> Change "Add New Trace"→ "Overplot New Trace" 

The rest looks all improved and fixed following CCB feedback.
Thank you!

ejoliet:
From my last comment, i was confused and Loi explained the differences in option. The 'Overplot New Trace' will appear for regular spectra which doens't include orders.
Also, i've raised ticket fo the bug `XY options: toggling reverse checkbox does not work reliably when log is also selected.`

PR #1081, opened 2021-04-21
FIREFLY-643: Resizing a column to zero makes it disappear forever

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-643

Set the minimum width of the table's column to 5px.  This prevents it from resizing to zero.
Also,
- Reset button in the Table option panel will also reset all columns back to theirs default width
- Adjust row height for several tables with input fields, i.e.  TAP constraint table, IRSA catalog constraint table, and Table Option panel
- Only show More Actions (...) icon when the column has a default text based renderer

Test URL: https://fireflydev.ipac.caltech.edu/firefly-643-col-resize-min/firefly/

### DIFF
--- a/src/firefly/js/tables/TableConnector.js
+++ b/src/firefly/js/tables/TableConnector.js
@@ -102,7 +102,7 @@ export class TableConnector {
         TblCntlr.dispatchTableUiUpdate(changes);
     }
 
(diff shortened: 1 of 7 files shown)


ejoliet (approved):
Works great!
Thanks.

PR #1082, opened 2021-04-27
IRSA-4042,FIREFLY-761,FIREFLY-765: Chart multi-trace related change requests

loitly:
Test URL: https://irsa-4042-doubleclick-zoom-reset.irsakudev.ipac.caltech.edu/irsaviewer/

https://jira.ipac.caltech.edu/browse/IRSA-4042
- Disable plotly's "double click to zoom out" feature.  
- Set `yLimitUI` defaults back to `false`.  Chart setting options will not show y-axis upper/lower limit fields unless it is explicitly set.

https://jira.ipac.caltech.edu/browse/FIREFLY-761
- Color picker doesn't update

### DIFF
--- a/src/firefly/js/ui/ColorPicker.jsx
+++ b/src/firefly/js/ui/ColorPicker.jsx
@@ -27,19 +27,16 @@ var lastEv;
 function ColorPickerWrapper ({callback,color,callbackOnOKOnly, callbackOnBoth}) {
     const [currentColor, setCurrentColor] = useState(color);
 
-    const updateColor= (ev) => {
+    const onChange = (ev) => {
+        lastEv=ev;
         const {r,g,b,a}= ev.rgb;
         setCurrentColor(`rgba(${r},${g},${b},${a})`);
-        callback(ev,false);
+        if (!callbackOnOKOnly && callback) callback(ev,false);
     };
 
     return (
         <div>
-            <SketchPicker
-                         onChangeComplete={ (ev) => {
-                             lastEv=ev;
-                             if (!callbackOnOKOnly) updateColor(ev);
-                         } }
+            <SketchPicker onChange={onChange}
                          color={currentColor} />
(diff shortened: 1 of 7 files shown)


lrebull:
I can't get it to load images - i get "Failed - could not parse results" for WISE images.  And now it works but i can't get catalogs. wth. trying again.
IRSA-4042 - issues described there indeed rendered obsolete!
FIREFLY-760 issues, though, still hanging around - making a movie to demonstrate and will put in that ticket. If i can ever get catalogs to load again.
can't test FIREFLY-761 until i can get it to load a catalog.
updates as events warrant...

lrebull:
I inadvertently left the catalog query on this build running overnight, and it still hasn't pulled a catalog despite trying for ~12 hours. A new query on a new load of the tool isn't coming back either. I can't test this until I can get a catalog loaded into the build. Should I load one from disk? Why is this build not pulling catalogs from gator?

lrebull:
Just to get over this hump, I'm loading catalogs from disk.
FIREFLY-760 - I think this is doing what we want. See notes in ticket.
FIREFLY-761 - yes, this is working.

src/firefly/js/ui/ColorPicker.jsx
```diff
@@ -27,19 +27,16 @@ var lastEv;
...
     };
 
     return (
         <div>
-            <SketchPicker
-                         onChangeComplete={ (ev) => {
-                             lastEv=ev;
-                             if (!callbackOnOKOnly) updateColor(ev);
-                         } }
+            <SketchPicker onChange={onChange}
```
robyww, line 39: All this code is fine, however, in working on firefly-762, I also fixed this bug. So the code will probably look a little different when I merge.

robyww (approved):
Look good.

lrebull:
I still can't get it to do a catalog search for me straightaway, but thanks to your suggestion from before, I placed that query in the background and the second query worked instantly (as you described). 

Everything works!

PR #1084, opened 2021-04-29
FIREFLY-531: Update supported-browser list in the README

gpdf:
Updated some of the help to clarify current capabilities, as well.  Regularized the use of MarkDown header levels.


robyww (approved):
Thanks @gpdf !

PR #1086, opened 2021-05-03
FIREFLY-722: helpid to point to location instead

ejoliet:
The change here is to fix how we get the panel prefix id to be used as HelpId when ObsCore is enabled.

It should be picking up the 'location' prefix instead of always 'spatial'.

Please, make sure that the change in 'ObsCore''s case doesn't affect the fieldgroup or other UI bindings and kept the difference between Spatial and Location panel.

This PR include minor label change - FIREFLY-773


robyww (approved):
Look good.

PR #1087, opened 2021-05-03
FIREFLY-763: bug in symbol size

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-763

When an expression starting with numbers is entered into Size Map field, plot fail to render the symbol size correctly.

test URL: https://firefly-763-symbol-size.irsakudev.ipac.caltech.edu/irsaviewer/
Follow ticket's steps to reproduce the bug.

### DIFF
--- a/src/firefly/js/charts/dataTypes/FireflyGenericData.js
+++ b/src/firefly/js/charts/dataTypes/FireflyGenericData.js
@@ -105,7 +105,7 @@ export function createChartTblRequest(chartId, traceNum, tablesource) {
     let sreq = cloneRequest(request, {
         startIdx: 0,
         pageSize: MAX_ROW,
-        inclCols: Object.entries(mappings).filter(([k, v]) => !numberOrArrayProps.includes(k) || Number.isNaN(parseFloat(v))).map(([k, v]) => {
+        inclCols: Object.entries(mappings).filter(([k, v]) => !(numberOrArrayProps.includes(k) && !isNaN(v))).map(([k, v]) => {   // if field can be a number and is a number, then skip
             // we'd like expression columns to be named as the paths to trace data arrays, ex. data[0].x
             // otherwise use column names to preserve original column attributes (type, format, etc.)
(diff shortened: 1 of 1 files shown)


src/firefly/js/charts/dataTypes/FireflyGenericData.js
```diff
@@ -105,7 +105,7 @@ export function createChartTblRequest(chartId, traceNum, tablesource) {
     let sreq = cloneRequest(request, {
         startIdx: 0,
         pageSize: MAX_ROW,
-        inclCols: Object.entries(mappings).filter(([k, v]) => !numberOrArrayProps.includes(k) || Number.isNaN(parseFloat(v))).map(([k, v]) => {
+        inclCols: Object.entries(mappings).filter(([k, v]) => !(numberOrArrayProps.includes(k) && !isNaN(v))).map(([k, v]) => {   // if field can be a number and is a number, then skip
```
robyww, line 108: A couple of thoughts just to make sure this is the change you want to make:

- `Number.isNan` I think requires a number while `isNaN` converts to a number.
- `parseFloat` will convert the number part of a string and throw out the rest so `’123.45.abc’` will be `123.45`
- so the change is a little more than simplifying the code to `&& !`

I don’t know if this really matters but it could affect the processing of the expression a little.

loitly: Yes.  `parseFloat` was the cause of the problem.  '2 * ra' returns 2, which is a number, therefore is excluded from the filter.  I use `isNaN` because it's more strict when parsing.  I also changed the condition to make it more precise.

ejoliet (approved):
Great! I was looking to also fix it there. 
Thanks for looking at it.
Good to go.

robyww (approved):

robyww:
_Note-_ this PR should be against `rc-2021.2`.

loitly:
> _Note-_ this PR should be against `rc-2021.2`.

you're right.  let me fix that.

loitly:
Merged in this PR:  https://github.com/Caltech-IPAC/firefly/pull/1088

PR #1089, opened 2021-05-06
Fix ADQL query generation

brianv0:
Table name changes were being properly dispatched, so building queries was being done on stale table names. This fixes that.

This also adds some newline breaking to generated ADQL text when the list of column names is very long, while also breaking on FROM and WHERE keywords.

_To test:_ https://fireflydev.ipac.caltech.edu/brian-fix-table/firefly/


robyww (approved):
Looks good. Still waiting on testing then we will merge.

robyww:
@gpdf confirm that the testing looks good.

PR #1090, opened 2021-05-07
Firefly-779: improve data product view with ZTF

robyww:
Firefly-779: improve data product view with ZTF

- Update the ztf WebPlotRequest creation to work with ZTF
- Make a way to set an "no products" message from a property
- Fix bug finder the dataProductsId
- Make coverage viewer to the right thing when there are tables not an active one


wmiipac (commented):
Just need a small change. 
Everything works as expected. Very nice.

PR #1091, opened 2021-05-13
[DM-30073] Handle redirects when retrieving TAP errors

rra:
Errors from an async TAP call are stored the same way as data
results and thus may return a redirect to an external URL that
does not understand our tokens.  However, they lacked special
redirect handling and the underlying request library would resend
the authentication token for the original request when following
the redirect.  This in turn led to sending authentication tokens
to sites (such as a GCS bucket URL) which did not understand them.

Copy the same code pattern from getData to getErrorMsg: disable
redirects, check the result for a redirect status, and then follow
the redirect manually, re-checking its URL for whether we should
send an authentication token.

I suspect the status of the secondary request may not propagate
back to the outer request for both this and getData, but couldn't
confirm that.  In the longer term, we should rethink how this
code works (and possibly use an underlying HTTP library that strips
Authorization headers when following redirects).


ejoliet:
@loitly heads up - i did `gl :firefly:bAD` and it failed compiling java:
```
> Task :firefly:compileJava
/private/tmp/firefly/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java:212: error: variable method is already defined in method getErrorMsg()
                                (method -> err.setSource(getErrResp(method, errorUrl))));
```

gpdf:
@ejoliet I think this is `bug`, not `enhancement`.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -199,11 +199,27 @@ public String getErrorMsg()  {
...
+                .setFollowRedirect(false);
+            HttpServices.Status status = HttpServices.getData(input, (method -> {
+                try {
+                    if (HttpServices.isOk(method)) {
+                        err.setSource(getErrResp(method, errorUrl));
+                    } else if (HttpServices.isRedirected(method)) {
+                        String location = HttpServices.getResHeader(method, "Location", null);
+                        if (location != null) {
+                            HttpServices.getData(HttpServiceInput.createWithCredential(location),
+                                (method -> err.setSource(getErrResp(method, errorUrl))));
```
loitly, line 212: Fail to compile.  `method` is already declared at line 204.

rra:
Apologies for that; I don't have a Java development environment set up and was afraid of something like that. :( I'm not sure what is the correct syntactical way to express that method callback in Java for the interior call to `getData`.

loitly (changes requested):
Logic looks good.  Please correct the compilation error, then it's good to go.

loitly (approved):

gpdf:
https://jira.ipac.caltech.edu/browse/FIREFLY-780 for fixing this in the 2021.2.x series.

gpdf:
Changing the base branch did not do what I expected.

gpdf:
This PR is probably a goner at this point.  I'll work on making a new one from scratch.

robyww:
Changing the branch base is rarely a good idea. That is why I asked you to close this one and open a new one to `rc-2021.2`. We have had bad experience with changing the branch base in the past.

gpdf:
Already done: #1092

gpdf:
Closing; no longer useful.

gpdf:
I wasn't sure that I could preserve @rra's authorship if I made a new branch from scratch, but in the end it worked out fine.

PR #1095, opened 2021-05-28
Firefly 785 fieldgroup

robyww:
Do the following with FieldGroup:

 

Change Field group to a function component
play with the FieldGroupConnector Effect for the component still works like before.
remove all sorts of functionality that was put into FieldGroup but not used or barely used.
related actions - Monitoring actions with a saga or watch
simplify `keepState` remove initValues functionality (never used)
Removed a lot of `validatorFunc`  properties from field group, I am not sure how this every got in there.

https://fireflydev.ipac.caltech.edu/firefly/

### DIFF
--- a/src/firefly/js/charts/ui/HistogramOptions.jsx
+++ b/src/firefly/js/charts/ui/HistogramOptions.jsx
@@ -299,7 +299,7 @@ export class HistogramOptions extends Component {
         // to avoid width change due to scroll bar appearing when full height suggest box is rendered
         return (
             <div>
-                <FieldGroup groupKey={groupKey} validatorFunc={null} keepState={false}
+                <FieldGroup groupKey={groupKey} keepState={false}
                             reducerFunc={columnNameReducer(colValStats,basicFieldsReducer)}>
 
                     <ColumnOrExpression {...xProps}/>
--- a/src/firefly/js/core/BootstrapRegistry.js
+++ b/src/firefly/js/core/BootstrapRegistry.js
@@ -16,7 +16,7 @@ import ComponentCntlr, {DIALOG_OR_COMPONENT_KEY} from './ComponentCntlr.js';
 import AppDataCntlr from './AppDataCntlr.js';
 import BackgroundCntlr from './background/BackgroundCntlr.js';
 import ImagePlotCntlr from '../visualize/ImagePlotCntlr.js';
-import FieldGroupCntlr, {addFieldGroupRelatedWatcher, MOUNT_COMPONENT} from '../fieldGroup/FieldGroupCntlr.js';
+import FieldGroupCntlr, {MOUNT_COMPONENT} from '../fieldGroup/FieldGroupCntlr.js';
 import MouseReadoutCntlr from '../visualize/MouseReadoutCntlr.js';
 import TablesCntlr from '../tables/TablesCntlr.js';
 import DrawLayerCntlr from '../visualize/DrawLayerCntlr.js';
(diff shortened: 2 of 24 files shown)


loitly:
Let me put down a comment before I forget.  I will do some testing after.  No need to review code, as I expect it to always be perfect!

There's a few of things I think would greatly improve the usability of FieldGroup/Input.

1. All components should have `className` and `style` props set to its container block.  And, label should use <label> tag. 
 Styling our InputField is tough because not all style attributes are exposed.  This change will allow us to style our component like any react/html component.  i.e.  `.myClsName > label { text-weight: bold; margin: 0 3em;}`
2. All components should have a `key` props.  This should work like any react component as well.  When `key` is changed, it should be remounted(unmount/mount).  This is especially important when you want to control what the field value should be, considering a remount will clear current value and populate default value(initialState).
3. initialState: I think this already works like it should be.  It's equivalent to input's `defaultValue`, except it can be anything else, not just value.  Which means, any props outside of initialState should render as requested.  I am afraid it may cause regression issues if enforced.

These are just things I would like you to consider.  Of course, I don't expect any of these changes to go into this PR.

robyww:
@loitly I agree with all of this. I will actually require getting out of field group core and more in to the users of the fieldgroup but I think we could do most of it.

robyww:
> initialState: I think this already works like it should be. It's equivalent to input's defaultValue, except it can be anything else, not just value. Which means, any props outside of initialState should render as requested. I am afraid it may cause regression issues if enforced.

I do think this works like this now. I made some effort on it about a year ago. If there is a case that does not work then we can fix iit.

robyww:
> All components should have a key props. This should work like any react component as well. When key is changed, it should be remounted(unmount/mount). This is especially important when you want to control what the field value should be, considering a remount will clear current value and populate default value(initialState).

We need to do some  testing with this. You don't usually add `key` props with react, you  just use them. When they are there then react does the right thing. Maybe we need to do some tesitng to make sure that `FieldGroupConnector` is doing the right thing.  If you have a test case that is not working then let me know.

loitly:
So far, I've only found IRSA Catalog now working.  After entering parameters, `Search` failed with exception in the console.
I've done some spot checking in `firefly`.  If we're serious about merging this PR, we have to build all of IRSA's apps and test them to make sure they don't have regression issues as well.  They also use a lot of FieldGroup/Form components.

loitly:
> > All components should have a key props. This should work like any react component as well. When key is changed, it should be remounted(unmount/mount). This is especially important when you want to control what the field value should be, considering a remount will clear current value and populate default value(initialState).
> 
> We need to do some testing with this. You don't usually add `key` props with react, you just use them. When they are there then react does the right thing. Maybe we need to do some tesitng to make sure that `FieldGroupConnector` is doing the right thing. Maybe that is a potential problem. If you have a test case that is not working then let me know.

You are right.  `key` is not a prop we expose.  It's a special props that React handles.  We just need to make sure our fields behave correctly when `key` is changed(mount/unmount cycles).

robyww:
@loitly I think i was finally able to make a test and experiment with what is involved by changing the `key` for a field. I can make this work as long as `keepState` is `false`. When `keepState` is `true` then the initialState is going to come from what ever was stored in the field group. I think that is what we want. It does give more control to the panel to force a reinit.

robyww:
> All components should have className and style props set to its container block. And, label should use <label> tag.
Styling our InputField is tough because not all style attributes are exposed. This change will allow us to style our component like any react/html component. i.e. .myClsName > label { text-weight: bold; margin: 0 3em;}

I do think this is a good idea but I think it is really  out of scope of the work I am trying to do which is more focused on the internal. I am all for doing it as a separate task. @loitly is right in the we did not approach labels they best way. Too much  inexperienced at the time. I am scare making this change will break a lot so we need to be careful.

loitly (approved):
I did some testing on the built apps and did not find anything regression issues.  Okay to merge.

PR #1097, opened 2021-06-11
Handle redirects when retrieving TAP errors

gpdf:
This PR is just to get this commit into the Caltech-IPAC/firefly repo, so that it can be accessed by Jenkins.

Errors from an async TAP call are stored the same way as data
results and thus may return a redirect to an external URL that
does not understand our tokens.  However, they lacked special
redirect handling and the underlying request library would resend
the authentication token for the original request when following
the redirect.  This in turn led to sending authentication tokens
to sites (such as a GCS bucket URL) which did not understand them.

Copy the same code pattern from getData to getErrorMsg: disable
redirects, check the result for a redirect status, and then follow
the redirect manually, re-checking its URL for whether we should
send an authentication token.

I suspect the status of the secondary request may not propagate
back to the outer request for both this and getData, but couldn't
confirm that.  In the longer term, we should rethink how this
code works (and possibly use an underlying HTTP library that strips
Authorization headers when following redirects).


gpdf:
Copying the key commit into a non-production branch.

PR #1100, opened 2021-06-15
IRSA-4090 ZTF migration add ancillary file download

wmiipac:
minor change for ZTF file retrieve, more changes are in [ife branch](https://github.com/IPAC-SW/irsa-ife/pull/160).


ejoliet (approved):
Is this change also will allow the dialog be updated in other cases than ZTF, like in SOFIA? 

IFE branch is tested too! 

Thanks!
Good to go from my side.

wmiipac:
"Is this change also will allow the dialog be updated in other cases than ZTF, like in SOFIA?"
No, Sofia download button need to be updated,  it has to be done in ife

loitly (approved):
looks fine.

PR #1101, opened 2021-06-17
Improve nested error handling in retrieval of TAP error documents

gpdf:
Ensure that errors in both the initial retrieval and in the retrieval of a
redirected error document are handled correctly.  Previous version was
incorrectly still reporting an error after a successful redirect.


loitly (approved):
Changes look good.

PR #1104, opened 2021-06-22
FIREFLY-805: Filter does not work on boolean type column

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-805

Get exceptions when filtering boolean column:
```
Table Load Error:Invalid statement: 
"w4sigmpro.mask" IN ('true') 

```
Also  fixed a couple of minor errors showing up in the console.

Test: https://fireflydev.ipac.caltech.edu/firefly-805-boolean-col-broken/firefly/
Use attached tables in the ticket to test.

### DIFF
--- a/src/firefly/js/tables/FilterInfo.js
+++ b/src/firefly/js/tables/FilterInfo.js
@@ -137,13 +137,8 @@ export class FilterInfo {
                 for (let i = 0; i < parts.length; i += 2) {
                     const [cname, op, val] = parseInput(parts[i]);
                     if (cname) {
-                        let isNumeric = false;
-                        if (columns) {
-                            const col = columns.find((c) => c.name === cname);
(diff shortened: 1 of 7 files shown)


gpdf:
I'll test this.  It's going to hit Data Preview 0.1 in Rubin hard - they have a lot of Boolean columns.

gpdf (approved):
Fixes boolean behavior.

Tested both via pull-down category-column filtering and by typing "false", "true", "=false", and "=true" into the column-head fields.

All work as expected.  Ordinary integer field filtering still works correctly, too.

Very nice!

gpdf:
You probably still want @ejoliet to do a normal code review; I only looked at the behavior.

ejoliet (approved):
Works like a charm!
Thanks!

PR #1107, opened 2021-07-01
FIREFLY-800:Make image search sections scroll all together

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-800

Make Sections 1-4 of the **Image Search** form scroll together.
In addition to the ticket (need approval?) :
- `4. Select Data Set` toolbar will remain visible when scrolled out of view.
- `Filter by:` and `Selection:` moved to toolbar so that it does not get scrolled off the top.

Also:
- convert to functional component

Test: https://fireflydev.ipac.caltech.edu/firefly-800-image-search-scroll/firefly/

### DIFF
--- a/src/firefly/js/ui/ImageSelect.css
+++ b/src/firefly/js/ui/ImageSelect.css
@@ -18,24 +18,40 @@
     color: saddlebrown;
     font-style: italic;
     font-size: smaller;
-    padding: 0 4px 2px;
+    padding-top: 1px;
 }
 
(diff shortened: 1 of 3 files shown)


ejoliet (approved):
Looks clean and usable in small screen now! Also, the bar stays on top which is an added feature to keep it always visible. Very good improvements!
Thanks!

robyww (approved):
It looks really good. Nice job.

PR #1109, opened 2021-07-07
Firefly-823: Implement the file saving standard

robyww:
### Firefly-823: Implement the file saving standard 

 - standard is referenced in the ticket
 - created a utility function for the standard: makeDefaultDownloadFileName in fetch.js

_Test build_: https://fireflydev.ipac.caltech.edu/firefly-823-save-files/firefly/
_Ticket_: https://jira.ipac.caltech.edu/browse/FIREFLY-823

#### Testing
_It is very important to look at the ticket first to understand what was implemented_

Save the following types of files:

Charts
- save a file from chart using the default
- save a file from chart using another name
- add a title and save a chart

Tables
- save a file from table using the defaults
- change the type and save
- edit the name then change the type and save

Images
- save a image or region or png using the default
- change the default then change the type and save
- save a three color
- save a three color but change the band first
- save a three color but change the name then change the band
- save a three color png.

### DIFF
--- a/src/firefly/js/charts/ui/PlotlySaveDialog.jsx
+++ b/src/firefly/js/charts/ui/PlotlySaveDialog.jsx
@@ -1,0 +1,79 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+import React from 'react';
+import {getWorkspaceConfig} from 'firefly/visualize/WorkspaceCntlr.js';
+import {PopupPanel} from 'firefly/ui/PopupPanel.jsx';
+import DialogRootContainer from 'firefly/ui/DialogRootContainer.jsx';
+import {dispatchHideDialog, dispatchShowDialog} from 'firefly/core/ComponentCntlr.js';
+import {isString} from 'lodash';
+import CompleteButton from 'firefly/ui/CompleteButton.jsx';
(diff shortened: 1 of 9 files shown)


ejoliet (approved):
I've tested all 3 in Firefly viewer and looks good and consistent. 
Couple of remarks:
- when changing the image filename, is not sticky to the dialog after i close it which is fine but user could be surprised.
- the anchor help id referring to the help for the chart saving dialog is not correct, is pointing to 'rotate' - the helpid should be a feedback from @lrebull or we need to create a new helpid in Firefly docs.
- I didn't test Gator, FinderChart or SOFIA GUI, i imagined that those changes will apply to inherit apps and HTMLs as well, would be great if someone test that too, in particular SOFIA for the imminent RC.

Firefly side looks better and more consistent now.
Thank you!

robyww:
> when changing the image filename, is not sticky to the dialog after i close it which is fine but user could be surprised.

This is a the way images were before. I did make table and charts to work the same way. I think it is best to reset to default if the dialog is closed. I could image certain use cases where a user might be unhappy but I think it is more an edge case.

> the anchor help id referring to the help for the chart saving dialog is not correct, is pointing to 'rotate' 

Thanks. I will fix.

loitly (approved):
I tested this PR from IRSA perspective.  It works as described.  Looks great.

PR #1110, opened 2021-07-08
FIREFLY-822: table link going to the wrong url

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-822
Ticket reported table link going to the wrong URL after switching between 'Featured' sources.
The root cause is its failure to keep track of the selected row.  The logic was overly complex.  It was too hard to understand.
Instead, I've done the following.  It should fix the reported issues.
- convert to server-backed table
- remove complex bookkeeping logic
- attempt to simplify numerous available options
  - `firefly.options.hips.defHipsSources` used to allow a csv string, object with {label,source}, or an array of object with {label,source} with some kind of auto-fix logic.  I removed the string option and remove the auto-fix logic.
  - `firefly.options.hips.hipsSources` defaults to `all`
  - although `HiPSDataType` is a parameter, it's always 'image' and 'cube'.  Not sure when or where 'catalog' is used.
- small changes to UI to make it consistent with the other Image Search options
  - the `IRSA Featured` checkbox is moved to the title area next to `4. Select Data Set` (not centered)
  - when ` Select Image Source = URL`, it just say `4. Enter URL` instead of `4. Select Data Set` then a `Enter URL` underneath.

Test: https://fireflydev.ipac.caltech.edu/firefly-822-hips-table-url/firefly/


robyww (approved):
The changes look good. After testing I don't see any problem. I had forgotten that we used this table to `resolveHiPSIvoURL` I am glad you caught that.

One small UI issue. the `IRSA Featured` checkbox was centered before.

ejoliet:
> although HiPSDataType is a parameter, it's always 'image' and 'cube'. Not sure when or where 'catalog' is used.

HiPSDataType catalog is for HiPS catalogs which we don't support yet and we won't for the near future.

loitly:
> One small UI issue. the `IRSA Featured` checkbox was centered before.

In `Change HiPS`, it's also left aligned.  Is this an issue? I prefer this over the centered look because it shifts when resizes.  But, I can also centered it if that's what people prefer.

ejoliet (approved):
Looks good!

Only thing i noticed and we discussed offline is the link URL for a 'blank hips' s brokien when clicking to the icon (i). Probably disble the click or remove the icon or leave it if it's too much of an effort. OPS is broken the same way for now anyway.

loitly:
> Looks good!
> 
> Only thing i noticed and we discussed offline is the link URL for a 'blank hips' s broken when clicking to the icon (i). Probably disble the click or remove the icon or leave it if it's too much of an effort. OPS is broken the same way for now anyway.

This is fixed and updated.

src/firefly/js/visualize/HiPSListUtil.js
```diff
@@ -212,107 +97,25 @@ export function resolveHiPSIvoURL(ivoOrUrl) {
...
+        if (ivoIdx<0 || urlIdx<1) return undefined;
+        const lowerIvo= ivoOrUrl.toLowerCase();
+        // now match the table
+        const foundRow= get(tableModel,'tableData.data', []).find( (row) =>
+            row?.[ivoIdx]?.toLowerCase().includes(lowerIvo) );
+        return foundRow?.[urlIdx] || ivoOrUrl;
+    };
+
+    if (FULL_HIPS_TABLE) {
+        return Promise.resolve(findInTable());
```
robyww, line 112: should be:
```
 return Promise.resolve(findInTable(FULL_HIPS_TABLE));
```

loitly: Thanks, it's fixed.  Test instance updated.

ejoliet:
If i select 'Blank hips' and 'search' nothing happens - i think this broken now. @loitly could you confirm?

loitly:
> If i select 'Blank hips' and 'search' nothing happens - i think this broken now. @loitly could you confirm?

@ejoliet Sorry, didn't test after I removed the info icon.  Should be fixed now.

PR #1111, opened 2021-07-09
IRSA-3867_show_middle_plane_of_any_sofia_data_cube_by_default

deannaji:
Jira: https://jira.ipac.caltech.edu/browse/IRSA-3867
K8s: https://irsa-3867.irsakudev.ipac.caltech.edu/applications/sofia/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
@@ -2,27 +2,22 @@ package edu.caltech.ipac.firefly.server.dpanalyze;
 
 import edu.caltech.ipac.firefly.core.FileAnalysisReport;
 import edu.caltech.ipac.firefly.data.sofia.Sofia1DSpectraExtractor;
+import edu.caltech.ipac.firefly.data.sofia.SofiaFitsConverterUtil;
 import edu.caltech.ipac.firefly.data.sofia.SofiaSpectraModel;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.table.DataGroup;
-import edu.caltech.ipac.table.io.IpacTableReader;
-import edu.caltech.ipac.table.io.IpacTableWriter;
 import edu.caltech.ipac.table.io.VoTableReader;
 import edu.caltech.ipac.table.io.VoTableWriter;
 import edu.caltech.ipac.util.FitsHDUUtil;
-import edu.caltech.ipac.util.download.FailedRequestException;
-import edu.caltech.ipac.util.download.URLDownload;
 import nom.tam.fits.BasicHDU;
 import nom.tam.fits.Fits;
 import nom.tam.fits.FitsException;
 import nom.tam.fits.Header;
-import edu.caltech.ipac.firefly.data.sofia.SofiaFitsConverterUtil;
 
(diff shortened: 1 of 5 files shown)


ejoliet (changes requested):
Doesn't seems to be showing the middle of a cube.
Example:
- 'all-sky' , only GREAT instrument, filter files by 'FITS' , i see the cube displayed starting from first plane out of 126. 
- 'all-sky', HAWK+, first row... 26 planes, starts at 1, when it should at 13.

Are you sure the build is correct?
Thanks!

src/firefly/js/metaConvert/PartAnalyzer.js
```diff
@@ -266,6 +266,9 @@ function analyzeImageResult(part, request, table, row, fileFormat, fileOnServer,
         newReq.setTitleOptions(TitleOptions.NONE);
         override= true;
     }
+    part['additionalImageParams'] && Object.entries(part['additionalImageParams']).forEach(([k,v]) => {
+        newReq.setParam(k,v);
+    });
```
robyww, line 271: You can simplify this to:
```js
Object.entries(part.additionalImageParams ?? {} ).forEach(([k,v]) => newReq.setParam(k,v);
```

`part['additionalImageParams']` is harder to read than `part.additionalImageParams` we rarely use that except if there is a variable in the brackets.  Using the syntax `part.additionalImageParams ?? {}` says if the object is `undefined` then use an empty object which makes an empty array which makes the forEach a noop.

src/firefly/java/edu/caltech/ipac/firefly/server/dpanalyze/SofiaAnalyzer.java
```diff
@@ -44,6 +39,9 @@ public FileAnalysisReport analyzeFits(FileAnalysisReport inputReport,
         String code = params.getOrDefault("product_type", "");
         String level = params.getOrDefault("processing_level", "").toUpperCase();
         String inst = params.getOrDefault("instrument", "").toUpperCase();
+        Map<String, String> additionalParams = new HashMap<>();
+        additionalParams.put("CUBE_FIRST_FRAME","50%");
+        inputReport.setAdditionalImageParams(additionalParams);
```
robyww, line 44: The code is fine but since you are just setting one parameter  you could just use a singleton map.

```java
inputReport.setAdditionalImageParams(Collections.singletonMap("CUBE_FIRST_FRAME","50%"));
```

robyww (changes requested):
I see the same issue that @ejoliet sees. It does not appear to be working. Do you have a search where it does work? 

So there are two possibilities

option 1. **The image viewer is not doing the right thing.** You are doing everything correctly and the image viewer is not honoring the parameter. If so then I need to debug the image viewer.

option 2. **There are cases where the '50%' parameter is not getting set.** If that is true then you need to look at sofia analyzer and figure out why.

If you have a test case where it does actually work then I think the overall problem is option 2. If you have never seen it work correctly then I have something to fix.

I also gave a couple of suggestions for the code.

robyww:
@deannaji I found another place on the javascript side that `additionalImageParams` had to be set. That was the problem that @ejoliet found. I pushed a commit to fix it.

Please rebuilt and you and @ejoliet confirm it is working.

ejoliet (approved):
Seems to be working , i tested GREAT and see the middle plane of cubes displayed.
Thanks!

robyww (approved):

PR #1115, opened 2021-07-27
FIREFLY-829: Input fields inside a table is not being re-rendered correctly

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-829
- Input fields inside a table fail to re-render due to bad comparator
- fix Image Search not rendering correctly in Safari
- fix HiPS table not rendering correctly when pop-up in Safari
- fix Image Search 3-color not rendering correctly in Safari

Also:
- auto-size TAP column constraints

Test:
https://fireflydev.ipac.caltech.edu/firefly-829-client-table-bug/firefly/

Make sure to test all 4 bug fixes in all browsers.
Also, I changed the TAP `Column Selection` table to auto-size the column's width.  In my opinion, it looks a lot better.  But, if CCB does not like it, it's easy to undo.

### DIFF
--- a/src/firefly/js/tables/ui/TableRenderer.js
+++ b/src/firefly/js/tables/ui/TableRenderer.js
@@ -340,7 +340,10 @@ export const CellWrapper =  React.memo( (props) => {
 }, skipCellRender);
 
 function skipCellRender(prev={}, next={}) {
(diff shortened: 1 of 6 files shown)


robyww (approved):
Tested all fixes. Looks Good.

PR #1118, opened 2021-08-03
Default Spatial Target with proper Errors validation

brianv0:
Fixes validation issues in TAP panel, both the error messages themselves and the bubbling up of those errors.

This adds the following logic to the UI when switching between tables:

• If ra/dec columns are identified in a table, ~and the table is NOT an ObsCore table~ or the table is an ObsCore table, then preselect the Spatial dropdown requiring a target in the TAP panel
• If you switch to a table without ra/dec, deselect Spatial dropdown when switching tables.
~• If you switch to ObsCore table (which does have an ra/dec), deselect "Spatial" Dropdown (same widget but it has a different title)~

The last case exists because we don't necessarily expect ra/dec to be the primary way to search an ObsCore table.

### DIFF
--- a/src/firefly/js/ui/tap/ObsCore.jsx
+++ b/src/firefly/js/ui/tap/ObsCore.jsx
@@ -185,19 +185,21 @@ export function ObsCoreSearch({cols, groupKey, fields, useConstraintReducer}) {
         const constraintsResult = makeConstraints(fields, fieldsValidity, panelActive);
         updatePanelFields(constraintsResult.fieldsValidity, constraintsResult.valid, fields, newFields, panelValue, panelPrefix);
         if (isPanelChecked(panelValue, panelPrefix, newFields)) {
-            if (constraintsResult.valid){
-                if (constraintsResult.adqlConstraint?.length > 0){
+            if (constraintsResult.valid) {
+                if (constraintsResult.adqlConstraint?.length > 0) {
                     adqlConstraint = constraintsResult.adqlConstraint;
                 } else {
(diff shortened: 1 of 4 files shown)


robyww:
_test build:_ https://fireflydev.ipac.caltech.edu/u-bvan-spatial-target-and-errors/firefly/

robyww:
>  If you switch to ObsCore table (which does have an ra/dec), deselect "Spatial" Dropdown (same widget but it has a different title)

I have real concerns about this one. This will promote all-sky searches. I thought this PR was to discourage that. Am I missing something?

gpdf:
It was primarily to discourage all-sky searches on regular catalogs, but I think you're right, limited-sky-area searches should be the default for (ObsCore) image metadata tables, too.

robyww (changes requested):
@brianv0 the code looks good and it works but I think @gpdf and I agree

> we should  limited-sky-area searches should be the default for (ObsCore) image metadata tables, too.

gpdf (changes requested):
To complete my review after interacting with the test build:

I agree with @robyww : it seems like the behavior for ordinary spatial catalogs and for non-spatial tables is as agreed.  Thanks!

Please do change the ObsCore default to enabling spatial search, with the target field left empty (and therefore the "empty entry" error displayed as for regular spatial catalogs).

brianv0:
Boolean has been reversed to make sure the panel is active if an ObsCore table has been selected.

robyww (approved):

robyww:
All looks good now. I will merge.

PR #1119, opened 2021-08-05
DM-30780: Fix ivoa.ObsCore detection

brianv0:
This makes a number of changes to improve ivoa.obscore detection and handling:

1. Rewrite query to go by table name 'ivoa.ObsCore' or 'ivoa.obscore' instead of columns, which also makes it so that the obscore-like tables are no longer considered for the radio button UI ( MAST in particular was not happy with column detection query). It seems like the two hardcoded names are most reliable (rather than relying on a lower casing function in ADQL to be implemented, which isn't guaranteed).
2. This pulls up async detection logic out of BasicUI to the parent component so that the radio button works better when advanced ADQL editor is selected (and also helps with a complicated timing issue).


robyww:
_testing:_ https://fireflydev.ipac.caltech.edu/u-bvan-dm-30780-obscore-detect/firefly/

robyww (approved):

PR #1120, opened 2021-08-09
Fix reducermap registration/deregistration

brianv0:
This fixes some useEffect hooks to do things on component mount/unmount, instead of every time.


brianv0:
Normally the function returned by `FieldGroupUtils.bindToStore` is directly returned by hooks, but I wanted to add some extra logic, so this returns a function that will call that on unmount instead.

If there's a slightly cleaner way to do this, let me know - this seems to work fine.

PR #1125, opened 2021-09-02
FIREFLY-732_custom_JSON_parser

loitly:
ticket: 
- https://jira.ipac.caltech.edu/browse/FIREFLY-732

test: 
- https://fireflydev.ipac.caltech.edu/firefly-732-custom-json-parser/firefly/
- https://firefly-732-custom-json-parser.irsakudev.ipac.caltech.edu/applications/sofia/

- fork json-bigint to fix its inability to handle scientific notation and long decimal.
- add flag to allow JSON with bigint support only if requested.
- rollback chart's support for bigint as it's not applicable

JSON parsing with BigInt support is limited to only a few Table's calls where it expects original table data with BigInt.
Although, JSON.stringify with BigInt support is available, it need to explicitly request it.

Fork of json-bigint is here: https://github.com/Caltech-IPAC/json-bigint

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "fixed-data-table-2": "~1.2",
     "prismjs": "~1.22",
     "blissfuljs": "~1.0",
-    "json-bigint-patch": "^0.0.4"
+    "json-bigint": "git://github.com/Caltech-IPAC/json-bigint.git"
   },
   "devDependencies": {
     "@babel/core": "~7.8",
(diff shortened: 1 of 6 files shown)


robyww (approved):
Changes look good. This has less impact on the code and less unexpected errors.

I also looked at https://github.com/Caltech-IPAC/json-bigint
The code looks good and i hope the PR will be accepted by the developer.

@loitly and I discussed this but I will say it as a  way of documentation. There does not appear to ever be a reason use scientific notation and expect it to be parsed into a 64 bit long (`BitInt`). It should always be parse into a double (`number`).

ejoliet:
But now the chart axis label is wrong if i plot the file attached in IRSA-3996. Before at leas the value was displayed with E format, now it shows 1,2,3,4,5,6? 
I believe we will have to fork plotly too to address the same parsing problem?

PR #1126, opened 2021-09-09
IRSA-3475: added template index.html and generator 

ejoliet:
The program is generating an html using properties file given a hips url root.
This is to answer the IRSA request IRSA-3475.
Should be a start for making a template HTML to be used for HiPS internally and replace CDS ones.
Should be safe to merge, no collision with other things.


robyww (approved):
Looks fine.

PR #1127, opened 2021-09-10
IRSA-4223: 'accept period' doesn't work in new build

loitly:
Ticket:  https://jira.ipac.caltech.edu/browse/IRSA-4223
Test   :  https://irsa-4223-timeseries-accept-period.irsakudev.ipac.caltech.edu/irsaviewer/timeseries

Phase folded table was originally written as formatted strings.  It did not get converted into real values when we moved to client-side formatting.  This PR fixes this problem by converting the client phase folded table into real values.
Follow instructions in the ticket to test.


robyww (approved):
I tested it and it works. I am surprised how much code was involved in the fix but all looks fine.

lrebull:
Works!!!

PR #1129, opened 2021-09-15
IRSA-4233: Add shared.work.directory concept to GWT codebase

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-4233

When the concept of `shared.work.directory` was introduced in 2018, it was not implemented in GWT.  As a result, GWT based applications are writing all of the 'shared' work files into its local filesystem and are not accessible by other instances of the same app.

Also:
- init stats.log.dir on start up as needed

@robyww  This is more like a FYI.  I don't expect you test it.


ejoliet (approved):
Agree.
Looks ok.

PR #1131, opened 2021-09-18
firefly-843_Add_wavelength_unit: add units for wavelength to spectral-chart unit conversion 

cwang2016:
https://jira.ipac.caltech.edu/browse/FIREFLY-843?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

- add units A (Angstroms), nm(nanometer), mm(millimeter) to support the wavelength units. 

Test:
https://fireflydev.ipac.caltech.edu/firefly-843-add-wavelength-unit/firefly/

- upload the test votable files (attached with above ticket link), each file containing a 'Wavelength' field for each supported wavelength unit, i.e.  A, nm, um, mm, cm, m. 
- pick one chart and open up the 'Plot Parameters' panel to check if the Spectral axis column(x) changes correctly when different unit is selcted from the drop-down list of  'Spectral axis units'. 
- click the setting and get the 'Plot Parameters' panel for each chart, 
- make the same selction to 'Spectral axis units' for each chart and check if all charts shows the same chart content.

### DIFF
--- a/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
+++ b/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
@@ -53,7 +53,7 @@ const UnitXref = {
             label: '𝛎 [%s]'
         },
         wavelength: {
-            options: [{value:'um'}, {value:'cm'}, {value:'m'}],
+            options: [{value: 'A'}, {value: 'nm'}, {value:'um'}, {value: 'mm'}, {value:'cm'}, {value:'m'}],
             label: 'λ [%s]'
         },
         flux_density: {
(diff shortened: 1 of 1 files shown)


loitly (approved):
Changes look good.  I like you using exponential notation for this.  Good to go.

gpdf (commented):
Reviewed code and tested with sample files.

Works as desired, thank you!

A couple of comments on the code:

- I agree that the use of scientific notation in the code makes it easier to read/review.  However, it should then be used consistently, not mixing "10000.0" with "1.00E+6"; this will facilitate review and future maintenance.
- Please say "1.0E+6" in preference to "1.00E+6" - it's more conventional, there's no need for the extra zero, and it raises a question in a reader's mind as to whether it's there for a reason.
- For future changes of this nature, code review would be easier if the change were divided into two commits: first one applying the notation change (or any other reformatting/refactoring) and then one making the substantive change in behavior (here, adding support for the new units).  The two commits can always be squashed before final merge if that's desired.  I am not asking for this PR to be re-done that way - it was easy enough to review - but it's a really helpful approach in general.

gpdf:
I wrote:
> I agree that the use of scientific notation in the code makes it easier to read/review. However, it should then be used consistently, not mixing "10000.0" with "1.00E+6"; this will facilitate review and future maintenance.

I didn't think about this from the user perspective - of course these formulas are actually being displayed to them in the plot options dialog.  With that in mind: users are probably going to be irritated by seeing "1.0E+1" instead of "10".

So I would modify my feedback like this: I would leave "10" and "100" alone in the code but I would convert the instances of "1000" and "10000" to "1.0E+3" and "1.0E+4", respectively.

robyww:
> For future changes of this nature, code review would be easier if the change were divided into two commits: first one applying the notation change (or any other reformatting/refactoring) and then one making the substantive change in behavior (here, adding support for the new units). The two commits can always be squashed before final merge if that's desired. I am not asking for this PR to be re-done that way - it was easy enough to review - but it's a really helpful approach in general.

I really don't want to ask the developers to do that. Also, it is against an established process. We usually do 1 commit for the initial work and 1 commit to respond to the feedback.

It is too much juggling to try to get the commits just right and it might not be how the developer is working.

cwang2016:
> I wrote:
> 
> > I agree that the use of scientific notation in the code makes it easier to read/review. However, it should then be used consistently, not mixing "10000.0" with "1.00E+6"; this will facilitate review and future maintenance.
> 
> I didn't think about this from the user perspective - of course these formulas are actually being displayed to them in the plot options dialog. With that in mind: users are probably going to be irritated by seeing "1.0E+1" instead of "10".
> 
> So I would modify my feedback like this: I would leave "10" and "100" alone in the code but I would convert the instances of "1000" and "10000" to "1.0E+3" and "1.0E+4", respectively.

yes, this is my original concern. 100.0 is bettern than 1.0E+2, and 1.0E+5 is better than 100000.0 for user to read. 

Here are the changes I am going to make, 
- I will change 1000.0 & 10000.0 to 1.0E+3 and 1.0E+4. 
- change 1.00E+xx to 1.0E+xx

Question, 
- for 10 (& 100),  should that be displayed like 10.0  or 10?

gpdf:
I think it would be better to have it less wordy, for readability, so I prefer "10" to "10.0".  

The only caveat is that I want to be sure that changing from "/10.0" to "/10" doesn't somehow end up triggering the actual computation to be done as integer division.

PR #1136, opened 2021-10-06
Firelfy-838: Extraction of data from a FITS image

robyww:
### Firelfy-838: Extraction of data from a FITS image  
  - Cube z-axis extraction
  - Line extraction across an imagae
  - Points extraction by click on points
  - Direct extraction from fits file, no extra data read into memory
  - UI syncing the images, charts and tables together after extraction
  - New icons for all three extractions types
  - All images are now uncompressed
  - Create a spectral data model for cube extraction with wavelength data
  - If there is a HDU name that begin with ERR use that for spectrum error
  - Improved Catalog drawing layer to handle catalogs of image pt

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-838

### Testing
 - url: https://fireflydev.ipac.caltech.edu/firefly-838-cube-extractor/firefly/
 - reading in a fits cube or plain FITS image.
 - Us tool pull down to find 3 types of extraction.
 - Line and point work on any fits 
 - Z-axis works on cubes
 - Results immediately 
 - Use Keep to create a table
 - Use save to save the table right away.
 - Spectral cubes will create a spectral chart when you keep

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/table/MetaConst.java
@@ -112,13 +112,6 @@ public class MetaConst {
      * */
     public static final String DEFAULT_CHART_Y_COL= "defaultChartYCol";
 
-    /**
-     * json parameters to add additional chart parameter to the default chart
-     * used with DEFAULT_CHART_X_COL and DEFAULT_CHART_Y_COL
-     * not used with DEFAULT_CHART_DEF
-     */
-    public static final String ADDITIONAL_CHART_DEF= "additionalChartDef";
-
     /** the column name with access rights info;  true if (public, 1, or true), otherwise false  */
     public static final String DATARIGHTS_COL = "DATARIGHTS_COL";
 
--- a/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
@@ -362,7 +362,7 @@ public final class FITSTableReader
                                                 List<FitsReadUtil.ExtractionResults> results,
                                                 String wavelengthColName, String fluxColName) {
 
-        dataGroup.getTableMeta().addKeyword("utype", "spec:Spectrum");
+        dataGroup.getTableMeta().addKeyword(TableMeta.UTYPE, "spec:Spectrum");
         GroupInfo wlGr = new GroupInfo("spec:Data.SpectralAxis", "");
         GroupInfo.RefInfo wlref= new GroupInfo.RefInfo(wavelengthColName, "", "spec:Data.SpectralAxis.Value");
         wlGr.setColumnRefs(Collections.singletonList(wlref));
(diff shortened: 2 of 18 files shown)


lrebull:
Do you have sample FITS files for us to try?

lrebull:
Wow, this is really cool.
trying the "drill" icon tool.

1. I know you probably don't want detailed comments like this, but a possible ambiguity in wording -- "download", empirically, refers to downloading the data table, but because this is a plot, i thought initially that it meant to download the plot. Maybe that can be a choice in the pop-up? or one of the choices in the download file format?
2. After I click "keep", and the pop-up window is closed, the main image pane jumps back to the first plane, as opposed to the plane I left it on (15).
3. If you try to make it go to a plane by typing in a number larger than the number of planes, it just drops you back in the first plane. (obviously not a deal breaker, but had me confused for a bit.)
4.  After messing around with it for a while, i have two target markers on my image, and i can't make them go away. Were they supposed to be the locations of the drill down? I've now deleted all the tables and they are still there. I tried again, and now those target markers sure seem like they are supposed to be the locations of the drilldown, but they are offset. (see movie attached next item)
5. when scrolling through the rows in my "kept" drilldown, the image jumps, or rather the markers jump? maybe both? see movie: https://user-images.githubusercontent.com/23038600/136466471-468b5fd8-9422-419a-99bf-898a77c27701.mp4

I haven't tried anything other than the "drill" icon tool yet. More later.

robyww:
> 1. I know you probably don't want detailed comments like this, but a possible ambiguity in wording -- "download", empirically, refers to downloading the data table, but because this is a plot, i thought initially that it meant to download the plot. Maybe that can be a choice in the pop-up? or one of the choices in the download file format?

- [x] Good Suggestion: I will make it `Download Table` and add a `Download Chart`.

> 2. After I click "keep", and the pop-up window is closed, the main image pane jumps back to the first plane, as opposed to the plane I left it on (15).

- [x] FIXED: I was aware of that but you comment helped me think a solution.

> 3. If you try to make it go to a plane by typing in a number larger than the number of planes, it just drops you back in the first plane. (obviously not a deal breaker, but had me confused for a bit.)

I would have to see this one, I though it gave you the error indicator. It does for me, maybe you are doing something different.

> 4. After messing around with it for a while, i have two target markers on my image, and i can't make them go away.


- [ ] They go away for me. I will study your movie.


> 5. when scrolling through the rows in my "kept" drilldown, the image jumps, or rather the markers jump? 

- [x] FIXED:  I am going to center the marker points.

_Details:_

The markings are the point that was extracted. They are on the bottom left of the pixel. I will make them center on the pixel. You will see this again at other times like line extraction. This is pretty standard for fits viewers. If I plot the point 5,5 on a image with a zoom level of like 20 you will see them on  the bottom left corner. To center them I have to make the point 5.5,5.5. That is fine for the markers. However, line extraction that shows up in the table will plot them on the bottom left, because it is plotting the table data.

The jumping is the image repositioning itself. Sometime it is slow enough to see. I think it ends up in the right place. So, for now at least, I am not going to do anything about it.

- [x] FIXED:  I am going to call this one fix since the only action item is to center the marker points.

lrebull:
video demonstrating my #3 which you said you couldn't reproduce
https://user-images.githubusercontent.com/23038600/136476082-e5cc515e-fee9-4cdd-bf72-2e139322644c.mp4

lrebull:
checking new version from slack. markers still seem to me to be offset but maybe it's just the big pixels ?

Ooh, a help anchor has appeared! I need to spend some time figuring out what the docs framework needs to be - i may very have to reorganize substantially to accomodate the rework (not just the new toolbar stuff but the reorg may mean the docs need to be comprehensively changed). I'm not sure I'll really be able to "see" the new org until i see more of the new tool, but i can try.

PR #1142, opened 2021-10-23
Firefly-874: only read cube planes as user request them.

robyww:
#### Firefly-874: only read cube planes as user request them. 
  - lazy read the cube planes as the user visualizes them
  - added optimizations in transferring and caching data
  - server foot print substantially smaller with large cubes
  - fixed: after changing color the flux readout begin to fail

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-874

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-874-lazy-cube/firefly/
 - read in a FITS file with any cube. It should work as before, performance will be a little better with large cubes


loitly (approved):
I have no idea how to review this.  The feature is definitely needed for large cube.  The code passes all of the tests.  Should be ok to merge.

PR #1145, opened 2021-11-03
Firefly-866: small tweeks and many bug fixes

robyww:
#### Firefly-866: small tweeks and many bug fixes

 - Clean up of text and help according to the ticket
 - Option in upload panel to treat a file as a spectrum
 - Convert to passing line data points to server, include world coordinates, compute offset
 - Clean up of offsets for line tool, rounding, units
 - Color of extraction markers the same, shows active tab
 - Pin extraction clearer, better message
 - FIREFLY-868 Add world-coordinate measurement to line-extraction tool
 -  fixed: slate problem with new window appearing below (need a ticket?)

_tickets:_
https://jira.ipac.caltech.edu/browse/FIREFLY-866
https://jira.ipac.caltech.edu/browse/FIREFLY-868

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-866-extraction/firefly/
 - Implement Changes listed in the ticket (866)
 - Implemented line extraction table saving and graphic according to the ticket (868)

### DIFF
--- a/src/firefly/js/visualize/ui/ExtractionDialog.jsx
+++ b/src/firefly/js/visualize/ui/ExtractionDialog.jsx
@@ -49,7 +49,7 @@ import {
     dispatchAttachLayerToPlot, dispatchCreateDrawLayer,
     dispatchDetachLayerFromPlot, dispatchModifyCustomField, getDlAry } from 'firefly/visualize/DrawLayerCntlr.js';
 import {makeImagePt} from 'firefly/api/ApiUtilImage.jsx';
-import {getXorYinEquation} from 'firefly/visualize/VisUtil.js';
+import {computeDistance, computeScreenDistance, getLinePointAry} from 'firefly/visualize/VisUtil.js';
 import {RadioGroupInputFieldView} from 'firefly/ui/RadioGroupInputFieldView.jsx';
 import {addZAxisExtractionWatcher} from 'firefly/visualize/ui/ExtractionWatchers.js';
 import {sprintf} from 'firefly/externalSource/sprintf.js';
@@ -58,6 +58,7 @@ import {Band} from 'firefly/visualize/Band.js';
 import {onTableLoaded} from 'firefly/tables/TableUtil.js';
 import {showTableDownloadDialog} from 'firefly/tables/ui/TableSave.jsx';
 import {getAppOptions} from 'firefly/api/ApiUtil.js';
+import {showPinMessage} from 'firefly/ui/PopupUtil.jsx';
 
 
 
@@ -117,7 +118,7 @@ function getStoreState(prevPv) {
 }
 
 const sizeOp=[];
-for(let i= 1; i<=5;i++) sizeOp.push({label:`${i}x${i}`, value:i});
+for(let i= 1; i<=7;i+=2) sizeOp.push({label:`${i}x${i}`, value:i});
 
 function afterZAxisChartRedraw(imPt, pv, chart) {
     chart.on('plotly_click', (ev) => {
@@ -132,17 +133,18 @@ function afterZAxisChartRedraw(imPt, pv, chart) {
     });
 }
 
-function afterLineChartRedraw(pv, chart,pl, pt1, pt2) {
+function afterLineChartRedraw(pv, chart,pl, imPtAry, pt1, pt2) {
     chart.on('plotly_click', (ev) => {
         setTimeout( () => {
             const plot= primePlot(pv);
-            const {x,y}= getXorYinEquation(pt1,pt2,ev.points[0].x);
-            const imPt= makeImagePt(Math.round(x)+.5, Math.round(y)+.5);
+            const imPt= makeImagePt(Math.round(imPtAry[ev.points[0].pointNumber].x)+.5, Math.round(imPtAry[ev.points[0].pointNumber].y)+.5);
             dispatchModifyCustomField( ExtractLineTool.TYPE_ID,
                 {activePt: hasWCSProjection(plot) ? CCUtil.getWorldCoords(plot,imPt) : imPt},
                 pv.plotId);
             dispatchAttributeChange({plotId:plot.plotId,overlayColorScope:false,toAllPlotsInPlotView:false,
-                changes:{[PlotAttribute.SELECT_ACTIVE_CHART_PT]: {x:ev.points[0].x,y:ev.points[0].y}}
+                changes:{
+                    [PlotAttribute.SELECT_ACTIVE_CHART_PT]: {x:ev.points[0].x,y:ev.points[0].y}
+                }
             });
         },5);
     });
@@ -258,6 +260,7 @@ function PointExtractionPanel({canCreateExtractionTable}) {
 
 function LineExtractionPanel({canCreateExtractionTable}) {
     const [{plotlyDivStyle, plotlyData, plotlyLayout},setChartParams]= useState({});
+    const [imPtAry,setImPtAry]= useState(undefined);
     const [pointSize,setPointSize]= useState(1);
     const [allRelatedHDUS,setAllRelatedHDUS]= useState(true);
     const [pv]= useStoreConnector( getStoreState);
@@ -279,15 +282,21 @@ function LineExtractionPanel({canCreateExtractionTable}) {
     useEffect(() => {
         const getData= async () => {
             if (ipt1 && ipt2 && plot) {
-                const {startValue,direction}= xLineData(ipt1,ipt2);
-                const dataAry= await callGetLineExtractionAry(plot.plotState, hduNum, plane, ipt1, ipt2, pointSize, allRelatedHDUS);
+                const {direction}= xLineData(ipt1,ipt2);
+                const newImPtAry= getLinePointAry(ipt1,ipt2);
+                if (!newImPtAry?.length) return;
+                const cc= CysConverter.make(plot);
+
+                const dataAry= await callGetPointExtractionAry(plot.plotState, hduNum, plane, newImPtAry, pointSize, allRelatedHDUS);
                 const chartTitle= makeLineExtractionTitle(pv,x1,y1,x2,y2);
-                const xDataAry= dataAry.map((d,i) => Math.trunc(startValue+(i)));
-                const x= !isNaN(chartX) ? chartX : usingXAxis(ipt1,ipt2) ? x1:y1;
-                const y= !isNaN(chartY) ? chartY : dataAry[direction>0?0:dataAry.length-1];
+                const pt0= hasWCSProjection(plot) ? cc.getWorldCoords(newImPtAry[0]) : newImPtAry[0];
+                const xOffAry= hasWCSProjection(plot) ?
+                    newImPtAry.map( (pt) => computeDistance(pt0,cc.getWorldCoords(pt))*3600) :
+                    newImPtAry.map( (pt) => computeScreenDistance(pt0.x,pt0.y,pt.x,pt.y));
                 const chartData=
-                    genSliceChartData(ipt1,ipt2,xDataAry, dataAry, x, y, pointSize, chartTitle, direction<0);
+                    genSliceChartData(plot, ipt1,ipt2,xOffAry, dataAry, chartX, chartY, pointSize, chartTitle, direction<0);
                 setChartParams(chartData);
+                setImPtAry(newImPtAry);
             }
             if (!plot) cancelLineExtraction();
         };
@@ -299,7 +308,7 @@ function LineExtractionPanel({canCreateExtractionTable}) {
             allRelatedHDUS, setAllRelatedHDUS, pointSize, setPointSize, canCreateExtractionTable,
             plotlyDivStyle, plotlyData: extractionData&&plotlyData, plotlyLayout,
             startUpHelp: 'Draw line on image to extract point on line and show chart',
-            afterRedraw: (chart,pl) => afterLineChartRedraw(pv,chart,pl,makeImagePt(x1,y1), makeImagePt(x2,y2)),
+            afterRedraw: (chart,pl) => afterLineChartRedraw(pv,chart,pl,imPtAry,makeImagePt(x1,y1), makeImagePt(x2,y2)),
             callKeepExtraction: (download) => keepLineExtraction(ipt1,ipt2, pv, plot, plot.plotState.getWorkingFitsFileStr(), hduNum, plane, pointSize, download),
             cancelFunc: () => cancelZaxisExtraction()
         }} /> );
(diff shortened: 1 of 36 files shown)


ejoliet (commented):
I started to upload test image but i'm facing a problem. If i use the upload url, it works fine but if i use the same file downloaded and upload it via 'use image' or via 'upload' file, it takes forever to load. No error in the console.

Example URL: https://irsa.ipac.caltech.edu/data/COSMOS/images/herschel/pacs/pep_COSMOS_100um_blind_Residuals.3sigma.DR1.fits

I'll continue testing from URL upload for now.

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -600,13 +628,12 @@ function makePlotlyDataObj(xDataAry, yDataAry,x,y,ttXStr,ttYStr, makeXDesc= (i)=
 }
 
 
-function genSliceChartData(ipt1,ipt2, xDataAry,yDataAry,x,y, pointSize, title, reversed) {
-    const xAxisTitle= usingXAxis(ipt1,ipt2) ? 'Image X' : 'Image Y';
-
+function genSliceChartData(plot, ipt1,ipt2, xDataAry,yDataAry,x,y, pointSize, title, reversed) {
+    const unitStr= hasWCSProjection(plot) ? ' (arcsece)' : '';
```
ejoliet, line 632: i believe this is a misspelled `arcsece`, should be `arcsec`.

ejoliet (commented):
I've played with many example from here: 
https://confluence.ipac.caltech.edu/display/FIREFLY/Example+MEF%2C+Cube%2C+Spectral+Image+FITS+Files

With one example, i couldn't see the plane updated when i highlight different rows. 
File example: https://irsa.ipac.caltech.edu/data/Herschel/KINGFISH/galaxies/NGC3077/PACS/NGC3077_CII_cube.fits

I'll continue a little bit to test other files but so far so good.

ejoliet:
Apart from the upload problem i was facing, i still thing there is something going on with the 'image search' component, once i hit the button, it takes time to render and then everyhting is slower than other components. Was there something that has changed in the way we render the 'image search' component? 

It appears to me really slower than before, enough that i can notice a lag in responsiveness which is not happening in irsadev in IRSAViewer for example. Specially after a result come back.

robyww:
> I started to upload test image but i'm facing a problem. If i use the upload url, it works fine but if i use the same file downloaded and upload it via 'use image' or via 'upload' file, it takes forever to load. No error in the console.

I have repeated this.  i will look at it.

robyww:
> It appears to me really slower than before, enough that i can notice a lag in responsiveness which is not happening in irsadev in IRSAViewer for example. Specially after a result come back.

Do you have a specific example?

lrebull:
Genuine question. In the upload panel, the "treat tables as a spectrum" box is supposed to do what, exactly? Elsewhere in the tool when it is in "plot a spectrum" mode, the points are connected by a line. I've ticked that box and read in a file, and the points are not connected by a line. What does it gain me by ticking the "treat tables as a spectrum" box?
Separately: table is singular, tables is plural. Spectrum is singular, spectra is plural. It should be one of these:
* treat table as spectrum
* treat tables as spectra

lrebull:
Bug:
Load in anything with multiple HDUs and planes (i am using the SOFIA product you gave me called _data_SOFIA_FIFI-LS_L4_2016-07-05_FI_F316_p3993_g10_F0316_FI_IFS_04015210_BLU_WXY_600284-700065.fits). Launch the drill tool. Change the plane to anything -- I'm using 25 out of 49. Click anywhere to drill down. Pin the table. The image reloads back at the first plane. It should stay in the plane I was using.

lrebull:
I am working through the stuff listed in FIREFLY-866 could have SWORN I saw something go by about this item:

> When you select a table that represents extracted data, how do you tell which target symbol that data corresponds to? Just make all the targets the same color 
>    Highlight the one that corresponds to the selected table tab.

Specifically, it was about the secondary bullet here -- highlight the one that corresponds to the selected table tab. It's not doing it for me right now. Should it be doing it?

lrebull:
Your solution to this one:
>The table tabs should have the same name appear in the layers dialog. Might need to remove the word "Point" to make it fit on the tab.
is really nice! I love it.

lrebull:
I think this means just in the table you save when you extract a line along the image plane?
> Create an simple angular distance column alone the line, create a column and slit into pieces, put angular distance in Meta
I don't see an angular distance in the metadata for the table that I download after drawing a line on the image.

lrebull:
I am confused about this one:
> “Selected point” - maybe a better wording that refer to the drill feature? “Drilled point”?

I see "point extract preview" and the resultant pinned tab is "Extract pts- FLUX-Plane: 25-17", which leads me to another question. I picked "1x1" aperture and clicked on several points in plane 25, so why is it giving me planes 25-17? Or does it mean "17 points out of plane 25"? This is confusing.

lrebull:
Changing colors or symbols doesn't work on the extracted points. Steps to reproduce:

1. load a multi-plane, multi-HDU image in
2. drill down and pin several locations.
3. go to layers menu
4. now i can turn off just one of them, but changing colors doesn't work. Click on "color" to change the point color, pick another color, click "OK" and nothing happens.
5. Symbols are the same thing.

it works just fine for extracted lines or points; it's just the saved drill downs that fail like this.

lrebull:
1. load in more than one multi-plane, multi-HDU file. I'm using _data_SOFIA_FIFI-LS_L4_2016-07-05_FI_F316_p3993_g10_F0316_FI_IFS_04015210_BLU_WXY_600284-700065.fits and F0316_FI_IFS_04015210_RED_WXY_600284-700027.fits. These happen to cover the **same patch of sky** and this turns out to be important.
2. Select one of the two. 
3. Pick the line extraction tool and draw a line on the active image.
4. the line appears on both of the images, leading to confusion over whether the extracted pop-up window is giving me a plot of the active image or the non-active image, or both.

let me know if you want me to make a ticket for this one.

lrebull:
1. load a multi-plane, multi-HDU image
2. add some drill downs, pin; add a line extraction, pin; do a point extraction, pin.
3. click on the coverage tab. Note that the point extraction points are on the coverage image, but none of the other points are. I think this is confusing and it should either all be carried over, or none of them should.

let me know if you want me to make a ticket for this one.

robyww:
> Genuine question. In the upload panel, the "treat tables as a spectrum" box is supposed to do what, exactly? Elsewhere in the tool when it is in "plot a spectrum" mode, the points are connected by a line. I've ticked that box and read in a file, and the points are not connected by a line. What does it gain me by ticking the "treat tables as a spectrum" box?

@lrebull 
This means that it will attempt to guess the column names and upload the table as a spectrum data model. If you were at the IVOA meeting, I demoed this. I upload an ipac table that was an IRS spectrum with the check box on. Otherwise it would not have known what to do with the table and would have plotted the first two columns against each other.

What text should I use for the check box?

lrebull:
> > Genuine question. In the upload panel, the "treat tables as a spectrum" box is supposed to do what, exactly? Elsewhere in the tool when it is in "plot a spectrum" mode, the points are connected by a line. I've ticked that box and read in a file, and the points are not connected by a line. What does it gain me by ticking the "treat tables as a spectrum" box?
> 
> @lrebull This means that it will attempt to guess the column names and upload the table as a spectrum data model. If you were at the IVOA meeting, I demoed this. I upload an ipac table that was an IRS spectrum with the check box on. Otherwise it would not have known what to do with the table and would have plotted the first two columns against each other.
> 
> What text should I use for the check box?

Ok. Hm. Well, at least it should be "treat tables as spectra" but maybe we need to add some wishy-washy words, like "attempt to read in tables as spectra" so that if the tables that are being uploaded don't follow the data model, then people shouldn't expect magic..? Maybe we should discuss at the CCB.

robyww:
@lrebull 
>  Launch the drill tool. Change the plane to anything -- I'm using 25 out of 49. Click anywhere to drill down. Save the table as a tbl. The image reloads back at the first plane. It should stay in the plane I was using.

I remember this issue from a while back and I fixed it. I can't duplicate it now, it works for me. I tried several different combination with the file you were using.

lrebull:
> @lrebull
> 
> > Launch the drill tool. Change the plane to anything -- I'm using 25 out of 49. Click anywhere to drill down. Save the table as a tbl. The image reloads back at the first plane. It should stay in the plane I was using.
> 
> I remember this issue from a while back and I fixed it. I can't duplicate it now, it works for me. I tried several different combinations with the file you were using.

The behavior when 'saving' the table is the new thing. It did it before, I agree, but when I "download as table", then it flips back.

robyww:
@lrebull 
> Or does it mean "17 points out of plane 25"?

That is what it means. I could say `Extract pts- FLUX - Plane: 25 - 17 pts` or `Points - FLUX - Plane: 25 - 17 pts`. As we have discussed, dynamic name is always difficult. 

How do you want to title it?

robyww:
Aww download as table!!
> The behavior when 'saving' the table is the new thing. It did it before, I agree, but when I "**download as table"**, then it flips back.

I see it now. thanks.

lrebull:
> @lrebull
> 
> > Or does it mean "17 points out of plane 25"?
> 
> That is what it means. I could say `Extract pts- FLUX - Plane: 25 - 17 pts` or `Points - FLUX - Plane: 25 - 17 pts`. As we have discussed, dynamic name is always difficult.
> 
> How do you want to title it?

ok I see now. I would take out the "17 pts" because that's clear by looking at the table itself. Even if it has >100 points, there is still an indication of the total number of the rows in the top center of the table. It's really likely that users (like me) will ahve enough tables that the table name gets truncated so you just see the "25-17" part and not the "pts" part.

robyww:
>  would take out the "17 pts 

ok - i still need to put a counter there. I guess I will make it: `Points 1- FLUX - Plane: 25`, then ``Points 2- FLUX - Plane: 25``

lrebull:
> > would take out the "17 pts
> 
> ok - i still need to put a counter there. I guess I will make it: `Points 1- FLUX - Plane: 25`, then `Points 2- FLUX - Plane: 25`

that seems good, i think, but may be good to run it past CCB

robyww:
> I don't see an angular distance in the metadata for the table that I download after drawing a line on the image. Maybe I'm not looking in the right place or for the right term?

The offset column is the distance in arcsec from the stating point. I think i did what the ticket wanted. Maybe I missed something.

robyww:
> load in more than one multi-plane, multi-HDU file. I'm using _data_SOFIA_FIFI-LS_L4_2016-07-05_FI_F316_p3993_g10_F0316_FI_IFS_04015210_BLU_WXY_600284-700065.fits and F0316_FI_IFS_04015210_RED_WXY_600284-700027.fits. These happen to cover the same patch of sky and this turns out to be important.
Select one of the two.
Pick the line extraction tool and draw a line on the active image.
the line appears on both of the images, leading to confusion over whether the extracted pop-up window is giving me a plot of the active image or the non-active image, or both.
let me know if you want me to make a ticket for this one.

This is working at intended. The extraction is on the highlighted image. I think it is a feature not a bug.

lrebull:
> > load in more than one multi-plane, multi-HDU file. I'm using _data_SOFIA_FIFI-LS_L4_2016-07-05_FI_F316_p3993_g10_F0316_FI_IFS_04015210_BLU_WXY_600284-700065.fits and F0316_FI_IFS_04015210_RED_WXY_600284-700027.fits. These happen to cover the same patch of sky and this turns out to be important.
> > Select one of the two.
> > Pick the line extraction tool and draw a line on the active image.
> > the line appears on both of the images, leading to confusion over whether the extracted pop-up window is giving me a plot of the active image or the non-active image, or both.
> > let me know if you want me to make a ticket for this one.
> 
> This is working at intended. The extraction is on the highlighted image. I think it is a feature not a bug.

But not every extraction, just one of them. That seems weird to me, but if that was what was desired by the CCB, so be it.

robyww:
> click on the coverage tab. Note that the point extraction points are on the coverage image, but none of the other points are. I think this is confusing and it should either all be carried over, or none of them should.

The idea here is that you are actually making a catalog on the fly. That is why I handled it differently.

lrebull:
> > click on the coverage tab. Note that the point extraction points are on the coverage image, but none of the other points are. I think this is confusing and it should either all be carried over, or none of them should.
> 
> The idea here is that you are actually making a catalog on the fly. That is why I handled it differently.

ok!

robyww:
- [x] FIXED: from @ejoliet -  i believe this is a misspelled arcsece, should be arcsec.
- [x] FIXED: from @ejoliet - With one example, i couldn't see the plane updated when i highlight different rows.
- [x] FIXED: from @lrebull - "download as table" put cube at plane 0
- [x] FIXED: from @lrebull - point tab title

robyww:
Changes in last commit

 - Pined Line extraction is now treated as a catalog if there is a WCS (like point extraction)
 - _Fixed_: User now can changed the color of a z-axis extracted point
 - Z-axis extraction points are grouped together in the layer dialog
 - _Fixed_: Line extraction now works with switching images
 - Point extraction now has a mode to switch images without adding a point
 - Extraction dialog now adds the image title to the dialog name
 - _Optimizations_:
     - chart with over 6000 points now chart with webgl
     - if fits uses 32 bit floats then data is return as 32 bit instread of 64
     - Tested with 60000 plane cube
 - _Not changed_: "Selected Point" in layer dialog will stay the same since it is used in other components

lrebull:
I am going to try to do this all in one comment (wish me luck).

Fixed stuff:

- tab names for pinned things
- pinned line pixels overlaid on coverage just like saved points
- saving table doesn't change what plane i'm on
- (my fault on the angular distance stuff; i see now)
- woo as far as i can tell, color changing works even though you said you thought it might be too hard!
- extract popup does have the image name now!
- i *think* it is behaving better with the line updating as I go between images!
- I love that the extraction points are grouped in the layer dialog.

honestly, I do not know how to explicitly test the optimizations you added.

as far as i can tell, this is the only thing that got lost in the shuffle. In the upload screen, the tickbox at the bottom got changed, but not all the way. It should be “Attempt to interpret tables as spectra”

woo hoo!

robyww:
> honestly, I do not know how to explicitly test the optimizations you added.

> as far as i can tell, this is the only thing that got lost in the shuffle. In the upload screen, the tickbox at the bottom got changed, but not all the way. It should be “Attempt to interpret tables as spectra”

-  Oops - yes, I forgot the text change. Thank you.
- You can't test the optimizations. I had a 60,000 plane cube that was not performing well and locking up the browser. It is radically better.
- There is a new build with the text fixed. - https://fireflydev.ipac.caltech.edu/firefly-866-extraction/firefly/

PR #1146, opened 2021-11-04
IRSA-4289_ztf_icon_in_the_upper_left_is_transparent

deannaji:
Jira: https://jira.ipac.caltech.edu/browse/IRSA-4289
K8s: https://irsa-4289.irsakudev.ipac.caltech.edu/applications/ztf/

### DIFF
--- a/src/firefly/js/ui/Banner.css
+++ b/src/firefly/js/ui/Banner.css
@@ -9,6 +9,9 @@
 
 .banner__left {
     flex-grow: 0;
+    background-color: #3d3d3d;
+    border-radius: 0 20px 0 0;
+    border-top: 1px solid #eee;
 }
 
 .banner__middle {


loitly (changes requested):
These changes should not be in firefly.  Instead, for ZTF, change it here:
https://github.com/IPAC-SW/irsa-ife/blob/6f0434d02d6abe174a65083a4b9ef9dd45f9b8fe/src/irsa/js/ztf/Ztf.jsx#L27

You can do the same for other projects at its corresponding places.

src/firefly/js/ui/Banner.css
```diff
@@ -9,6 +9,9 @@
 
 .banner__left {
     flex-grow: 0;
+    background-color: #3d3d3d;
```
loitly, line 12: Also, I think #3e3e3e matches the black better.

robyww (changes requested):
This is going to affect others app beyond ZTF so we can't just change this just to make ZTF look better. In general, I don't really like the way this looks and I am concerned it will not look right in other apps.  We need to discuss exactly what the goal is.

ejoliet:
> This is going to affect others app beyond ZTF so we can't just change this just to make ZTF look better. In general, I don't really like the way this looks and I am concerned it will not look right in other apps. We need to discuss exactly what the goal is.

@deannaji What was the final proposal to improve the icon UI?

ejoliet:
(BTW, i think it looks good in any app - where is not going to fit well? TimeSeries? )

robyww:
I gave @deannaji an Icon that works well with ZTF.  It looks quite good without having to mess with any CSS layout issue.
Screen Shot 2021-11-08 at 4 57 29 PM

I think if we have icons that don't look right this is more the way to go.

PR #1147, opened 2021-11-04
Correct typos in File Analyzer error messages

gpdf:
Corrected "Cannot not" to "Cannot" in two error messages


robyww (approved):
Thank you!

PR #1149, opened 2021-11-10
Firefly-841: UI improvements

robyww:
Firefly-841: UI improvements

  - Expose more zoom buttons to the image embedded zoom menu
  - Enabled options to control and turn off the active tool on the image (distance, extract, area select)
  - Shift key will always allow for only scrolling

_ticket:_
https://jira.ipac.caltech.edu/browse/FIREFLY-879

_testing_
 - Firefly: https://fireflydev.ipac.caltech.edu/firefly-879-ui-improve/firefly/
 - ZTF: https://firefly-879-ui-improve.irsakudev.ipac.caltech.edu/applications/ztf/
 - FinderChart:  https://firefly-879-ui-improve.irsakudev.ipac.caltech.edu/applications/finderchart/

### DIFF
--- a/src/firefly/js/drawingLayers/ExtractLineTool.js
+++ b/src/firefly/js/drawingLayers/ExtractLineTool.js
@@ -211,7 +211,7 @@ function start(drawLayer, action) {
     const {imagePt,plotId,shiftDown}= action.payload;
     const plot= primePlot(visRoot(),plotId);
     const mode= getMode(plot);
-    if (!plot) return;
+    if (!plot || shiftDown) return;
     const cc= CsysConverter.make(plot);
     if (!cc.pointInData(imagePt)) return;
     let retObj= {};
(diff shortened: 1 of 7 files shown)


lrebull:
Really cool!
suggestion for wording -- will need to be discussed, I think, but here are my votes:

- "finish extraction tool"
- "finish distance tool"
- "finish select tool"

robyww:
It really needs to be shorter. There are cases where there is not much room.

lrebull:
ok, can at least omit "tool" from each one, so it'd be (note last one is different)

- "finish extraction"
- "finish distance"
- "finish selection"

but i do think this should be discussed tomorrow with ccb-science

ejoliet:
I really like the minimal change that gives now a full visibility and control of what is enabled and less click to disable the modes. Great suggestion and implementation.

PR #1151, opened 2021-11-16
FIREFLY-872: display job information in background monitor or in mask panel if not backgrounded.

loitly:
ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-872
- remove unused backgrounding code
- use log4j to output logs to console as well as files when in a docker container

Test: 
https://fireflydev.ipac.caltech.edu/firefly-872-show-job-info/firefly/
https://firefly-872-show-job-info.irsakudev.ipac.caltech.edu/applications/sofia/

**What to expect.**
Job information is available on submitted jobs, like package requests and backgroundable searches.
If you 'send to background', it will appear in the background monitor.  Clicking on the job's header to display 'job info'.
If you left it running in the foreground, you can click on 'progress description' to show 'job info'.  Job info popup will update itself while the job is running.

Job info popup is resizable.  Some key/value pair can be very long, like parameters.  If it's chopped off, a context menu will appear if you mouse over it(similar to table cell).  You can copy to clipboard, display the full text, or display it as JSON formatted text.


robyww:
Feedback on UI:

- It is really great. I think power TAP user will will really like it.

_Important_:
- After the table is loaded `Table Options` -> `Table Meta` -> `ID: IVOA_UWS_JOB` url should open to a new tab not replace firefly

_Small_:
- I really like the `Job Information`, I got to it by clicking on `working/fetch data...`  or the background monitor,  but I can't see how to get to it one the table is loaded. Could it be under the `i` on the table toolbar?
- There should probably be some sort of `i` in the background monitor.  using`js/html/images/info-icon.png`

robyww (approved):
All looks good. See my UI comments.

loitly:
@robyww , I did the other two.
 -  I really like the Job Information, I got to it by clicking on working/fetch data... or the background monitor, but I can't see how to get to it one the table is loaded. Could it be under the i on the table toolbar?
Only backgroundable tables have JobInfo.  So, I can either add the icon to the toolbar when there's JobInfo, or I can add another Tab to the TableOptions.  It would be next to the `Table Meta`.  I think the 2nd option makes more sense, since JobInfo is not that important to be on the toolbar.  What do you think?

robyww:
> What do you think?
 
The makes sense. Maybe we should just remove the `i` altogether?

ejoliet (changes requested):
Ok, i tested as a simple user who doesn't really know what to expect. I have several comments/questions to make it better but we can talk further about solutions to my questions. Bare with me.

The I did a TAP query and i tested the link in the 'metadata' part of the table settings. When i clicked it would got to the link but not in a different tab. I think it should go to a new tab otherwise we will have problem when we go back to the app from it. Actually i did and the table didn't rendered back.

About the background monitor: 
I didn't know that 'AsyncTapQuery' label is clickable! I just clicked looking for more just by surprise. This should a link or a more visual guide to the user to know that there is more info to the job. And again the links in the 'Job information' dialog should go to new tab.
Can the jobId be also copyable like the links? Maybe that would be useful, isnt' what Gregory wanted to put in Python code if you already have the link base root?

Is the 'parameters' object at the bottom very useful? Maybe ok to remove? This is really exposing program jargon which can cause user to ask why all those parameters means - documentation will be terrible.

The 'result' part is confusing: it should be '.../result/rsults'. instead i get an ascii file that i can't parse. What is this link about? What is the purpose? 
Example:this is a result table of a query:
https://irsa.ipac.caltech.edu/TAP/async/39955556/results/result

Instead the job info shows: 
https://fireflydev.ipac.caltech.edu/firefly-872-show-job-info/firefly/CmdSrv/async/1637178698793/results/result

This is not the same and can't be used directly.

Can the job monitor give the high level information about the query, like table name and schema name? Is that useful? Is that available there? Just mentioning because the background monitor has the same name 'AsyncTapQuery' and 'show results', if you have several jobs, how do you know which one you one to click to get more info?

Thanks!

gpdf:
I'm going to do systematic testing this afternoon, after the SPHEREx systems engineering meeting (at which I have to present something).  Sorry for the delay.

robyww:
@ejoliet  you might want to update your comments with @loitly 's new build. Some overlap with mine and are reflected in the build.

robyww:
Screen Shot 2021-11-17 at 2 31 54 PM

This is the type of result that @ejoliet was talking about. There are 2 URL's. The top URL is the backend job UWS xml. I think this is fine. If you want the backend results it does require opening the UWS xml and the selecting the result URL out of that. It might be night to have the UWS `results/result` URL directly available. However, I am concerned that this might be a lot more work and very little benefit. 

The bottom one is the table model json. Our table model is very stable at this point so I don't think it is a problem exposing it and it could help a python developer pick up our results. It would be easy to do. Also I expect we will start using these result URL's in `firefly_client`

@loitly after some testing I did notice that the firefly `results/result` url that returns the table model json only gives a page (first 100 rows). It this intentional? I am not sure what it _should_ do but I wanted to make sure that is what you intended.

ejoliet:
> The bottom one is the table model json. Our table model is very stable at this point so I don't think it is a problem exposing it and it could help a python developer pick up our results. It would be easy to do. Also I expect we will start using these result URL's in firefly_client

I don't think is useful to expose internal data. User won't do anything with that and will cause confusion. Plus, if you expose, you have to add to the docs and explain format, which can change again next time.

This should be changed or removed from the UI in my opinion.

loitly:
Below are my answers to the questions above.  I also built SOFIA so you can see what a download request job looks like.
https://firefly-872-show-job-info.irsakudev.ipac.caltech.edu/applications/sofia/

I am open to limiting the information we show.  Per Trey request, I also added a `Job Info` tab to the table's options popup.  We can decide whether or not to keep that as well.

- link open in current window

This is fixed in `Table Meta` tab.  But, I will also fix existing code that open the link in a `Links` window.  This is fine until you reuse that window/tab to launch firefly.  I will change it to open a new(`_blank`) window instead.

-I didn't know that 'AsyncTapQuery' label is clickable!

I've taken Trey recommendation to have an 'info' icon next to it instead.  My original intention is to have it not so prominent in the UI.  Like you said, it's not normal use-case.

- Can the jobId be also copyable like the links? 

This is not necessary since you can select any text on there and copy it using default browser feature.  This is also NOT the upstream or TAP service jobId.  It is firefly's jobId. 

- Is the 'parameters' object at the bottom very useful 

Like most of the exposed information, they are only useful for debugging or advance user.  If there's an error, you will see `error:  some_error_messages`.  Maybe, the parameters will be useful then?

- The 'result' part is confusing: it should be '.../result/rsults'

This is the results of firefly request.  When it's a table search, then you will get a firefly table in JSON format.  If it's a download package request, you will get one or more url to the zip file(s).

- Can the job monitor give the high level information about the query, like table name and schema name

For TAP search, this should be done at the form submit time.  Background monitor just display what is set as the job's label.  In this case, it's the table's title.

- I did notice that the firefly `results/result` url that returns the table model json only gives the first page

It was not intentional.  The result url is suppose to return the submitted request.  In this case, the request was only for 100 rows.  I will not do anything for now.  This is correct as far as UWS is concern.  If we have use for it, I will change it then.

ejoliet:
> Below are my answers to the questions above. I also built SOFIA so you can see what a download request job looks like. https://firefly-872-show-job-info.irsakudev.ipac.caltech.edu/applications/sofia/
> 
> I am open to limiting the information we show. Per Trey request, I also added a `Job Info` tab to the table's options popup. We can decide whether or not to keep that as well.
> 
> * link open in current window

Great!

> 
> This is fixed in `Table Meta` tab. But, I will also fix existing code that open the link in a `Links` window. This is fine until you reuse that window/tab to launch firefly. I will change it to open a new(`_blank`) window instead.
> 
> -I didn't know that 'AsyncTapQuery' label is clickable!
> 
> I've taken Trey recommendation to have an 'info' icon next to it instead. My original intention is to have it not so prominent in the UI. Like you said, it's not normal use-case.
> 
> * Can the jobId be also copyable like the links?
> 
> This is not necessary since you can select any text on there and copy it using default browser feature. This is also NOT the upstream or TAP service jobId. It is firefly's jobId.
> 
Ok, understood. Agree.

> * Is the 'parameters' object at the bottom very useful
> 
> Like most of the exposed information, they are only useful for debugging or advance user. If there's an error, you will see `error: some_error_messages`. Maybe, the parameters will be useful then?
>
I think this is too much developer syntax and very difficult to a user to get the info from there.
 
> * The 'result' part is confusing: it should be '.../result/rsults'
> 
> This is the results of firefly request. When it's a table search, then you will get a firefly table in JSON format. If it's a download package request, you will get one or more url to the zip file(s).
>

I still don't see how to use it. Plus, on an old search, the link doesn't work anymore (current testing)
 
> * Can the job monitor give the high level information about the query, like table name and schema name
> 
> For TAP search, this should be done at the form submit time. Background monitor just display what is set as the job's label. In this case, it's the table's title.
>
Ok, we can lave that for now. Agree.
 
> * I did notice that the firefly `results/result` url that returns the table model json only gives the first page
> 
> It was not intentional. The result url is suppose to return the submitted request. In this case, the request was only for 100 rows. I will not do anything for now. This is correct as far as UWS is concern. If we have use for it, I will change it then.

robyww:
I think I agree with @ejoliet about the parameters data. I don’t really see a use case. However I disagrees about the `results/result` I see how this will be useful in the future.

gpdf (changes requested):
Initial feedback here: https://confluence.ipac.caltech.edu/display/FIREFLY/Feedback+on+FIREFLY-872%3A+Job+status+tracking+improvements

I will do more testing after dinner.

Most crucially: I would strongly prefer keeping the "(i)" button in the table toolbar and in fact moving both "Table Meta" and "Job Info" to tabs under that (read-only) dialog, keeping the setting ("gears") dialog limited to editable settings.

loitly:
@robyww @gpdf @ejoliet 
I've made all of the change requests we talked about in the meeting.  Let me know if I missed any.

Changes from PR reviewer recommendations
|     - Move 'Table Meta' and 'Job Info' tabs from table options into table info and make table info show by default for all table
|     - Remove parameters and results from Job Info dialog
|     - Rename labels
|     - Change Copy to Clipboard to the one used by Image Readout
|     - Rearrange Mask Panel when there's error
|     - Add help icons to Job Info and Table Info dialogs
|     - Fix: package summary not showing
|     - Fix: Job Info not updated when a new job is selected from background monitor
|     - Fix: job info link is always available with it's a backgrounable search

While testing in the meeting, we noticed a failed TAP query from IRSA did not return `Job Link`.  This is an IRSA TAP service bug. 
IRSA TAP immediately run the job when submitted.  This is ok as long as it returns the `location` of the submitted job.  However, in this case, it returns a `400 Bad Request` instead of the `location` where we can get the error from defined `/error` endpoint.
Example:
`curl -v "https://irsa.ipac.caltech.edu/TAP/async?QUERY=SELECT+TOP+1+*+FROM+fp_pscBAD_TABLE"`

gpdf:
I'll research the TAP-standard issues involved a bit and help put together an IRSA ticket if indeed this is unacceptable behavior for a TAP server.

gpdf:
Firefly CCB-D feedback from previous Friday meeting (apologies for the long delay!):

* In the tabbed "Table Info" dialog reached from the "(i)" button in the table toolbar, the "Job Info" tab, when present, should be the leftmost one and selected by default.

* In the tabbed "Table Info" dialog, the "Table Meta" tab should be renamed "Table Metadata".

* In the tabbed "Table Info" dialog, the "(?)" help icon is on the lower left; it is almost always on the lower right in Firefly.

* In the tabbed dialog's "Table Meta" tab, the "ID: IVOA_UWS_JOB Universal Worker Service Job" entry in the "Links" disclosure section is redundant with the "Job Info" tab and should be removed, because it's potentially confusing to users, who won't know that it's the same thing, or which one to choose.  Out of curiosity, what other information might appear in this "Links" section?  (Note that DataLink service descriptors appear under "Resources", not "Links".)

* In the untabbed "Job Information" dialog, the "(?)" help icon is on the lower left; it is almost always on the lower right in Firefly.

* In both the tabbed and untabbed versions of the Job Info display, the italicized URL has a bit of the final character truncated (at least, on Safari).

* In both the tabbed and untabbed versions of the Job Info display, the "ID" should have a mouseover that says "Internal query identifier, usable for diagnostics".

loitly:
> * In the tabbed "Table Info" dialog, the "(?)" help icon is on the lower left; it is almost always on the lower right in Firefly.
It was placed on the left because when it's on the right corner, it's very close to the 'resizable' handle.  I went ahead and moved it to the right.  Let me know if it's too close and need adjustment.

> * Out of curiosity, what other information might appear in this "Links" section?  (Note that DataLink service descriptors appear under "Resources", not "Links".)
VOTable specs allow for LINKs under TABLE.  Code is written to display if it exists.

Test URL updated.  Let me know if any additional changes is needed.  Please approve PR once all requests are met.

robyww:
> It was placed on the left because when it's on the right corner, it's very close to the 'resizable' handle. I went ahead and moved it to the right. Let me know if it's too close and need adjustment.

It will be fine if you just add a `padding-right: 10px;`

robyww:
I am still seeing the large white border around the info data. 

I think it comes from a scroll bar
```css
.JobInfo__main {
    display: flex;
    flex-direction: column;
    padding: 10px;
    overflow: scroll;
}
```

I think it could be fixed with changing `overflow` to `auto`
```css
   overflow: auto;
```

loitly:
> It will be fine if you just add a `padding-right: 10px;`

I will add this, although I don't like how it's not symmetric.

> I think it could be fixed with changing `overflow` to `auto`

I will add this, although I cannot duplicate it in any of my 3 browsers.  Must be my system settings.

PR #1154, opened 2021-12-02
FIREFLY-855: Improve the firefly tab bar to handle large number of tabs

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-855
Test: https://fireflydev.ipac.caltech.edu/firefly-855-improved-tabpanel/firefly/

- Add 'Search open tabs' button  option to TabPanel.jsx

This is modeled after the Chrome browser 'Search Tabs' feature.  When this option is enabled, a `down-arrow` will appears on the right side of the tab panel.  The drop down panel consist of only a single column containing the Tab's title.  
From this dropdown, you can: 
- select the active tab
- sort and filter

`Allow the user to edit the name of a tab` is not implemented, yet.  It requires the tab panel to maintain states, which was not in the original design.  May need to refactor/rewrite `TabPanel.jsx` to accommodate more advance features in the future.

Things to consider:
- This feature is only enabled for the multi-tables displayed as tabs.
- It appears even if there's only one tab.  Do we need to set a min-count before this appears?

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -291,6 +291,22 @@ div.rootStyle img {
     font-weight: bold;
 }
 
+.round-btn {
+    width: 16px;
+    height: 16px;
+    display: flex;
+    align-items: center;
+    padding: 1px;
+    justify-content: center;
+}
+
+.round-btn:hover {
+    background-color: rgba(0,0,0, .15);
+    border-radius: 50%;
+    cursor: pointer;
+    border: 1px solid rgba(0,0,0, 0.15);
+}
+
 
 .keyword-label  {
     font-weight: bold;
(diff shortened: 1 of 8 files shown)


robyww (approved):
The code look good and the UI for the part of ticket you implemented looks good. 

If you think we should push tab renaming to the next release, you should update firefly-855 to remove the renaming part and make a new ticket with tab renaming.

lrebull:
OK, that's pretty slick.

ejoliet (approved):
Looks great and works very nice within the tables area.
One thing minor i noticed that if i open the list of tabs and close one tab, the list is not updated until i close and reopen, leaving the list with orphan table that i can still sort or filter. Is there a way to update the list or close the panel if i close one tab?

I agree that the rename should be in another ticket for next release. Was there a request to add a 'Close all or close others' feature? Maybe for the future too when is necessary. 

Looks good!

PR #1155, opened 2021-12-03
IRSA-4355: Fix Firefly common IBE processing for multi-position search

loitly:
ticket: https://jira.ipac.caltech.edu/browse/IRSA-4355
test: https://irsa-4355-multiposition-ztf.irsakudev.ipac.caltech.edu/applications/ztf/
- changes include fixes to all missions: atlas, ptf, 2mass, wise, ztf

`makeQueryParam` failed to detect Multi-Object Search parameters because `UserTargetWorldPt` parameter is null.  Add logic to include `filename` as well.

After changes, Multi-Object works fine when Search Region is involved.  When `Image contains target` or `Any pixel overlaps search region`  is select, `Science Exposure` timed out.  This is the same behavior found in OPS.  @wmiipac , you should report this to back-end.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/ibe/BaseIbeDataSource.java
@@ -4,11 +4,16 @@
 package edu.caltech.ipac.astro.ibe;
 
 import edu.caltech.ipac.util.StringUtils;
+import edu.caltech.ipac.visualize.plot.CoordinateSys;
+import edu.caltech.ipac.visualize.plot.Plot;
+import edu.caltech.ipac.visualize.plot.WorldPt;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Map;
+
+import static edu.caltech.ipac.util.StringUtils.isEmpty;
 
 /**
  * Date: 4/18/14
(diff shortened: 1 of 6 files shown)


wmiipac (commented):
I see you moved makeQueryParam from individual mission's DataSource into the common Ibe DataSource, that's nice and clean.

"After changes, Multi-Object works fine when Search Region is involved. When Image contains target or Any pixel overlaps search region is select, Science Exposure timed out. This is the same behavior found in OPS. @wmiipac , you should report this to back-end."

I don't understand why the ibe search type "Image contains target" or "Any pixel overlaps search region" works for single object/position, not working for Multipostion search. Could be network or database connection issue?

loitly:
> I don't understand why the ibe search type "Image contains target" or "Any pixel overlaps search region" works for single object/position, not working for Multipostion search. Could be network or database connection issue?

This is not something Firefly can fix.  It's in the backend IBE service and should be reported since it's affecting OPS as well.  Would you like me to report it?

```
cat upload.tbl 
| object | ra        | dec       |
| char   | double    | double    |
  m51      202.48417   47.23056    
```

Single POS search works:
```
curl -F 'POS=202.48417,47.23056' -F 'INTERSECT=COVERS' -F 'mcen=' https://irsadev.ipac.caltech.edu/ibe/search/ztf/products/sci
```

Multi POS search failed(timed out), even though it has only that same one position above:
```
curl -F 'POS=@upload.tbl' -F 'INTERSECT=COVERS' -F 'SIZE=0.0336' https://irsadev.ipac.caltech.edu/ibe/search/ztf/products/sci
```

But, if `SIZE` is used instead of `mcen`(most centered), then it will return results:
```
curl -F 'POS=@upload.tbl' -F 'INTERSECT=COVERS' -F 'SIZE=0.0336' https://irsadev.ipac.caltech.edu/ibe/search/ztf/products/sci
```

This is an issue for `Science Exposure` data only.  `Reference Image` works fine.
I am guessing the join between the uploaded temp table and the sci table is too big causing it to timed out.
I am sure the developer will find out the real issue once you look into it.

wmiipac (approved):
Works as the PR claimed. Code clean up  looks good. 
Thanks for the fix! 
I will follow up with Anastasia on the ibe problem.

wmiipac:
Yes, it's not ui problem. I asked Luisa to confirm the problem on irsa ops, will send the issue to Anastasia.

ejoliet (approved):
Tested. Looks as expected now apart from the IBE backend problem. 
Thanks for taking time to clean up also the code!

PR #1156, opened 2021-12-06
Firefly-886: add user agent based on tag to firefly network calls

robyww:
### Firefly-886: add user agent based on tag to firefly network calls
 
 - gets the firefly tag 
 - show it in the version dialog
 - use the tag to make a user agent for net work calls

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-886

#### Testing

- add a tag to the branch such as `release-2033.44.33` (that you will delete later)
- build the branch
- You should be the tag in the about dialog and any network call will log the user agent as `firefly/2033.44.33`

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/download/URLDownload.java
+++ b/src/firefly/java/edu/caltech/ipac/util/download/URLDownload.java
@@ -5,7 +5,9 @@ package edu.caltech.ipac.util.download;
 
 import edu.caltech.ipac.firefly.data.FileInfo;
 import edu.caltech.ipac.firefly.data.HttpResultInfo;
+import edu.caltech.ipac.firefly.data.Version;
 import edu.caltech.ipac.firefly.server.util.Logger;
+import edu.caltech.ipac.firefly.server.util.VersionUtil;
 import edu.caltech.ipac.util.Base64;
 import edu.caltech.ipac.util.FileUtil;
 import edu.caltech.ipac.util.StringUtils;
@@ -30,6 +32,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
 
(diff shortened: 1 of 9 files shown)


gpdf:
A clarifying question: if an application (like SOFIA, or the Rubin Portal a/k/a "suit") is built on top of Firefly, will it just be the Firefly tag that's reported?

robyww:
> A clarifying question: if an application (like SOFIA, or the Rubin Portal a/k/a "suit") is built on top of Firefly, will it just be the Firefly tag that's reported?

It will be `firefly/2033.44.33 (suit/1.2.3)`

src/firefly/java/edu/caltech/ipac/util/download/URLDownload.java
```diff
@@ -45,6 +49,30 @@
...
+        String tag = v.getBuildGitTagFirefly();
+        if (StringUtils.isEmpty(tag)) return unknown;
+        Matcher m = Pattern.compile("[1-9]").matcher(tag);
+        if (!m.find()) return unknown;
+        int firstIdx= m.start();
+        int lastIdx= m.start();
+        while (m.find()) lastIdx= m.start();
+        try {
+            return "Firefly/" + tag.substring(firstIdx,lastIdx+1);
+        } catch (Exception e) {
```
loitly, line 71: This logic may leads to wrong value.  It ignores the trailing 0s which is significant.  Try something like this:
`jshell> Pattern.compile("\\d+(\\.\\d+)+").matcher("release-2021.13.30.dev").results().findFirst().get().group();`

loitly (approved):
Changes look fine.  Look at my comment about pattern.  If you disagree, it's fine, too.  Tag is internally controlled.  Not a big deal.

src/firefly/java/edu/caltech/ipac/util/download/URLDownload.java
robyww: I did forget the `0` but you are so much better at regular expressions. I am sure this is much better but I will play with it to understand.

gpdf:
"It will be firefly/2033.44.33 (suit/1.2.3)" 

+1, I like that.

PR #1172, opened 2022-01-31
Firefly-922: fixed: MOC should just be added to HiPS and not images

robyww:
#### Firefly-922: Fixed: MOC should just be added to HiPS and not images

 - Also Fixes an occasional but that is hard to duplicate. Once an image request fails it continues to fail until the server is restarted.
 
_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-922

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-922-moc-on-fits/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/LockingVisNetwork.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/LockingVisNetwork.java
@@ -26,7 +26,7 @@ import java.util.Map;
 import java.util.Objects;
 
 /**
- * This class will download files via VisNetwork.  However it also locks so that two of the same request do not happen
+ * This class will download files via VisNetwork.  However, it also locks so that two of the same request do not happen
  * at the same time. This way a file will not overwrite itself during download.
  *
  * @author Trey Roby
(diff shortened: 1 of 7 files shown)


lrebull:
Hm. I'll grant you that my CPU isn't brought to its knees but the MOC still appears on the FITS and the fans are still running. Is that what you expected me to see? I can't tell from here or in the ticket.

Screen Shot 2022-01-31 at 3 54 22 PM

lrebull:
Also, since i see the note here about image failures, I am getting image failures for the 6 degree 2MASS mosaics on this build. They work on ops. (my target is M16, but 2MASS covers the whole sky, so should have coverage everywhere.)

robyww:
Something must be wrong. MOC should not appear on images at all.

lrebull:
I tried it again in a 'new private window' to try to get a movie for you. I didn't do "new private window" before, because I naively thought that starting from your build, I wouldn't need to do a new private window. Silly me.

I get the three polygon red thing in the FITS images (the red you can see in the screen shot above) again, but not the full green SDSS MOC. So maybe that's ok? The red polygon does appear on the FITS images, though; is that expected? what is the red polygon, and why does it have the sub-rectangles?

lrebull:
It looks like maybe it is the relevant MOC tile when the MOC is first loaded, THAT is what's showing up on the FITS images?

robyww:
@lrebull: I missed a case in the last build. I think it is fixed. 

- Just to be clear. My solution for this ticket is just not to add the MOC layers on the images. I think that is the core of the issue. 
- The bug I fixed with the images is a case I have seen in production as well but is hard to duplicate.  I managed to get my development version in that state and I found the issue. I don't think you can test that because I don't know what causes it. The bug fix is allowing the firefly server to recover.

#### New Build: For your peace of mind: The following three lines should show on the version information dialog
Version:    2022.1-DEV-firefly-922-moc-on-fits
Built On:   Tue Feb 01 08:09:35 PST 2022
Git Commit: c7092d76f


_URL:_ https://fireflydev.ipac.caltech.edu/firefly-922-moc-on-fits/firefly/

lrebull:
I agree, this is working now!

ejoliet (approved):
I confirm too. Looks great. Thanks!

PR #1181, opened 2022-02-11
FIREFLY-883: "Data Type" column in Radar results no longer have filter choices

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-883

The `Data Type` column in Radar table results no longer have an enumerated filter choices to pick from.

Test: https://firefly-883-choiced-column-fix.irsakudev.ipac.caltech.edu/frontpage/
Do any Radar search -> turn on filter to see the drop-down option under `Data Type` column.


lrebull:
YAY YAY YAY yes this is fixed ! makes me so happy to see this fixed!

wmiipac (approved):
it's fixes. good to merge.

PR #1182, opened 2022-02-11
Firefly-935: Add quotes to table or columns where needed

robyww:
#### Firefly-935: Add quotes to table or columns where needed 
  - check table and columns using regular expressions
  - also fixed small bug in ObsCoreSRegionParser

_ticket:_ [FIREFLY-935](https://jira.ipac.caltech.edu/browse/FIREFLY-935)

#### Testing the TAP Panel
  - Version: 2022.1-DEV:firefly-935-quotes_0d46
  - URL: https://fireflydev.ipac.caltech.edu/firefly-935-quotes/firefly/
  - Do a VizieR Search
    - @lrebull Search: VizieR, collections: J_ApJ, table: 752/127  (no other constraints)
  - try various searches with the advance panel

### DIFF
--- a/src/firefly/js/ui/tap/AdvancedADQL.jsx
+++ b/src/firefly/js/ui/tap/AdvancedADQL.jsx
@@ -12,7 +12,7 @@ import {cloneDeep, defer, isArray, isObject} from 'lodash';
 
 import {InputAreaFieldConnected} from '../InputAreaField.jsx';
 import {SplitContent} from '../panel/DockLayoutPanel';
-import {loadTapSchemas, loadTapTables, loadTapColumns, getTapServices} from './TapUtil';
+import {loadTapSchemas, loadTapTables, loadTapColumns, getTapServices, maybeQuote} from './TapUtil';
 import {getColumnIdx} from '../../tables/TableUtil.js';
 import {dispatchValueChange} from '../../fieldGroup/FieldGroupCntlr.js';
 import {getFieldVal} from '../../fieldGroup/FieldGroupUtils.js';
@@ -115,11 +115,11 @@ export function AdvancedADQL({adqlKey, defAdqlKey, groupKey, serviceUrl, style={
             if (taVal) {
                 insertAtCursor(textArea, tname, adqlKey, groupKey, prismLiveRef.current);
             } else {
-                dispatchValueChange({fieldKey: adqlKey, groupKey, value: `SELECT TOP 1000 * FROM ${tname}`, valid: true});
+                dispatchValueChange({fieldKey: adqlKey, groupKey, value: `SELECT TOP 1000 * FROM ${maybeQuote(tname,true)}`, valid: true});
                 window.setTimeout( () => prismLiveRef.current.syncStyles?.(), 10);
             }
         } else if (type === 'column') {
-            const val = ffcn.current.checked ? `${tname}.${cname}` : cname;
+            const val = ffcn.current.checked ? `${maybeQuote(tname,true)}.${maybeQuote(cname)}` : maybeQuote(cname);
             insertAtCursor(textArea, val, adqlKey, groupKey, prismLiveRef.current);
         }
     };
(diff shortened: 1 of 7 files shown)


loitly (changes requested):
I don't think the logic should be spread out into all of these places.  The logic should be at where we query the TAP schema.  

Table here: https://github.com/Caltech-IPAC/firefly/blob/0d4669031ec008aa2db070c866aca3bd8f898093/src/firefly/js/ui/tap/TapUtil.js#L125
Columns here: https://github.com/Caltech-IPAC/firefly/blob/0d4669031ec008aa2db070c866aca3bd8f898093/src/firefly/js/ui/tap/TapUtil.js#L158

You may need to look at this, too.  I didn't examine the code to see what it's doing.
https://github.com/Caltech-IPAC/firefly/blob/0d4669031ec008aa2db070c866aca3bd8f898093/src/firefly/js/ui/tap/TapUtil.js#L84

Here, we check to see if a table or column name fail validation.  If so, we'll fix it by enclosing it with double-quotes.  There may be table or column names with quotes already.  We are not introducing anything here that may causes regression issues.

An issue you may encounter after this fix is how we should display identifiers with quotes.  Do we remove the quotes when showing them to the users for readability reason?  Or, we leave the quotes there to be precise.  I don't remember what we did.  But, this is a separate existing issue.  It does not need to be solved in this PR.

robyww:
@loitly I looked at that idea and it did not work. It seems to make the safest place to quote is at point the sql parts are being built. Otherwise is will affect other logic. The advantage of doing this is that columns are consistently treated just as the are now. The quotes don’t happen to the very last step in the process. I think this is the safest change.

I don’t understand the rest of you comments so we should talk.

lrebull:
For the record, I'm totally good with "this is too expensive in resources for us to fix their problem" and instead doing something in documentation to make Vizier work.

It occurs to me - does anyone know if the Vizier folks even realize that their service isn't compliant?

loitly:
> @loitly I looked at that idea and it did not work. It seems to make the safest place to quote is at point the sql parts are being built. Otherwise is will affect other logic. The advantage of doing this is that columns are consistently treated just as the are now. The quotes don’t happen to the very last step in the process. I think this is the safest change.
> 
> I don’t understand the rest of you comments so we should talk.

According to Gregory and I confirmed, TAP standard specifically say that table and column name should follow ADQL and that if quotes are needed, it should be included in the name.
If what you say is true, that our code does not handle names with quotes, then it's a bug we need to fix.  I am still convinced that the comment I made is true.  But, I am okay with you merging your changes if you think this is the best solution for now.

loitly (approved):

robyww:
I think it is the safest. There is a lot of processing that goes on. 

>According to Gregory and I confirmed, TAP standard specifically say that table and column name should follow ADQL and that if quotes are needed, it should be included in the name.
If what you say is true, that our code does not handle names with quotes, then it's a bug we need to fix. I am still convinced that the comment I made is true. But, I am okay with you merging your changes if you think this is the best solution for now.

I did take your advice for doing the check for all columns and tables. So it is a general fix and not just for VizieR. I did include a function that would process it just for vizieR but it is not used. So it does fix the bug in a general. It just takes a downstream approach. For better or worse, the code is building sql fragments in several places. Each of those areas are responsible for building valid sql. So that is where the fix is. Maybe in the future we find a way to centralize the building of fragments or at least refactor them out to a set of small utility functions.

robyww:
@lrebull 
>For the record, I'm totally good with "this is too expensive in resources for us to fix their problem" and instead doing something in documentation to make VizieR work.

I guess what we have decided is that we can implement a general improvement to the code. I expect this issue might fall into a gray area of the standard and they might argue that it is fine. It has not taken that much time. I am glad you brought up the issue. VizieR is a big archive, it would be nice to have our TAP search work with it.

lrebull:
OK! I'm glad to have it work, don't get me wrong; just giving you an 'out' if it was proving to be prohibitively tough.

PR #1185, opened 2022-02-16
Firefly-323: integrate the FITS and HiPS info dialog

robyww:
#### Firefly-323: integrate the FITS and HiPS info dialog

_ticket:_ https://jira.ipac.caltech.edu/browse/FIREFLY-323

#### Testing
 - Version: 2022.1-DEV:firefly-323-info_8b12
 - URL: https://fireflydev.ipac.caltech.edu/firefly-323-info/firefly/
 - Bring up multiple FITS and HiPS, show the info dialog, click between them
 - The dialog will change to show information with the plot that is active

### DIFF
--- a/src/firefly/js/ui/PopupPanel.jsx
+++ b/src/firefly/js/ui/PopupPanel.jsx
@@ -9,17 +9,35 @@ import {getDefaultPopupPosition, humanStart, humanMove, humanStop} from './Popup
 import './PopupPanel.css';
 import DEL_ICO from 'html/images/blue_delete_10x10.png';
 
+/** @typedef {Object} LayoutType
+ * enum can be one of
+ * @prop CENTER
+ * @prop TOP_EDGE_CENTER
+ * @prop TOP_CENTER
+ * @prop TOP_LEFT
+ * @prop TOP_RIGHT
+ * @prop NONE
+ * @prop USER_POSITION
+ * @type {Enum}
+ */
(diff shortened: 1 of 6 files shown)


lrebull:
I confirm that the pop-up does dynamically change now when going between FITS and HiPS.
I note that the gears and filters appear in the FITS header but not in the HiPS header. I can sort but not filter the HiPS keywords. This is true on ops as well, and I honestly cannot remember if there is a ticket on this or not; I just don't know anymore. I (still) think this is odd.

robyww:
> note that the gears and filters appear in the FITS header but not in the HiPS header. 

I might be able to add that, I will look.

robyww:
@lrebull Just rebuilt, can you test again: I added the filters and I added a second ticket fix to it. Firefly-705

lrebull:
i can see the filters! but still not the gears, FWIW.
(FTR, i had to restart Firefox AND do a 'new private window' to see the changes.)

robyww:
>  not the gears, FWIW.

I don't think the gear are necessary for a two column table. do you really want them?

lrebull:
fair point! I just noticed the difference as I was in essence blinking between the two tables. If all HiPS headers are only two columns, then yeah, probably won't ever need the gears.

ejoliet (approved):
I confirm Firefly-323 is fixed. I will let Gregory confirm the Python/Slate related one.
Thanks!

PR #1186, opened 2022-02-18
firefly-45,firefly-543,Firefly-633: Multiple tickets

robyww:
#### Multiple Tickets

 - [Firefly-45](https://jira.ipac.caltech.edu/browse/FIREFLY-45): api option to hide image toolbar
 - [Firefly-543](https://jira.ipac.caltech.edu/browse/FIREFLY-543): better computation of  minimum size for coverage
 - [Firefly-633](https://jira.ipac.caltech.edu/browse/FIREFLY-633): Improve terminology for zoom max

Testing:
- Version: 2022.1-DEV:firefly-45_543_633_e575
- URL: https://fireflydev.ipac.caltech.edu/firefly-45-543-633/firefly/
- [Firefly-543](https://jira.ipac.caltech.edu/browse/FIREFLY-543)
   - do any search and then sort down to 1 object
   - now sort down to 3 to 5 objects that are fairly close together
   - The image size should be the same.
 - [Firefly-633](https://jira.ipac.caltech.edu/browse/FIREFLY-633)
   - zoom all the way to make the image as big as possible. 
   - Clicking the zoom plus one more time will show a maximum zoom message that does not include the level.
   - If this message is not quite right then maybe someone (@lrebull?) could give me better working. 
-  [Firefly-45](https://jira.ipac.caltech.edu/browse/FIREFLY-45)
   - URL: https://fireflydev.ipac.caltech.edu/firefly-45-543-633/firefly/test/test-image.html?test=Show+Image+with+no+toolbar
   - The image will not show the toolbar
   - Expanding it will have a toolbar

### DIFF
--- a/src/firefly/html/test/test-image.html
+++ b/src/firefly/html/test/test-image.html
@@ -77,8 +77,41 @@
         onFireflyLoaded = function (firefly) {
 
             const req= {
-                url: 'https://irsa.ipac.caltech.edu/cgi-bin/Subimage/nph-subimage?origfile=/irsadata/AKARI//images/N60/l000.85_b+65.00_ecl_6deg_N60_sigma.fits&ra=317.778840&dec=56.472080&xsize=2.000000',
+                url: 'https://web.ipac.caltech.edu/staff/roby/many-images/mips_IER_7735552_ier600_A24_P24_10s.cal.fits',
                 Title: 'Image Title Here'
+            };
+
+            firefly.showImage( 'show-image',req);
+        }
+    </script>
+</template>
+
+<template title="Show Image with no toolbar" class="tpl" >
+    <div id="expected" style="position: relative" >
+        <div class="source-code indent-3" style="position: absolute; bottom: 0; left: 0">
+            show image with no toolbar.
+            <code>
+                window.firefly= {
+                       options: {
+                            MenuItemKeys: {
+                                showImageToolbar:false
+                            }
+                        }
+                    };
+            </code>
+        </div>
+    </div>
+    <div id="actual" class="flow-h">
+        <div id="show-image" class="box"></div>
+    </div>
+    <script>
+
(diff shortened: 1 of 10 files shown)


lrebull:
The issue described in FIREFLY-45 seems to work.

The issue described in FIREFLY-543 doesn't seem to work. I searched on 02 00 32.08 -58 10 41.0 in NED with 1 arcmin and got 17 objects, and was given a 7.4' WISE-1 image. The .1 arcmin search never comes back -- i never get a table, I never get a plot, and I never get a coverage image.  AND now even the 1 arcmin search is hosed. So I don't know what's going on, but I can't test this, because it's not working for me at all.

FIREFLY-633: it says : ...You reached you maximum... please change to "...You reached your..." and then I think it's fine.

(I still think we need to fix the apparent second angular measure that isn't what it says it is, but that's another ticket.)

lrebull:
The issue described in FIREFLY-543 is failing in the same way it did last week for me... i'm going to close this browser window now, so @ me when i should try again.

cwang2016:
Here are the results for the tests on 
- Firefly-45:  I tried the provided api option to set window.firefly in an sample app. to hide the image tool bar => api option works. 
- 
- Firefly-543: I tried the above same sample app. with no tool bar setting (failed to get image from url https://fireflydev.ipac.caltech.edu/firefly-45-543-633/firefly/) to do NED search on object 'NGC 4842' and '1 arcmin', I got 58 objects in the table and the image (2MASS_K_s) with FOV 14' and 8.6' before and after the expansion, and then change the search radius to include 1 or 4~5 objects, the FOV stays the same.  
When I change api option to make the sample app. to show the image with the took bar, I got FOV 16' and 9.7' before and after the expansion. 

- Firely-633, do the zoom all the way to make the image as big as possible and click zoom plus to get the image back to the original one with maximum FOV  (14' and 8.6'). 

All tests works as the testing instruction.

PR #1189, opened 2022-02-23
FIREFLY-938: ZTF proprietary metadata is not returned in GUI

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-938
Test: https://firefly-938-ztf-proprietary.irsakudev.ipac.caltech.edu/applications/ztf/

Should be fixed now.  Follow ticket description to test.


lrebull:
I confirm, this works; I can see data from 2/20/2022. I had some trouble (error messages) when logging in, but I did finally successfully log in, and I can see those data.

lrebull:
(and before logging in, i could not see those data)

wmiipac (approved):
Code make sense to me for each ibe query url pass the authentication credential. 
Tested searches and download with the proprietary data, works well.
Thanks!

ejoliet (approved):
Tested. Looks good! 

(Was the Deprecated ibe file uploader class meant for IBE multi position search or was for something else)

loitly:
> Tested. Looks good!
> 
> (Was the Deprecated ibe file uploader class meant for IBE multi position search or was for something else)

It was just an unnecessary layer of abstraction design element that's not used.  I would've removed it.  But, since this is a hot patch fix, I don't want to put too much changes into it.

src/firefly/java/edu/caltech/ipac/astro/ibe/IbeFileUploader.java
```diff
@@ -12,6 +12,7 @@
  *
  * @author loi
  * @version $Id: $
+ * @deprecated
```
robyww, line 15: If this file is not used why not just removed it.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ibe/IbeFileUploaderImpl.java
```diff
@@ -27,6 +27,7 @@
  * Implementing IbeFileUploader using HttpClient via MultiPartPostBuilder
  * @author loi
  * @version $Id: $
+ * @deprecated
```
robyww, line 30: If this file is not used why not just removed it.

robyww (approved):
Looks good. I don't see any reason not to remove the deprecated files.

PR #1191, opened 2022-02-26
IRSA-4468: include irsa4468, irsa-4467 and irsa-4466 added datasets

ejoliet:
This PR changes the image config file and adds 3 more datasets to the image search WISE and Spitzer contributed data.
See the tickets IRSA-4466, IRSA-4467, and IRSA-4468 to use the coordinates sample and the labelling requested for testing.

[No build link added yet as i can't enter to Jenkins - the link will be added later. Build locally for now]


lrebull:
Works pretty well on all three! Minor thing on one of themm z0mgs:
In the other programs with many very different wavelengths, we had listed the shorter ones first. So, that would mean that this would be FUV, NUV, W1, W2, W3, W4. Currently in this build, it's W1, W2, W3, W4, NUV, FUV. Is it possible to change the order?

ejoliet:
> Works pretty well on all three! Minor thing on one of them, z0mgs:
> In the other programs with many very different wavelengths, we had listed the shorter ones first. So, that would mean that this would be FUV, NUV, W1, W2, W3, W4. Currently in this build, it's W1, W2, W3, W4, NUV, FUV. Is it possible to change the order?

Yes, i will try to change the order before merging. Thanks!

PR #1200, opened 2022-03-22
Multiple tickets PR(FIREFLY-960, FIREFLY-962, FIREFLY-963)

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-960
FIREFLY-960: cancel filters from table doesn't work
- fixed.  regression bug.

https://jira.ipac.caltech.edu/browse/FIREFLY-962
FIREFLY-962: can't turn on filters in catalogs if they're loaded from the background
- fixed.  regression bug.

https://jira.ipac.caltech.edu/browse/FIREFLY-963
FIREFLY-963: strange behavior of tab keys to navigate among filters
- I noticed this behavior in OPS as well.  The table re-render after applying the filter then focus is lost.  I thought it was expected
- But, I went ahead and add the code needed to keep focus after re-rendering.
- Also, I don't understand the "lock by click" test case.  Please test to see if my fix also fixes that.

Test: https://fireflydev.ipac.caltech.edu/firefly-960-962-963-table-filters/firefly/

### DIFF
--- a/src/firefly/js/tables/TableConnector.js
+++ b/src/firefly/js/tables/TableConnector.js
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {isEmpty, omitBy, isUndefined, cloneDeep, get, range, memoize} from 'lodash';
+import {isEmpty, omitBy, isUndefined, cloneDeep, get, range} from 'lodash';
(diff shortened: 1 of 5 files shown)


lrebull:
All works!
The "lock by click" note just meant that the focus changed from the images to the "lock by click" area, then down to the table. But it's fixed now.

robyww (approved):
Looks good.

PR #1204, opened 2022-04-04
FIREFLY-976: TablePanel failed to pickup new changes when tbl_ui_id changes

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-976
more changes here: https://github.com/IPAC-SW/irsa-ife/pull/208

After TablePanel was converted to a functional component([FIREFLY-930](https://jira.ipac.caltech.edu/browse/FIREFLY-930)), it failed to pickup new changes when tbl_ui_id changes.  This was reported by Mike Lund.

@robyww , there a few shortcomings in the popular `useStoreConnector` function I would like to change.
1. The interface allow for multiple state getters but is rarely used.  This introduces unnecessary complexity in the code.
2. It uses `Object.is` by default when comparing `old state` vs `new state`.  But, in most cases, we want `shallowequal`.
3. No way to update the getter function once its dependencies changes.

The new `useStoreState` is what I would like to replace it with.  But, I don't want to introduce a new function.  I would like to update the old function and fix any necessary code that uses it.  There will be many of changes(file touched), and regression testing is needed.  How do you think?  

Test: https://fireflydev.ipac.caltech.edu/firefly-976-tablepanel-rerender/firefly/

### DIFF
--- a/src/firefly/js/charts/ui/ChartSelectPanel.jsx
+++ b/src/firefly/js/charts/ui/ChartSelectPanel.jsx
@@ -162,7 +162,7 @@ ChartSelectPanel.contextType= RenderTreeIdCtx;
 
 function ChartAction({chartId, chartActions, chartAction, chartActionChanged}) {
 
-    const [activeTrace=0] = useStoreConnector(() => getChartData(chartId)?.activeTrace);
+    const activeTrace = useStoreConnector(() => getChartData(chartId)?.activeTrace ?? 0);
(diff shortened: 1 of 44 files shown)


robyww (approved):
I am fine with all that you want to do. I think it will an improvement.

robyww (approved):
All looks good. `useStoreConnector` code is easier to follow. I like defaulting to `shallowequals`

PR #1205, opened 2022-04-07
Firefly-968: Support HiPS Aitoff projection

robyww:
#### Test

- _URL:_ https://fireflydev.ipac.caltech.edu/firefly-968-hips-aitoff/firefly/
- Load a HiPS, choose the AITOFF options and play around
- To a wise catalog search and play around with the coverage
- _Version:_ 2022.2-DEV:firefly-968-hips-aitoff_ed03

### DIFF
--- a/src/firefly/html/test/test-hips-mocs.html
+++ b/src/firefly/html/test/test-hips-mocs.html
@@ -288,6 +288,27 @@
     </script>
 </template>
 
+<template title="Demo show a aitoff hips at about 210 degrees" style='height: 400px' class="tpl lg " >
+    <div id="expected" >
+        <div class="source-code indent-3">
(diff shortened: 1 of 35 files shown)


ejoliet (changes requested):
The image projection works great. I found that the grid is now broken somehow.
When adding the grid overlay to Aitoff it works fine, when switching to HiPS it gets wonky. Grid lines are going out of the image and canvas.
See picture :  
Screen Shot 2022-04-07 at 3 14 05 PM

ejoliet:
Something maybe to discuss with CCB for improvement: I would change the FoV when switching to Aitoff so you can see that you have all sky projected maybe centered on the target

PR #1206, opened 2022-04-07
Firefly 978 big table performance

loitly:
ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-978
test: https://fireflydev.ipac.caltech.edu/firefly-978-big-table-performance/firefly/

Improve reading IPAC table
- Replace the use of DataGroup and DataObject with indexed values when reading IPAC table to DataGroup
- Greatly reduces read time for large table. (~30%)
- The sample file used to take over 7s now takes around 4.5s to read.

Improve loading IPAC table info DB
- turn off DB logging.  reduce DB ingestion time from 6s to 3s

Lower memory requirement
A sample 1 million rows WISE catalog is around 360mb file; requiring 550mb of RAM in memory.
To ingest the table into the DB requires 3 copies in memory; DataGroup, PrepareStatement, and the DB tables in memory.
  - Instead of loading the full table in one batch commit, separate the table into multiple smaller batches(set to 10000)
  - A PrepareStatement of 10k rows take less memory than having the 3rd copy
  - This slightly decreases loading performance, but uses less memory.
   
Also:
- expand Test functionalities
  - added util functions
  - enhance logging capability
- remove and clean unused flag ‘debug.mode’

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
@@ -457,6 +457,7 @@ public final class FITSTableReader
                 if (colUnits!=null && colUnits.length>1) idxDT.setUnits(colUnits[1]);
                 dataTypes.add(dataDT);
                 DataGroup dataGroup = new DataGroup(desc, dataTypes);
+                dataGroup.setInitCapacity(data.length);
                 for (int i = 0; (i < data.length); i++) {
                     DataObject aRow = new DataObject(dataGroup);
                     aRow.setDataElement(idxDT, i);
@@ -486,6 +487,7 @@ public final class FITSTableReader
                 }
 
                 DataGroup dataGroup = new DataGroup(desc, dataTypes);
+                dataGroup.setInitCapacity(data[0].length);
 
                 DataType[] dd;
                 for (int row = 0; row < data[0].length; row++) {
@@ -580,6 +582,7 @@ public final class FITSTableReader
 
         // creating DataGroup rows.
         DataGroup dataGroup = new DataGroup(table.getName(), dataTypes);
+        dataGroup.setInitCapacity((int)table.getRowCount());
         for (long row = 0; row < table.getRowCount(); row++){
             addRowToDataGroup(table, dataGroup, colIdxMap, row, strategy);
         }
--- a/config/app.config
+++ b/config/app.config
@@ -12,7 +12,6 @@ work.directory = ""
 stats.log.dir = '${sys:catalina.base}/logs/statistics'
 alerts.dir = ""
 
-debug.mode = false
 ignore.auth=true
 
 /*  keep vis.shared.mem.size small if pct is used */
--- a/config/common.prop
+++ b/config/common.prop
@@ -14,9 +14,6 @@ DbAdapter.type = hsql
 # default to System.getProperty("java.io.tmpdir") + "/workarea"
 work.directory=@work.directory@
 
-# set this to false to disable debug related code, like logging, etc.
-debug.mode=@debug.mode@
-
 # Allow JMX to monitor ehcache's status.
 # Enabling this feature may incur a small performance hit.
 ehcache.jmx.monitor=true
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/ServerContext.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/ServerContext.java
@@ -204,7 +204,6 @@ public class ServerContext {
         Logger.info("",
                 "CACHE_PROVIDER : " + EhcacheProvider.class.getName(),
                 "WORK_DIR       : " + getWorkingDir(),
-                "DEBUG_MODE     : " + AppProperties.getBooleanProperty("debug.mode", false),
                 "Available Cores: " + getAvailableCores() );
 
          if (!getWorkingDir().equals(getSharedWorkingDir())) {
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/EmbeddedDbUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/EmbeddedDbUtil.java
@@ -169,6 +169,7 @@ public class EmbeddedDbUtil {
         sql = dbAdapter.translateSql(sql);
         String ddSql = refTable == null ? null : dbAdapter.getDDSql(refTable);
 
+        logger.trace("execQuery => SQL: " + sql);
         try {
             DataGroup dg = (DataGroup)JdbcFactory.getTemplate(dbInstance).query(sql, rs -> {
                 return dbToDataGroup(rs, dbInstance, ddSql);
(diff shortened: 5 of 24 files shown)


ejoliet (approved):
I tested it in IV but frankly i cannot really check if it made the difference with my samples.

Where i could see the differences i think is in the TimeSeries tool, for an uploaded light-curve of 3000 rows, the periodogram calculation will result in 30k rows or something and there i think i can see a difference. Though probably the biggest one might be on a million rows table but i don't have a sample or query for that, do you? I tried WISE 1000' on single exposure but the job never finished somehow while i was spending time testing so i stopped it after 5mins or so. 
I think disabling the log for the database jobs is also a good idea.

Kudos for cleaning up and add/refactor tests!

Thanks!

robyww:
> This slightly decreases loading performance, but uses less memory.

It this our goal? Would should try to keep long term memory load  down but the app should be sized for transient memory during a load.

gpdf:
For Rubin use and for future IRSA use we are going to need to make sure that we are also performant on reading in large VOTables.  Million-row query results are to be expected.  Firefly already performs reasonably well on one-million-row tables, once loaded, so that should be sort of a benchmark to aim at for improving the performance of loads.

loitly:
> > This slightly decreases loading performance, but uses less memory.
> 
> Is this our goal? Would should try to keep long term memory load down but the app should be sized for transient memory during a load.

I've played around with batchSize on this 1m rows sample file to find what's best and settled at 10k.
Below is the best result from 3 runs for each case.  As you can see, `ingestDataIntoDb` time is pretty much identical and yet the 10k size requires 300mb less to load.  Also, this is a single run.  In the multi users runtime environment, execution time is less important then unpredictable hard limit where the application would error on out-of-memory.

`10k batchSize` 
```
09:18:38.348 [main] INFO  edu.caltech.ipac.util.download.URLDownload - Download Complete: IpacTableFromSource5687858460056933396.tbl : 375256168 bytes, time: 252 ms
    file:/Users/loi/dev/cm/firefly_test_data/large-ipac-tables/wise-950000.tbl
09:18:42.576 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - fetchDataGroup: IpacTableFromSource ran with elapsed time of 4.4870 SECONDS
09:18:45.525 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - ingestDataIntoDb: IpacTableFromSource ran with elapsed time of 2.9350 SECONDS
09:18:45.527 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - createDbFile: IpacTableFromSource ran with elapsed time of 7.7500 SECONDS
09:18:45.541 [main] TRACE edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil - execQuery => SQL: select * from DATA   limit 100 offset 0
09:18:45.585 [main] TRACE edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil - converting a 100 rows ResultSet into a DataGroup
09:18:45.595 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - getDataset: IpacTableFromSource ran with elapsed time of 0.0670 SECONDS
```

`single batch (1m rows)`
```
09:24:20.381 [main] INFO  edu.caltech.ipac.util.download.URLDownload - Download Complete: IpacTableFromSource830938347332589139.tbl : 375256168 bytes, time: 232 ms
    file:/Users/loi/dev/cm/firefly_test_data/large-ipac-tables/wise-950000.tbl
09:24:24.887 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - fetchDataGroup: IpacTableFromSource ran with elapsed time of 4.7460 SECONDS
09:24:27.885 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - ingestDataIntoDb: IpacTableFromSource ran with elapsed time of 2.9830 SECONDS
09:24:27.887 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - createDbFile: IpacTableFromSource ran with elapsed time of 8.0600 SECONDS
09:24:27.902 [main] TRACE edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil - execQuery => SQL: select * from DATA   limit 100 offset 0
09:24:27.948 [main] TRACE edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil - converting a 100 rows ResultSet into a DataGroup
09:24:27.958 [main] DEBUG edu.caltech.ipac.firefly.server.util.StopWatch - getDataset: IpacTableFromSource ran with elapsed time of 0.0710 SECONDS

```

loitly:
@ejoliet 950k rows sample file.  firefly_test_data/large-ipac-tables/wise-950000.tbl

loitly:
@gpdf I agree on VOTable.  In this round, I looked at reading IPAC table, mostly because I know there's room for improvement.  But, I also looked at loading the in-memory object(DataGroup) and ingesting it into the Database which is common to all formats.

In a later round, I will look into reading FITS table, ascii VOTable, and binary VOTable.  But, those relies on 3rd party libraries.  I am not sure how much we can improve without drastic changes.  It's still worth looking into, though.

gpdf:
Thanks!

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -486,6 +487,7 @@ private static DataGroup getFitsImageAsTable(BasicHDU<?> hdu,
                 }
 
                 DataGroup dataGroup = new DataGroup(desc, dataTypes);
+                dataGroup.setInitCapacity(data[0].length);
```
robyww, line 490: What did you not add a `setInitCapacity` after new `new DataGroup` in this file?

robyww (approved):
It looks good to me. It is fine to merge.

>Replace the use of DataGroup and DataObject with indexed values when reading IPAC table to DataGroup


The only optimization is see is setting the `initCapacity` for the `PrimitiveList`. Where are the other optimizations?

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -580,6 +582,7 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
 
         // creating DataGroup rows.
         DataGroup dataGroup = new DataGroup(table.getName(), dataTypes);
+        dataGroup.setInitCapacity((int)table.getRowCount());
```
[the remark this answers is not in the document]
loitly: Similar to ArrayList, capacity is increased by `oldCapacity * 1.5` when it needs to grow.  At that time, an arrayCopy is performed.  Setting a good initial capacity can avoid unnecessary arrayCopy and unused memory allocation.  It didn't help too much, but every little thing adds up.

The biggest improvement came from using Object[] to represent a row during parsing instead of DataObject, which relies on a HashMap  keyed by column's name to resolve column's data.
It also uses a Map to store column parsed info.  So, 2 Map.get() was needed for each cell value.  In my sample file, a total of 45 million calls were made.  Avoiding these calls were a big gain.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
loitly: When not set, it defaults to 1000.

PR #1207, opened 2022-04-07
Firefly-968: Support HiPS Aitoff projection

robyww:
#### Test

- _URL:_ https://fireflydev.ipac.caltech.edu/firefly-968-hips-aitoff/firefly/
- Load a HiPS, choose the AITOFF options and play around
- To a wise catalog search and play around with the coverage
- _Version:_  2022.2-DEV:firefly-968-hips-aitoff_b632


robyww:
From @ejoliet 
The image projection works great. I found that the grid is now broken somehow.
When adding the grid overlay to Aitoff it works fine, when switching to HiPS it gets wonky. Grid lines are going out of the image and canvas.
See picture

robyww:
from @ejoliet 

Something maybe to discuss with CCB for improvement: I would change the FoV when switching to Aitoff so you can see that you have all sky projected maybe centered on the target

robyww:
@ejoliet I tested the grid code a lot and never got the display you got. Can you do it again and give the the (approximately) center ra and dec?

robyww:
@ejoliet  never mind, I reproduced it with safari

robyww:
the grid issue is fixed.

src/firefly/js/drawingLayers/ComputeWebGridData.js
```diff
@@ -73,6 +73,8 @@ import { getDrawLayerParameters} from './WebGrid.js';
 import {Regrid} from '../util/Interp/Regrid.js';
 import {getPointMaxSide} from  '../visualize/HiPSUtil.js';
 import {convert} from '../visualize/VisUtil.js';
+import {isHiPS, isHiPSAitoff} from 'firefly/visualize/WebPlot.js';
+import {getFoV} from 'firefly/visualize/PlotViewUtil.js';
```
loitly, line 77: I think you should use relative path within firefly instead of the `firefly` alias.

src/firefly/js/visualize/saga/MouseReadoutWatch.js
```diff
@@ -400,7 +398,7 @@ function makeWLResult(plot,imagePt= undefined) {
         }
     }
     else {
-        return;
+        return undefined;
```
loitly, line 401: Is there any difference is the 2 methods?

loitly (approved):
I only looked at a few files since I am not familiar with most if not all of these files.
I can only comment on a few tests I did.

`/test/tests-viewer -> Load Six HiPS Maps`
Initial load is drastically different between `dev` and this PR.
I think the default FOV and coord sys has changed?

`/test/tests-viewer -> Load 1 table to Tri-View`
In `dev`, there 3 points plotted outside the image.  They are not in your PR.

`/test/tests-viewer -> Load many tables Tri-View`
Coverage is totally different, but I think that's expected.

`/test/tests-viewer -> Load 1 image to Tri-View`
This failed in both cases.

`/test/test-coverage`
Test 6&7, grid does not look right.

src/firefly/js/visualize/saga/MouseReadoutWatch.js
robyww: i like `return undefined` because it is better documented.

src/firefly/js/drawingLayers/ComputeWebGridData.js
robyww: yes, that is a mistake.

robyww:
@loitly Thanks for going the the viewer test. I forgot they were some complete. I did not look at them.

> /test/tests-viewer -> Load Six HiPS Maps
Initial load is drastically different between dev and this PR.
I think the default FOV and coord sys has changed?

This is all fine. it is centered on 0,0, GAL. There is a little race condition that determine which order they are loaded and I changed the defaults on the field of view slightly.

> /test/tests-viewer -> Load 1 table to Tri-View
In dev, there 3 points plotted outside the image. They are not in your PR.

Not sure about this one. The test look at same to me. I don't think this is affected by the PR.

> /test/tests-viewer -> Load many tables Tri-View
Coverage is totally different, but I think that's expected.

_Fixed_: It is still different, but it was too small. I fixed it. I am glad you tested that one. 
Before there was a all sky FITS that is galactic based. The all sky HiPS is set to be J2000. 

> /test/tests-viewer -> Load 1 image to Tri-View

_Fixed_: This has nothing to do with the PR but I changed the IRSA call to retrieve a different image.

> /test/test-coverage
Test 6&7, grid does not look right.

These are fine. You are seeing Galactic  vs J2000. I can ask the scientist if they would prefer galactic in the AITOFF 360 case.

PR #1210, opened 2022-04-12
Firefly-984: clean up of server side visualization code

robyww:
#### [Firefly-984](https://jira.ipac.caltech.edu/browse/FIREFLY-984): clean up of server side visualization code 

  - over 5000 lines of code removed
  - Since we are not longer generating image tiles a lot of code is now deprecated
  -  Zoom, rotate, color is not longer computed on the server
  - Most necessary data now kept in FitsRead, not WebPlot
  - Clean up some code so it is more maintainable
  - Change data objects records, much easier to understand their use
  - Attempted to collapse some level if indirection that are no longer necessary
  - Removed: many unused methods, java classes and exceptions
  - fixed: mask not drawing when first come up
  - fixed: cube plane color change should affect all the planes of the cube

#### Testing: 
 - _url:_ https://fireflydev.ipac.caltech.edu/firefly-984/firefly/
 - everything should work as before


loitly (approved):
Very hard to review due to the many changes. I can only run tests to see if I can spot any regression issues.  I found a couple.  See if you can fix before merging.  Thanks.
1. compilation error(s) in irsa-ife.
2. /test/tests-main.html
   - error in console, but does not seem to affect results

>

PR #1214, opened 2022-04-21
FIREFLY-982: Add the capability to a create persistent resource from a table request

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-982

This is taken from ticket under implementation notes.
```
Resources are persistent data taken from a unique Table Request. Uniqueness is based on getUniqueID.
Once a resource is created, it will remain unchanged until deleted.
firefly.util.table.makeResourceRequest is created to help work with resources.
See "Table as a Resource" in tests-table.html for sample usage.

Resource comes with basic security features. This info is saved as a .acl file next to the database files.

scope: This control the visibility of a resource. It can be 'global', 'user', or 'protected'.
    global is public. Anyone can access it. If this resource is created with a secret, then that secret is needed to delete it.
    user is visible to only the user who created it. User is based on a 'usrkey' cookie set on the browser.
    protected resource is one created with a secret token. It can only be accessed using the same secret token.
Any resource created without scope is considered global.
```

See test for an example of how it's used:
https://fireflydev.ipac.caltech.edu/firefly-982-table-as-resource/firefly/test/tests-table.html?test=Table+as+a+Resource

### DIFF
--- a/src/firefly/html/test/tests-table.html
+++ b/src/firefly/html/test/tests-table.html
@@ -355,39 +355,34 @@ added in this test.  Open console, then
     <script>
         function onFireflyLoaded(firefly) {
 
-            const {makeFileRequest, makeResourceRequest} = firefly.util.table;
+            const {makeFileRequest, makeResourceRequest, createResource, deleteResource} = firefly.util.table;
 
-            // define 3 resources
-            const gReq = makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/WiseDemoTable.tbl');
-            const global = {request: gReq};     // default to global
+            // simplest use case.  create a public resource then query it.
+            const simplest = {request: makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/WiseDemoTable.tbl')};
+            // createResource(simplest);        // this is optional because a default global scope resource will be created if one does not exists.
+            const simplestTbl = makeResourceRequest(simplest, 'Simplest');
+            firefly.showTable('actual', simplestTbl);
 
-            const uReq = makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/WiseDemoTable.tbl', null, {dummy:'user'});    // insert dummy param to make the request different from global
-            const user = {request: uReq, scope: 'user'};
-
-            const pReq = makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/WiseDemoTable.tbl', null, {dummy:'protected'});
-            const secured = {request: pReq, scope: 'protected', secret: 'mysecret'};
-
-            // create the resources
-            // makeResourceRequest({resource: global, action: 'create'});   // by default, a Resource is created with 'global' scope if one does not exists.  This line is not needed, but is nice for clarity.
-            makeResourceRequest({resource: user, action: 'create'});
-            makeResourceRequest({resource: secured, action: 'create'});
-
-
-
-            const globalTbl =   makeResourceRequest({resource: global, title: 'Global Resource'});
-            const userTbl =     makeResourceRequest({resource: user, title: 'User Resource'});
-            const protectedTbl = makeResourceRequest({resource: secured, title: 'Protected Resource'});
-            const noAccess =    makeResourceRequest({resource: {...secured, secret:''}, title: 'No Access'});        // this should fail.  access protected resource without secret token
-
-            firefly.showTable('actual', globalTbl);
+            // create a 'user' resource
+            const user = {scope: 'user', request: makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/fp_2mass.fp_psc29179.tbl')};
+            createResource(user);
+            const userTbl = makeResourceRequest(user, 'User Resource');
             firefly.showTable('actual', userTbl);
-            firefly.showTable('actual', protectedTbl);
-            firefly.showTable('actual', noAccess);
+
+            // create a 'protected' resource with secret token
+            const secured = {scope: 'protected', secret: 'mysecret', request: makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/wise-orbit-allsky.tbl')};
+            createResource(secured);
+            const securedTbl = makeResourceRequest(secured, 'Protected Resource');
+            firefly.showTable('actual', securedTbl);
+
+            // Accessing 'protected' resource without secret token
+            const accessDeniedTbl = makeResourceRequest({...secured, secret: 'wrong secret'}, 'Access Denied');
+            firefly.showTable('actual', accessDeniedTbl);
 
             firefly.deleteResourceTests = function () {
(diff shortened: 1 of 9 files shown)


src/firefly/js/tables/TableRequestUtil.js
```diff
@@ -160,6 +165,45 @@ export function makeVOCatalogRequest(title, params={}, options={}) {
...
+ * @param {TableRequest} [p.options] table request options.  see TableRequest for details.
+ * @param {string} p.action  one of 'query', 'create', 'delete'.  defautls to 'query'.
+ *                         when action is 'create' or 'delete', it will submit the request.
+ *                         no further action is needed.
+ * @returns {TableRequest}
+ * @public
+ * @func  makeResourceRequest
+ * @memberof firefly.util.table
+ */
+export function makeResourceRequest({resource={}, title, options={}, action='query'}) {
```
robyww, line 192: I think it function is too overloaded. It makes it API confusing. I think you should break it up it 

1. `makeResourceRequest` for query, which is consistent with how we do things now.
2. `createResource`
3. `deleteResource`

if you could come up with a good name then `createResource` and `deleteResource` could be combined (`putResource`?)

I had to really study your example to understand what was happning.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ResourceProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+ * Any resource created without scope is considered global.
+ */
+@SearchProcessorImpl(id = PROC_ID, params =
+        {   @ParamDoc(name = "action",          desc = "required.  one of 'create', 'query', 'delete'"),
+            @ParamDoc(name = "searchRequest",   desc = "required.  the table request to create this resource from"),
+            @ParamDoc(name = "scope",          desc = "visibility of this resource.  one of 'global', 'user', 'protected'.  defaults to 'global'"),
+            @ParamDoc(name = "secret",          desc = "secret token used to access 'protected' resources"),
+        })
+public class ResourceProcessor extends EmbeddedDbProcessor {
+    public static final String PROC_ID = "ResourceProcessor";
```
robyww, line 50: I like how this just functions as a wrapper around other request.

robyww (changes requested):
Overall it looks great. Code is clean and I like the fact you included ACL features.

_Changes_: I am most concern the API naming. Breaking up `makeResourceRequest` into two or three functions will make it clearer.

Some smaller comments.
- Using different tables URLs in the examples - I included a list.
- I added some general comments in `ResourceProcessor.java` about things I have become more aware of in all my cleaning up recently.
- I don't see how you know to keep this DB permanently. How is that done?

src/firefly/java/edu/caltech/ipac/firefly/server/query/ResourceProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * return a string
+     * @param request
+     * @return
+     */
+    public String getUniqueID(ServerRequest request) {
+        try {
+            TableServerRequest sreq = QueryUtil.getSearchRequest((TableServerRequest) request);
+            SortedSet<Param> params = sreq.getSearchParams();
+            return StringUtils.toString(params, "|");
```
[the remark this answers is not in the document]
loitly: Not only that.  Many of the util functions are out dated.  I noticed `combine` a few days ago while looking at your PR.  The overloaded functions with '[]' wrapped is so niched case that it should not be in a separate functions.
On top of that, with java 8.. you can do it like this
```
jshell> Arrays.asList("a","b").stream().collect(Collectors.joining("|"));
$2 ==> "a|b"
```
The `toString` function do have an advantage of not having to check for null, which is nice.

src/firefly/html/test/tests-table.html
```diff
@@ -331,6 +331,67 @@ <h4>Show how to inject table meta using API</h4>
...
+</pre>
+        </div>
+    </div>
+    <div id="actual"  class="box x3"/>
+    <script>
+        function onFireflyLoaded(firefly) {
+
+            const {makeFileRequest, makeResourceRequest} = firefly.util.table;
+
+            // define 3 resources
```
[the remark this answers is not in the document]
loitly: Yes, but I purposely reuse the same URL to demonstrate how one can create multiple resources from the same URL.  I can use different tables if you think that info is not needed.

src/firefly/js/tables/TableRequestUtil.js
loitly: That's what I had before.  `makeResourceRequest `,  and `createResource`.  
Then, I added the `delete` feature.  In order to not expose the `action` param, I need to add a third function, `deleteResource`.  That's when I decided to combine hem all and expose the `action` param.

In the simplest case, you can do this.  (original request)
```
const aResource = {request: makeFileRequest(null, 'https://web.ipac.caltech.edu/staff/roby/demo/WiseDemoTable.tbl')};
const myTbl =   makeResourceRequest({resource: aResource, title: 'My Table'});
firefly.showTable('actual', aResource);
```
I have no problem separate them into 3 functions.  After seeing my explanation, do you still think I should separate them?

robyww: I talk to @loitly, he will break it into three calls, after that merge.

robyww (approved):

ejoliet (approved):
I think i understood that the added function needed for API has been addressed.
Question i have is what is the difference between 'user' and 'protected'?
Regardless, is fine with me.

loitly:
> I think i understood that the added function needed for API has been addressed. Question i have is what is the difference between 'user' and 'protected'? Regardless, is fine with me.

`user`:  only the user who created the resource can access it.
`protected`: when created with a password, anyone with the password can access it.

loitly:
@ejoliet @robyww I added these after your feedbacks.  Take a look if you have a chance, otherwise I'll merge later in the afternoon.
- split `makeResourceRequest` into 3 function, adding `createResource` and `deleteResource`.
- add `ResourceProcessorTest`
- rewrite  tests in `tests-table.html` to make it easier to understand

src/firefly/html/test/tests-table.html
```diff
@@ -331,6 +331,62 @@ <h4>Show how to inject table meta using API</h4>
...
+            createResource(secured);
+            const securedTbl = makeResourceRequest(secured, 'Protected Resource');
+            firefly.showTable('actual', securedTbl);
+
+            // Accessing 'protected' resource without secret token
+            const accessDeniedTbl = makeResourceRequest({...secured, secret: 'wrong secret'}, 'Access Denied');
+            firefly.showTable('actual', accessDeniedTbl);
+
+            firefly.deleteResourceTests = function () {
+                makeResourceRequest({resource: global, action: 'delete'});
```
robyww, line 383: I think you meant `deleteResource` here

robyww (approved):
Found a small thing in the example but otherwise looks good.

PR #1217, opened 2022-05-02
Firefly-980: support HiPS based target selection

robyww:
#### [Firefly-980](https://jira.ipac.caltech.edu/browse/FIREFLY-980): support HiPS based target selection

   - Option 1: HiPS above target panel
   - Option 2: HiPS as pop up on target panel

#### Testing
- _url:_ https://fireflydev.ipac.caltech.edu/firefly-980-hips-target/firefly/firefly-dev.html
- click on `Example Js Dialog`
- Example 1
  - click on `HiPS/Target`  tab
  - click on HiPS to change target
  - change target see in change in HiPS
  - change radis, see if reflected on HiPS
- Example 2
   - click `HiPS/Target - popup`  tab
   - click on 20x20_click on target panel
   - do same testing but with the dialog


ejoliet (approved):
Nice!
I've interacted and tested the new 'click to set target' and i think it has all the potential to be integrated with a search but also with options menus as described in FIREFLY-217.

One thing we might have to do is that the click could be a single point/coordinate but also it should allow an area selection. And that goes together with what Vandana's design showed (pdf in the ticket FIREFLY-217). You might already have that in mind for a follow-up PRs.

Vandana's (and CCB) design was little bit more advanced in a sense that she requested to
enhance the existing click-and-drag rectangular selection by adding to the current options upon selection are “Zoom to fit selected area” and “Recenter image to selected area”, a “Search for additional data in selected area”.
From this option, you get a submenu with search type 'images...' or 'catalogs...'.

On IRSA, each type would bring each search accordingly, with the target coordinates and size of the area (or polygon/rectangle if area). On Rubin/LSST , they might want to bring TAP directly (to be confirmed).

Also, it was mentioned in the same PDF an option to upload a file with targets for multi-position search instead of a point-like or area selected.

At this point, because this looks like 80% already of what we need, you might want to merge and continue on another PR for the submenu options and connect to IRSA image search and catalog search panel.
Thanks!

Thanks!

robyww:
>  Vandana's (and CCB) design was little bit more advanced in a sense that she requested to
enhance the existing click-and-drag rectangular selection by adding to the current options upon selection are “Zoom to fit selected area” and “Recenter image to selected area”, a “Search for additional data in selected area”.
From this option, you get a submenu with search type 'images...' or 'catalogs...'.

I don't think all those options make sense in the Atlas use-case. When you are already on a search panel it does not make sense to have an option to such as “Search for additional data in selected area”. I think that is for some future search mode when you are looking at results.  And... that has all sorts of undefined question such as what are we actually searching.

However, I could go ahead have an option to use the area select to define the search area. I will go ahead and implement that with this PR and then we can to the other stuff when we know more.

ejoliet:
On ATLAS case, i agree. 
I agree also that the search using what the user click will depend on the app, so for Euclid/Spherex might be different.
But on the IRSAViewer it make sense and the ticket originally was meant for IRSAviewer but we can postpone it for now and focus on ATLAS case only.

The area selection will be a common need so that is ok to try to add it to this PR. I agree.
Thanks!

PR #1218, opened 2022-05-09
Firefly-992: Improve status page, improve the following

robyww:
#### Firefly-992: Improve status page, improve the following 

- Messanger information
- Cache replicating information
 - Server Event queue information

#### test: https://fireflydev.ipac.caltech.edu/firefly-992/firefly/admin/status

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/messaging/Messenger.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/messaging/Messenger.java
@@ -33,6 +33,7 @@ import java.util.concurrent.Executors;
  * @version $Id: $
  */
 public class Messenger {
+    public enum Status {CONNECTED, FAIL_TO_CONNECT, NOT_USING};
     private static final String REDIS_HOST = AppProperties.getProperty("redis.host", "127.0.0.1");
     private static final int REDIS_PORT = AppProperties.getIntProperty("redis.port", 6379);
     private static final int MAX_POOL_SIZE = AppProperties.getIntProperty("redis.max.poolsize", 25);
@@ -41,6 +42,7 @@ public class Messenger {
 
     // message broker..  Jedis
     private static JedisPool jedisPool;
+    private static Status status = Status.NOT_USING;
 
     // to limit one thread per topic
     private static ConcurrentHashMap<String, SubscriberHandle> pubSubHandlers = new ConcurrentHashMap<>();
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/messaging/Messenger.java
```diff
@@ -33,6 +33,7 @@
  * @version $Id: $
  */
 public class Messenger {
+    private enum ConnectAttempt {CONNECTED, FAIL_TO_CONNECT, NOT_USING};
```
loitly, line 36: I think ConnectAttempt should be called Status.

src/firefly/java/edu/caltech/ipac/firefly/messaging/Messenger.java
```diff
@@ -73,7 +77,7 @@ static boolean init() {
     public static String getStats() {
 
         JsonHelper stats = new JsonHelper();
-        if (!init()) {
+        if (connectAttempt!=ConnectAttempt.CONNECTED || jedisPool==null) {
             return stats.setValue("Messenger is offline").toJson();
```
loitly, line 81: `init()` is needed before every call.  It's a lazy-load approach to ensure it's called before it's used.
Also, since you added status, you can do something like this:
```
        init();

        stats.setValue(status, "status");
        if (status == Status.CONNECTED) {
            ....
        }
        return stas.toJson();
```

src/firefly/java/edu/caltech/ipac/firefly/messaging/Messenger.java
```diff
@@ -93,6 +97,14 @@ public static String getStats() {
         }
     }
 
+    public static String getRedisHostPortDesc() {
+        String detail= connectAttempt==ConnectAttempt.CONNECTED ? "" :
+                connectAttempt==ConnectAttempt.FAIL_TO_CONNECT ? " (failed to connect)" : " (not using)";
+        return REDIS_HOST + ":" + REDIS_PORT + detail;
```
loitly, line 103: Since the Status enum already contains the description, a toString() of it should be good enough.  init() is needed to ensure it's initialized.
```
init();
return String.format("%s:%s (%s)", REDIS_HOST, REDIS_PORT, status);
```

loitly (changes requested):
Please see the use of `init()`.  It's needed unless we remove lazy-loading approach.

PR #1222, opened 2022-05-19
add advanced internal scheme to support secrets, quotes, and spaces in envVar value

loitly:


docker/launchTomcat.sh
```diff
@@ -107,11 +85,18 @@ export CATALINA_OPTS="\
...
+  export CATALINA_OPTS="${CATALINA_OPTS} ${jvmProps}"
+fi
+# envVar with names matching 'FIREFLY_OPTS_*'.
+# A more advanced internal scheme to support secrets, quotes, and spaces in values
+# Use '__' in key, to sub for '.' since '.' is not allowed in envVar.
+for var in "${!FIREFLY_OPTS_@}"; do
+  prop=`sed 's/FIREFLY_OPTS_//g;s/__/./g' <<< "$var"`
+  export CATALINA_OPTS="${CATALINA_OPTS} -D$prop=${!var}"
+done
+#----- eval FIREFLY_OPTS
```
robyww, line 99: So the docker line to send firefly options would be something like
```
-e 'FIREFLY_OPT_firefly_options={tap : { additional:{ services: [ { "label":"LSST RSP", "url":"https://my.lsst.org/tap"}]}}'
```
Is that the idea?

loitly: Yes, if you prop is `firefly_options`.  If your prop name contains '.', escape that with '__'(double-underscore).

robyww (approved):

PR #1223, opened 2022-05-19
Firefly-981: UI generation and other task

robyww:
#### [Firefly-981](https://jira.ipac.caltech.edu/browse/FIREFLY-981): UI generation and other task
   - UI generation from a list of field defs
   - _Caching:_ cached object can now report their size instead of ehcache inspecting them
   - _Caching:_ removed two unusued caches
   - improve image recovery when server restarts
   - moved some JS code to mark it unused

#### Testing:
   - _URL:_ https://fireflydev.ipac.caltech.edu/firefly-981-ui-gen/firefly/firefly-dev.html
   - click on Test Searches
   - click on each of Dynamic 1, Dynamic 2, Dyn Search Panel
   - Dynamic 2 has its own ok button
   - Dyn Search Panel has it own search button

#### Key file to review
- UI Generation
   - `DynamicUISearchPanel.jsx` 
   - `TestQueriesPanel.jsx`

- Cacheing
   - `ObjectSizeEngineWrapper.java`
   - `HasSizeOf.java`
   - `Cache.java`
   - `ehcache.xml`


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/FitsCacher.java
```diff
@@ -95,12 +94,18 @@ private static void addToCache(FileInfo fitsFileInfo, FitsDataEval fitsDataEval)
...
+    }
+
 
     private static void prepareCacheSpace(FileInfo fitsFileInfo) {
-        if (memCache != null) {
-            memCache.put(fitsFileInfo, new ObjectSizeEngineWrapper.BluffSize(fitsFileInfo.getFile().length()));
-            memCache.put(fitsFileInfo, null);
-        }
+        if (memCache == null) return;
+        memCache.put(fitsFileInfo, new ObjectSizeEngineWrapper.BluffSize(fitsFileInfo.getFile().length()));
```
loitly, line 107: A good place to use anonymous class.  BluffSize is no longer needed.
`memCache.put(fitsFileInfo, (HasSizeOf) () -> fitsFileInfo.getFile().length());`

loitly (approved):
Cache related changes look good.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/FitsCacher.java
robyww: I did not realize you could use a cast for force a certain type of interface.

ejoliet (approved):
I played with the test search panel, looks good.

I've also looked at the code how is the panel dynamically generated. I think it covers all the features but radio button & checkboxes. Do we need those? Might be needed for single/multi target search if needs to be set dynamically? Or checkbox for type of search (cone or image coordinate?)

I'm unable to comment on the ehcache though, sorry. I believe Loi has reviewed it.

Other than that, good to be merged.

robyww:
> I've also looked at the code how is the panel dynamically generated. I think it covers all the features but radio button & checkboxes. Do we need those? Might be needed for single/multi target search if needs to be set dynamically? Or checkbox for type of search (cone or image coordinate?)

It support option menu which is equivalent to a radio button. If we want to have both we can do that too, but I think we can decide that later.  I did forget checkboxes so I am going to add them before I merge.

PR #1231, opened 2022-06-27
FIREFLY-599_1002: Multi tickets changes

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-599
- move user-info to the right of Background Monitor
 
https://jira.ipac.caltech.edu/browse/FIREFLY-1002
- add docker support to change basic auth path, including the ability to disable it

Manually added `USE_ADMIN_AUTH` override after deployed:
```
    kubectl edit deploy firefly-599-1002-userinfo-loc-admin-path-suit
    # add USE_ADMIN_AUTH env var to deployment
      - env:
        - name: USE_ADMIN_AUTH
          value: false
```

Test: https://firefly-599-1002-userinfo-loc-admin-path.irsakudev.ipac.caltech.edu/suit/
- check user-info is on the right of Background Monitor
- user's name is limited to 150px.  mouse-over to see the complete text
- suit/admin/status does not require user-id/password

### DIFF
--- a/docker/launchTomcat.sh
+++ b/docker/launchTomcat.sh
@@ -2,6 +2,7 @@
 
 NAME=${BUILD_TIME_NAME:-"ipac/firefly"}
 ADMIN_PASSWORD=${ADMIN_PASSWORD:-`echo $RANDOM | base64 | head -c 8`}
+USE_ADMIN_AUTH=${USE_ADMIN_AUTH:-"true"}
 
 echo -e "\n!!============================================================"
 echo "!!============================================================"
@@ -88,8 +89,8 @@ export CATALINA_OPTS="\
   -Dvisualize.fits.search.path=${VIS_PATH} \
 	"
 
-#----- override ADMIN_PROTECTED path if set
-if [ ! -z ${ADMIN_PROTECTED+x} ]; then export CATALINA_OPTS="${CATALINA_OPTS} -DADMIN_PROTECTED='${ADMIN_PROTECTED}'"; fi
+#----- remove ADMIN_PROTECTED path so it no longer restricted by basic auth
+if [ "${USE_ADMIN_AUTH,,}" = "false" ]; then export CATALINA_OPTS="${CATALINA_OPTS} -DADMIN_PROTECTED="; fi
 
 #----- eval PROPS if exists.  key-value pairs are separated by spaces. therefore, it does not support values with spaces in it.
 if [ ! -z "${PROPS}" ]; then
--- a/docker/launchTomcat.sh
+++ b/docker/launchTomcat.sh
@@ -88,6 +88,9 @@ export CATALINA_OPTS="\
   -Dvisualize.fits.search.path=${VIS_PATH} \
 	"
 
+#----- override ADMIN_PROTECTED path if set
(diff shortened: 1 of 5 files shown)


docker/launchTomcat.sh
```diff
@@ -88,6 +88,9 @@ export CATALINA_OPTS="\
   -Dvisualize.fits.search.path=${VIS_PATH} \
 	"
 
+#----- override ADMIN_PROTECTED path if set
+if [ ! -z ${ADMIN_PROTECTED+x} ]; then export CATALINA_OPTS="${CATALINA_OPTS} -DADMIN_PROTECTED='${ADMIN_PROTECTED}'"; fi
```
robyww, line 92: Are we really want to allow the admin path to be changed. I don't know if there is a use case for this.
We could do something like the following: defined a USE_ADMIN (which default to true) if false set the path to a empty string. 
```bash

export CATALINA_OPTS="${CATALINA_OPTS} -DADMIN_PROTECTED=$([ "$USE_ADMIN" == "FALSE" ] && echo '' || echo '/admin/*')
```

This seem less confusing in the config.

loitly: I basically allow the `url-pattern` value of the ADMIN protected location to be overridden by an env.
It currently defaults to `/admin/*`.   Firefly have 2 endpoints under admin, `/admin/status` and `/admin/db`.
In practice, you can set ADMIN_PROTECTED=/admin/db, to still use basic auth on that endpoint.

But, if you rather create a single-use `{USE_ADMIN_AUTH:-true}` to set all or nothing, I am fine it with.

robyww: I think it will be better to just have a boolean. I think it will be easier to support.

loitly: done.

robyww (approved):
Looks good.

docker/launchTomcat.sh
gpdf: That change is fine with me, and I think the RSP team would be more comfortable just turning something on or off rather than attempting to make distinctions between the two.  This way you can add whatever you want under `/admin` in the future and it will all just work.

firefly-docker.env
```diff
@@ -1,7 +1,11 @@
-#---- Set environment variables for the docker startup
-#---- Examples-
+##---- Set environment variables for the docker startup
+##---- Examples-
 #PROPS_FIREFLY_OPTIONS=$'{ "coverage":  {"hipsSourceURL" : "ivo://CDS/P/2MASS/color"} }'
 #ADMIN_PASSWORD=reset-me
 #CLEANUP_INTERVAL=3h
-#---- Empty declarations: environment will come from the shell
+
+## Use Tomcat Admin basic auth on protected resources.  Defaults to true.
```
gpdf, line 7: I would add a comment here that says what endpoint(s) is/are being protected and that clarifies that the intent here is to permit a deployer to take care of authorization externally (as Rubin will).

E.g. "Use Tomcat Admin basic auth on protected resources, including the Firefly administration endpoints under .../admin/*. Defaults to true.  May be disabled in deployments where authorization is handled externally, e.g., by applying authorization-dependent redirects in a Kubernetes environment."

gpdf (commented):
Having VPN difficulties, so unable to review the login-display change until later today when I'm on site at Caltech.

It might be good to add a few words about the admin-endpoint changes to one of the Firefly documentation files.

PR #1234, opened 2022-06-30
Add Firefly Docker local development environment

loitly:
See added section [Run Firefly in DevMode](https://github.com/Caltech-IPAC/firefly/blob/firefly_local_env/docker/README.md#run-firefly-in-devmode) of the document for more details.
There also a small change in GWT as well even though there's not DevMode in GWT.

Other PRs related to this change.
https://github.com/Caltech-IPAC/firefly/pull/1233
https://github.com/IPAC-SW/irsa-ife/pull/212
https://github.com/IPAC-SW/irsa-ife/pull/213
https://github.com/lsst/suit/pull/40

> UPDATED: Additional changes

- Upgrade Node.js to v18.  v12 is end-of-life and forces wait during build
- GWT: Remove buildClient and therefore remove Node.js dependency
- Update docs to reflect dependency changes
- Minor tweak to JavaScript test(Jest) output


docker/launchTomcat.sh
```diff
@@ -74,8 +75,8 @@ fi
 
 #---------   CATALINA_OPTS must be exported for catalina.sh to pick them up
 export CATALINA_OPTS="\
-  -XX:InitialRAMPercentage=${INIT_RAM_PERCENT} \
-  -XX:MaxRAMPercentage=${MAX_RAM_PERCENT} \
+  -XX:InitialRAMPercentage=${INIT_RAM_PERCENT:-10} \
+  -XX:MaxRAMPercentage=${MAX_RAM_PERCENT:-90} \
```
robyww, line 79: I thought you wanted to make this like 97 and use almost all of the memory.

docker-compose.yml
```diff
@@ -10,17 +10,25 @@ services:
...
     command: bash -c "cd firefly && gradle firefly:test"
-    volumes:
+    volumes: &volInfo
       - ../firefly_test_data:/opt/work/firefly_test_data
       - ../firefly:/opt/work/firefly
       - /opt/work/firefly/.gradle
       - /opt/work/firefly/build
       - /opt/work/firefly/jars/build
       - /opt/work/firefly/node_modules
       - ./build/dist:/opt/work/firefly/build/dist/
```
robyww, line 25: I don't understand why these have to be listed as volumns. Maybe we can talk.

docker/Dockerfile
```diff
@@ -30,6 +30,39 @@ COPY firefly/package.json firefly/yarn.lock ./
 RUN yarn install --ignore-platform --frozen-lockfile
 
 
+FROM node_module AS dev_env
+
+RUN apt-get update \
+# install tomcat9 for dev mode \
+    && curl -O https://dlcdn.apache.org/tomcat/tomcat-9/v9.0.64/bin/apache-tomcat-9.0.64.tar.gz \
```
robyww, line 37: If you  are installing tomcat directly here. Should just do it in a base layer and not inherit `FROM tomcat:9.0-jdk17-openjdk-buster`

docker/README.md
```diff
@@ -30,62 +28,80 @@ See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
...
-    docker-compose push
+    docker-compose push firefly
 
 **To run Firefly from DockerHub**
 
 If you want to run a version available from DockerHub.  
 Set BUILD_TAG to the version you want, then 
 
-    docker-compose pull && docker-compose up -d
+    export BUILD_TAG=latest
```
robyww, line 103: is there a way to put `BUILD_TAG` on the command line?

docker/Dockerfile
```diff
@@ -30,6 +30,39 @@ COPY firefly/package.json firefly/yarn.lock ./
...
+
+# Tomcat scripts, config files, etc
+COPY firefly/docker/*.sh .
+COPY firefly/docker/tomcat-users.xml conf/
+
+# set up directory protections, copy stuff around, add tomcat user and group
+RUN mkdir -p /firefly/workarea /firefly/logs/statistics /firefly/alerts \
+    && chmod a+x *.sh
+
+CMD /opt/tomcat/launchTomcat.sh && cd /opt/work/${work_dir} && gradle ${project}:bAD && gradle ${project}:dev
```
robyww, line 63: This appears that it will honor all of the `PROPS_*` environment variables. right?

robyww (commented):
I have not tested yet, but I did have a few comments and questions.

docker/launchTomcat.sh
loitly: You're right.  I should go ahead with the change.

docker-compose.yml
loitly: Good question.  This is tricky.
The syntax is <host-path>:<container-path>
The first line.. I mount all of ./firefly into the container
`../firefly:/opt/work/firefly`

But, I don't want to do that.  I only want the source files, not any of the files generated from build (dirty).
So, I exclude the files I don't want by forcing an empty mount on those paths where `:` is not given.
i.e `/opt/work/firefly/build`

docker/Dockerfile
loitly: We use a multi-stage Dockerfile.  This part of the build is in the `dev-env` stage.
Docker is smart enough to know that if you request to build the `final` stage, it will traverse the stage's dependencies and skip this `dev-env` stage.  Therefore, this tomcat9 install line will not be executed during a normal firefly build, only for DevMode.

docker/README.md
loitly: Sure, I can suggest a one-line approach.

docker/Dockerfile
loitly: It should.  I tried to use the same startup script so that DevMode is as close to the normal deployment as possible.

PR #1235, opened 2022-07-05
Firefly-1011: HiPS and other fixes

robyww:
#### [Firefly-1011](https://jira.ipac.caltech.edu/browse/FIREFLY-1011): HiPS and other fixes 
 - fixed: HiPS caching too much and using too much memory - now smarter, limit size per hips with a maximum
 - fixed: HiPS not loading correct tile after color/bias change and then back
 - add lsst service access to hips select dialog 
      - refactored  `...server.visualize.hips` package
      - Most work is now done in `HiPSListUtil.java`
      - for now lsst hips server is defaulted to irsa 
      - java property `lsst.hips.masterUrl`
 - log service descriptor URL
 - fixed dropdown bug with MultiProductViewer

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1011-hips-fixes/firefly/
 - most things can't be tested.
 - Can't test
      - hips file caching only calls problem with the large tiles the lsst creates
      - lsst hips server can only be tested in lsst environment
 - test: 
     - change the hips color along with bias and then go back to default
     - do a TAP ztf search:  
         - ztf -> ztt_objects_dr11 -> m31 -> 100 arcsec
         - do the service descriptor search with the defaults, is should work
     - the ztf search should show the service descriptor url in the console


ejoliet (approved):
Tested successfully.

(I've also tested "show table" button which i didn't lasttime and that's a great feature!)

All ok form the test i could make.
Thanks!

PR #1236, opened 2022-07-06
FIREFLY-1008: migrate sortByCols feature from GWT

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1008
- added test to test/tests-table.html

To test feature via API: 
https://fireflydev.ipac.caltech.edu/firefly-1008-sortbycols/firefly/test/tests-table.html?test=Table+sortByCols
From view source, you can see column's meta was added.

To test WISE:
https://firefly-1008-sortbycols.irsakudev.ipac.caltech.edu/applications/wise/

Search: Position -> m51 -> Any pixel overlaps search region -> Search Region=600 arcsec -> Single Exposure -> all bands
Verify that sorting `scan_id` also sorts frame_num and band.

### DIFF
--- a/src/firefly/html/test/tests-table.html
+++ b/src/firefly/html/test/tests-table.html
@@ -393,10 +393,10 @@ added in this test.  Open console, then
     <div id="expected" >
         <div style="margin: 10px">
             <h4>use sortByCols for multi-columns sorting</h4>
-            <p>
-                scan_id column is sorted by: 'scan_id,band',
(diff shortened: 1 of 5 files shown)


ejoliet (approved):
Tested on postion search succesfully. 
Moving object is not sorted but it might need more logic to add for the MOST case as we have discussed earlier.

Ok to merge! Thanks!!

PR #1240, opened 2022-07-12
FIREFLY-1020-1021-1022: Chart multi-ticket UI changes

loitly:
Test: https://firefly-1020-1021-1022-chart-ui-changes.irsakudev.ipac.caltech.edu/irsaviewer/

https://jira.ipac.caltech.edu/browse/FIREFLY-1020
persistent split pane position  

https://jira.ipac.caltech.edu/browse/FIREFLY-1021
plural/singular pinned chart label

https://jira.ipac.caltech.edu/browse/FIREFLY-1022
location of 'pin chart' button

Beside the requested changes, I've also added these:
- Increase max pinned charts to 12 so you can see how crowded it can get.  If you want me to remove the limit completely, I can do that as well.
- Decreased the gap between split panels.  This is a global change.  All split panels will look like this.
- Change Active and Pinned Charts panel UI.  They look like separate tabs now.

I am open to changes.  Just leave comments below.

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -113,7 +113,7 @@ div.rootStyle img {
     border: 1px solid #acacac;
     border-radius: 2px;
     white-space: nowrap;
-}
+    font-weight: normal;}
 
 .button.text:hover {
     border: 1px solid #878787;
@@ -179,19 +179,17 @@ div.rootStyle img {
 .Pane.vertical .content, .Pane.horizontal  .content{
     overflow: hidden;
     position: absolute;
-    top: 5px;
-    bottom: 5px;
-    left: 5px;
-    right: 5px;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
 }
 
 .Pane.vertical .wrapper, .Pane.horizontal .wrapper{
     flex: auto;
     display: flex;
     overflow: hidden;
-    /*border: 1px solid #a3aeb9;*/
-    border-radius: 5px;
-    margin: 3px;
+    margin: 5px;
     position: relative;
 }
 
--- a/src/firefly/js/charts/ui/ChartPanel.css
+++ b/src/firefly/js/charts/ui/ChartPanel.css
@@ -64,27 +64,32 @@
 
 .ChartPanel__section--title .label {
     border: 1px solid #b3b3b3;
-    height: 16px;
+    height: 17px;
     padding: 3px 15px 0;
     border-top-left-radius: 5px;
     background-color: #c8c8c8;
     z-index: 1;
-    margin-bottom: -1px;
-    border-bottom: none;}
+    border-bottom: none;
+}
 
 .ChartPanel__section--title {
     display: inline-flex;
     justify-content: space-between;
     font-weight: bold;
-    align-items: center;
-    height: 20px;
+    align-items: flex-start;
+    height: 21px;
     margin-left: 5px;
+    margin-bottom: -1px;
+}
+
+.ChartPanel__section--title .button.text {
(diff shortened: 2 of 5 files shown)


lrebull:
https://jira.ipac.caltech.edu/browse/FIREFLY-1020 - **persistent split pane position** - this works!

https://jira.ipac.caltech.edu/browse/FIREFLY-1021 - **plural/singular pinned chart label** - this works!

https://jira.ipac.caltech.edu/browse/FIREFLY-1022 - **location of 'pin chart' button** - this works! although the spacing looks.. "crowded" for lack of a better term? on my Firefox. I don't know how else to describe it. It's hardly a critical thing, though. This is a great improvement in usability, I think, over the prior version.

**Increase max pinned charts to 12** - I tried pushing the limits here. When I got up near 12, the 10 extractions I had were fine, but the 2 normal (traditional? original?) charts I had were spinning and failing to update; see screenshot:
Screen Shot 2022-07-12 at 2 44 24 PM

**Decreased the gap between split panels**. - hm. I think maybe we're talking about different things. In the "active chart" here, there is space left (see magenta oval) for a greyscale legend that isn't there or relevant because this isn't a heatmap. In the admittedly ridiculous number of charts i have pinned, the extraction plots are mostly margin, not plot. (I only did magenta ovals on one of them.)
Screen Shot 2022-07-12 at 2 46 38 PM

**Change Active and Pinned Charts panel UI** - I admit I wasn't 100% sure now what it looked like before, so I had to go look, but I agree this is an improvement!

loitly:
> Decreased the gap between split panels. - hm. I think maybe we're talking about different things
I am talking about these gaps...

image

lrebull:
yeah, i figured you were talking about those, not the chart margins, which is why i added the screen shots...

loitly:
> location of 'pin chart' button

I don't really know where else to put them.  If someone want to mock up how the buttons should look, I can change it.

> charts I had were spinning and failing to update

Although I can't reproduce, I do noticed performance is a lot worse in Firefox than Chrome or Safari.  So, maybe having a max limit is a good idea after all.

> space left (see magenta oval) for legend

I am not sure why there's extra space there.  If you add a new trace, you will see that the legend takes more space than that.
Like the persistent split position, it's existing behavior.  I think we should move this request to future tickets so we can wrap up this round first.


Toolbar space is also an issue.  If you add a couple of new charts and traces, then switch to single chart mode, this is the minimum width of the toolbar before it gets clipped.

image

lrebull:
Let's leave it as a max of 12 charts for now (because who else other than testers is likely to hit that limit?)


location of 'pin chart' button: I agree i wasn't clear at all in what i was noticing. here is another attempt to explain. There is space around the words "active charts" before you get to the edge of the tab. There is space around all of the icons before you get to the edge of the interface. There is space around (in this case) "ba saljkflsf" before the edge of the menu. There is space above, to the left, and to the right of the words "Pin Chart". There is less space below, and less space between the bottom of that button and the top of the bar where the icons start. Am I making any sense or am I making mountains out of molehills? If it's the latter, that's ok too...
Screen Shot 2022-07-12 at 4 52 39 PM
 

I agree on the plot margin stuff (that's why i'd originally tagged it on the confluence page as "too hard for right now").

loitly:
> There is less space below, and less space between the bottom of that button and the top of the bar where the icons start

Ahh.. I get it now.  The 'Pin Chart' text vertical alignment looks good in Chrome and Safari.  But yes, it's a little off in Firefox.  I'll see what I can do.

As for the 'less space between the bottom of that button', I can try to give it 1 more pixel.
That's probably the most I can do before the top of the button is clipped.

image

PR #1242, opened 2022-07-20
Firefly-1030: improve visual target selection, (Also Firefly-1025)

robyww:
#### [Firefly-1030](https://jira.ipac.caltech.edu/browse/FIREFLY-1030): improve visual target selection
 - Visual target selection
   - make fields yellow that are being editied
   - set FOV smarter
   - add help text at top of visual selection, and help icon
   - show s_region

 - Other stuff:
   - revamp aitoff coverage so if AITOFF if always comes up at 0,0 gal
   - add a 0,0, galactic button if showing aitoff
   - if coverage start with aitoff have option of different image
   - fixed: distance tool turn off button sometime shows when it should not
   - fixed: remove distributed data” from tool tip
   - image will show EXTNAME or EXTTYPE
- Also Fixes: [Firefly-1025](https://jira.ipac.caltech.edu/browse/FIREFLY-1025): improve visual target selection

#### Testing

  - https://fireflydev.ipac.caltech.edu/firefly-1030-map-selection/firefly/
  - TAP:
     - Goto tap dialog and select an area
     - do a search with CADC, ivoa/ivoa.obscore - target m1, 
         - with first line of result select, choose cutout under the `more` dropdown
         - click on the pointer, you should see the s_region overdraw 
  - Coverage:
     - load a search with all sky
     - the coverage will come up with always centered on 0,0 GAL with a galactic projection
     - If the first result is all sky the default hips will be a DIRBE
     - any time hips is switch to Aitoff a button will appear to center on Galactic 0,0 with a galactic projection

### DIFF
--- a/src/firefly/js/core/background/BgMaskPanel.jsx
+++ b/src/firefly/js/core/background/BgMaskPanel.jsx
@@ -88,7 +88,7 @@ export const BgMaskPanel = React.memo(({componentKey, onMaskComplete, style={}})
                 <div style={{display: 'flex', alignItems: 'center'}}>
                     <div className='BgMaskPanel__content'>
                         <div style={{display: 'inline-flex', alignItems: 'center', justifyContent: 'center'}}>
-                            <di style={{marginRight: 5, fontSize: 18}}> {jobInfo.phase} </di>
+                            <div style={{marginRight: 5, fontSize: 18}}> {jobInfo.phase} </div>
                             <img className='JobInfo__items--link' onClick={showInfo} src={INFO} style={{height: 20}}/>
                         </div>
                         <div className='BgMaskPanel__msg'>{msg}</div>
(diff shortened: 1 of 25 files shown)


ejoliet (approved):
I've tested most of the fixes.

What i realized is that if you select a radius via the selection target from the HiPS and then you change the TAP search (for example CADC), the radius (hidden) is set to the previous selected radius from HiPS. Is that the regular behavior?

The search i did you mentioned was giving error because it was too big and that's how i noticed when i got back to a 'cone' search, the radius was the one i chose with big radius.

Maybe the radius should be reset if the user change TAP registry?

Rest seems to be improved or fixed!
Thanks!

lrebull:
I love the Aitoff stuff! Others may have opinions on which HiPS map to use but I think you've picked a great one. (It hadn't been one of the ones I'd been playing with, but doing some research into it, you've picked a better one than I'd been using!!)

I think the click-to-search TAP stuff looks great. I think it's much clearer than it was as to what the user is supposed to do.

PR #1248, opened 2022-08-10
Firefly-1045: Fixed several bugs

robyww:
#### [FIREFLY-1045](https://jira.ipac.caltech.edu/browse/FIREFLY-1045): Fix several bugs

- Fixed: Expanding/restoring an image in MultiProductViewer causes a reload stretch and color are reset
- Fixed: Exceptions happening when FItsHeaderView is up
- Fixed: Exceptions happening with StretchDropDownView
- Fixed: draw layer panel is hidden at times (zIndex was not correct)
- Fixed: on resize or when popping up mouse readout displayed zoomed to fix when it should not have
- Fixed: Moc validation was not accepting a valid FITs TFORM header
- Fixed: Better support for authorization in URLDownload.java
- Fixed: Expose MOC in Target Hips Panel
- Fixed: double load of web api
- Fixed: TAP search api did not init table correctly, also- fixed some bugs in examples
- Stretch should be sticky on MultiProductViewer
- Stretch should be stick for multi HDU images, if the same extension type
- [Firefly-1048](https://jira.ipac.caltech.edu/browse/FIREFLY-1049): Fixed: tooltip

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1045-bugs/firefly/
 - Some issues are too hard to test
 - Do an obscore search with TAP. (this test same bug found in wise, ztf, sofia)
     - MAST - ivoa - ivoa.obscore - m1  
    - change color on an image and expand then colapse - color should not change
    - change stretch and change image (by click on another row), stretch type should be sticky
 - on the tap panel - bring up visual target chooser - show layer dialog, you should be able to see it.
 - show a HiPS image and zoom in, resize the window and the HiPs will not rezoom, it should just stay the same
 - If you have an multi-hdu image with the same EXTTYPE keyword, upload, then changed the stretch, it will apply to all with that extension type
 - load https://fireflydev.ipac.caltech.edu/firefly-1045-bugs/firefly/?api=hips then click one of the the examples.  There should just be one HiPS loaded (not 2, fixes the double load)

### DIFF
--- a/firefly-docker.env
+++ b/firefly-docker.env
@@ -8,6 +8,11 @@
 ## May be disabled in deployments where authorization is handled externally, e.g., by applying authorization-dependent redirects in a Kubernetes environment.
 #USE_ADMIN_AUTH=false
 
+# to set up personal access tokens
+#PROPS_sso__framework__name=PAT
+#ROPS_sso__req__auth__hosts=.ncsa.illinois.edu,.lsst.cloud
+#ROPS_sso__access__token=xxx
+
 ##---- Empty declarations: environment will come from the shell
 ADMIN_PASSWORD
 env
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/tables/IpacTableFromSource.java
@@ -144,7 +144,7 @@ public class IpacTableFromSource extends IpacTablePartProcessor {
                 } else if (checkForUpdates) {
                     FileUtil.writeStringToFile(nFile, "workaround");
                     nFile.setLastModified(res.lastModified());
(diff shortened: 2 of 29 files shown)


gpdf (changes requested):
- HiPS double-load checked: OK
   - It would be nice if there were a URL API parameter to force on the visualization of a MOC, if available
- Loading a HiPS from the UI gives a faulty error message about a missing fill-in-field value that is explicitly optional:
Screen Shot 2022-08-11 at 20 08 37

- Image scaling resets on window-resize and expand/close seem fixed
- Color scale resets in data-product viewer seem fixed
- Layers control in visual-target dialog in TAP present and works

robyww:
- @gpdf I fixed the  faulty error message issue you found.
- also fixed tap api not finding the correct table

PR #1249, opened 2022-08-11
Irsa 4720 wise full image search with blank image size

wmiipac:
see ife PR: https://github.com/IPAC-SW/irsa-ife/pull/217


ejoliet (approved):
Looks good. Thanks!

PR #1255, opened 2022-08-25
IRSA-4756 fix source id searches

wmiipac:
This PR fix the WISE source id searches and add a helper anchor in the 3color image popup.
There is also a PR in ife for multiple bug fixes. Test instructions  are in ife PR.


ejoliet (approved):
Works! Thanks!

PR #1258, opened 2022-08-30
IRSA-4789 Fix the imageset constraints with wise merged table searches.

wmiipac:
Fixed them imageset constraints when wise ibe search uses merged table. 
Please test with k8s build:  https://irsa-4789.irsakudev.ipac.caltech.edu/applications/wise/


lrebull:
> On a different topic: shouldn't be the "Return most centered" option be "Yes" by default - i believe that's how it is in OPS?

Steve wants it to match ZTF, not match ops, so that the default for both tools is the same. so that's what it is.

PR #1267, opened 2022-09-09
Firefly 1063 PointShapeSizePicker class to functional component

kpuriIpac:
#### [Firefly-1063](https://jira.ipac.caltech.edu/browse/FIREFLY-1063): converting the ShapePickerWrapper class to a functional component 
 - Converted the ShapePickerWrapper class to a functional component (ShapePicker) 
 - The **if** condition inside **componentDidUpdate** in the class component was never being reached (none of the 3 conditions were true for any of the cases I tested: size, symbol and width were always equal to their previous values), so I did not convert that using **useEffect** and only did so for **componentDidMount** and **componentWillUnmount**.  
 - The second **if** condition (dlColor !== drawingDef.color) inside the **storeUpdate** (which is called using useStoreConnector) is also never reached, but I have kept this code in place for now. It may come in handy if we modify a shape that is not using the draw layers DrawingDef (but I couldn't figure out how to test that). 

#### Testing
 -  https://fireflydev.ipac.caltech.edu/firefly-1063-pointshapesizepicker-functional/firefly/
 - go to TAP -> service: IRSA/NED -> Coordinates or Object Name: m1 -> Radius: 100 
 - Open the color/symbol picker dialog (4th from the right hand side) 
 - Click on Symbol
 - Change the Symbol shape, change the symbol size by manually entering values and using the up/down array to move through values and ensure size is changing for every symbol type
 - Test out of bound values (less than 3 and greater than 100) 
 - Ensure behavior of symbol picker and symbol size is the same as that on Firefly (as there have should not have been any  UI changes, only refactoring of code from class to functional component). 

#### Bug Fix 
- follow the testing steps above 
- Try to enter a number in the valid range [3,100] followed by a space or other non-numeric characters. This should be disallowed
- The fix: The updateSize function takes the user entered "**size**" and uses parseFloat followed by Math.floor on it to get "**isize**". But when updating the state, we were setting (setSize) **size** to **size**, as opposed to **isize**. Setting it to **isize** fixes this issue. Try copy pasting "100 abc" into the input field. **size** = "100 abc" gets converted to "100" which is what the user sees since we update the state with this value. 
- Caveat: Because of Math.floor and setting **size** to **isize**,  this also means that the user won't be able to enter numbers with decimal points (try entering **42.5**). Is this functionality acceptable?

### DIFF
--- a/src/firefly/js/ui/PointShapeSizePicker.jsx
+++ b/src/firefly/js/ui/PointShapeSizePicker.jsx
@@ -130,7 +130,7 @@ export function ShapePicker({drawingDef, displayGroupId, plotId, update, getColo
             const symbolSize = DrawUtil.getSymbolSize(isize, isize, drawingDef.symbol);
             const newDD = {...drawingDef, size: symbolSize};
 
-            setSize(size);
+            setSize(isize);
             setValidSize(validSize);
             setDrawingDef(newDD);
             const dl = getDrawLayersByDisplayGroup(getDlAry(), displayGroupId);
--- a/src/firefly/js/ui/PointShapeSizePicker.jsx
+++ b/src/firefly/js/ui/PointShapeSizePicker.jsx
@@ -2,10 +2,9 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import React, {PureComponent} from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 import {get} from 'lodash';
-import {flux} from '../core/ReduxFlux.js';
 import {dispatchShowDialog} from '../core/ComponentCntlr.js';
 import {PopupPanel} from './PopupPanel.jsx';
 import {DrawSymbol} from '../visualize/draw/DrawSymbol.js';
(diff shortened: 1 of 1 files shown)


kpuriIpac:
I will remove imports for PureComponent, flux and clone. Also need to fix the double quotes in importing useStoreConnector, that got past me.

robyww (approved):
This looks really good. I think you have mostly nailed it.

- clean up the stuff I noted.
- look for any `var` and change to `let` or `const` since that better documents what is happening.
- The biggest issue is just recognizing that you don't need the `iAmMounted` any more.

Clean up this stuff, squash, and merge.

loitly (approved):
Trey already did a thorough review of the code.  So, I did some testing, instead.  I noticed an existing bug.  If it's easy to fix, will you go ahead and fix it?  Otherwise, just create a ticket for it, and we'll fix it at a later time.

Enter `100 abc` in `Symbol Size`.  You'll notice it shows as valid and the symbol to the right is much bigger than it should.

robyww (approved):
Latest change looks good. Go ahead and merge.

robyww:
Remember to rebase and squash.

PR #1272, opened 2022-09-20
[FIREFLY-1079] Update Plotly to v2.18.0

tgoldina:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1079

Test deployment: https://fireflydev.ipac.caltech.edu/firefly-1079-plotly2-18/firefly/

- Updated Plotly from v1.58.4 to v 2.18.0
- Plotly now supports persistent selections: additional selection areas can be added with Shift-MouseDrag. However Firefly filtering from a chart currently works for a single rectangular selection. If multiple areas are selected, Firefly will issue a warning, but preserve the selections layer. Selecting points in multiple areas is allowed. (The points then can be filtered from the table.) 
- Removing selections layer now requires removing `layout.selections`.


src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -260,15 +263,21 @@ function onSelect(chartId) {
                 } else {
                     dispatchChartUpdate({
                         chartId,
-                        changes: {'selection': {points, range: {x: [xMin, xMax], y: [yMin, yMax]}}}
+                        changes: {
+                            'selection': {
```
robyww, line 267: you don't need the single quotes just `selection : {`

src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -239,6 +239,9 @@ function onClick(chartId) {
 function onSelect(chartId) {
     return (evData) => {
         if (evData) {
+            // multiple selections are not supported
+            const selections = get(evData, 'selections', [])
```
robyww, line 243: we are trying to get away from `get`, instead use the new syntax.
```js
const selections = evData?.selections ?? []
```

robyww (commented):
Just took a quick look at the code and left a couple of small comments. I will let @loitly do the real review.

Thanks for the work.

loitly (approved):
Changes look good.

PR #1273, opened 2022-09-21
FIREFLY-1084: converting Download Options Dialog to a functional component

kpuriIpac:
#### [Firefly-1084](https://jira.ipac.caltech.edu/browse/FIREFLY-1084): converting the DownloadOptionsDialog class to a functional component 
 - converted DownloadOptionsDialog to a functional component 
 - used useStoreConnector instead of flux.addListener directly 
 - made changes to PointShapeSizePicker.jsx's ShapePicker functional component to update state in useEffect instead of (incorrectly) in useStoreConnector itself 

#### Testing
 -  https://firefly-1084-downloadoptionsdialog-functional.irsakudev.ipac.caltech.edu/irsaviewer/
 - go to Catalogs -> NED -> m1 -> Radius: 100 
 - Click on the "Save" icon for the image and for the table (chart save does not use this Download Options Dialog box) 
 - For image, try switching between FITS Image, PNG File and Region File and change the file name
 - For table, try changing the file format and file name
 - Click on "Save" to make sure it downloads the image/table in the selected format 
 - Switch between **Local File** and **Workspace** (**Workspace** may not work in this build and will give an error, since I don't think you can actually login here - but it behaves as it would in ops/dev when you're not logged in)
 - Click on "Save" after switching to **Workspace**, then switch back to **Local File** and make sure the 'Workspace access error' feedback goes away

### DIFF
--- a/src/firefly/js/ui/DownloadOptionsDialog.jsx
+++ b/src/firefly/js/ui/DownloadOptionsDialog.jsx
@@ -1,19 +1,20 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import React, {PureComponent} from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
-import {flux} from '../core/ReduxFlux.js';
-import {isEmpty, get, isNil} from 'lodash';
+import {isEmpty ,get} from 'lodash';
 import {ValidationField} from './ValidationField.jsx';
 import {RadioGroupInputField} from './RadioGroupInputField.jsx';
-import {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
-import {getWorkspaceList, isExistWorspaceFile, getWorkspacePath, getWorkspaceErrorMsg,
+import FieldGroupUtils, {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
+import {getWorkspaceList, getWorkspaceErrorMsg,
         dispatchWorkspaceUpdate, isAccessWorkspace} from '../visualize/WorkspaceCntlr.js';
-import {WorkspaceSave, workspacePopupMsg} from './WorkspaceViewer.jsx';
+import {WorkspaceSave} from './WorkspaceViewer.jsx';
 
 
 import LOADING from 'html/images/gxt/loading.gif';
+import {useStoreConnector} from 'firefly/ui/SimpleComponent';
+import {flux} from "firefly/core/ReduxFlux";
 export const LOCALFILE = 'isLocal';
 export const WORKSPACE = 'isWs';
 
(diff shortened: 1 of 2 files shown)


loitly (changes requested):
I recommend making my suggested changes before merging.

src/firefly/js/ui/DownloadOptionsDialog.jsx
```diff
@@ -28,152 +28,138 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
...
-                                 padding: 10,
-                                 boxSizing: 'border-box',
-                                 width: 'calc(100%)',
-                                 textAlign: 'center',
-                                 border:'1px solid #a3aeb9'}}>
-                        {message}
-                    </div>
-                );
-            };
+    }, [wsSelectTemp]);
```
robyww, line 78: Using separate effects are fine. It is you call, but you could also so something like this:

```js
    useEffect(() => {
        if (isUpdating !== isUpdatingTemp) setIsUpdating(isUpdatingTemp);
        if (wsList !== wsListTemp) setWsList(wsListTemp);
        if (wsSelect !== wsSelectTemp) setWsSelect(wsSelectTemp);
    }, [wsSelectTemp,isUpdatingTemp,wsListTemp]);
```
In general we don't want to load too much into an effect because if can become difficult to follow and support. However, this one is so straight forward it would be OK to combine.

robyww (commented):
The code mostly looks good. The `storeUpdate` code is exactly how I suggested. However, I think @loitly is right. We keep having better ways to do this stuff so maybe it is worth one more look.

We will decide what to do next after we meet.

src/firefly/js/ui/DownloadOptionsDialog.jsx
```diff
@@ -28,152 +26,98 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
...
+                wrapperStyle={{marginTop: 10}}
+                size={50}
+                fieldKey={'fileName'}
+            />
+
+            {workspace && showLocation()}
+
+            <div  style={{width: dialogWidth, height: dialogHeight}}>
+                {where === WORKSPACE && ShowWorkspace()}
+            </div>
```
loitly, line 118: I think you misunderstood my comment on using uppercase for component name.  Take a look at this:
https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized

The idea of naming functional components with uppercase is so you can use it as JSX.  i.e.
`<ShowWorkspace />`.  Otherwise, function name should be in lower camel case standard.

loitly (approved):
Looks much better now.  Good to go.

src/firefly/js/ui/DownloadOptionsDialog.jsx
kpuriIpac: Got it, that make sense. I'll change this back to showWorkspace then since it's not being used as JSX. Did you want me to change anything else in your original comment though? Or was it meant to informative? Thanks!

kpuriIpac: Nevermind...I got it now. I can leave it with uppercase and call it as JSX instead, right?

kpuriIpac: also just to clarify, then should I go ahead and rename other functions that are returning React components and call them using JSX as well? (loading, showSave, etc. -> Loading, ShowSave and `loading()`, `showSave()` to` <Loading />`, `<ShowSave />`)

robyww: It has to be uppercase. Or JSX will not recognize it.

src/firefly/js/ui/DownloadOptionsDialog.jsx
```diff
@@ -28,152 +26,98 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
...
+                <div style={{marginTop: 10,
+                             boxSizing: 'border-box',
+                             width: 'calc(100%)', height: 'calc(100% - 10px)',
+                             overflow: 'auto',
+                             padding: 5,
+                             border:'1px solid #a3aeb9'
+                             }}>
+                    <WorkspaceSave fieldKey={'wsSelect'}
+                                   files={wsList}
+                                   value={wsSelect}
```
robyww, line 66:
```js
const showSave= (   
                <div style={{marginTop: 10,
                             boxSizing: 'border-box',
                             width: 'calc(100%)', height: 'calc(100% - 10px)',
                             overflow: 'auto',
                             padding: 5,
                             border:'1px solid #a3aeb9' }}>
                    <WorkspaceSave fieldKey={'wsSelect'}  files={wsList} value={wsSelect} \>
               </div>);
```

src/firefly/js/ui/DownloadOptionsDialog.jsx
```diff
@@ -28,152 +26,98 @@ export function getTypeData(key, val='', tip = '', labelV='', labelW) {
...
 
+    const ShowWorkspace = () => {
+
+        const loading = () => {
             return (
-                (isUpdating) ? loading() :
-                    (!isEmpty(wsList) ? showSave() : showNoWSFiles('Workspace access error: ' + getWorkspaceErrorMsg()))
+                <div style={{width: '100%', height: '100%', display:'flex', justifyContent: 'center', alignItems: 'center'}}>
+                    <img style={{width:14,height:14}} src={LOADING}/>
+                </div>
```
robyww, line 51:
```js
const loading= (
           <div style={{width: '100%', height: '100%', display:'flex', justifyContent: 'center', alignItems: 'center'}}>
                    <img style={{width:14,height:14}} src={LOADING}/>
          </div>
);

**Review by robyww — CHANGES_REQUESTED** [review]
Now that you have this code so clean I we able to concentrate on `ShowWorkspace`

The component has a closure component, with more closure functions. It is hard to read and unnecessary. All the functions could just be variables.

It is unnecessary to make `showSave()`, `loading()` and probably  `showNoWSFiles()` functions. There is no performance gain, or reuse, and it just make the code more cluttered. 

Something to remember about react. JSX is just an object, it does not get turned into html and attach to the DOM until render time. So creating and not using a JSX object is not big deal.

Also- some of the more simple components, just make one line

make the final render to look like:

```js
return isUpdating ? loading : !isEmpty(wsList) ? showSave : showNoWSFiles;
```

So it is looking much better, but since you in the middle of it you might as well do a little more clean up.

robyww (approved):

PR #1274, opened 2022-09-22
FIREFLY-1041: add feature to combine charts(spectra)

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1041

Add the ability to combine multiple charts.  Limited to spectra with same axis unit.
Sample data added to the ticket.

To test: https://fireflydev.ipac.caltech.edu/firefly-1041-combine-spectrum/firefly/
Upload 2 or more spectra, then pin them.
Select a spectrum to combine, then click on `Combine Chart`.
A list of 'suitable' charts will appear.  Select 1 or more to continue.
If none show up, then there are no suitable charts to combine with the one you selected.  To test the ability to combine any chart(experimental), check the box `Show All Charts`.

Also added these small changes:
- create source map for both local and dev env
- Docker DevMode get verbose log
- update Table API doc to reflect showInfoButton now default to true
- move and define additional chart jsdoc into chart-typedefs.jsdoc

### DIFF
--- a/buildScript/webpack.config.js
+++ b/buildScript/webpack.config.js
@@ -35,7 +35,7 @@ export default function makeWebpackConfig(config) {
     const localBuild= BUILD_ENV === 'local';
 
     if (!process.env.NODE_ENV) {
-        process.env.NODE_ENV = localBuild ? 'development' : 'production';
+        process.env.NODE_ENV = ['local', 'dev'].includes(BUILD_ENV) ? 'development' : 'production';
     }
 
     console.log('ENV_DEV_MODE: ' + ENV_DEV_MODE);
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,6 +33,7 @@ RUN yarn install --ignore-platform --frozen-lockfile
 FROM node_module AS dev_env
 
 RUN apt-get update \
+    && apt-get install -y vim procps \
 # install tomcat9 for dev mode \
     && curl -O https://dlcdn.apache.org/tomcat/tomcat-9/v9.0.65/bin/apache-tomcat-9.0.65.tar.gz \
     && mkdir /opt/tomcat && tar xzf apache-tomcat-9*tar.gz -C /opt/tomcat --strip-components=1 \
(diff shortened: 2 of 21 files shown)


robyww (approved):
I am approving the PR, but as you said we need to limit the functionality some. I do think it is a good idea to let the science - CCB to see it before you do.

I tested with extraction and it works really well. Maybe nicer than your tested data the the data is more homogeneous.

UI Suggestions:

- In combine charts dialog show the name of the chart that is selected, in some bold text at the top.
- If there is only one chart then show it checked.
- I find `show table` button a little confusing. I think you should make color insensitive the the table is already showing (`color: rgba(0,0,0,.3)`)
- I might require a little more book keeping, but I don't think you should list table that have already been combined as options. I.e. I have 4 tables, I combined 1 and 2. I want to make a chart with that combine one and table 3 or 4. Don't list 1 and 2.
- try to name the trace after the table so that the user does not always have to rename. If you can't then don't use `trace` use `table` that make more sense in this context
- A unit change should change it for all traces in the combined context.

Those are my immediate suggestions. I might do more experimenting and have some more

ejoliet (approved):
I think this will need couple of iterations as we already discussed earlier. Plus feedback from CCB-D now that they actually can play around.
So far it's great!!

Apart from the bugs already shared and discussed that are inherited by the combined axis and probably need attention on plotly level, i have tested the features in many cases and the flow and the different button actions are very intuitive and works all together very well.

Only minor feedback i would add to the early discussion are:
- can we have the 'name' be updated with running number 'trace 0', 'trace 1', etc. Currently is always default to 'trace 0' and user have to manually change it anyway. Suggestion , make it running number increasing? Also i'd like to hear scientist give a feedback on all those default labelling we inherit from plotly, like instead of 'trace', calling it maybe more like 'dataset 0' or 'series 0', but that's up to CCB-D.
- In my test, i used the drill tool and also uploading files separately, when i combined 2 spectra, all the tests i did they all came up with same color so it was hard to visualize different spectras, might be a bug only related to this feature?
- When i try to expand, is not expanding on the active chart, but on the initial first one. I was expecting to see the one i highlight to be expanded though

Rest looks good.

PR #1275, opened 2022-09-23
FIREFLY-1086: Click to Action

robyww:
#### [FIREFLY-1086](https://jira.ipac.caltech.edu/browse/FIREFLY-1086): Click to Action
 - defined data structure to support actions
 - made UI to load searches
 - made a Search Refinement tool

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1086-click/firefly/
 - https://firefly-1086-click.irsakudev.ipac.caltech.edu/irsaviewer/
 - Load an image or a HiPS
   - Select an area
   - Click on searches to see options
 - Do the same thing but this time click on search and choose "search refinement tool"
 - This time use the IrsaViewer url. Load a hips, select an select and 1 degree area, and choose search sofia


lrebull:
ooh cool!!!

I'm not sure what kinds of feedback you are looking for, but I'll just write everything down and you can take or leave it as need be. Numbered for easy rejection. ;)

1. it seems weird to me that the sigma icon (the stats thing) is now off by itself. I mean, i could get used to it, but it seems odd off by itself.
2. searches menu: Actions base on Center: --> Actions based on center: (note verb tense of "based" and capitalization of "center")
3. searches menu: Search using TAP with radius of 0.0159 degrees (need "using" to be clear and you have the space; need units on radius)
4. searches menu: Search using TAP around an area (4 points) - i've tried this and i'm still not sure what it's doing that is different than the prior - is it cone vs. polygon? then use those words.  "Search (cone) using TAP..." or "Cone search using TAP..." and "Search (polygon) using TAP..." etc. -- make it parallel 
5. searches menu: Search Fits -> Search FITS
6. first search I did took me out of the select mode, which seemed weird to me. When the TAP search came back, I was no longer in select mode, so I had to go initiate it again. All subsequent searches didn't do this, so maybe it was 'just me' (again)?
7. because the additional "select options" menu is so long, when i did an additional image search, i then promptly lost the searches menu. I had to drag the window pane way over to be able to see the searches menu at all. (see screenshot) i don't know exactly what the right answer to this is, but if you want the searching from images to be as central a feature as I think you (we?) do, this can't be as buried as this.
8. additional image searches don't have the "select options" menu or region overlaid. They do have the catalogs overlaid. This seems weird to me, not that i have a clean/clear solution for you (see screenshot)
9. there are (still?) weird interactions between the select-a-region and the search refinement tool. I clicked something and exited the select-a-region while still in the search refinement tool and now all i can to to change it is type in the yellow boxes in the pop-up. selecting the region tool again clears the confusion (and it does tell me to do that, so it's fine; it just was initially confused). 
10. searches menu from the search refinement tool needs similar changes to the suggestions above.
Screen Shot 2022-09-23 at 12 08 33 PM

robyww:
### @lrebull I just did an update with the easy stuff you mentioned. I also tried to respond to the hard items.

- [x]  1. it seems weird to me that the sigma icon...

- [x] 2.  Actions based on center

- [x] 3. searches menu: Search using TAP with radius of 0.0159 degrees (need "using" to be clear and you have the space; need units on radius)

- [x] 4. searches menu: Search using TAP around an area (4 points)...

- [x] 5. searches menu: Search Fits -> Search FITS

- [ ] 6. first search I did took me out of the select mode, which seemed weird to me. When the TAP search came back, I was no longer in select mode, so I had to go initiate it again. All subsequent searches didn't do this, so maybe it was 'just me' (again)?

   - _from Trey: I don't quite understand this one._

- [ ] 7. because the additional "select options" menu is so long,...

   - _from Trey: I am also concerned about this. I think the searches dropdown needs to be more prominent_

- [ ] 8. additional image searches don't have the "select options" menu or region overlaid....

   - _from Trey: When select is on and more images are added, select is not automatically set on for those images. I am not sure this is worth doing. You are going to need to convince me._

- [ ] 9. there are (still?) weird interactions between the select-a-region and the search refinement tool....

  - _from Trey: This is the way I intended to to work. When to refinement turns on the select does on select and then turns off and the refinement tool takes over. Anything else would probably create too much competition between the two layers._

- [x] 10. searches menu from the search refinement tool needs similar changes to the suggestions above.

src/firefly/js/Firefly.js
```diff
@@ -206,15 +214,18 @@ function installOptions(options) {
  * @param {AppProps} props
  * @param {FireflyOptions} options
  * @param {Array.<WebApiCommand>} webApiCommands
+ * @param {Array.<ClickToActionCommand>} clickToActionCommands
  */
-function fireflyInit(props, options={}, webApiCommands) {
+function fireflyInit(props, options={}, webApiCommands, clickToActionCommands) {
```
loitly, line 219: These `webApiCommands, clickToActionCommands` are special where they cannot be in `options`?  My concern is the growing trend of these added features.

loitly (approved):
Looks good.

src/firefly/js/Firefly.js
robyww: oh I meant to take that parameter out. I was just experimenting. I put in options.

lrebull:
number 3 - you need a units on the radius

number 6 - here is a video with more words to be clearer:
https://user-images.githubusercontent.com/23038600/192058189-16e24df6-7c58-40d6-86ad-28200df43762.mp4

robyww:
New push and build: added units and turned off select when a search action is clicked.

lrebull:
nice!

ejoliet (approved):
I think all looks very close to what Vandana had in mind or at least from the slides shown for IRSAViewer and Euclid/SPHEREx.

The only search missing i would suggest to add is the search for "Catalogs" with cone, and polygon - that would complete the design request for IRSAViewer. (Will differ for Euclid/SPHEREx).
But you can add it in another next PR.

Other than that, works smoothly and the popup to follow-up on the positioning of the search is also a good idea for scientist to look at the potentials or new ways to interact with the original design.
Thanks!

PR #1276, opened 2022-09-30
FIREFLY-1085: removing reducer from fitsDownloadDialog

kpuriIpac:
#### [Firefly-1085](https://jira.ipac.caltech.edu/browse/FIREFLY-1085): removing reducer from FitsDownloadDialog
 - Took a bit longer than I expected, but got rid of a lot of code and encountered a couple of bugs and partially fixed them (see below) 
 - Got rid of the reducer and and used useFieldGroupValue to get the file type, name and location and then monitor changes in any of those via a useEffect (which basically sets a new filename when the user changes the file name or extension/file type or picks a new filename from the workspace) 
 - made MakeFileOptions a react component

#### Testing
 -  https://firefly-1085-fitsdownloaddialog-remove-reducer.irsakudev.ipac.caltech.edu/irsaviewer/
 - go to TAP -> service: IRSA/NED -> Coordinates or Object Name: m1 -> Radius: 100 
 - open the fits download dialog by clicking on the **save** icon
 - ensure this works as expected. try changing file type, file name, and file location (the only thing I wasn't able to test was workspace - I am having trouble logging in on dev/ops so I haven't been able to mimic being logged in on the test build yet to test saving files to the workspace - if you can test this, please make sure it works - I'll try and test this myself too) 
 - go to Images ->  Create 3-Color Composite -> Select any image for at least one of red/green/blue and search
 - Now open the fits download dialog box again, you should see "**Color Band**" this time 
 - toggle between file types (if you select png or reg, the color band goes away - is this expected behavior? this is also related to the bug fix below) 

#### Bug Fixes
1. In irsadev or ops, open a fits download dialog box after searching for a 3-color composite image. You will not be able to switch any of the radio buttons or edit the file name (you'll get uncaught TypeError in your console)
- I was able to re-create this locally, and fixed it in the **makeFileName** function by checking that **req** exists before we try to access its properties (line 286) 
- In the test build, when you open a fits download dialog box from a 3-color composite image search, you should now be able to toggle the radio buttons and edit the file name (you'll not be able to backspace from the end, but you can select the entire string and delete and enter a new file name - this is related to the 2nd bug below) 
2. This is a partial fix: In irsadev or ops, open a fits download dialog box on a "View FITs Image" (not 3-color composite). 
- If you try to edit the file name here, you won't be able to backspace from the end but can select and edit the file name directly (before the extension). 
- But if you select the entire string and delete it, you'll still be left with the extension ("**.fits**" or "**.png**" or "**.reg**"). If you click "**Save**" a file named "fits.txt" will be saved if you had file type fits selected. This does not seem right. In the test build, I made it so that you can select and delete the entire string. If you try to save now (with the file name field completely empty), a file with a default name and the correct extension will be downloaded. e.g.: image_SEIP-MIPS24.fits 
- And if you start typing a new string, the appropriate extension is added at the end of the string. 
-  on the test build, open a fits download dialog box from a 3-color composite image search. Here, with "fits" selected as the file type, if you try to select and delete the existing file name, you'll be left with a ".fits" and if you click save, it'll be saved as "fits.txt". Essentially, this is similar to the existing bug on dev/ops for a regular FITS image search (or any non 3-color search). The fix for this wasn't as straightforward but I can spend some more time looking at it and fixing it, but I just wanted to check the behavior of the download dialog for a 3-color composite image search. Should the user still be able to switch to **png** and **reg** here? And when they switch to **png** or **reg**, should the "**Color Band**" field still be visible? I think yes for both - and in that case I can fix this and push again. But since on dev/ops you can't switch radio buttons at all (due to the TypeError), I wanted to check what the behavior should look like.

### DIFF
--- a/src/firefly/js/ui/FitsDownloadDialog.jsx
+++ b/src/firefly/js/ui/FitsDownloadDialog.jsx
@@ -1,35 +1,38 @@
 /*
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
-import React, {memo,useState} from 'react';
+import React, {memo, useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
-import {set, isEmpty, capitalize} from 'lodash';
+import {isEmpty, capitalize} from 'lodash';
 import {dispatchShowDialog, dispatchHideDialog, isDialogVisible} from '../core/ComponentCntlr.js';
 import {Operation} from '../visualize/PlotState.js';
-import {getRootURL, getCmdSrvSyncURL, encodeUrl, updateSet, replaceExt} from '../util/WebUtil.js';
+import {getRootURL, getCmdSrvSyncURL, encodeUrl, replaceExt} from '../util/WebUtil.js';
 import {RadioGroupInputField} from './RadioGroupInputField.jsx';
 import CompleteButton from './CompleteButton.jsx';
 import {FieldGroup} from './FieldGroup.jsx';
 import DialogRootContainer from './DialogRootContainer.jsx';
 import {PopupPanel} from './PopupPanel.jsx';
-import FieldGroupUtils, {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
-import {primePlot, getActivePlotView, getAllCanvasLayersForPlot, isThreeColor} from '../visualize/PlotViewUtil.js';
+import {getFieldVal} from '../fieldGroup/FieldGroupUtils.js';
+import {
+    primePlot,
+    getActivePlotView,
+    getAllCanvasLayersForPlot,
+    isThreeColor
+} from '../visualize/PlotViewUtil.js';
 import {Band} from '../visualize/Band.js';
 import {visRoot} from '../visualize/ImagePlotCntlr.js';
 import {isImage} from '../visualize/WebPlot.js';
 import {makeRegionsFromPlot} from '../visualize/region/RegionDescription.js';
 import {saveDS9RegionFile} from '../rpc/PlotServicesJson.js';
-import FieldGroupCntlr from '../fieldGroup/FieldGroupCntlr.js';
-import {DownloadOptionsDialog, fileNameValidator, getTypeData, WORKSPACE, LOCALFILE} from './DownloadOptionsDialog.jsx';
-import {isValidWSFolder, WS_SERVER_PARAM, getWorkspacePath, isWsFolder, dispatchWorkspaceUpdate} from '../visualize/WorkspaceCntlr.js';
+import {DownloadOptionsDialog, WORKSPACE, LOCALFILE} from './DownloadOptionsDialog.jsx';
+import {isValidWSFolder, WS_SERVER_PARAM, getWorkspacePath, dispatchWorkspaceUpdate} from '../visualize/WorkspaceCntlr.js';
 import {doDownloadWorkspace, workspacePopupMsg, validateFileName} from './WorkspaceViewer.jsx';
 import {ServerParams} from '../data/ServerParams.js';
 import {INFO_POPUP, showInfoPopup} from './PopupUtil.jsx';
 import {getWorkspaceConfig} from '../visualize/WorkspaceCntlr.js';
 import {upload} from '../rpc/CoreServices.js';
 import {download, downloadBlob, makeDefaultDownloadFileName} from '../util/fetch.js';
-import {useBindFieldGroupToStore} from './SimpleComponent.jsx';
-
+import {useFieldGroupValue} from './SimpleComponent.jsx';
 import HelpIcon from './HelpIcon.jsx';
 
 const STRING_SPLIT_TOKEN= '--STR--';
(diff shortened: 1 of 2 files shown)


src/firefly/js/ui/FitsDownloadDialog.jsx
```diff
@@ -147,19 +128,47 @@ function getInitState()  {
...
+                //checking if filename is !empty string allows user to delete the input
+                if (fileName) fName = replaceExt(fileName, fileType);//replaceExt(fields.fileName.value, fType);
+            } else {
+                fName = matchPossibleDefaultNames(plot, fileName) ?
+                    makeFileName(plot, fileName, fileType) :
+                    replaceExt(fileName, fileType);
+            }
+        }
+        setFileName(fName);
+    }, [getFileType, getFileName, getLocation]);
```
loitly, line 161: Personally, I find the behavior very awkward.
I bring up the dialog and wanted to change the name.  So, I select all, then begin to type... 'a', immediately it jumped to 'a.fits', at this point, I cannot backspace(delete) because my cursor is at the end and the field will not let me enter anything that does not end with '.fits'.  But, I don't know that.  I thought my keyboard is stuck.
But if I leave it blank, then I can process to save?  This is odd.  It let me save, but not a file name I chose.
Anyways, I understand this is existing behavior.  Therefore, this is just a comment and not a change request.

src/firefly/js/ui/FitsDownloadDialog.jsx
```diff
@@ -319,7 +283,7 @@ function makeFileName(plot, band='NO_BAND', ext= 'fits') {
             case Band.BLUE.key: root= 'image-blue'; break;
         }
     }
-    const title= req.getDownloadFileNameRoot() ? req.getDownloadFileNameRoot() : plot.title;
+    const title= req?.getDownloadFileNameRoot() ? req?.getDownloadFileNameRoot() : plot.title;
```
loitly, line 286: This can be shorten ..
```
    const title= req?.getDownloadFileNameRoot() ?? plot.title;     
```
Or || instead of ?? depending on whether or not you allow blank.

loitly (changes requested):
Save to workspace did not work for me.  Please investigate.

src/firefly/js/ui/FitsDownloadDialog.jsx
```diff
@@ -147,19 +128,47 @@ function getInitState()  {
...
+    useEffect(() => {
+        const fileType = getFileType();
+        const fileName = getFileName();
+        const fileLocation = getLocation();
+        let fName = '';
+
+        // change the filename if a file is selected from the file picker
+        if (fileLocation === 'isWs' && isValidWSFolder(fileLocation, false).valid) {
+            fName = fileLocation.substring(fileLocation.lastIndexOf('/') + 1);
+        }
```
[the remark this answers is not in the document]
kpuriIpac: Yes that makes sense, I think I missed this. I'll fix it and try logging into workplace again to see if I can test saving it myself too.

robyww:
Also- one more thing. After you get the workspace working on you machine. Please see if line 59 and 60 are even working.
```js
  const fileLocation = getFieldVal(fitsDownGroup, 'fileLocation', LOCALFILE);
  if (fileLocation === WORKSPACE) dispatchWorkspaceUpdate();
```
I am concerned that `fileLocation` is always `undefined`

kpuriIpac:
@loitly @robyww can you please look at my changes?

- the default name for 3-color images now reflects the color band (for instance, previously it was "image_3-color.fits". Now if the color band is "red", the name will be "image-red_3-color.fits"). 
 
- You can now delete the entire input string for the file name even with the download dialog open for a 3-color image (previously if you tried to do this in my PR, you would be left with a default extension like ".fits". And in ops/dev, you can't edit the name at all). If you try to save the file with an empty file name, the default file name and extension will be used as the file name. 

- If the user still manages to click save with the file name set to ".fits" or ".png" or ".reg", previously in my PR (and also on ops/dev) this would be saved as "fits.txt" or "png.png" or "reg.txt" respectively. Fixed it so that now if the file name is, for instance, ".fits", it'll be saved as "defaultName.fits" instead. (The fix for this was a check that I added, where if the file name is of the format "." + extension (".fits")- then i set file name to empty - because in the following line I have a check where makeFileName is called to set a default file name when the given file name is empty). 

About workspace: 
- I was making progress on workspace yesterday (was initially able to access workspace locally). But halfway through the day I lost access again (I am only able to access it on ops and on my Jenkins build sometimes) but not on irsadev or on my localhost after logging into irsadev. Now I simply get "workspace access error". Tried to fix it further without being able to debug locally. 

- I think workspace still works (mostly), the only thing that doesn't work is: if you try to access workspace on ops, enter a new file name and click save, it'll give you an error that says "please select a workspace folder" and when you click on "WS_Home" and then click save again, your file will be saved. On my build, you won't get this error if you don't have the workspace selected (but your file won't be saved as will, which isn't intuitive for the user). I would like to fix this but it's hard to do without being able to debug/study what's going on here.

src/firefly/js/ui/FitsDownloadDialog.jsx
```diff
@@ -147,19 +127,51 @@ function getInitState()  {
 }
 
 const FitsDownloadDialogForm= memo( ({isWs, popupId, groupKey}) => {
-    const [{pv,plot,hasOperation,threeC,colors}]= useState(getInitState);
-    useBindFieldGroupToStore(groupKey); // this will force a rerender when anything changes
-    const band = threeC ? getFieldVal(groupKey, 'threeBandColor', colors[0]) : Band.NO_BAND.key;
-    const totalChildren = (isWs ? 3 : 2) +  (hasOperation ? 1 : 0) + (threeC ? 1 : 0);// fileType + save as + (fileLocation)
-    const childH = (totalChildren*(20+mTOP));
+    const [{pv, plot, hasOperation, threeC, colors}] = useState(getInitState);
```
robyww, line 130: I took one more look at this line and realized that is it completely unnecessary and wrong. We are doing a `useState` without a set. That means that it does not function as state at all. All these fields could be computed in line. 

Remove `getInitState` and replace with.

```js
const pv= getActivePlotView(visRoot());
const plot = primePlot(pv);
const colors= getColors(plot);
const threeC= isThreeColor(plot);
const hasOperation= plot?,plotState.hasOperation(Operation.CROP) ?? false;
```

If we wanted to dialog to update when the user switches plots (under the dialog) we would use `useStoreConnector` and the replace the `const pv=...` with
```js
 const pv= useStoreConnector( () => getActivePlotView(visRoot()));
```
but that is not how it works right now (maybe if we ever get that request we will do that.)

I think I put this line in when I converted it to a functional component, I should have recognized the problem then.

robyww (approved):
Your changes looked good. I found one more preexisting issue that I would like you to fix that I mentioned below.  After you do that I am fine for you to merge. I don't need to look at it again.

loitly (approved):
Works fine.  Good to go.

PR #1278, opened 2022-10-06
firefly-1103: Fix firefly fITS reader so it can read Multiple large HDUs for a 14G file

robyww:
#### [Firefly-1103](https://jira.ipac.caltech.edu/browse/FIREFLY-1103) Allows firefly fits ready to read even bigger files

- fix the `TableUtil.guessFormat()` so it only looks ahead 1k
- work with longs for the file offset header `SPOT_OFF`
- change property to increate the max fits file size

This is a small PR. You can really test it unless you start you server at about 20G and have a 14G fits file. So, just look at the code.

### DIFF
--- a/config/app.config
+++ b/config/app.config
@@ -26,7 +26,7 @@ sso.user.profile.url = "https://irsa.ipac.caltech.edu/account/uman/uman.html//id
 __$help.base.url = "onlinehelp/"
 
 
-visualize.fits.MaxSizeInBytes= 10737418240
+visualize.fits.MaxSizeInBytes= 21474836480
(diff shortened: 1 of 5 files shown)


loitly (approved):
I didn't realize how confusing `TableUtil.guessFormat` is.  There are many assumptions, like startsWith without checking for whitespaces and the logic for guessing TSV vs CSV is confusing.  But, since there's no complains, it must be good enough.

robyww:
I is a little hard to follow. But the big probably is that was readying a ton of data for any fits file. Basically the whole thing.

PR #1280, opened 2022-10-14
Firefly-1060: Support datatype and other features in extract

robyww:
#### [Firefly-1060](https://jira.ipac.caltech.edu/browse/FIREFLY-1060): clean and support datatype in extract
  - Also added some features mentioned in firefly-1059, combinations types average, add and bitwise OR
  - Some general clean up, refactored some code around to make it cleaner
     - new files: `FitsExtract.java` and `FITSExtractToTable.java`

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1060-extract/firefly/
 - Do an extraction and pin a table - notice the unit type should be correct with you FITS
 - For a over `1x1` size you have access to `average`, `add`, or `OR` (`OR` only available to int or long HDUs)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ExtractFromImage.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ExtractFromImage.java
@@ -4,9 +4,10 @@ import edu.caltech.ipac.firefly.data.ServerParams;
 import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.server.SrvParam;
 import edu.caltech.ipac.table.DataGroup;
-import edu.caltech.ipac.table.io.FITSTableReader;
+import edu.caltech.ipac.table.io.FITSExtractToTable;
 import edu.caltech.ipac.visualize.plot.ImagePt;
 import edu.caltech.ipac.visualize.plot.WorldPt;
+import edu.caltech.ipac.visualize.plot.plotdata.FitsExtract;
 import nom.tam.fits.FitsException;
 
 import java.io.IOException;
@@ -48,6 +49,7 @@ public class ExtractFromImage extends EmbeddedDbProcessor {
         int extractionSize = req.getIntParam(EXTRACTION_SIZE, 1);
         int refHduNum = req.getIntParam(REF_HDU_NUM, -1);
         int plane = req.getIntParam(ServerParams.PLANE, -1);
+        FitsExtract.CombineType ct= Enum.valueOf(FitsExtract.CombineType.class,req.getParam(ServerParams.COMBINE_OP,"AVG"));
         boolean allMatchingHDUs = req.getBooleanParam(ALL_MATCHING_HDUS, true);
         try {
             if (extType == null || extType.equals("z-axis")) {
@@ -57,18 +59,21 @@ public class ExtractFromImage extends EmbeddedDbProcessor {
                 String wlUnit= req.getParam(ServerParams.WL_UNIT);
                 Map<Integer,String> fluxUnit= makeMapOfUnitsFromParam(req);
                 double[] wlAry= SrvParam.getDoubleAryFromJson(req.getParam(ServerParams.WL_ARY));
-                return FITSTableReader.getCubeZaxisAsTable(pt, wpt, filename, refHduNum, allMatchingHDUs, extractionSize, wlAry,wlUnit,fluxUnit);
+                return FITSExtractToTable.getCubeZaxisAsTable(pt, wpt, filename, refHduNum, allMatchingHDUs,
+                        extractionSize, ct, wlAry,wlUnit,fluxUnit);
             }
             else if (extType.equals("line")) {
                 ImagePt[] ptAry= SrvParam.getImagePtAryFromJson(req.getParam(ServerParams.PTARY));
                 WorldPt[] wptAry= SrvParam.getWorldPtAryFromJson(req.getParam(ServerParams.WPT_ARY));
                 checkPointParams(ptAry, plane, filename, refHduNum);
-                return FITSTableReader.getLineSelectAsTable(ptAry, wptAry, filename, refHduNum, plane, allMatchingHDUs, extractionSize);
+                return FITSExtractToTable.getLineSelectAsTable(ptAry, wptAry, filename, refHduNum, plane, allMatchingHDUs,
+                        extractionSize, ct);
             } else if (extType.equals("points")) {
                 ImagePt[] ptAry= SrvParam.getImagePtAryFromJson(req.getParam(ServerParams.PTARY));
                 WorldPt[] wptAry= SrvParam.getWorldPtAryFromJson(req.getParam(ServerParams.WPT_ARY));
                 checkPointParams(ptAry, plane, filename, refHduNum);
-                return FITSTableReader.getPointsAsTable(ptAry, wptAry, filename, refHduNum, plane, allMatchingHDUs, extractionSize);
+                return FITSExtractToTable.getPointsAsTable(ptAry, wptAry, filename, refHduNum, plane,
+                        allMatchingHDUs, extractionSize, ct);
             }
         } catch (IOException | FitsException e) {
             throw new IllegalArgumentException("Could not make a table from extracted data");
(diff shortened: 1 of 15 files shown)


lrebull:
OK, it seems to be working, and is really cool! But the tooltip for the new average/sum/logical OR hasn't been updated and as such no longer applies. And I am really confused about what exactly the logical OR is doing. Could you explain it to me in more words?

robyww:
I logical OR does a binary OR on the fields  (use `|` for `or`)
- 2 | 3= 3
- 2 | 8 = 10
- 2 | 2 = 2
- 2| 5 | 5 = 7
try it at the python prompt

src/firefly/java/edu/caltech/ipac/firefly/server/query/ExtractFromImage.java
```diff
@@ -57,38 +59,38 @@ public DataGroup fetchDataGroup(TableServerRequest req) throws DataAccessExcepti
...
-                if (s.length==2) {
-                    try {
-                        int hdu= Integer.parseInt(s[0]);
-                        fluxUnit.put(hdu,s[1]);
-                    } catch (NumberFormatException ignore) { }
-                }
+        if (sAry==null) return fluxUnit;
+        for(String entry : sAry) {
+            String[] s= entry.split("=");
+            if (s.length==2) {
```
loitly, line 90: FYI, If you use split("=", 2), then you can accept '=' as part of the value.  Not sure what the requirement is.

robyww:
Is this the tip: `'Extract the average pixel values in the specifed aperture centered on the pixel closest to where you clicked.'`

What should it be?

src/firefly/java/edu/caltech/ipac/table/io/FITSExtractToTable.java
```diff
@@ -0,0 +1,265 @@
...
+        if (drillSize<2) return desc;
+        return desc+ " ("+drillSize+"x"+drillSize+","+ct.toString()+")";
+    }
+
+    private static Class<?> getDataType(List<Number> valueList) {
+        Number n= getFistNonNaN(valueList);
+        if (n instanceof Double) return Double.class;
+        if (n instanceof Float) return Float.class;
+        if (n instanceof Integer) return Integer.class;
+        if (n instanceof Long) return Long.class;
```
loitly, line 63: n.getClass() is not good enough?

ejoliet (approved):
yes! Tested. Great!
Thanks!

src/firefly/java/edu/caltech/ipac/table/io/FITSExtractToTable.java
```diff
@@ -0,0 +1,265 @@
...
+    private static String makeMetaEntryForHDUs(List<FitsExtract.ExtractionResults> results) {
+        StringBuilder str= new StringBuilder();
+        for(FitsExtract.ExtractionResults r : results) {
+           if (str.length()>0) str.append(";");
+           str.append(makeKeyforHDUTab(r)).append("=").append(r.hduNum());
+        }
+        return str.toString();
+    }
+
+    private static Number getFistNonNaN(List<Number> valueList) {
```
gpdf, line 46: Is this a typo for "First"?

src/firefly/java/edu/caltech/ipac/table/io/FITSExtractToTable.java
```diff
@@ -0,0 +1,265 @@
...
+        return valueList.get(0);
+    }
+
+    private static String addSize(String desc, int drillSize, FitsExtract.CombineType ct) {
+        if (drillSize<2) return desc;
+        return desc+ " ("+drillSize+"x"+drillSize+","+ct.toString()+")";
+    }
+
+    private static Class<?> getDataType(List<Number> valueList) {
+        Number n= getFistNonNaN(valueList);
```
gpdf, line 59: "First"?

src/firefly/java/edu/caltech/ipac/table/io/FITSExtractToTable.java
```diff
@@ -0,0 +1,265 @@
+package edu.caltech.ipac.table.io;
```
gpdf, line 1: This whole file was just refactored out of FITSTableReader, right?  So it's not a big a mod as it looks...

gpdf (commented):
Thank you, this is great.  It's really pretty useful to be able to work accurately with flags, and this will benefit both Rubin and SPHEREx, at least.

Can we get the names of the image planes - when available - used as the column names in the table?  E.g., "FLUX", "MASK", "VARIANCE" (which you are already able to display in the image title bar), instead of "HDU #1", "HDU #2", "HDU #3".

Comments about the Average/Sum/OR menu: 

1. This is a really great feature when someone uses the larger apertures (and will also be super-useful if/when we add the ability to extract from 1xN regions)!  But, one complication:
2. This menu only appears when you start with the mask image as the place where you draw your line - but it's (much) more likely that someone will draw a line on the flux image and then use "Pin Table" to take advantage of the tool's ability to extract from all the planes at ones.

I think at the moment it might be asking for too much to find a solution for when/where the user would see the additional UI element to control the processing of other integer-valued planes.

So this is my recommendation:

1. Make the default pixel-combining action for integers be "OR" when this is done "offscreen".  Rationale: if someone is working directly on an integer image, it might be a real integer-ADU flux image, for which OR is the wrong answer, but if someone is working on a different image plane (like a floating-point flux) and an integer plane is processed _in addition_ only as part of "Pin Table", it is far, far more likely to be a mask image, and OR is almost certainly the best default.
2. Make sure the documentation makes this clear, and explains to the user that if they don't like this default, they can start on the integer plane, and still have "Pin Table" processing the/any floating-point plane(s).

Finally, can the mouse-over for the Average/Sum/OR menu have a different tooltip depending on which menu item is currently selected?

gpdf:
If it's possible to have menu-selection-specific tooltips, I'd suggest:
* "Average the pixel values in the aperture; result is always floating-point"
* "Sum the pixel values in the aperture"
* "Compute the bitwise OR of the pixel values in the aperture"

gpdf:
Also, if the current plane is floating-point, and the "Average/Sum/OR" menu is not displayed, then the tooltip for the aperture size menu should still say "Extract and average", rather than "Extract and manipulate".

lrebull:
We need to get into the tooltip something about "centered on the pixel closest to where you clicked" because that is still a big deal - it's not interpolating between pixels.

gpdf:
@lrebull You have that in the tooltip that you recommended for the aperture-size menu.  I'm trying to avoid having the two duplicate each other.

lrebull:
right, it just created so much ...drama in testing, resulting in, e.g., this figure (and associated text) in the docs: https://irsa.ipac.caltech.edu/onlinehelp/irsaviewer/irsaviewer/img/notinterpolating.png 
that i want to make sure it's easily discoverable.

gpdf:
It's not really "closest to where you clicked" for the line-extract tool, of course.  But that gets hard to explain in a tooltip!

gpdf:
That figure is also pretty problematic and misleading, because of the half-pixel offsets (which I recognize are in the data, not originating in the figure).

gpdf:
I've been trying to investigate exactly what coordinates are reported in the table, and I _think_ they are inconsistent with what we report in the coordinate readout dialog, by a half-integer offset.

gpdf:
Another issue that I think was missed in the first round of work on this tool:

If "lock by click" is selected, drawing a line-extract changes the _values_ shown in the coordinates dialog to the end-point of the drawn line, *but does not erase the previous lock-by-click mark*, making them inconsistent.

I would prefer that the previous lock-by-click location be preserved, but if that is too hard, then we can discuss what would be a fallback - either move the mark to the end of the line, or clear the locked click altogether.

robyww:
@gpdf @lrebull My responses to the rather overwhelming amout of discussion.
- Some of the issues are out of scope of the ticket. I think there needs to be bug ticket written.
- I think the tools tips are too much. However, they are easy to implement. If you really want them at that detail then give me the exact tooltip you want for all 4 combinations. If I don't hear back I am just going with @lrebull original single tooltip suggestion.
- I think I can do the xor default for integer offscreen HDU's

src/firefly/java/edu/caltech/ipac/table/io/FITSExtractToTable.java
robyww: fixed - thanks

robyww:
@gpdf @lrebull I updated the build.
https://fireflydev.ipac.caltech.edu/firefly-1060-extract/firefly/

gpdf:
I think we might have to ask the CCB-D about the tooltips.  I'll submit bug tickets for the other odd things I've seen.  I'll review the updated build tomorrow (out of time for tonight).

gpdf:
Actually I did a quick check just now.  Looks good.  Verified: column naming from extension name works for SPHEREx images; automatic OR for "offscreen" 7x7 sampling of integer mask works as expected; Luisa's tooltip in place.  Have not tested Rubin images.  I still think we need more clarity in the tooltips, but I'll discuss with CCB-D.

robyww:
I will go ahead and merge we can add the tooltips at any time.

PR #1282, opened 2022-10-24
[FIREFLY-1095] Combine Spectra with unit conversion

tgoldina:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1095
Test deployment: https://fireflydev.ipac.caltech.edu/firefly-1095-combinedunits/firefly/

- On combined chart creation, the units are converted to the unit of the base trace for all traces with compatible units
- On combined chart update, the units are applied to all traces with compatible units 
- Errors in non-active traces are preserved.

### DIFF
--- a/src/firefly/js/charts/ui/CombineChart.jsx
+++ b/src/firefly/js/charts/ui/CombineChart.jsx
@@ -303,19 +303,11 @@ function doUnitConversion({chartId, chartData, axisType, to}) {
                 traceNum: activeTrace,
                 axis
             });
-            // changes for fields must be converted to changes for state
+            // field changes must be converted to state changes
             changes = evalChangesFromFields(chartId, tbl_id, changes);
             Object.entries(changes).forEach(([key, val]) => {
                 set(chartData, key, val);
             });
-            // for (let [k, v] of Object.entries(changes)) {
-            //     if (k.startsWith('_tables')) {
-            //         // the unit conversion applies to chart state, not to input fields
-            //         k = k.replace('_tables.', '');
-            //         v = `tables::${v}`;
-            //     }
-            //     set(chartData, k, v);
-            // }
             unit = to;
             set(fireflyData, [activeTrace, `${axisType}Unit`], unit);
         }
--- a/src/firefly/js/charts/ui/CombineChart.jsx
+++ b/src/firefly/js/charts/ui/CombineChart.jsx
@@ -3,7 +3,7 @@
  */
 
 import React, {useEffect, useState} from 'react';
-import {cloneDeep, pick, set} from 'lodash';
+import {cloneDeep, get, pick, set} from 'lodash';
 
 import {getMultiViewRoot, getViewerItemIds} from '../../visualize/MultiViewCntlr.js';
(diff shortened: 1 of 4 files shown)


src/firefly/js/charts/ui/CombineChart.jsx
```diff
@@ -256,12 +256,16 @@ function combineChart(chartIds, props={}) {
...
-        // failed when only one of the two are mapped(??).  apply map to both(workaround)
         const xmapped = (data?.[activeTrace]?.x || '').startsWith?.('tables:');
         const ymapped = (data?.[activeTrace]?.y || '').startsWith?.('tables:');
         if (xmapped ^ ymapped) {
-            if (!ymapped) set(data, `${activeTrace}.y`, `tables::${tablesources?.[activeTrace]?.mappings?.y}`);
-            if (!xmapped) set(data, `${activeTrace}.x`, `tables::${tablesources?.[activeTrace]?.mappings?.x}`);
+            // all mapped fields should be handled: x, y, error_y.array, etc.
+            Object.entries(tablesources?.[activeTrace]?.mappings || {}).forEach(([key, val]) => {
+                // need get to handle compound keys, like 'error_y.array'
+                if (!get(data, `${activeTrace}.${key}`)?.startsWith?.('tables:')) {
```
loitly, line 265: We're trying to do way with lodash's `get`.  Can you use optional chaining (?.) like you did already?
`!data?.[activeTrace]?.[key]?.startsWith.('tables:')`

src/firefly/js/charts/ui/CombineChart.jsx
```diff
@@ -298,9 +303,19 @@ function doUnitConversion({chartId, chartData, axisType, to}) {
...
                 set(chartData, key, val);
             });
+            // for (let [k, v] of Object.entries(changes)) {
+            //     if (k.startsWith('_tables')) {
+            //         // the unit conversion applies to chart state, not to input fields
+            //         k = k.replace('_tables.', '');
+            //         v = `tables::${v}`;
+            //     }
+            //     set(chartData, k, v);
+            // }
```
loitly, line 318: if you want to keep this, add comment to explain when it's needed.

loitly (approved):
Tested and changes look good.  Thanks for fixing this.

src/firefly/js/charts/ui/CombineChart.jsx
tgoldina: In table source `mappings` the key is compound: "error_y.array", while in `data` it is not compound, it is a path to the field.

tgoldina:
I found what triggered "Warning: Can't perform a React state update on an unmounted component." It was caused by active trace change in chart's fetchData, which was triggering a failure in FieldGroupConnector:204 on `setFieldChange({mounted: false})` on a displayed field. I've fixed it and rebuilt.

src/firefly/js/charts/ui/CombineChart.jsx
tgoldina: Removed.

PR #1285, opened 2022-11-14
FIREFLY-1113: Converted CollapsiblePanel to functional component

jaladh-singhal:
Fixes [FIREFLY-1113](https://jira.ipac.caltech.edu/browse/FIREFLY-1113)

- Replaced class component by functional one:
  - using setState hook for handling local state 
  - and setting local state if it has changed globally [during rendering](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops) to implement getDerivedStateFromProps 
- Used spread syntax in objects and assignments to simplify code, like setting default arguments

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1113-func-colpan/firefly/

- Click on the collapsible panels throughout the app
- Click on "Expand/Collapse All" on Image Search page (to check if all panels respond to global state changes)

### DIFF
--- a/src/firefly/js/ui/panel/CollapsiblePanel.jsx
+++ b/src/firefly/js/ui/panel/CollapsiblePanel.jsx
@@ -3,107 +3,47 @@
  */
 
 import './CollapsiblePanel.css';
-import React, {memo, PureComponent} from 'react';
+import React, {memo} from 'react';
 import PropTypes from 'prop-types';
-import {isBoolean, isFunction} from 'lodash';
+import {isFunction} from 'lodash';
 import {dispatchComponentStateChange, getComponentState} from '../../core/ComponentCntlr.js';
-import {clone} from '../../util/WebUtil.js';
 import {useFieldGroupConnector} from '../FieldGroupConnector.jsx';
+import {useStoreConnector} from 'firefly/ui/SimpleComponent';
 
 
-export const CollapseBorder = {
-                            Noborder: 0,
-                            Oneborder: 1,    // left side
-                            Twoborder: 2,    // left and right sides
-                            Threeborder: 3   // left, bottom and right sides
-                        };
+export function CollapsiblePanel(props){
+    const isOpenState = useStoreConnector(()=>(getComponentState(props.componentKey)?.isOpen ?? props.isOpen));
+    const newProps = {...props,
+            isOpen : isOpenState,
+            onToggle: (newIsOpen)=>{
+                dispatchComponentStateChange(props.componentKey, {isOpen: newIsOpen});
+                props.onToggle?.(newIsOpen);
+        }};
 
-export const CollapseHeaderCorner = {
-                            TopLeft: 0b0001,
-                            TopRight: 0b0010,
-                            BottomRight: 0b0100,
-                            BottomLeft: 0b1000
-                        };
-
-function collapsibleStateFromProps(props) {
-    // component key needs to be defined if the state needs to be saved though unmount/mount
-    let isOpen = props.componentKey && getComponentState(props.componentKey).isOpen;
-    if (!isBoolean(isOpen)) {
-        // use property to initialize state
-        isOpen = props.isOpen? props.isOpen: false;
-    }
-
-    return {isOpen};
+    return <CollapsiblePanelView {...newProps} />;
 }
 
-export class CollapsiblePanel extends PureComponent {
 
-    constructor(props) {
-        super(props);
+export function CollapsiblePanelView({header, isOpen=false, headerStyle, contentStyle, style,
+                                         onToggle, children }) {
+    let headerClassName = 'CollapsiblePanel__Header';
+    if (isOpen) headerClassName += ' CollapsiblePanel__Header--is-open';
+    const headerContent = isFunction(header) ? header() : header;
 
-        this.state = collapsibleStateFromProps(props);
+    const contentClassName = 'CollapsiblePanel__Content';
+    const contentHeight = isOpen ? 'initial' : 0;
+    contentStyle = {display: 'block', ...contentStyle, height: contentHeight};
 
-        this.handleClick = this.handleClick.bind(this);
-        this.getContentHeight = this.getContentHeight.bind(this);
-    }
-
-    static getDerivedStateFromProps(props) {
-        return collapsibleStateFromProps(props);
-    }
-
-    handleClick() {
-        const {componentKey, onToggle} = this.props;
-        const isOpen = !this.state.isOpen;
-
-        if (componentKey) {
-            dispatchComponentStateChange(componentKey, {isOpen});
-        }
-        this.setState({isOpen});
-
-        if (onToggle) {
-            onToggle(isOpen);
-        }
-    }
-
-    getContentHeight() {
-        return this.state.isOpen ? {height: 'initial'} : {height: 0};
-    }
-
-    render () {
-        const {header, headerRoundCorner, borderStyle, wrapperStyle, children} = this.props;
-        var {headerStyle, contentStyle} = this.props;
-        const contentBorderClassName = ['', ' CollapsiblePanel__Content-oneborder',
-                              ' CollapsiblePanel__Content-twoborder', ' CollapsiblePanel__Content-threeborder'];
-        let headerClassName = 'CollapsiblePanel__Header';
-        let contentClassName = 'CollapsiblePanel__Content';
-        let headerCorner = '';
-
-       ['TopLeft', 'TopRight', 'BottomRight', 'BottomLeft'].forEach((corner) => {
-            CollapseHeaderCorner[corner]&headerRoundCorner ?
-                                            headerCorner += ' 0.5em' : headerCorner += ' 0em';
-       });
-
-        headerStyle = Object.assign({}, headerStyle, {'borderRadius': headerCorner});
-
-        if (this.state.isOpen) {
-            headerClassName += ' CollapsiblePanel__Header--is-open';
-            contentClassName += contentBorderClassName[borderStyle];
-        }
-
-        contentStyle = Object.assign({display: 'block'}, contentStyle, this.getContentHeight());
-        const headerContent = isFunction(header) ? header() : header;
-
-        return (
-            <div style={{display: 'flex', flexDirection: 'column', ...wrapperStyle}}>
-                <div style={headerStyle} onClick={this.handleClick} className={headerClassName}>
-                    {headerContent}
-                </div>
-                <div style={contentStyle} className={contentClassName}>
-                    {children}
-                </div>
+    return (
+        <div style={{display: 'flex', flexDirection: 'column', ...style}}>
+            <div style={headerStyle} onClick={()=>{onToggle(!isOpen);}} className={headerClassName}>
+                {headerContent}
             </div>
-        );
-    }
+            <div style={contentStyle} className={contentClassName}>
+                {children}
+            </div>
+        </div>
+    );
 }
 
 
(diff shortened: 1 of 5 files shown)


src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,87 @@ export const CollapseHeaderCorner = {
...
-        return (
-            <div style={{display: 'flex', flexDirection: 'column', ...wrapperStyle}}>
-                <div style={headerStyle} onClick={this.handleClick} className={headerClassName}>
-                    {headerContent}
-                </div>
-                <div style={contentStyle} className={contentClassName}>
-                    {children}
-                </div>
+    return (
+        <div style={{display: 'flex', flexDirection: 'column', ...wrapperStyle}}>
```
robyww, line 99: We made a mistake a few years make by using the term `wrapperStyle`.  Change prop `wrapperStyle` to `style`. And change any code that users it.

To do this you will need use intelliJ to do a find usages on both `CollapsiblePanel` and `FieldGroupCollapsible`. I think there are only two or three files that actually use `wrapperStyle`

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,87 @@ export const CollapseHeaderCorner = {
...
-
-        headerStyle = Object.assign({}, headerStyle, {'borderRadius': headerCorner});
-
-        if (this.state.isOpen) {
-            headerClassName += ' CollapsiblePanel__Header--is-open';
-            contentClassName += contentBorderClassName[borderStyle];
-        }
+    function getContentHeight() {
+        return isOpenState ? {height: 'initial'} : {height: 0};
+    }
```
robyww, line 92: This really does not need to be a function, just simply:
```js
const contentHeight= isOpenState ? {height: 'initial'} : {height: 0}
```

robyww (changes requested):
Good job so far. I looks like you understood what was going on. My comments are mostly around clean up. Some of this was pre-existing. However, since you are making changes it is good clean up anything that needs it.

Just to help you understand the flow:
There  are two ways that this panel is use. Stand alone, in which case is needs a defined `componentKey` and as part of a field group using `FieldGroupCollapsible`. In that case `componentKey` will be undefined and the state will come from the field group.

If you have questions I can go over the change with you on zoom.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,87 @@ export const CollapseHeaderCorner = {
...
+    header,
+    isOpen=false,
+    headerRoundCorner=0,
+    headerStyle,
+    contentStyle,
+    wrapperStyle,
+    borderStyle=CollapseBorder.Oneborder,
+    onToggle,
+    children
+}) {
```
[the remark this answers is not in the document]
jaladh-singhal: Is there any specific style guide that we follow that I can refer to?

I've been using [Prettier](https://prettier.io/)

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,87 @@ export const CollapseHeaderCorner = {
...
+    function getIsOpenFromGlobalState() {
+        // component key needs to be defined if the state needs to be saved though unmount/mount
+        let isOpen_ = componentKey && getComponentState(componentKey).isOpen;
+        if (!isBoolean(isOpen_)) {
+            // use property to initialize state
+            isOpen_ = isOpen? isOpen: false;
+        }
+        return isOpen_;
```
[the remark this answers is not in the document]
jaladh-singhal: Thanks. `?.` and `??` makes it look so neat!

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,87 @@ export const CollapseHeaderCorner = {
...
     }
+    const isOpenGlobalState = getIsOpenFromGlobalState();
 
-    static getDerivedStateFromProps(props) {
-        return collapsibleStateFromProps(props);
-    }
+    const [isOpenState, setIsOpenState] = useState(isOpenGlobalState);
```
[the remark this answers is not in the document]
jaladh-singhal: Yes I thought of doing that but we also need this in L55 (`if (isOpenGlobalState !== isOpenState) setIsOpenState(isOpenGlobalState);`), that's why I stored it in a variable so that it can be reused.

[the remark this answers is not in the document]
robyww: ok let me look that line.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,64 @@ export const CollapseHeaderCorner = {
...
+    }
+    const isOpenGlobalState = getIsOpenFromComponentState();
 
-    constructor(props) {
-        super(props);
+    const [isOpenState, setIsOpenState] = useState(isOpenGlobalState);
 
-        this.state = collapsibleStateFromProps(props);
+    // if global state has updated, update local state
+    if (isOpenGlobalState !== isOpenState) setIsOpenState(isOpenGlobalState);
```
robyww, line 40: This line can't be here. You should not set state in the render. I should only be in a callback or an effect.

```js
useEffect( () => {
   if (isOpenGlobalState !== isOpenState) setIsOpenState(isOpenGlobalState);
},[isOpenState,isOpenGlobalState]
```

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +24,64 @@ export const CollapseHeaderCorner = {
...
-        }
-        this.setState({isOpen});
+    const contentBorderClassName = ['', ' CollapsiblePanel__Content-oneborder',
+        ' CollapsiblePanel__Content-twoborder', ' CollapsiblePanel__Content-threeborder'];
+    let headerClassName = 'CollapsiblePanel__Header';
+    let contentClassName = 'CollapsiblePanel__Content';
+    let headerCorner = '';
+    ['TopLeft', 'TopRight', 'BottomRight', 'BottomLeft'].forEach((corner) => {
+        headerCorner += (CollapseHeaderCorner[corner]&headerRoundCorner ? ' 0.5em' : ' 0em');
+    });
```
robyww, line 60: I had a comment on this `forEach` last time but I must have forgot to push the add comment button.

This is a classic `reduce` and using `reduce` makes the code more clean. Also an array defined and the operated on also is confusing. Put the array outside of the component at the top around line 26.
`const corners= ['TopLeft', 'TopRight', 'BottomRight', 'BottomLeft']`
Then...
```js
 const headerCorner= corners.reduce((str,corner) => {
            str += (CollapseHeaderCorner[corner]&headerRoundCorner ? ' 0.5em' : ' 0em');
            return str;
        },'');
```

robyww (changes requested):
A couple of more things that I listed below.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
jaladh-singhal: Well actually I did that earlier but then removed it (see [this diff](https://github.com/Caltech-IPAC/firefly/pull/1285/commits/5426ab1fe20c825e4198e4223404c6ba62b1d912)) because according to [react docs](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops), `getDerivedStateFromProps` that was doing this job in class component, is in fact setting the state during rendering. I know getDerivedStateFromProps is discouraged but wasn't sure if setting state after rendering is more appropriate.

Also this made me thinking, shouldn't we pass state from parent component to child component as props rather than passing it to global redux state and then checking in child component (with a hook) if a change has occurred in that piece of global state to update local state of child? I'd appreciate if you can point me to our code or redux docs section that explain the best way to implement such data flow.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
jaladh-singhal: I agree this is more clean.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
robyww: I am glad you pointed that out because that is what has been bothering me the whole time about this component. 

It made to work on two modes, one way is as part of a FieldGroup (`FieldGroupCollapsible`) which effectively gets it state from redux via out field group (see `FieldGroupCntrl.jsx`).

But that is not the problem, it is the other mode which is getting the state from the store _if_ componentKey is defined.

There are two modes but to make this work correctly there should be a view component that only works with props.

So do this.

Make a `CollapsiblePanelView` component takes `isOpen` but not `componentKey`. change `FieldGroupCollapsible` to use it.  Then make a new component small `CollapsiblePanel` component that uses `useStoreConnector` which is a hooks that update when the store updates.  

It would look something like this.
```js
export function CollapsiblePanel({componentKey, isOpen, ...rest }) {

    const isOpenState= useStoreConnector(() => {
        return getComponentState(componentKey)?.isOpen ?? isOpen;
    });

    return (
        <CollapsiblePanelView {...{ ...rest, isOpen:isOpenState,
            onToggle: (newIsOpen) => dispatchComponentStateChange(componentKey, {isOpen: newIsOpen}) }}
        />);

}
```

This way we have  pure view component and two redux connect component that use it.

jaladh-singhal: @robyww Thank you for this explanation and suggestion. It took me a while to understand this but I understood it eventually. And data flow in the component is definitely more readable now after splitting out view as pure component.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -25,85 +25,53 @@ export const CollapseHeaderCorner = {
...
 
-        this.state = collapsibleStateFromProps(props);
+export function CollapsiblePanelView({header, isOpen=false, headerRoundCorner=0, headerStyle, contentStyle,
+                                     style, borderStyle=CollapseBorder.Oneborder, onToggle, children }) {
+    const contentBorderClassName = ['', ' CollapsiblePanel__Content-oneborder',
+        ' CollapsiblePanel__Content-twoborder', ' CollapsiblePanel__Content-threeborder'];
+    let headerClassName = 'CollapsiblePanel__Header';
+    let contentClassName = 'CollapsiblePanel__Content';
+    const headerCorner = Object.values(CollapseHeaderCorner).reduce(
+        (headerCornerStr, corner) => headerCornerStr + (corner&headerRoundCorner ? ' 0.5em' : ' 0em'), '');
```
robyww, line 50: Nice- better than my suggestion!

src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -113,28 +81,32 @@ export const FieldGroupCollapsible= memo( (props) => {
...
     })
 };
 
-CollapsiblePanel.defaultProps= {
-    headerRounderCorner: 0,
-    borderStyle: CollapseBorder.Oneborder,
-    isOpen: false
+CollapsiblePanel.propTypes = {
+    componentKey: PropTypes.string,
+    ...CollapsiblePanelView.propTypes
```
robyww, line 103: Good. `componentKey` probably should be `isRequired`

Which don't you add `propTypes`  for `FieldGroupCollapsible` as well but in the case it would be
in that case instead of it should be `fieldGroupKey` as `isRequired`

robyww (approved):
The code looks good. It all works in my testing.

Now do the following.
 1. Make the same changes I mentioned, I don't need to look at it again
 1. squash and rebase. We prefer this done using the command line. The easiest way...
     - checkout `dev` and pull
     - checkout `FIREFLY-1113-func-colPan`
     - use `git rebase -i`
     - force push
  1. use the github `Merge pull request`
  2. Wait to after the stand up to merge so we can address any questions.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
loitly: I scanned through this PR and was puzzled at this block of code.  It has nothing to do with Jaladh changes.  I am referring to 
```
export const CollapseBorder = {
                            Noborder: 0,
                            Oneborder: 1,    // left side
                            Twoborder: 2,    // left and right sides
                            Threeborder: 3   // left, bottom and right sides
                        };
export const CollapseHeaderCorner = {
                            TopLeft: 0b0001,
                            TopRight: 0b0010,
                            BottomRight: 0b0100,
                            BottomLeft: 0b1000
                        };
```
It's unnecessarily complex.  And imo, not how you style a component.  `borderRadius` is now fixed(cannot override) to either 0 or 0.5em.  The good part is, it's not used anywhere.  I suggest we remove this complex logic and add the `Oneborder` caes into `CollapsiblePanel__Header` as default.  Of course, one can override it by setting `headerStyle` already.

loitly (commented):
Good job on the changes.  See my one comment which is not related to your changes.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
jaladh-singhal: I don't think we can make `componentKey` because if we do so then for using the collapsible panel without component state, one has to use `CollapsiblePanelView` that doesn't have any state so user has to make sure of changing the `isOpen` prop being passed to the component each time `onToggle` is called. This made me realize that we should make `CollapsiblePanelView` as internal component by removing the `export`. What do you think?

I have added `propTypes` for `FieldGroupCollapsible` too.

robyww (approved):
Looks good.

loitly (commented):
Looks much better.  Thank you.

src/firefly/js/ui/panel/CollapsiblePanel.jsx
robyww: It should be required. We don't have a use case for a simple state. We had that for awhile but it is lost every time you unrender it. There for the two was to use it is with `componentKey` or as part of a field group. Also- it you let it be optional you will put and `undefined` key in the store.

jaladh-singhal: Ohk

PR #1286, opened 2022-11-16
firefly-1115: cleanup tap panel

robyww:
#### [Firefly-1115](https://jira.ipac.caltech.edu/browse/FIREFLY-1115): cleanup tap panel
  - break up some large file
  - simplify flow
  - use new FieldGroup constructsw
  - make components work consistently

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1115-cleanup-tap/firefly
 - everything should work like before

#### Review - @loitly 
 - File to look at
   - FieldGroup hooks `useFieldGroupValue`, there are a couple of other but this is the one that is most important.
   - users of field group: `TemporalSearch.jsx`, `SpatialSearch.jsx` - there is only one `export` in each, scoll down to it and you will see what to look at
   - A couple of notes about `useFieldGroupValue`. it uses a `[getterFunc,setterFunc]` instead of `[value,setterFunc]` like `getState`. 
      - Primary because it is easy to get into the problem that you occasionally encounter with get state that another places has set the value but you can't get at it. 
      - Also using a getter allows more flexibility about what you can get back. Most the time you want the getter to turn just the value but there are times you need the whole field, such as when working with the time panel.


ejoliet (approved):
I've tested and didn't find any problem.
Good for me.
Thanks!

src/firefly/js/ui/FieldGroup.jsx
```diff
@@ -19,6 +36,7 @@ export const FieldGroup = memo( ({keepMounted, reducerFunc=undefined, groupKey,
         unregister:wrapperUnregister,
         registeredComponents:wrapperRegisteredComponents}= useContext(FieldGroupCtx);
 
+
```
loitly, line 39: Forcing a FieldGroup re-render when `registeredComponents` changes seem problematic.  Imagine having a complex page/form with many components with only one small component uses this feature, it will trigger the whole page/form to re-render.  Something to be aware of if there is regression issues in the future.

loitly (commented):
Thank you for adding `useFieldGroupWatch`.  It did not have the effect I was hoping for.  I hope it was not a waste of your time.

src/firefly/js/ui/FieldGroup.jsx
robyww: I will look

robyww: OK- looked. This was already in place before this PR, and it is being used by another PR so I can't just change it without trying it in the other branch. I can look at it in a week or two after everything gets merge. 

That said, I don't think it is a big deal. Just because `FieldGroup` re-renders it does not mean that any of the children re-render. In fact, they almost never do because it does not affect their props or state.

gpdf (changes requested):
Initial review from earlier today.

See also the thread at https://caltech-ipac.slack.com/archives/CG82RK4EM/p1669751695539389

Points 1-4 below apply to ObsTAP searches, Points 5&6 to regular table searches.

1. The “Spectral Coverage” section in the ObsTAP search screen is initially coming up selected, which it previously was not, and which it should not be unless the user explicitly wants that.
2. Starting from that point, “Populate and edit ADQL” brings up a “Constraints Error” dialog, but nothing in the UI is highlighted in red or otherwise flagged.  Unchecking “Spectral Coverage” fixes that, but there’s no way for a user to know that from the UI.
3. Even with a value filled in for the wavelength (with the “wavelength coverage” menu at “contains”) the “Constraints Error” still appears.
4. Changing the “wavelength coverage” menu to “overlaps” no longer has any effect - it is supposed to change the UI to show a wavelength range.  See screenshot from the current production version: image (1)
5. Selecting the `allwise_p3as_mep` table on IRSA (the Multi-Epoch Photometry catalog) and trying to use the “Temporal” search helper (not the ObsTAP “Timing” one, but the ordinary-table one), entering the (valid) column name “mjd” as the “Temporal Column” produces an error: image (2)  (Note the “.” after “mjd” in the error message - it shouldn’t be there anyway, as it’s not a sentence, but I wonder if somehow by accident it’s actually trying to use “mjd.” as the column name.)
6. It doesn't seem to be possible to uncheck the "Temporal" box once it's in this state.

robyww:
New update with multiple fixes:
 - Fixed issues with `Spectral Coverage` panel
 - Fixed issues column recognition in `Temporal ` panel
 - Fixed `Observation Type and Source` not giving warning with nothing selected
 - Improved message and down load for unrecognized data in data product viewer. In the future it should learn to recognized JSON

This alternate entry point can be used to test the ADQL and includes a `filterDefinitions` test.
 -  https://fireflydev.ipac.caltech.edu/firefly-1115-cleanup-tap/firefly/firefly-dev.html

PR #1287, opened 2022-11-18
FIREFLY-1091: changed layers popup text to be consistent

kpuriIpac:
#### [Firefly-1091](https://jira.ipac.caltech.edu/browse/FIREFLY-1091)
Design ticket:   [Firefly-758](https://jira.ipac.caltech.edu/browse/FIREFLY-758)

- changed HiPS and FITS text in layers popup to "Image (HiPS)" and "Image (FITS)" respectively
- in the code, I've hardcoded `showSize` to `false` and commented out the code next to it that determines if `showSize's` value. This is in case we ever decide to go back and show "Search Size" next to "Image (FITS)" in the layers popup as we previously did. I can get rid of this code if this is not needed. 

**Testing**: https://fireflydev.ipac.caltech.edu/firefly-1091-layers-popup-text/firefly
- Search for something that'll have both FITS / HiPS. From "TAP/Table Searches", under "IRSA Catalogs", I selected "Spitzer" and searched for "m1". 
- Then with "FITS" selected, open the layers popup, and you should see "Image (FITS)" next to the title on the first line (in this case "WISE Atlas W1")
- Switch to either "HiPS" or "HiPS/Aitoff" and click on the layers popup again. Now you should see "Image (HiPS)" on the first line next to the title ("DSS Colored" in this case) 

**Updates 11/21**: 
- Search size is back for FITS images in the layers popup. Test build updated.

### DIFF
--- a/src/firefly/js/drawingLayers/ImageRoot.js
+++ b/src/firefly/js/drawingLayers/ImageRoot.js
@@ -80,7 +80,7 @@ function getTitle(pv, plot, drawLayer) {
     const titleEmLen= Math.min(plot.title.length+2,24);
     const minWidth= showSize || !drawLayer.isPointData ? (titleEmLen+6)+'em' : titleEmLen+'em';
     const {blank=false}= plot;
-    const hipsStr= blank ? '': 'HiPS';
+    const hipsStr= blank ? '': 'Image (HiPS)';
     return () => {
        return  (
             <div style={{
@@ -93,7 +93,7 @@ function getTitle(pv, plot, drawLayer) {
             } }
                  title={plot.title}>
                 <div>{plot.title}</div>
-                <div style={{paddingLeft: 10, fontSize:'80%'}}>{`${isHiPS(plot) ? hipsStr : 'Image'}${showSize?',':''}`}</div>
+                <div style={{paddingLeft: 10, fontSize:'80%'}}>{`${isHiPS(plot) ? hipsStr : 'Image (FITS)'}${showSize?',':''}`}</div>
                 {showSize &&
                 <div  style={{paddingLeft: 5, fontSize:'80%'}}>
                     {`Search Size: ${sprintf('%.4f',r.getSizeInDeg())}${String.fromCharCode(176)}`}


lrebull:
looks great to me!

lrebull:
I tried to get, e.g., a png image in there to see if it would work in Vandana's proposed miscellaneous case, but it looks like pngs are loaded just in the data product viewer and as such don't have the image toolbar with the layers icon. :)

kpuriIpac:
> I tried to get, e.g., a png image in there to see if it would work in Vandana's proposed miscellaneous case, but it looks like pngs are loaded just in the data product viewer and as such don't have the image toolbar with the layers icon. :)

Yes I forgot to add a comment for that. The layers popup applies only to FITS/HiPS, right?

lrebull:
Yes, evidently. I wasn't sure myself! Maybe it won't be this way in the future?

kpuriIpac:
Got it. We can handle that case in the future then if things change.

robyww:
@lrebull i just want to confirm. You not longer want to see the size of the image, for `FITS` images?

lrebull:
Ah, you're right that I didn't comprehend that the size was missing. The size probably should be back in there. I was focused on the FITS and HiPS parallel label structure.

robyww (approved):
Looks good.

lrebull:
Screen Shot 2022-11-21 at 4 21 27 PM

looks ok, but on my screen (Mac OS 11.7, Firefox 107.0), the last digit on the size is truncated. See attached screenshot.

kpuriIpac:
Do you see that even on ops @lrebull? (https://irsa.ipac.caltech.edu/) 

> Screen Shot 2022-11-21 at 4 21 27 PM
> 
> looks ok, but on my screen (Mac OS 11.7, Firefox 107.0), the last digit on the size is truncated. See attached screenshot.

lrebull:
no; on ops i get the whole number, and its units (degrees). on your test build, i get what you see in the screenshot (truncated number and no units).

kpuriIpac:
Okay thanks for letting me know. I’ll look into this tomorrow.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #1292, opened 2022-12-02
Firefly-1120: Convert NaifidPanel to functional component and cleanup

jaladh-singhal:
Fixes [FIREFLY-1120](https://jira.ipac.caltech.edu/browse/FIREFLY-1120)

Besides converting class to functional component, it does cleanup of:
- multiple one-time use functions
- global variables (searchHistory, naifNameValue) using useRef (see [this](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables))

## Testing
https://firefly-1120-naif-panel-func.irsakudev.ipac.caltech.edu/applications/sofia

In Sofia Search, select "Solar system target", enter object name or id. 

Similarly in Wise, SHA, etc interfaces.

### DIFF
--- a/src/firefly/js/ui/NaifidPanel.jsx
+++ b/src/firefly/js/ui/NaifidPanel.jsx
@@ -2,99 +2,89 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import React, {PureComponent, memo} from 'react';
+import React, {memo, useRef} from 'react';
 import PropTypes from 'prop-types';
-import {get, sortBy} from 'lodash';
+import {sortBy} from 'lodash';
 import {resolveNaifidObj} from  './NaifidPanelWorker.js';
 import {SuggestBoxInputFieldView} from './SuggestBoxInputField';
 import {useFieldGroupConnector} from './FieldGroupConnector.jsx';
 import {TargetFeedback} from './TargetFeedback';
 
 
-const LABEL_DEFAULT='Moving Target Name:';
-const makeValidRet= (valid,message='') => ({valid,message});
-//const defValidator= (val) => val ? makeValidRet(true): makeValidRet(false,'Naif name not found');
-const defValidator= () => makeValidRet(true);
-const searchHistory =[];
-let naifNamevalue = '';
+const LABEL_DEFAULT = 'Moving Target Name:';
+const searchHistory = []; // defining as global to persist it throughout the lifetime of app
 
 
-class NaifidPanelView extends PureComponent {
-    constructor(props) {
-        super(props);
-        this.state = {suggestions: ''};
-        this.getSuggestions = this.getSuggestions.bind(this);
-        naifNamevalue = props.value;
-    }
+function NaifidPanelView({showHelp, valid, message, examples, feedback, value, labelWidth, feedbackStyle, popStyle,
+                             label= LABEL_DEFAULT, fireValueChange, updateNaifNameValue}){
+    const getSuggestions = (val= '') => {
+        if (!val) return [];
 
-    componentWillUnmount() {
-        if (this.removeListener) this.removeListener();
-        this.iAmMounted = false;
-    }
-
-    componentDidMount() {
-        this.iAmMounted = true;
-    }
-
-    getSuggestions(val= '') {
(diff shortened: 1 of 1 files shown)


robyww:
I have not looked at the code closely yet. I just wanted to make a quick comment. 

You actually want the `searchHistory` to be a global variable, it is a cache that should build over the life of the application. The problem with not having it global is that you lose all the history when the component gets unrendered.

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -2,100 +2,92 @@
...
-    getSuggestions(val= '') {
+
+
+function NaifidPanelView({showHelp, valid, message, examples, feedback, value, labelWidth, feedbackStyle, popStyle,
+                             label= LABEL_DEFAULT, fireValueChange, updateNaifNameValue}){
+    const searchHistory = useRef([]); // ref instead of state so that we can mutate it without causing re-renders
+
+
+    const getSuggestions = (val= '') => {
         if (val.length<1) return [];
```
robyww, line 23: This could be...
```js
if (!val) return [];
```

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -108,80 +100,44 @@ NaifidPanelView.propTypes = {
...
-        }
-    }
-    fireValueChange(newPayload) ;
-}
 
 export const NaifidPanel= memo( (props) => {
+    const {fieldKey='NaifId'} = props;
+    const {viewProps, fireValueChange} = useFieldGroupConnector({fieldKey, ...props});
+
+    const naifNameValue = useRef(props.value ?? ''); // ref instead of state so that we can mutate it without causing re-renders
```
robyww, line 112: I had a little trouble understanding this. So I think `naifNameValue` was a global before and you are now keeping track of it as a ref. OK - this makes sense. `naifNameValue` should not have been a global.

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -108,80 +100,44 @@ NaifidPanelView.propTypes = {
...
             ...viewProps,
             visible: true,
             label: viewProps.label || LABEL_DEFAULT,
             tooltip: 'Enter a target',
             value: viewProps.displayValue,
-            feedback: viewProps.feedback|| '',
-            showHelp: get(viewProps,'showHelp', true),
-            fireValueChange: (payload) => handleValueChange(payload,fireValueChange)
+            feedback: viewProps.feedback || '',
+            showHelp: viewProps?.showHelp ?? true,
```
robyww, line 138: glad you removed the `get`

src/firefly/js/ui/NaifidPanel.jsx
```diff
@@ -2,100 +2,92 @@
...
-    componentDidMount() {
-        this.iAmMounted = true;
-    }
-
-    getSuggestions(val= '') {
+
+
+function NaifidPanelView({showHelp, valid, message, examples, feedback, value, labelWidth, feedbackStyle, popStyle,
+                             label= LABEL_DEFAULT, fireValueChange, updateNaifNameValue}){
+    const searchHistory = useRef([]); // ref instead of state so that we can mutate it without causing re-renders
```
robyww, line 19: See my previous comment about `searchHistory` we _do_ want that to be a global since it is a cache that should never empty.

robyww (changes requested):
This all looks good and clean.  This `searchHistory` change is the only major thing.

src/firefly/js/ui/NaifidPanel.jsx
jaladh-singhal: Yes that's right! See [this diff](https://github.com/Caltech-IPAC/firefly/commit/c072d51a2eab2ac1bc759dd4de71a2adeb7f306f)

src/firefly/js/ui/NaifidPanel.jsx
jaladh-singhal: Yes I'm reverting this, it makes sense - we want search history throughout the app's lifetime, not just when this component is mounted.

robyww (approved):

PR #1294, opened 2022-12-06
Firefly-1124 hips tabs

kpuriIpac:
#### [Firefly-1124](https://jira.ipac.caltech.edu/browse/FIREFLY-1124): 
  - added upload panel as a FieldGroupTab (alongside "Search" / "Featured MOC") in the HiPS/MOC **Add MOC Layer** dialog 
  - keepState for this FieldGroup is currently set to true. Should I change it to false? (it was false when we were launching the upload dialog separately from the HiPS/MOC **Add MOC Layer** dialog)

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1124-hips-tabs/firefly
 - open a HiPS image (search for m1, choose image type "View HiPS Image") 
 - From the **HiPS/MOC** dropdown, select **Add MOC Layer**. Toggle between "Search" and "Use my MOC" tabs. 
 - In the "Use my MOC" tab, attempt to upload a non-MOC file (should give you warning), and if you attempt to load it, you should get a popup error/warning. Now attempt to upload a moc fits file, this should work as expected. Make you selections and "Add MOC" (the dialog should close now). 
 - Add MOC from the "Search" tab as well. 
 - try and change the size of the dialog (for both tabs, the content should stretch and fit within the dialog)
 -  From the **HiPS/MOC** dropdown, select **Change HiPS**, this should work as expected as well (the only thing I changed for this one was its size - to make it consistent with the size of the **Add MOC Layer** dialog's size). 

**Bug fixes**
- There was a bug from my previous work on the upload panel, where switching to 'Upload from URL' would not work (it would show the same prompt as that for 'Upload file'. This is now fixed in the latest build (as of 12/7). 

**Updates 12/6**: 
- used acceptMoc in FileUploadViewPanel to restrict showing "Load as MOC Overlay" & "Load as Table" options, and also "Attempt to interpret tables as spectra" (if you attempt to upload a .tbl file)
    - I should perhaps change **acceptMoc** to **onlyMoc** or **acceptOnlyMoc** as I think that's more indicative of what we're trying to do here? 
- fixed the padding between "Search" and the Featured MOC checkbox under it 
- made the changes in FieldGroup and  setSubmitText as requested 
- **Testing**: 
    - Same build link 
    - Use the regular upload panel to make sure it works as expected (upload non-moc fits file, tbl file and moc fits file - you should see radio button options as expected). 
    - Attempt to do the same from the **Add MOC Layer** dialog, and you should only be able to upload a moc fits file (and you should not see any options here).  

**Updates 12/9**: 
- the "Add MOC Layer" dialog size is bigger and should not change much now even with the additional of the URL input and/or 'Clear File'. The same height/width have been applied to the "Change HiPS" dialog too to keep things consistent. 
   - there is a small issue: if the user tries to resize this dialog and make it smaller, we lose the "Add MOC" text (as it gets smaller than minHeight). This is a pre-existing issue, but with the default dialog size even bigger now, it's probably more likely that a user may try and resize the dialog to make it smaller. 
- In the "Use my MOC" tab, before you upload a file, the supported format list has been downsized to include only MOC FITS files 
- some cleanup based on feedback
- - **Testing**: 
    - Same build link 
    -  Upload moc fits and non-fits files from the **Add MOC Layer** dialog, and you should only be able to upload a moc fits file. Also attempt to upload from the URL (test URL you can use: https://irsa.ipac.caltech.edu/data/hips/ESAC/PACS-color/Moc.fits)

### DIFF
--- a/src/firefly/js/ui/HiPSImageSelect.jsx
+++ b/src/firefly/js/ui/HiPSImageSelect.jsx
@@ -89,7 +89,7 @@ export function showHiPSSurveysPopup(pv, moc= false) {
             if (rootUrl) {
                 const plot = pv ? primePlot(pv) : primePlot(visRoot());
                 // update the table highlight of the other one which is not shown in table panel
-                moc ? createHiPSMocLayer(getIvoaId(), getTitle(), rootUrl, primePlot(pv), true).then() :
+                moc ? createHiPSMocLayer(getIvoaId(), getTitle(), rootUrl, primePlot(pv), true) :
                     dispatchChangeHiPS({plotId: plot.plotId, hipsUrlRoot: rootUrl});
                 dispatchHideDialog(DIALOG_ID);
             }
@@ -109,7 +109,7 @@ export function showHiPSSurveysPopup(pv, moc= false) {
                             help_id='visualization.changehips'>
                     {moc &&
                         <FieldGroupTabs initialState= {{ value:'search' }} fieldKey='mocTabs'
-                                            style={{minWidth:600, minHeight:500, width:'100%', height: '100%'}}>
+                                            style={{minWidth:715, minHeight:500, width:'100%', height: '100%'}}>
                             <Tab name='Search' id='search'>
                                 <div className='ImageSearch__HipsPopup' style={{padding: '5px 0px 0px 0px'}}>
                                     <SourceSelect moc={moc}/>
@@ -118,14 +118,14 @@ export function showHiPSSurveysPopup(pv, moc= false) {
                             </Tab>
 
                             <Tab name='Use my MOC' id='uploadMoc'>
-                                <div style={{width: '100%'}} >
+                                <div style={{width:'100%', minWidth:715, minHeight:500}} >
                                     <FileUploadViewPanel acceptMoc={true}/>
                                 </div>
                             </Tab>
                         </FieldGroupTabs>}
 
                     {!moc &&
-                        <div className='ImageSearch__HipsPopup' style={{minWidth:600, minHeight:500}}>
+                        <div className='ImageSearch__HipsPopup' style={{minWidth:715, minHeight:500}}>
                             <SourceSelect moc={moc}/>
                             <HiPSSurveyTable groupKey={groupKey} moc={moc}/>
                         </div>}
--- a/src/firefly/js/ui/HiPSImageSelect.jsx
+++ b/src/firefly/js/ui/HiPSImageSelect.jsx
@@ -29,9 +29,10 @@ import {primePlot} from '../visualize/PlotViewUtil.js';
 import {BLANK_HIPS_URL} from '../visualize/WebPlot.js';
 import {createHiPSMocLayer} from 'firefly/visualize/task/PlotHipsTask.js';
 import {getAppOptions} from 'firefly/core/AppDataCntlr.js';
-import CompleteButton from 'firefly/ui/CompleteButton.jsx';
 import {getDefaultMOCList} from 'firefly/visualize/HiPSMocUtil.js';
-import {showUploadDialog} from 'firefly/ui/FileUploadDropdown.jsx';
+import {FieldGroupTabs, Tab} from 'firefly/ui/panel/TabPanel';
+import {FileUploadViewPanel} from 'firefly/visualize/ui/FileUploadViewPanel';
+import {resultSuccess} from 'firefly/ui/FileUploadProcessor';
 
 const useSourceHiPS = 'useSourceHiPS';
 const useSourceMOC = 'useSourceMOC';
@@ -69,7 +70,6 @@ HiPSImageSelect.propTypes = {
 
 const DIALOG_ID = 'HiPSImageSelectPopup';
 
-
 /**
  * show HiPS survey info table in popup panel
  * @param {PlotView} pv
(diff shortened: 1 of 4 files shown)


kpuriIpac:
Also, @robyww, I think we don't need the showUploadDialog method in FileUploadDropdown.jsx, correct? it was previously being used to load upload panel as a new dialog HiPSImageSelect.jsx. But now it's a tab, so show showUploadDialog isn't being called.

src/firefly/js/ui/HiPSImageSelect.jsx
```diff
@@ -80,38 +80,57 @@ export function showHiPSSurveysPopup(pv, moc= false) {
...
+                           style={ { resize:'both', overflow: 'hidden', zIndex:1}}
+                            groupKey={groupKey}
+                            submitText={moc ? 'Add MOC' : 'Change HiPS'}
+                            onSubmit={onSubmit}
+                            onCancel={() => dispatchHideDialog(DIALOG_ID)}
+                            params={{disabledDropdownHide: true}}
+                            help_id='visualization.changehips'>
+                    {moc &&
+                        <FieldGroupTabs initialState= {{ value:'search' }} fieldKey='mocTabs'
+                                            style={{minWidth:600, minHeight:500, width:'100%', height: '100%'}}>
```
robyww, line 112: good: glad you use `minWidth` and `minHeight` instead of `width` and `height`

robyww (changes requested):
I will test a closer look tomorrow, but I have to leave now.  Here is the first cut at my feedback. 

Overall this is looking good. There are a few things I would like you to look at.

UI Testing:

- When you are forcing MOCs then the option: `Load as MOC Overlay, Load as Table` should not show
- it I try to load a table file that is not a moc i see the option: 
`Attempt to interpret tables as spectra` the should also be disable in moc only mode
- on the search tab there should be a little padding above Featured check box.


You also should get @lrebull to look at it.

lrebull:
Ah, I was seconds too early! In the build you *just* made, I no longer see the "attempt to interpret..."

lrebull:
(copying from email: )
I agree, the DES one works and the wise one doesn’t seem to do anything(?) ..

kpuriIpac:
> (copying from email: ) I agree, the DES one works and the wise one doesn’t seem to do anything(?) ..

Also, the wise one also does nothing if you try to load it from the upload panel. It just goes back to the image search page  if you try to "load" it (if you don't already have any other images/tables loaded). 

There is either something off with this file, or if it's an issue with us not being able to load it, then at least the issue is not specific to this implementation of the upload panel as a tab in the HiPS/MOC dialog.

ejoliet (approved):
Looks great!
Very picky here:
- supported format list is probably not true, we only support MOC.fits. under 
- selecting the option "from url" makes the panel to 'jump' to grow and to allow the list of supported format to fit into the fixed panel width. Maybe there is a way to use the empty white space on the left and avoid that happen? 

Thanks!

loitly:
I have similar comment to Emmanuel about styling.
- left margin is too much, making it looks off aligned.
- the 'Clear File' appearing/disappearing should be accounted for when creating the default popup size so that it does not resize the dialog causing 'flashes'.

This comment is maybe for @robyww or @ejoliet 
Why when you select 'Upload file', you do not need to click 'Upload', but you need to do that when you choose 'Upload from URL'?  This should be done 'onblur' or off-focus event so that the flow remain consistent between the 2 options.

src/firefly/js/ui/HiPSImageSelect.jsx
```diff
@@ -80,38 +80,57 @@ export function showHiPSSurveysPopup(pv, moc= false) {
...
+    const onSubmit = (request) => {
+        if (request.mocTabs === 'uploadMoc') {
+            if (resultSuccess(request)) dispatchHideDialog(DIALOG_ID);
+        }
+        else {
+            const rootUrl = getHipsUrl();
+            if (rootUrl) {
+                const plot = pv ? primePlot(pv) : primePlot(visRoot());
+                // update the table highlight of the other one which is not shown in table panel
+                moc ? createHiPSMocLayer(getIvoaId(), getTitle(), rootUrl, primePlot(pv), true).then() :
```
loitly, line 92: Interesting.  I know it's existing code, but why is `.then()` needed?

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -79,6 +78,7 @@ export function FileUploadViewPanel({setSubmitText, acceptMoc}) {
     const {message, analysisResult, report, summaryModel, detailsModel, prevAnalysisResult} =
         useStoreConnector(() => {
             const loadingOp= getLoadingOp();
+            console.log('loadingOp: ' + loadingOp);
```
loitly, line 81: Use Logger.js for debug logging, or remove.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -362,14 +362,14 @@ function getDetailsModel(tableModel, report, detailsTblId, UNKNOWN_FORMAT) {
...
-function TableDisplayOption({isMoc, summaryTblId, currentReport, currentSummaryModel}) {
+function TableDisplayOption({isMoc, summaryTblId, currentReport, currentSummaryModel, acceptMoc}) {
 
     const selectedTables = getFileFormat(currentReport) ?
         getSelectedRows('Table', summaryTblId, currentReport, currentSummaryModel) : [];
 
     if ( selectedTables.length < 1) return null;
 
-    if (isMoc) {
+    if (isMoc && !acceptMoc) {
```
loitly, line 372: I am confused by `acceptMoc` flag.  What does it means?  Is it allow MOC file, or only take MOC file?
Also, there's a concept of `SUPPORTED_TYPES` in this panel, e.g.. REGION, Image, Table, etc.
Shouldn't MOC be added to them, instead of treated differently?
If `acceptMoc` is a way to define what this upload panel is allowed to accept, then should there be a more generic parameter, like `allowTypes` which takes an array of types it will allow.

loitly (approved):
I only have questions and suggestions.  I defer to other reviewers for final approval.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
robyww: This is one of the problems that we have not yet gotten to with this component. It was planned with @kpuriIpac last PR but we put it off.

`SUPPORTED_TYPES` is barely used and it is wrong, since `SUPPORTED_TYPES` is really what the FileAnalyzer recognizes. For example, the FileAnalyzer does not recognize spectrums.

The upload panel additionally recognizes the following tables:  
- MOC
- spectrum tables 
- Datalink tables (and I am adding in another branch) 


So we do need a list of what is allowed.

- region
- any table
- spectrum tables
- moc tables
- service descriptor tables (in another PR)
- images
- VO UWS xml file (future)

It seems to me this needs to be a stand alone task. I have not yet made a ticket for it so I probably should.

robyww (commented):
More details for Loi

lrebull:
as far as i can tell, this is all the same behavior as before -- the DES one works, the WISE one doesn't .. but i now get "you are no longer connected to the server" (!!) when i try to load it. 
other than that, it looks good -- the interface changes look good.

kpuriIpac:
Thanks @lrebull. Does the suggestion text before you upload a file look okay? (the one that reads "Note: You can only load a MOC FITS file from this dialog"). @ejoliet suggested this change because we can only upload moc fits files from this dialog, as opposed to the regular upload panel which is a longer list of all file types supported for upload.

lrebull:
Yes, that looks good.

robyww (approved):
Looks good.

PR #1296, opened 2022-12-07
FIREFLY-1097: Combined chart color enhancement

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1097

Test: https://fireflydev.ipac.caltech.edu/firefly-1097-combined-chart-color/firefly

Expected chart behavior:
- New chart start with a set of predefined colors
- Each new trace(data series) is assigned the next color in that set

Pin Chart:
Pinning a chart is basically cloning the current chart and placing it into the Pinned Chart panel. 
Therefore, it should have the exact same color as the current chart.

Combine Chart:
- Start by selecting a chart in the Pinned Chart panel.  This is considered the 'base' chart.  We will keep the color of the traces as is.
- New trace(s) created by this combining process will get the next color from the set of predefined colors.  This is to ensure each trace has a good unique color.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -1140,5 +1140,5 @@ export function getChartProps(chartId, tbl_id, activeTrace) {
 export function getTblIdFromChart(chartId, traceNum) {
     const {data, fireflyData, activeTrace} = getChartData(chartId) || {};
     traceNum = traceNum ?? activeTrace;
-    return data?.[traceNum]?.tbl_id || fireflyData?.[traceNum].tbl_id;
+    return data?.[traceNum]?.tbl_id || fireflyData?.[traceNum]?.tbl_id;
 }
--- a/src/firefly/js/charts/ui/CombineChart.jsx
+++ b/src/firefly/js/charts/ui/CombineChart.jsx
@@ -7,7 +7,7 @@ import {cloneDeep, get, pick, set} from 'lodash';
 
 import {getMultiViewRoot, getViewerItemIds} from '../../visualize/MultiViewCntlr.js';
 import {dispatchChartAdd, getChartData} from '../ChartsCntlr.js';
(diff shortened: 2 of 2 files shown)


ejoliet (approved):
I've tested and all looks good to me! Thank you!

lrebull:
sorry for the delay; i lost track of this yesterday.

I'm not 100% sure what i am supposed to be seeing, or testing.

(1) When i make a bunch of traces and try to combine them, the bug I had before where I couldn't select the traces is gone, but now I can no longer see the "ok" buttons unless I make the pop-up long. See movie/screenshot. 

(2) Am i supposed to be able to control the set of predefined colors it draws from when merging plots? because i can't figure out how to do that, if so. 

(3) In the set of colors I get (see 2nd attached screenshot), the colors repeat but not in an obvious pattern. like, i would have expected it to go, if you had 5 colors, abcdeabcde. but trace 2 and 9 seem to be the same color, as are 7 and 10. I don't get it.

Screen Shot 2022-12-08 at 1 06 30 PM
https://user-images.githubusercontent.com/23038600/206567460-06eb4b95-8e0a-4f94-ab72-55cdbf7fedfd.mp4

Screen Shot 2022-12-08 at 1 08 32 PM

lrebull:
Sorry; I again lost track of this. (The December crunch is grinding me down!)

The basic functionality is all still there and seems fine. 

It wasn't clear to me before what was going on with the colors, how it was picking them. We may need suggestions from people with strong opinions about the ordering of the pre-defined colors. I can see now that 2&9 and 7&10 are not technically the same color but boy oh boy they sure look the same unless you look _really_ closely. It may be that we should adopt some color-blind-friendly sequence instead? I found this -- look near the bottom at the "wong" or "tol" color palettes. https://davidmathlogic.com/colorblind/#%23D81B60-%231E88E5-%23FFC107-%23004D40

PR #1297, opened 2022-12-10
add base path settings to Firefly docker

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1121

Add Firefly docker support for setting a base deployed path(baseURL).
`baseURL` is passed into the container as an environment variable.  It will be picked up on first start and cannot be changed for the duration of the container's lifetime.

To test, 
- checkout this branch
- uncomment `#baseURL=/my/version/` at firefly-docker.env:6
```bash
cd ./firefly
docker compose up firefly --build
```
Then, goto http://localhost:8080/my/version/firefly/ to see Firefly deployed at the defined baseURL.


robyww (approved):
Looks good.

PR #1299, opened 2022-12-16
FIREFLY-1024: Pinning charts from extraction dialog 

kpuriIpac:
#### [Firefly-1024](https://jira.ipac.caltech.edu/browse/FIREFLY-1024): 
  - the major changes are in ChartUtil.js. Added an ActionWatcher to watch for chart updates
  - the callback of this action watcher then ensures that the chart is fully loaded before calling pinChart to pin the chart 
  - temporarily commented out the call to DefaultChartsContainer in ChartsContainer.jsx and calling PinnedChartPanel directly, so that you can test pinning charts from both the extraction dialog and the pin chart button to ensure both work as expected 
  - some ESLint suggested cleanup in PlotlyChartArea.jsx 

**Updates 12/20**
- When allowPinnedCharts() is set to true, the extraction dialog button will read "Pin Chart/Table", otherwise it will just read "Pin Table" 
- Better code segmentation and separation of concern based on @loitly's feedback and code 

**Updates 12/21 - 1.0**
- determined isChartLoaded was not really needed, removed its usage
- clean up of imports (small change, please look at the last 2 commits for code changes) 
- test build updated (try pinning Zaxis, line and points from the extraction dialog; for Zaxis, use the file in the JIRA ticket) 

**Updates 12/21 - 2.0**
- added logic back for checking if chart is fully loaded (and modified it a bit to rather check if chart is *still* loading) 
- look at the last 2 commits, though the main changes are in the commit titled "readded logic for checking if chart is still loading" (the one after that was because I forgot to remove export from doPinChart function) 

**Updates 12/23**
- cleaned up and changed some code as per @robyww's comments
- moved doPinChart out of the ExtractionPanelView component, but instead of sending it the callKeepExtraction prop as an argument (resulting in some error messages), I called callKeepExtraction within the onSuccess of the CompleteButton, got the tbl_id from it, and sent the tbl_id as an argument instead 
- new error/warning message text and title updated (for max charts) 

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1024-ext-pin-plot/firefly
 - load a fits file (from image search, search for "m1") and use the extraction dialog (line & point) and try to pin charts 
 - upload the file from the ticket link above, and use this to test Z-axis extraction and pinning those charts as well 
 - try adding more than 12 charts, you should get a warning each time after that about not being able to pin any more charts but tables can still be added
 - the only thing I'm not sure if how you could test the extraction dialog to say "Pin Table" when allowPinnedCharts() is set to false (in the test build). You could go into my branch though to set it to false and test it locally.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -13,11 +13,19 @@ import {assign, cloneDeep, flatten, get, has, isArray, isEmpty, isObject, isStri
 import shallowequal from 'shallowequal';
 
 import {getAppOptions} from '../core/AppDataCntlr.js';
-import { COL_TYPE, getColumnType, getMetaEntry, getTblById, isColumnType, isFullyLoaded, isTableLoaded,
-    stripColumnNameQuotes, watchTableChanges } from '../tables/TableUtil.js';
+import {
+    COL_TYPE, getColumnType, getMetaEntry, getTblById, isColumnType, isFullyLoaded, isTableLoaded,
+    stripColumnNameQuotes, watchTableChanges
+} from '../tables/TableUtil.js';
 import {TABLE_HIGHLIGHT, TABLE_LOADED, TABLE_SELECT, TABLE_SORT} from '../tables/TablesCntlr.js';
 import {dispatchLoadTblStats, getColValStats} from './TableStatsCntlr.js';
-import {dispatchChartHighlighted, dispatchChartSelect, dispatchChartUpdate, dispatchSetActiveTrace, getChartData} from './ChartsCntlr.js';
+import {
+    dispatchChartHighlighted,
+    dispatchChartSelect,
+    dispatchChartUpdate,
+    dispatchSetActiveTrace,
+    getChartData
+} from './ChartsCntlr.js';
 import {Expression} from '../util/expr/Expression.js';
 import {quoteNonAlphanumeric} from '../util/expr/Variable.js';
 import {flattenObject} from '../util/WebUtil.js';
@@ -517,6 +525,17 @@ export function setupTableWatcher(chartId, ts, idx) {
         [TABLE_LOADED, TABLE_HIGHLIGHT, TABLE_SELECT],
         (action) => updateChartData(chartId, idx, ts, action),
         uniqueId(`ucd-${ts.tbl_id}-trace`)); // watcher id for debugging
+}
+
+
+/**
+ * Get feedback as boolean on whether chart is fully loaded or not
+ * @param chartId - chartId of the current chart
+ */
+export function isChartLoading(chartId) {
+    const {fireflyData=[]} = getChartData(chartId);
+    const isChartLoading = fireflyData.some((e)=>  e.isLoading);
+    return isChartLoading; //true when chart is still loading
 }
 
 function tablesourcesEqual(newTS, oldTS) {
--- a/src/firefly/js/charts/ui/ChartsContainer.jsx
+++ b/src/firefly/js/charts/ui/ChartsContainer.jsx
@@ -97,7 +97,6 @@ function doUpdateViewer(viewerId, tblGroup, chartId, useOnlyChartsInViewer) {
 export const ChartsContainer = (props)  =>{
     const {chartId, expandedMode} = props;
 
-    // eslint-disable-next-line react-hooks/rules-of-hooks
     return expandedMode || chartId || !allowPinnedCharts() ?
         <DefaultChartsContainer {...props}/>:
         <PinnedChartPanel {...props}/> ;
--- a/src/firefly/js/charts/ui/PinnedChartPanel.jsx
+++ b/src/firefly/js/charts/ui/PinnedChartPanel.jsx
@@ -15,7 +15,9 @@ import {
     dispatchAddViewer, dispatchViewerUnmounted, dispatchUpdateCustom,
     getMultiViewRoot, getViewer, getLayoutType, PLOT2D, getViewerItemIds, dispatchRemoveViewerItems, dispatchAddViewerItems, NewPlotMode
 } from '../../visualize/MultiViewCntlr.js';
-import {getExpandedChartProps, getChartData, CHART_ADD, CHART_REMOVE, getChartIdsInGroup, dispatchChartAdd, dispatchChartRemove} from '../ChartsCntlr.js';
+import {
+    getExpandedChartProps, getChartData, CHART_ADD, CHART_REMOVE, getChartIdsInGroup, dispatchChartAdd, dispatchChartRemove, CHART_UPDATE
+} from '../ChartsCntlr.js';
 import {LO_VIEW, LO_MODE, dispatchSetLayoutMode} from '../../core/LayoutCntlr.js';
 import {MultiChartToolbarStandard, MultiChartToolbarExpanded} from './MultiChartToolbar.jsx';
 import {RenderTreeIdCtx} from '../../ui/RenderTreeIdCtx.jsx';
@@ -23,16 +25,17 @@ import {useStoreConnector} from '../../ui/SimpleComponent';
 import {Logger} from '../../util/Logger.js';
 import {findGroupByTblId, getActiveTableId, getTblById, monitorChanges} from '../../tables/TableUtil';
 import {TABLE_LOADED, TBL_RESULTS_ACTIVE, TABLE_REMOVE, dispatchActiveTableChanged} from '../../tables/TablesCntlr';
-import {getTblIdFromChart, uniqueChartId} from '../ChartUtil';
+import {getTblIdFromChart, isChartLoading, uniqueChartId} from '../ChartUtil';
 import {getActiveViewerItemId} from './MultiChartViewer';
 import {DefaultChartsContainer} from './ChartsContainer';
 import {StatefulTabs, switchTab, Tab} from '../../ui/panel/TabPanel';
 import {HelpIcon} from '../../ui/HelpIcon';
 import {getComponentState, dispatchComponentStateChange} from '../../core/ComponentCntlr';
 import {SplitPanel} from '../../ui/panel/DockLayoutPanel';
-import {hideInfoPopup, showInfoPopup, showYesNoPopup, showPinMessage} from '../../ui/PopupUtil.jsx';
+import {hideInfoPopup, showInfoPopup, showPinMessage} from '../../ui/PopupUtil.jsx';
 import {TextButton} from '../../ui/TextButton.jsx';
 import {CombineChart} from './CombineChart.jsx';
+import {dispatchAddActionWatcher} from 'firefly/core/MasterSaga';
 
 export const PINNED_CHART_PREFIX = 'pinned-';
 export const PINNED_VIEWER_ID = 'PINNED_CHARTS_VIEWER';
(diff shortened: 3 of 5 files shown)


kpuriIpac:
@loitly there may be a better way for me to enabling pinning - even just for the PR than commenting out DefaultChartsContainer, let me know if you have any suggestions.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -13,14 +13,23 @@ import {assign, cloneDeep, flatten, get, has, isArray, isEmpty, isObject, isStri
...
-    stripColumnNameQuotes, watchTableChanges } from '../tables/TableUtil.js';
+import {
+    COL_TYPE, getColumnType, getMetaEntry, getTblById, isColumnType, isFullyLoaded, isTableLoaded, monitorChanges,
+    stripColumnNameQuotes, watchTableChanges
+} from '../tables/TableUtil.js';
 import {TABLE_HIGHLIGHT, TABLE_LOADED, TABLE_SELECT, TABLE_SORT} from '../tables/TablesCntlr.js';
 import {dispatchLoadTblStats, getColValStats} from './TableStatsCntlr.js';
-import {dispatchChartHighlighted, dispatchChartSelect, dispatchChartUpdate, dispatchSetActiveTrace, getChartData} from './ChartsCntlr.js';
+import ChartsCntlr, {
+    CHART_ADD, CHART_HIGHLIGHT, CHART_MOUNTED, CHART_UPDATE,
```
loitly, line 23: Is normal to have unused imports during development.  Please clean them up before creating PR.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -519,6 +530,46 @@ export function setupTableWatcher(chartId, ts, idx) {
...
+ * Get feedback as boolean on whether chart is fully loaded or not
+ * @param chartId - chartId of the current chart
+ */
+export function isChartLoaded(chartId) {
+    const {fireflyData=[]} = getChartData(chartId);
+    const isLoading = fireflyData.every((e)=>get(e, 'isLoading'));
+    return !isLoading; //true when chart is fully loaded (isLoading: false)
+}
+
+const onChartLoad = (action, accept, cancelSelf) => {
```
loitly, line 543: It turns out this is not needed.  This logic is not generic to be a util function.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -519,6 +530,46 @@ export function setupTableWatcher(chartId, ts, idx) {
...
+            cancelSelf && cancelSelf(); //ActionWatcher cleanup
+        }
+    }
+    else { //if we possibly get a different tbl_id/chartId - cancel this ActionWatcher
+        cancelSelf && cancelSelf();
+    }
+};
+
+
+export function setupChartWatcher(tbl_id) {
```
loitly, line 557: Same for this one.

src/firefly/js/charts/ChartUtil.js
```diff
@@ -519,6 +530,46 @@ export function setupTableWatcher(chartId, ts, idx) {
...
+    return watchChartChanges([CHART_UPDATE],
+        (action, cancelSelf) => onChartLoad(action, accept, cancelSelf));
+}
+
+/**
+ * this function invokes the given callback when changes are made to a Chart (in this case, CHART_UPDATE from setupChartWatcher)
+ * @param {Object}   actions  an array of chart actions to watch
+ * @param {function} callback  callback to execute when table is loaded.
+ */
+export function watchChartChanges(actions, callback) {
```
loitly, line 568: There's no added value for this one.  It's exactly the same as `dispatchAddActionWatcher` since it will also create a uniqueID if not given.

src/firefly/js/charts/ui/ChartsContainer.jsx
```diff
@@ -98,9 +98,11 @@ export const ChartsContainer = (props)  =>{
...
 
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    return expandedMode || chartId || !allowPinnedCharts() ?
-        <DefaultChartsContainer {...props}/>:
-        <PinnedChartPanel {...props}/> ;
+    return <PinnedChartPanel {...props}/>;
+
+    //return expandedMode || chartId || !allowPinnedCharts() ?
+       // <DefaultChartsContainer {...props}/>:
+        //<PinnedChartPanel {...props}/> ;
```
loitly, line 105: I don't know the reason why this is needed.  As is, it breaks expanded mode.  Maybe you can show me the case when this is not working.

loitly (changes requested):
Instead of fully explaining complex suggestions, I went ahead and create a branch with my pseudo code and comment as suggestions.  Reach out to me for any part you find unclear, or just to understand more why I think it's a better approach.
In summary, I used your logic, but move them around, combining or breaking them out to different pieces to achieve better encapsulation and separation of concern.

https://github.com/Caltech-IPAC/firefly/pull/1300

src/firefly/js/charts/ChartUtil.js
```diff
@@ -519,6 +527,16 @@ export function setupTableWatcher(chartId, ts, idx) {
...
 }
 
+/**
+ * Get feedback as boolean on whether chart is fully loaded or not
+ * @param chartId - chartId of the current chart
+ */
+export function isChartLoaded(chartId) {
+    const {fireflyData=[]} = getChartData(chartId);
+    const isLoading = fireflyData.every((e)=>get(e, 'isLoading'));
+    return !isLoading; //true when chart is fully loaded (isLoading: false)
```
loitly, line 537: Are you sure this logic is right?  Shouldn't it be ...
```js
    return ! fireflyData.some((e) => e?.isLoading);
```

src/firefly/js/charts/ui/PinnedChartPanel.jsx
```diff
@@ -139,7 +140,30 @@ export const PinChart = ({viewerId, tbl_group}) => {
...
+    // there are cases when chart is not fully loaded.  if so, wait before pinning the chart
+    if (isChartLoaded(chartId)) {
+        doPinChart({chartId,autoLayout});
+    } else {
+        dispatchAddActionWatcher({actions:[CHART_UPDATE], callback: (action, cancelSelf) => {
+                const aChartId = action.payload?.chartId;
+                if (chartId === aChartId) {
+                    if (isChartLoaded(chartId)) {
+                        doPinChart({chartId, autoLayout});
+                        cancelSelf && cancelSelf();
```
loitly, line 154: It's from my `copy/paste`.  Can you change `cancelSelf && cancelSelf()` to `cancel?.()`?

loitly (approved):
Tested and it works fine.  But, can you double-check `isChartLoaded` logic?

loitly (approved):
Looks good.  Ok to merge

src/firefly/js/charts/ui/PinnedChartPanel.jsx
```diff
@@ -139,7 +149,32 @@ export const PinChart = ({viewerId, tbl_group}) => {
...
+                    // fallback to ensure there's no orphan watchers
+                    cancelSelf?.();
+                }
+            }
+        });
+    }
+    else {
+        doPinChart({chartId,autoLayout});
+    }
+}
```
robyww, line 173: This function is a little hard to read it does down 4 levels. It would be easier if you just started it with:
```js
if (!isChartLoading(chartId)) {
      doPinChart({chartId,autoLayout});
      return;
}
```
and the not `else`

In the same way inside of call back function, start with (no `else`):
```js
if (chartId !== aChartId) {
  cancelSelf?.()
  return;
}
```
The whole thing will be less deep and easier to follow.

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -481,8 +515,10 @@ function ExtractionPanelView({pointSize, setPointSize, afterRedraw, plotlyDivSty
                 textAlign:'center', alignSelf: 'stretch', flexDirection:'row',  display:'flex',
                 justifyContent:'space-between', padding: '15px 15px 7px 8px' }}>
                 <div style={{display:'flex', justifyContent:'space-between'}}>
-                    {plotlyData && canCreateExtractionTable &&
-                    <CompleteButton style={{paddingLeft: 15}} text='Pin Table' onSuccess={()=> callKeepExtraction(false, true)} />}
+                    {plotlyData && canCreateExtractionTable && !allowPinnedCharts() &&
```
robyww, line 518: Glad you found and check for this options `allowPinnedCharts()`

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -587,6 +623,7 @@ function keepZAxisExtraction(pt,pv, plot, filename,refHDUNum,extractionSize, com
     save ? doDispatchTableSaving(dataTableReq, doOverlay) : doDispatchTable(dataTableReq, doOverlay);
     idCnt++;
     titleCnt++;
+    return {tbl_id};
```
robyww, line 626: I don't see why you are putting the tbl_id in an object. just
```js
return tbl_id;
```

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -616,6 +653,7 @@ function keepLineExtraction(pt, pt2,pv, plot, filename,refHDUNum,plane,extractio
...
```
robyww, line 656: ditto - see above

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -657,6 +695,7 @@ function keepPointsExtraction(ptAry,pv, plot, filename,refHDUNum,plane,extractio
     save ? doDispatchTableSaving(dataTableReq,doOverlay) : doDispatchTable(dataTableReq,doOverlay);
     idCnt++;
     titleCnt++;
+    return {tbl_id};
```
robyww, line 698: ditto - see above

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -448,6 +451,37 @@ function ExtractionPanelView({pointSize, setPointSize, afterRedraw, plotlyDivSty
...
                                  bottomUI, combineOp, setCombineOp, hasFloatingData}) {
+
+    const extractAddedChartInfo = (action) => {
+        const chartId = action.payload.chartId;
+        const tbl_id =   action.payload.data?.[0].tbl_id;
+        return {chartId, tbl_id};
+    };
+
+    const doPinChart = () => {
+        const {tbl_id} = callKeepExtraction(false, true);
```
robyww, line 462: see my comment about just returning tbl_id and not an object.

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -448,6 +451,37 @@ function ExtractionPanelView({pointSize, setPointSize, afterRedraw, plotlyDivSty
...
+                    // this flow assumes CHART_ADD will happen immediately after table is pinned
+                    // have a fallback to cancel this watcher in cases when this is not true.
+                    // since we have tbl_id, we will watch for any TABLE_UPDATE to catch cases where CHART_ADD was not called
+                    const aTblId = action?.payload?.tbl_id;
+                    if (aTblId !== tbl_id) cancelSelf?.();
+                }
+            }
+        });
+    };
+
```
robyww, line 484: I don't think this function should be a closure. I makes the component hard to read. 
just make a function to take `callKeepExtraction` as a parameter.
```js
function doPinChart(callKeepExtraction) {
```

src/firefly/js/visualize/ui/ExtractionDialog.jsx
```diff
@@ -448,6 +451,37 @@ function ExtractionPanelView({pointSize, setPointSize, afterRedraw, plotlyDivSty
                                  plotlyData, canCreateExtractionTable,
                                  plotlyLayout, startUpHelp, callKeepExtraction,
                                  bottomUI, combineOp, setCombineOp, hasFloatingData}) {
+
+    const extractAddedChartInfo = (action) => {
+        const chartId = action.payload.chartId;
+        const tbl_id =   action.payload.data?.[0].tbl_id;
+        return {chartId, tbl_id};
+    };
```
robyww, line 459: This function does not need to be a closure. Move it out of the component.
Also you can simply it.
```js
const extractAddedChartInfo = ({payload}) => ( {chartId: payload.chartId, tbl_id: payload.data?.[0].tbl_id});
```

robyww (changes requested):
This is looking really good. I don't see a good reason not to put this in the release. Make the clean  up changes I requested and see it you can change branch in the PR to `rc-2022.3`

UI Comment:

Change max pin message to:
```
Only pinning table: You have reached the maximum number of allowable pinned charts
```
title- `Max Pinned Charts`

lrebull:
Nice! It seems to work as desired. 
In looking briefly at the text above, is it supposed to tell me that it's pinning the table but not the plot when it hits the max plots pinned? Because it doesn't -- it just tells me it's hit the max number of pinned plots, nothing about the table.

kpuriIpac:
> Nice! It seems to work as desired. In looking briefly at the text above, is it supposed to tell me that it's pinning the table but not the plot when it hits the max plots pinned? Because it doesn't -- it just tells me it's hit the max number of pinned plots, nothing about the table.

@lrebull yes that's a change I'm working on right now. It'll then read "Only pinning table: You have reached the maximum number of allowable pinned charts".

robyww:
@lrebull Just let us know what the text should be, if you don't like my suggested change.

lrebull:
OK, that's fine. I just wasn't sure if I should be already seeing it or not. There should be a period at the end of the sentence too. :)

lrebull:
works great!

robyww (approved):
Looks good.

PR #1301, opened 2022-12-16
Firefly-1070: Remove SimpleComponent and replace its uses with functional components

jaladh-singhal:
Fixes [FIREFLY-1070](https://jira.ipac.caltech.edu/browse/FIREFLY-1070)

- Removed SimpleComponent class, the `useStoreConnector` hook can do what it was doing.
- Converted all 5* components that were using it to functional components that use this hook instead

*The 6th component as per ticket, SingleTraceUIToolbar wasn't used anywhere so removed it.


loitly (approved):
Looks good.  Spot checked chart functions and they worked fine.

src/firefly/js/charts/ui/ChartSelectPanel.jsx
```diff
@@ -106,60 +106,46 @@ function getGroupKey(chartId, chartAction) {
...
-                </FormPanel>
-            </div>
-        );
-    }
+export function ChartSelectPanel({tbl_id, chartId, chartAction, inputStyle={}, hideDialog, style={}}) {
+    const {renderTreeId} = useContext(RenderTreeIdCtx);
+
+    const chartActions = getChartActions({chartId, tbl_id});
+    const [chartActionState, setChartActionState] = useState(
+        (!chartAction || !chartActions.includes(chartAction)) ? chartActions[0] : chartAction);
```
robyww, line 114: I think 

```js
(!chartAction || !chartActions.includes(chartAction))
```
Can be
```js
(!chartActions?.includes(chartAction))
```

src/firefly/js/charts/ui/ChartSelectPanel.jsx
```diff
@@ -106,60 +106,46 @@ function getGroupKey(chartId, chartAction) {
...
+
+    return (
+        <div style={{padding: 10, ...style}}>
+            <FormPanel
+                groupKey={groupKey}
+                submitText={chartActionState===CHART_TRACE_MODIFY ? 'Apply' : 'OK'}
+                onSuccess={onChartAction({chartAction: chartActionState,
+                    tbl_id, chartId, hideDialog, renderTreeId})}
+                cancelText='Close'
+                onError={() => {}}
```
robyww, line 128: I don't think this is correct. It returns an undefined not an empty object. Maybe you want to return undefined. Either way it should be clear.

for object
```js 
onError={() => ({})}
```
for undefined
```js
onError={() => undefined}
```

src/firefly/js/charts/ui/PlotlyToolbar.jsx
```diff
@@ -155,46 +121,37 @@ function isSelectable(tbl_id, chartId, type) {
...
-class BasicToolbar extends SimpleComponent {
+function BasicToolbar({chartId, expandable}) {
+    const basicToolbarState = useStoreConnector(() => getToolbarStates(chartId), [chartId]);
 
-    getNextState(np) {
-        const {chartId} = np || this.props;
-        return getToolbarStates(chartId);
-    }
+    const {activeTrace, hasFilter, hasSelection, tbl_id, dragmode} = basicToolbarState;
+    const type = get(getChartData(chartId), `data.${activeTrace}.type`, '');
```
robyww, line 128: replace with
```js
const type= getChartData(chartId)?.data?.[activeTrace]?.type ?? ''
```

robyww (approved):
This looks good. Just make the clean up I suggested, test,  and merge.

src/firefly/js/charts/ui/ChartSelectPanel.jsx
jaladh-singhal: Well it's actually a void function that does nothing. The FormPanel component also supplies it as default for onError so I think we can omit it from here altogether.

PR #1302, opened 2022-12-20
Firefly-1109: Updates to Click to Action

robyww:
#### Firefly-1109: Updates to Click to Action

 - added to spatial and table: Search NED and Simbad
 - circumscribe polygon
 - separators in dropdowns
 - change some text as requested in the ticket
 - fixed TAP "selection type" to be "selection"
 - search area showing after search

#### Testing
 - https://firefly-1109-click-to-action.irsakudev.ipac.caltech.edu/irsaviewer
 - Try NED and Simbad from click to action

### DIFF
--- a/src/firefly/js/FFEntryPoint.js
+++ b/src/firefly/js/FFEntryPoint.js
@@ -61,5 +61,8 @@ let apiCommands;
 if (template==='FireflyViewer' || template==='FireflySlate') apiCommands= getFireflyViewerWebApiCommands();
 else if (template==='LightCurveViewer') apiCommands= getLcCommands();
 
+
+if (!template || template==='LightCurveViewer') options.searchActions= [];
+
 firefly.bootstrap(props, options, apiCommands);
 
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -35,7 +35,9 @@ import {setDefaultImageColorTable} from './visualize/WebPlotRequest.js';
 import {initWorkerContext} from './threadWorker/WorkerAccess.js';
 import {getTAPServices} from './ui/tap/TapKnownServices.js';
 import {loadAllJobs} from './core/background/BackgroundUtil.js';
-import {makeDefImageSearchActions, makeDefTapSearchActions} from './ui/DefaultSearchActions.js';
+import {
+    makeDefImageSearchActions, makeDefTableSearchActions, makeDefTapSearchActions, makeExternalSearchActions
+} from './ui/DefaultSearchActions.js';
 
 let initDone = false;
 const logger = Logger('Firefly-init');
@@ -173,6 +175,8 @@ const defFireflyOptions = {
         },
     },
     searchActions : [
+        ...makeExternalSearchActions(),
+        ...makeDefTableSearchActions(),
         ...makeDefTapSearchActions(),
         ...makeDefImageSearchActions(),
     ],
@@ -190,9 +194,10 @@ const defFireflyOptions = {
 
 /**
  * add options to store and setup any options that need specific initialization
- * @param {Object} options
+ * @param {Object} appSpecificOptions
  */
-function installOptions(options) {
+function installOptions(appSpecificOptions) {
+    const options=  mergeObjectOnly(defFireflyOptions, appSpecificOptions); // app specific will override default
     // setup options
     dispatchAppOptions(options);
     options.disableDefaultDropDown && dispatchUpdateLayoutInfo({disableDefaultDropDown:true});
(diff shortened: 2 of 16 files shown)


lrebull:
This is really cool! I'm looking forward to having this on ops.

OK, I have feedback, but I don't know if this is easily fixable or not. The first thing I did was literally the thing I'd been showing you, Trey, that I was doing the other day, that is, searching, e.g., on Sh 2-172 and looking for Simbad hits in this region. But here is what happened that I didn't expect. The default search radius for Simabad (and therefore also for you) is 2 arcminutes. When I defined a search region, **if I don't choose the pop-up** to refine the search radius, it looks like it defaults to that 2 arcmin radius, which (a) isn't what I expected (I expected it to take instead the size that I gave it  (but see update below)), and (b) wasn't big enough to give me any Simbad results, but what it actually told me was "Table Load Error: Failed to retrieve data" which means something different to me than "Simbad doesn't have any sources in this area". 

Update: Additional more methodical testing suggests that maybe it really is taking the size of the cone I'm giving it without refining that search (it is hard for me to tell, exactly, because everything it is doing is in units of degrees), but the error message really is throwing me for a loop. I'm not sure that's fixable, I realize.

robyww:
@lrebull I do sent it the search radius of the selection. I think that works because I tested it.

Really the error message is outside the scope the this ticket since it goes though general code. It is also very different since the code is just asking for a table via a url. Different servers do different things in error cases. The only thing we know is that there are no rows in the returned table.

lrebull:
Yeah, i figured, on both cases. I figured I was just getting decoyed by the units and that it really was doing what I thought it should be doing, and I figured the error message was a bigger issue than could be easily fixed. :)

lrebull:
Re: Emmanuel's comment in Slack on specific wording. OK, fair point. I assumed that I was just missing something. Let's pull this apart a little if for no other reason than someone might care in a month when we come back to this.

In the build I'm looking at, the top part of the menu reads:
Actions based on center: x, y equ z
Search (cone) using NED get with radius of x degrees
Search (cone) using Simbad get with radius of x degrees
Goto Simbad and search (cone) with radius of x degrees

All three of the choices after 'actions' seem weird to me, but I assumed that these are jargon that I just wasn't familiar with.
"NED get" and "Simbad get" -- does that really mean something specific?
"Goto Simbad" -- does that really mean something specific?

If it does NOT mean something specific, then I'd suggest changing to something that is less awkward, like this:
Search (cone) using NED with radius of x degrees
Search (cone) using Simbad with radius of x degrees
Go to Simbad and search (cone) with radius of x degrees

If it does mean something specific, please let me know and I'll put it into the docs when we get there.

robyww:
@lrebull can you can look at the table base searches and give feedback on the wording as well?

robyww:
@lrebull the word `get` in the description is a typo

robyww:
Just created a new build with feedback so far:

- https://firefly-1109-click-to-action.irsakudev.ipac.caltech.edu/irsaviewer

lrebull:
I made a movie because (1) it seemed efficient; (2) i had a question about menus and propagation of wording; (3) i found a bug and wanted to show you how i made it happen. let me know if you need me to type any/all of this up

https://user-images.githubusercontent.com/23038600/208784252-6a73e427-58dd-4783-967c-00a8faca210e.mp4

robyww:
I will fix the dropdown issue in this ticket. Thanks for finding it.

robyww:
@lrebull When I said tables, There is a dropdown on the tables menu as well. Do a tap or catalog search and you will see it.

robyww:
@lrebull Just made a new build with drop down issue fixed.

 - when reload make sure you don't have an old cache version. the version should show:
Version:    v5.3.2021.3.2
Built On:   Wed Dec 21 00:24:14 UTC 2022
Git Commit: c3386461
Firefly Library Version: 2022.3-DEV:HEAD_89c2
Firefly Git Commit:      89c29148e

lrebull:
sorry, trey, i'm still not understanding what it is you want me to check. here is a video explaining my confusion... maybe i'm not hitting the right server? you can see me try to check, and the server i am checking has ever so slightly different numbers for the git commit -- it has more characters than you have written above (see movie).

https://user-images.githubusercontent.com/23038600/208793711-103fabce-e67c-4449-a092-4e8b816bd2db.mp4

lrebull:
OHHHH *after* a catalog search! i get it ! sorry. looking now

lrebull:
OK, each of these options specify "row" except the middle ones, which may confuse users by the omission. suggest changing them as follows:

- Search NED at row with 5" radius
- Search Simbad at row with 5" radius
- Go to and search Simbad at row with 5" radius

note that third option also has not capitalized "search" (this is like the currently reworked image pulldown)

sorry for my confusion before!!

PR #1303, opened 2022-12-21
FIREFLY-250: Lock position and size across RGB for 3 color image

jaladh-singhal:
Fixes [FIREFLY-250](https://jira.ipac.caltech.edu/browse/FIREFLY-250)

Stored target position and size in its own field group, such that all channels: red, green, blue (as well as single and hips as suggested by @robyww) use the same values.

## Testing
https://fireflydev.ipac.caltech.edu/firefly-250-3color-lock-size/firefly/

Make archive selection in all different channels and ensure that target position and size remains same - changing it in one should change it in others too. Although in hips, size is allowed to be set differently.

### DIFF
--- a/src/firefly/js/visualize/ui/ImageSearchPanelV2.jsx
+++ b/src/firefly/js/visualize/ui/ImageSearchPanelV2.jsx
@@ -54,7 +54,8 @@ const FG_KEYS = {
     red: 'ImageSearchPanel_red',
     green: 'ImageSearchPanel_green',
     blue: 'ImageSearchPanel_blue',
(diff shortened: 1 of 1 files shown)


robyww (commented):
This looks fine. I did notice a little bug when input starts with a dot (`.`). I also looked at the FieldGroup approach. I think you are  right. It is better.

robyww (approved):
This all looks good. Don't merge yet though. I want to evaluate if we want to do a part two.

robyww:
After some evaluation I have changed my mind. Go ahead and merge and will make a second ticket to do the default 3 color

PR #1311, opened 2023-01-19
FIREFLY-1136: embedded DB customization via props

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1136
- define cleanup parameters as props
- fine tune cleanup policy
- added better visibility via /admin/status

Not much to test.  Review policy and cleanup parameters to make sure it's sensible.
`1 million max memory rows for every 1GB of memory at startup.` is a conservative default setting.  The idea is to use custom props for special use case.


robyww (approved):
Looks fine.

PR #1312, opened 2023-01-20
Firefly 1130 Modify FileUploadViewPanel to take an accept list

kpuriIpac:
#### [Firefly-1130](https://jira.ipac.caltech.edu/browse/FIREFLY-1130):
- really a draft PR, more work needs to be done but wanted to check if it looks okay so far
- most of the work that's left is for **FileUploadProcessor** (**FileAnalysis** in **FileUploadViewPanel** takes care of giving users warnings for file types not allowed - but they're still able to load files of types not in acceptList - that'll need to be taken care of in the **resultSuccess** function of **FileUpoadProcessor**)
- added checks in **ImageDisplayOption** and **TableDisplayOption** to show appropriate checkboxes (for spectra, moc, images, etc.) based on what's in acceptList 
- Currently in **FileUploadViewPanel's** **getLoadButtonText** function, I added warnings for when the user attempts to select tables and images (when either one is not present in the acceptList) 
    - this is not ideal. As long as the user is allowed to select items that are not in acceptList, my current check will throw warnings even if you pick images (and say, in this instance, images are allowed but tables are not) because tblCnt will be >= 1 if you have previously selected any 
      - I'll work on improving this
- I used **summaryModel** to check the fileTypes of the uploaded file (and not report.parts.dataTypes - reason being that the latter classifies even Table entries as Images - this is taken care of in **makeSummaryModel** function by looking at the number of axes to decide if an entry is Table or Image - which is why I decided to use summaryModel to get uniqueTypes) 
- The types I checked for were **REGION**, **Image** and **Table** (from the various fits, vot, moc fits, reg, etc. files that I had, I only saw these and "**HeaderOnly**" show up as types)
  - then added further checks for spectrum-tabels, moc-tables, and data-link-tables under Table
  - the only thing I wasn't sure about was if "**Spectrum**" needs a separate check? Are there fileTypes where it could show up as a type? If so, I can add a check for that as well..
  
**Updates 1/30**
- PR is now ready for review 
- Based on feedback, separated out some of the logic into functions (such as determineAccepted, buildAllowedTypes) 
- On firefly-dev, moved the buttons created for different acceptLists from the top navigation bar to the tabs within 'Test Searches' (new tab is called 'Uploads') 
- Added checks in FileUploadProcessor's resultSuccess to make sure even if user uploads a file type that is not allowed (FileAnalysis will show warning text and not upload the file), they still shouldn't be able to "Load" the file 
- If either tables or images or not allowed, and the uploaded file type has both tables and images as entries, then either tables or images will be pinked out. Additionally, added a warning line that reads "Any selected entries with highlighted pink lines will not be loaded."
   - If, for instance, images are not in the defaultAcceptList, but the user selects a few, the load button text will not change to reflect any selected images. 
   - Also, if the user tries to load this file with any images selected, we'll still load any selected tables but give a warning dialog that'll let the user know we're only loading tables 
- made some changes to the **getFirstExtWithData** function in FileUploadViewPanel. It would not have worked previously because we were checking for reports.parts.type, but having tested that I realized it always returns 'Image' for any FITS file (even for 'Table' entries). Fixed it to now look for type in summaryModel.tableData.data, which returns 'Image' or 'Table' accurately (makeSummaryModel takes care of that by looking at the number of axes for each entry to determine whether it's a 'Table' or 'Image'). 

**Updates 1/31**
- Cleaned up FileUploadProcessor - moved all showInfoPopup's from resultSuccess to a new determineValidity function
    - I tried hard to see if I could re-use determineAccepted from FileUploadViewPanel with a few tweaks instead, but it was getting a lot more complicated since the logic to determine validity before loading a file is a bit more complex (relies on tableIndices, imageIndices to check what selections the user made, etc.) than determining if the file type uploaded is allowed or not
- added a showYesNoPopup dialog for when user selects any image(s) or table(s) when either is not allowed - this will allow user to click Yes/No to proceed or cancel loading if they have some incorrect selections made that they'd like to edit per @loitly's suggestion  
- got rid of the tablesOnlyResultSuccess function as it's been made redundant now by the other checks we have in place  

**Updates 2/2**
- note: some of these changes that haven't been fully reviewed are from **Updates 3/31** above, if it helps 
- quite a few changes, further separation of logic into functions in FileUploadProcessor (determineValidity and determineLoadType do most of the work here) 
- also accounted for separation of logic when accept list has something like *only* DATA_LINK_TABLES or *only* MOC_TABLES...or when accept list includes either TABLES or SPECTRUM_TABLES (either of those indicate that we accept *all* tables - including data link and moc tables) 
    - had to make changes for this in FileUploadProcessor, but also in FileUploadViewPanel (for the text displayed before user uploads a file, and also after they upload it - to account for difference in rows being highlighted *pink* - we don't want to do that for moc/DL tables - so added a check for that as well) 

#### Bugs (some pre-existing):
- if you upload a moc fits file (in the regular upload panel, not the HiPs/MOC dialog), you'll see 2 radio button options: "Load as MOC Overlay" and "Load as Table". The Load button text reads "Load MOC" - and it doesn't change when you switch to "Load as Table" either (I'm not sure that it should, but if it needs to change to, say, "Load Table" - then this is a bug). 
- If you upload a file with only tables (like this one: http://web.ipac.caltech.edu/staff/roby/data-products-test/objsearch_ext.xml), then the load button text reads "Load Table", and it does not change when you start selecting more than 1 table. 
- Select a few tables/images in a fits file. If you then attempt to unselect an already selected/highlighted row, the load button text does not change to reflect this. In fact, you can now attempt to select and unselect this highlighted row multiple times, but the load button text will not change. it'll only be updated when you select/unselect a new row. 
- Looking into these bugs - and trying to see if I can fix them within this ticket. Let me know which ones (or all) seem most pressing. 

#### Testing (Updated 2/2): 
  - https://fireflydev.ipac.caltech.edu/firefly-1130-fuvp-acceptlist/firefly and https://fireflydev.ipac.caltech.edu/firefly-1130-fuvp-acceptlist/firefly/firefly-dev.html
  - **Testing Changes from 2/2**: Note: follow the steps below, but I also added 2 new buttons on the firefly-dev test build link: one for MOC_TABLES only and one for DATA_LINK_TABLES only 
  - on the firefly-dev testing link, under the "Test Searches" tab, theres'a new "Uploads" tab. The Uploads tab has 4 options: 'Tables Only Upload', 'All Tables Upload', 'Region Upload', 'Image Upload' and 'Accept Everything (Default)'. 
  - click on each one of those, make sure the text for what types of files the user is allowed to upload from each looks good
  - After changes on **1/31**: for files with multiple parts (fits or table files), attempt to select nothing and click on 'Load'. You should get a popup warning saying no extension was selected. Also attempt to select only the 'HeaderOnly' extension and load, this should give a popup warning as well. 
  - attempt to upload the allowed file type, and also file types not allowed from each. Attempt to 'Load' in both cases (should work if file type is allowed, and should give a warning if file type is not allowed). Example: On firefly-dev, under 'Test Searches' -> 'Uploads', click on 'Region Only' - and attempt to upload and load any non-region file type here. 
  - From the upload panel, also attempt to upload and load a file type that is not allowed/supported. Eg: try a PDF file. You should get a popup warning when you try to load. 
  - From the 'Tables Only Upload', upload a fits file with both tables and images (image entries should be pinked out - and a warning should show up right above the table) 
     - The first table entry should be the one selected by default (and not the first image entry) 
     - select a few images and tables and 'Load' - make sure the load button text changes only for tables, and upon loading you should see a warning dialog that says only tables will be loaded
     - if you select only images and unselect all table entries and then attempt to load, you should get a warning dialog that prompts you to select at least one table 
 - Repeat this process on the 'Image Upload' tab (here, the table entries should be pinked out)  
- Repeat this process from the Hips/Moc dialog (from "Images", choose Image Type: View HiPS Images - search for target m1, then click on "Add MOC" and go to the upload panel tab - repeat the same steps as above - attempt to upload both a moc fits file, and any non-moc fits file as well).

### DIFF
--- a/src/firefly/js/ui/FileUploadProcessor.jsx
+++ b/src/firefly/js/ui/FileUploadProcessor.jsx
@@ -1,10 +1,9 @@
 import {FileAnalysisType, Format} from 'firefly/data/FileAnalysis';
-import {showInfoPopup} from 'firefly/ui/PopupUtil';
+import {showInfoPopup, showYesNoPopup} from 'firefly/ui/PopupUtil';
 import {getAppHiPSForMoc, isMOCFitsFromUploadAnalsysis} from 'firefly/visualize/HiPSMocUtil';
 import {MetaConst} from 'firefly/data/MetaConst';
 import {isLsstFootprintTable} from 'firefly/visualize/task/LSSTFootprintTask';
 import {makeFileRequest} from 'firefly/tables/TableRequestUtil';
-import {getAppOptions} from 'firefly/core/AppDataCntlr';
 import {getFieldVal} from 'firefly/fieldGroup/FieldGroupUtils';
 import {createNewRegionLayerId} from 'firefly/drawingLayers/RegionPlot';
 import {dispatchCreateImageLineBasedFootprintLayer, dispatchCreateRegionLayer, getDlAry} from 'firefly/visualize/DrawLayerCntlr';
@@ -19,6 +18,12 @@ import {getAViewFromMultiView, getMultiViewRoot, IMAGE} from 'firefly/visualize/
 import {PlotAttribute} from 'firefly/visualize/PlotAttribute';
 import {isAnalysisTableDatalink} from '../util/VOAnalyzer.js';
 import {fetchDatalinkUITable} from './dynamic/FetchDatalinkTable.js';
+import {dispatchHideDialog} from 'firefly/core/ComponentCntlr';
+import React from 'react';
+import {
+    acceptAnyTables, acceptDataLinkTables, acceptImages, acceptMocTables, acceptNonDataLinkTables,
+    acceptNonMocTables, acceptRegions, acceptTableOrSpectrum, IMAGES, REGIONS, TABLES
+} from 'firefly/ui/FileUploadUtil';
 
 const FILE_ID = 'fileUpload';
 const uploadOptions = 'uploadOptions';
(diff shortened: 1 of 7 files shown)


src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -201,12 +219,27 @@ const LoadingMessage= ({message}) => (
...
+        if (tblCnt >= 1) {
+            showInfoPopup( 'Warning: Cannot select tables.', 'Warning');
+        }
+    }
+
+    if (!acceptList.includes('Image')) {
+        if (imgCnt >= 1) {
+            showInfoPopup( 'Warning: Cannot select images.', 'Warning');
+        }
+    }
```
robyww, line 242: cleaner if this could be 1 if 
```js
if (!acceptList.includes('Image')) && imgCnt >0)
```

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -173,9 +191,9 @@ export function FileUploadViewPanel({setSubmitText, acceptMoc}) {
                                                        }}/> }
                         </div>
                     </div>
-                    <FileAnalysis {...{report, summaryModel, detailsModel,tablesOnly, isMoc, UNKNOWN_FORMAT, acceptMoc}}/>
-                    <ImageDisplayOption summaryTblId={summaryTblId} currentReport={report} currentSummaryModel={summaryModel}/>
-                    <TableDisplayOption {...{isMoc, isDatalink, summaryTblId, currentReport:report, currentSummaryModel:summaryModel, acceptMoc}}/>
+                    <FileAnalysis {...{report, summaryModel, detailsModel, isMoc, UNKNOWN_FORMAT, acceptList}}/>
+                    <ImageDisplayOption summaryTblId={summaryTblId} currentReport={report} currentSummaryModel={summaryModel} acceptList={acceptList}/>
```
robyww, line 195: you might want to make this component to use the object style of passing props just to keep in more consistent with the line before and after.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -565,59 +604,123 @@ function Details({detailsModel}) {
...
+        You can load any of the following types of files:
+        <ul>
+            {list.map((fileType, index) => {
+                return <li key={index} style={liStyle}>{fileType}</li>;
+            })}
+        </ul>
+    </div>);
+};
+
+const NotAccepted = (props) => {
```
robyww, line 631: good idea

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -565,59 +604,123 @@ function Details({detailsModel}) {
...
             <div style={{flexGrow: 1, marginTop: 40, textAlign:'center', fontSize: 'larger', color: 'red'}}>
                 Error: Unrecognized Format
             </div>
         );
     }
-    return <AnalysisTable {...{summaryModel, detailsModel, report, isMoc, UNKNOWN_FORMAT}} />;
+    return <AnalysisTable {...{summaryModel, detailsModel, report, isMoc, UNKNOWN_FORMAT, acceptList}} />;
 }
 
+const AcceptedList = (props) => {
```
robyww, line 618: good idea

robyww (commented):
I think you are on the right track. I did not test but did look closely at the code.

- it might be easier to read to have a bunch of small functions to check the acceptList stuff
    - `acceptImage (al) -> al.includes('Image')`
    - you can also use this approach to check multiple type like with all the tables.
- spectrum does not need a separate check
- export and use the constants you defined.

src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -27,26 +27,64 @@ const SUPPORTED_TYPES=[
...
     const tableIndices = getSelectedRows(FileAnalysisType.Table, summaryTblId, currentReport, summaryModel);
     const imageIndices = getSelectedRows(FileAnalysisType.Image, summaryTblId, currentReport, summaryModel);
 
+    if (isRegion(summaryModel) && !acceptRegions(acceptList)) {
+        showInfoPopup('You may not load a Region file from here.', 'File Type Mismatch');
+        return false;
+    }
+
+    //uniqueTypes are the types of the parts of the uploaded file (retrieved in FileAnalysis)
+    if (uniqueTypes.includes(IMAGES) && uniqueTypes.includes(TABLES)) {
```
loitly, line 50: I got stuck here.  The logic is very complicated.  In this block, `uniqueTypes` is only checked against `IMAGES` and `TABLES`.  But elsewhere, it may contains other values, like `uniqueTypes.includes(REGIONS)`.  Also, is `uniqueTypes` good for validation?  
I have a hard time distinguishing validation logic versus processing logic.

For validation, I would expect to see `acceptList` match against the type returned from each selected row.
Of course, if the file is a primary data and not a table of data products, then it should validate against that.

The type returned from each row should be the most detailed type, i.e.  Spectrum, Region, etc.  
The type in `acceptList` can be a superclass, like Tables.

Once pass validation, then I would expect to see a loop to fire off action(s) based on the selected type.
I believe if done right, it would simplify the code, at least for me.

loitly (commented):
I will only leave a comment, since I may not fully understand the complexity of this PR.

When there are a mixed of valid and invalid selections, it goes ahead and process what it can, then show a popup warning saying some bad selections were ignored.
IMO, it should show the warning popup first with a `Continue anyways` button.  This way, if the user select cancel, he/she can see what selections were wrong.  Both approach requires the same number of clicks.

robyww (changes requested):
I have not been able to review everything yet, and since I might be committed for this rest of the day I will give you my partial comments.

- These long if blocks need to be broken into functions that just return the acceptableness.
- use the accept... functions you made
- break up using analysis functions and the UI.

I included examples of what I mean.

loitly (approved):

robyww (approved):
Looks good. Fix the yes/no issue and merge.

PR #1313, opened 2023-01-24
Firefly-1144: add better support tap search params for click-to-action

robyww:
#### [Firefly-1144](https://jira.ipac.caltech.edu/browse/FIREFLY-1144)
  - added more parameter checking in TAP
  - Fixed a small bug in ServerContext with choosing working directory
  - added "click-to-search" masking to control which features are avilable.

#### Testing - can only test on suit data-int

- Test on https://data-int.lsst.cloud/
- search a table 
    - Use click to search feature
    - try searching the truth table or obscore from image after selecting an area
    - try searching the truth table or obscore  from table

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/ServerContext.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/ServerContext.java
@@ -207,10 +207,12 @@ public class ServerContext {
             log.info(desc +  "; Using path from property: " + fromProp);
             File workDir = new File(fromProp);
             initDir(workDir);
-            if (!workDir.canWrite()) {
+            if (workDir.canWrite()) {
+                return workDir;
+            }
+            else {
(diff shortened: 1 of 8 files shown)


loitly (approved):
Changes look good.  
I do have a comment on the idea of `searchActions`.  I was confused at first when I saw it being reference in TablePanel, thinking why would there be searches here.  After reviewing the usage, then I understand.  In my opinion, it should have a more generic name since it's not only searching.  It can display things, navigate to a page with prefilled form fields, or anything else we wish to include.

Naming it something like, `shortcuts`, or even `clickActions`.  
Then, a component that is `shortcuts-aware`, can include it by adding it to its options, like this.
``` 
# Firefly.js
options:
    table : {
        pageSize: 100,
        shortcuts: {includes: "a,b,c"},
    }
```

Of course, this is a lot of changes and has nothing to do with this PR.  Something you may want to consider for the future.

PR #1315, opened 2023-01-31
FIREFLY-613: remember table options selection

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-613
Test: https://fireflydev.ipac.caltech.edu/firefly-613-table-col-sticky/firefly/

Remember user's selection from Table Options, then apply the same settings to the same table.  This includes columns, showUnits, showFilters, and showTypes.
`same` is defined with a `tbl_pref_key` tableMeta key, or the uniqueness of the combined column names in the same order.  
When previous settings(preferences) are applied to a table, show an option to clear it.

What is the difference between `Reset` and `clear preferences`.  Both will clear preferences and reset columns, showUnits, showFilters, and showTypes to default settings.  `Reset` also reset the `Page Size` and any filters added to the table.

Things to consider:
- Is there another way of separating `clear` from `Reset`?
- Currently, this feature is apply to all table across all applications.  Meaning, if you change Table Options to an `AllWISE Source Catalog` in IRSAViewer, it will be remembered when you open the same table in SOFIA.  Is this desirable?
- Always on.  Is this a good thing?  Do we need to add a flag so that it can be disabled?
- `tbl_pref_key` tableMeta can be used to pin down a specific table.  Are there any table we need to add this to?
    - Take HiPS image table.  Do we want to remember preferences here?

### DIFF
--- a/src/firefly/js/tables/TablePref.js
+++ b/src/firefly/js/tables/TablePref.js
@@ -1,0 +1,82 @@
+/*
+ * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
+ */
+
+
+import md5 from 'md5';
+import {getColumns, getTableUiById, getTblById} from './TableUtil.js';
+import BrowserCache from '../util/BrowserCache.js';
+import {dispatchTableUiUpdate} from './TablesCntlr.js';
+import {getModuleName} from 'firefly/util/WebUtil.js';
+
+export const PREF_KEY = 'tbl_pref_key';
+export const PREF_APPLIED = 'prefApplied';
+
+/**
+ *
+ * @param tbl_ui_id
+ * @param uiData
+ * @return {boolean} returns true if column preferences were applied
+ */
+export function applyTblPref(tbl_ui_id, uiData) {
+    const prefKey = getTblPrefKey(tbl_ui_id);
+    if (!prefKey) return false;
+
+    const cached = BrowserCache.get(prefKey);
+    if (cached?.colPref) {
+        const colPref = new Map(JSON.parse(cached.colPref));
+        uiData.columns?.filter((c) => c.visibility !== 'hidden')
+            .forEach((c) => {
+                if (colPref.has(c.name))    c.visibility = colPref.get(c.name);
+            });
+        uiData[PREF_APPLIED] = true;
+    }
+    return uiData[PREF_APPLIED];
+}
+
+/**
+ * set the visible columns in this table as the preferred visible columns
+ * @param {string} tbl_ui_id
+ */
+export function setTblPref(tbl_ui_id) {
+    const prefKey = getTblPrefKey(tbl_ui_id);
+    if (!prefKey) return;
+
+    const {columns=[]} = getTableUiById(tbl_ui_id) || {};
+    const colPref = JSON.stringify(columns.map((c) => [c.name, c.visibility || 'show']));  //
+
+    BrowserCache.put(prefKey, {colPref});
+}
+
+/**
+ * clear this table preferences
+ * @param {string} tbl_ui_id
+ */
+export function clearTblPref(tbl_ui_id) {
+    const prefKey = getTblPrefKey(tbl_ui_id);
+    if (!prefKey) return;
+
+    BrowserCache.remove(prefKey);
+    dispatchTableUiUpdate({tbl_ui_id, [PREF_APPLIED]:false});
+}
+
+/**
+ * @param tbl_ui_id
+ * @return {string} the key used to get column's preferences for the given table.  If a PREF_KEY is not
+ * in TableMeta, then return undefined to indicate Preferences is not used for this table.
+ */
+export function getTblPrefKey(tbl_ui_id) {
+    const {tbl_id} = getTableUiById(tbl_ui_id) || {};
+    const tableModel = getTblById(tbl_id);
+    if (!tableModel) return undefined;
+
+    const prefKey = tableModel.tableMeta?.[PREF_KEY];
+        // disable default impl.  will only apply to table with PREF_KEY
+        // md5(getColumns(tableModel)
+        //     .map( (c) => c.name)
+        //     .join('|')
+        // );
+    return prefKey && `colPref-${getModuleName()}-${prefKey}`;
+}
+
+
--- a/src/firefly/js/tables/ui/TablePanelOptions.jsx
+++ b/src/firefly/js/tables/ui/TablePanelOptions.jsx
@@ -4,15 +4,15 @@
 
 import React, {useEffect} from 'react';
 import PropTypes from 'prop-types';
-import {get, cloneDeep, isEmpty} from 'lodash';
+import {cloneDeep, get, isEmpty} from 'lodash';
 
-import {SqlTableFilter, setSqlFilter} from './FilterEditor.jsx';
+import {setSqlFilter, SqlTableFilter} from './FilterEditor.jsx';
 import {InputField} from '../../ui/InputField.jsx';
 import {intValidator} from '../../util/Validate.js';
-import {getTableUiById, isClientTable} from '../TableUtil.js';
 import {useStoreConnector} from '../../ui/SimpleComponent.jsx';
-import {dispatchTableAddLocal, dispatchTableRemove, TABLE_SELECT, dispatchTableFilter} from '../TablesCntlr.js';
-import {COL_TYPE,  getTblById, isOfType, parsePrecision, getColumnValues, watchTableChanges, getFilterCount, getColumn} from '../TableUtil.js';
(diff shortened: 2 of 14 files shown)


src/firefly/js/tables/ui/TablePanelOptions.jsx
```diff
@@ -244,6 +246,20 @@ export const ColumnOptions = React.memo(({tbl_id, tbl_ui_id, ctm_tbl_id, onChang
...
 
+function TablePref({tbl_id, tbl_ui_id}) {
+
+    const prefApplied = useStoreConnector(() => getTableUiById(tbl_ui_id)?.[PREF_APPLIED]);
+    if (!prefApplied) return null;
+
+    return (
+        <div className='TablePanelOptions__pref'>
+            Preferences were applied to this table.
+            Click <div className='ff-href' style={{margin: '0 4px'}} onClick={() => resetTblOptions(tbl_id, tbl_ui_id)}> here </div> to clear it.
```
robyww, line 257: I thought we were getting away from links to act as buttons and only using them to open other pages.  You probably should use a button here. 

I see you are trying to make it low key, which is a good idea. I sort of like the link better but I think we made a decision a couple of years ago to use buttons.

robyww (approved):
Looks good. I tested an it appears to work though it is a little hard to test beyond a reload.

- The most important comments are probably the links as buttons and the expiration.

src/firefly/js/tables/TablePref.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * set the visible columns in this table as the preferred visible columns
+ * @param {string} tbl_ui_id
+ */
+export function setTblPref(tbl_ui_id) {
+
+    const {showUnits, showTypes, showFilters, columns=[]} = getTableUiById(tbl_ui_id) || {};
+    const visiCnames = columns.filter((c) => (c.visibility || 'show') === 'show').map((c) => c.name);
+
+    BrowserCache.put(getColsPrefKey(tbl_ui_id), {showUnits, showTypes, showFilters,visiCnames});
```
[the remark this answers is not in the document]
loitly: Yes.  We should gather feedback on this.  But in my opinion, having an expiration time only complicate matters.  Clearing is straight forward and easy to do.

src/firefly/js/tables/ui/TablePanelOptions.jsx
loitly: I think a button between words or in a sentence is not very clean.  But, I will change to whatever is decided at the end of the review.
Another option is to not have the 'clear' function at all.  Just use 'Reset'.  It will clear filters and pagesize as well, but it will most likely not have any if it's the first thing you do after seeing the table.

src/firefly/js/tables/TablePref.js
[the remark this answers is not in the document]
robyww: ok, makes sense

src/firefly/js/tables/ui/TablePanelOptions.jsx
robyww: The button would have to be something link: 
```js
Preferences were applied to this table: <button> Remove</button>
```

robyww: I don't done know about just using reset.

lrebull:
Only noting problems in the interest of efficiency:

1. loaded images, then catalogs. pulled up gears for table. deselecting columns reloads table within pop-up window, so that if i scroll down to see some of the columns later in the table, as soon as i click on the tickbox, it reloads the window so that any scrolling down i did is lost. super annoying for unclicking a lot of columns in a row. Same thing happens when i click them all off and then selectively click some on.
2. my table options were NOT remembered. things i did: search on images first (WISE), then catalogs (2mass). did not select columns on search screen (is that where i did things wrong?) but instead selected columns from the gears in the search results (see prior item). searched 2mass catalogs a second time and it still gave me the default columns for a 2mass catalog, not my selected subset.
3. my table options were only sort of remembered when i selected columns from the search screen. i tried to select just ra,dec, j_m, j_msigcom, h_m, h_msigcom, k_m, and k_msigcom. additional columns (clon, clat, rd_flag, j_h, h_k, j_k) were returned even though I did not select them. These all were remembered for the next query.
4. description says something about 'clear' as being distinct from 'reset' but i only have 'reset', not 'clear', fwiw.

not sure what else i am supposed to look at or break...

loitly:
> Only noting problems in the interest of efficiency:

@lrebull reloading after a checkbox is clicked is an existing problem, but I agreed it's very annoying.  I will see if it's an easy fix.

Other than that, I don't understand why you're still getting the 'old' version.  I tested my PR instance in all 3 browsers and it worked fine.  We may have a caching bug since it happens to you quite often.  It's hard for me to duplicate since I clear cache all the time during development.  But, we should look into this @robyww.

@lrebull, if you can try again what you did but in a new private or incognito window.  Don't select columns from the search screen.  Use the default.   Select a few columns from the Table Options(gear).  Then when you do the same catalog search the 2nd time, it should remember the columns you selected.

lrebull:
problem seems to be that i was doing a cone search after a polygon search -- see video.

https://user-images.githubusercontent.com/23038600/216173823-39478205-b031-4ad3-9936-98e7902baf3a.mp4

loitly:
> problem seems to be that i was doing a cone search after a polygon search -- see video.

Ok.  After watching the video, I was able to duplicate the problem and make sense out of it.  As I mentioned in the PR, without specifically defining a `tbl_pref_key` to a search, I use the column names to determine equality of two tables.

The reason why Cone search and Polygon search were not considered the same results is because Gator added 2 additional columns to Cone search results; `dist` and `angle`.   
I also noticed Gator would add additional columns even though I did not ask for them.  When I requested for only `ra`, `dec` and `designation`, I get back `clon`, `clat` and several other columns.

To make this more usable, I will use catalog table name as the preference key(`tbl_pref_key`).  I will do the same for `TAP Searches`.  I'll let you know when I am done.

loitly:
@lrebull @robyww code and test link updated.  For TAP, I applied `tbl_pref_key` to `Single Table (UI assisted)` only, NOT `advanced ADQL`.

ejoliet (commented):
Thanks for implementing this most wanted and waited feature!

It kind of work as i would expect. I've played a little bit with Gator and TAP.

I think results are expected apart from when selecting other columns by default from the search screen. 
I think what should happen is always do the "intersection" (merge) of the columns list from the search and the columns of the preferences stored. Columns previously not displayed and new to the result should be displayed imo.
That way if there is new column they should be visible unless they were already present and not selected or displayed.

> Is there another way of separating clear from Reset?

We should clear this and merge the function, 'reset' and 'clear' should be doing the same imo.

> Currently, this feature is apply to all table across all applications. Meaning, if you change Table Options to an AllWISE Source Catalog in IRSAViewer, it will be remembered when you open the same table in SOFIA. Is this desirable?

I would expect so but if it's not the agreement among CCB, we would add 'app' name or something to the key to be able to separate app context? Looks like a lot of bookkeeping.
 I'd guess users would see natural to have same behavior across apps, imo.

> Always on. Is this a good thing? Do we need to add a flag so that it can be disabled?

No, i think we want table preferences always on.

> tbl_pref_key tableMeta can be used to pin down a specific table. Are there any table we need to add this to? Take HiPS image table. Do we want to remember preferences here?

Yes, table preferences should be applied to any table.

With those different use cases and questions we have, I think we should ask CCB to confirm and they should discuss and agree on the behavior and expectations.

loitly:
@lrebull @robyww @ejoliet I updated the test url to reflect the changes we talked about in the meeting.
https://fireflydev.ipac.caltech.edu/firefly-613-table-col-sticky/firefly/

Here's a summary of the changes made:
- Show newly added columns
- Column selection will only be applied to the current application.
- Replace current `Reset` button with 2 buttons
  - `Reset column selection` and `Clear all filters`
- Fixed table reloading after checkbox is clicked issue reported by @lrebull 
- Don't remember `Show` flags( show units, show filters, etc)
- Show message to indicate selection will be used in future results
  - Don't have the official message, yet.

TBD:  Should this feature be applied to all tables?
Currently, yes.  If no, then we need to add a table meta to every table we want to enable this feature on.

lrebull:
nice! 

- change button slightly: "Reset columns selection" --> "Reset column selection"
- Clicking "remove all filters" actually does remove the filters but they are left shown in place in the pop-up. (see screenshot)

Screen Shot 2023-02-07 at 5 48 38 PM

loitly:
> * change button slightly: "Reset columns selection" --> "Reset column selection"
> * Clicking "remove all filters" actually does remove the filters but they are left shown in place in the pop-up. (see screenshot)

Thank you!  fixed.

PR #1316, opened 2023-02-03
Add checkoutFirefly to gwt branch

loitly:
So irsa-ife can use it.


ejoliet (approved):
looks ok. Thanks for adding this to gwt.

PR #1318, opened 2023-02-07
Firefly-1147: Modify upload panel to accept/allow one item

kpuriIpac:
#### [Firefly-1147](https://jira.ipac.caltech.edu/browse/FIREFLY-1147): Modify upload panel to allow/accept only one item 
- added a new prop in FileUploadViewPanel: acceptOneItem (when it's true, we only allow user to select and load with one entry) 
- Default value of acceptOneItem is false (so we accept any number of entries) but FileUploadDropdown can be called with acceptOneItem set to true 
- some small changes to in FileUploadViewPanel and FileUploadProcessor to accommodate for acceptOneItem being true
   - To hide checkboxes, I call TabelPanel with selectable={false} when acceptOneItem is true

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1147-acceptonlyoneitem/firefly/firefly-dev.html
-  Go to Test Searches -> Uploads
   - All of the buttons here call the upload panel with acceptOneItem set to true 
- Click on "Tables Only" or "All Tables" upload, and upload a FITS file with both tables and images
    - ensure that checkboxes are hidden
    - first highlighted entry should be a table 
    - if you attempt to select an Image or HeaderOnly and "Load", you should get an error dialog 
- Repeat the above step from the "Image Upload" button (image should be the first highlighted item here) 
- From the MOC/DataLink buttons, attempt to upload a non-MOC/non-DataLink file and "Load" - this should give you an error dialog 
- Test uploading a MOC fits file from the "Moc FITS Only" button - you should not see any options here ("Load as MOC" / "Load as Table")
    - if you upload a MOC fits file from the "Tables Only" or "All Tables" upload button, though, you should see these 2 options show up  
-  Make sure other uploads work as expected (e.g.: region)

### DIFF
--- a/src/firefly/js/ui/FileUploadProcessor.jsx
+++ b/src/firefly/js/ui/FileUploadProcessor.jsx
@@ -49,12 +49,14 @@ const tableWarning = 'Only loading the image(s), ignoring any selected table(s).
 
 export function resultSuccess(request) {
 
-    const {currentReport, currentDetailsModel, summaryModel, groupKey, acceptList, uniqueTypes}= request.additionalParams ?? {};
+    const {report, detailsModel, summaryModel, groupKey, acceptList, uniqueTypes,
+        acceptOneItem}= request.additionalParams ?? {};
     const summaryTblId = groupKey; //FileUploadAnalysis
     const fileCacheKey = getFileCacheKey(groupKey);
 
     //determine if the file type or selections are valid to be loaded
-    const {valid, errorMsg, title} = determineValidity(acceptList, uniqueTypes, summaryModel, summaryTblId, currentReport);
+    const {valid, errorMsg, title} = determineValidity(acceptList, uniqueTypes, summaryModel,
+        summaryTblId, report, acceptOneItem);
 
     if (!valid) {
         showInfoPopup(errorMsg, title);
@@ -62,18 +64,18 @@ export function resultSuccess(request) {
     }
 
     const {loadType, tableIndices, imageIndices} = determineLoadType(acceptList, uniqueTypes, summaryModel,
-        summaryTblId, currentReport, fileCacheKey, request, currentDetailsModel);
+        summaryTblId, report, fileCacheKey, request, detailsModel, acceptOneItem);
 
     switch(loadType) {
         case LOAD_REGION:
-            sendRegionRequest(fileCacheKey, currentReport);
+            sendRegionRequest(fileCacheKey, report);
             return true;
 
         case LOAD_MOC:
             const mocMeta= {[MetaConst.PREFERRED_HIPS]: getAppHiPSForMoc()};
             if (request.mocOp==='table') mocMeta[MetaConst.IGNORE_MOC]='true';
             //loadToUI = true if request.mocOp==='table', else loadToUI=false
-            sendTableRequest(tableIndices, fileCacheKey, Boolean(request.tablesAsSpectrum==='spectrum'), currentReport, Boolean(request.mocOp==='table'), mocMeta);
+            sendTableRequest(tableIndices, fileCacheKey, request.tablesAsSpectrum==='spectrum', report, request.mocOp==='table', mocMeta);
             return true;
 
         case LOAD_DL:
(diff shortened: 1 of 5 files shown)


src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -49,31 +49,39 @@ const tableWarning = 'Only loading the image(s), ignoring any selected table(s).
...
+    //highlighted row and entryType is used when acceptOneItem === true
+    const highlightedRowIdx = summaryModel?.highlightedRow;
+    const highlightedRow = summaryModel?.tableData?.data?.[highlightedRowIdx];
+    const entryType = highlightedRow?
+        highlightedRow[1]: summaryModel?.tableData?.data?.[0]?.[1];
+
     //determine if the file type or selections are valid to be loaded
-    const {valid, errorMsg, title} = determineValidity(acceptList, uniqueTypes, summaryModel, summaryTblId, currentReport);
+    const {valid, errorMsg, title} = determineValidity(acceptList, uniqueTypes, summaryModel,
+        summaryTblId, report, acceptOneItem, highlightedRowIdx, entryType);
```
robyww, line 65: you are passing `determineValidity` the summary model already. Pull this out in the function and keep result sucess clean.

src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -49,31 +49,39 @@ const tableWarning = 'Only loading the image(s), ignoring any selected table(s).
...
 
     if (!valid) {
         showInfoPopup(errorMsg, title);
         return false;
     }
 
     const {loadType, tableIndices, imageIndices} = determineLoadType(acceptList, uniqueTypes, summaryModel,
-        summaryTblId, currentReport, fileCacheKey, request, currentDetailsModel);
+        summaryTblId, report, fileCacheKey, request, detailsModel, acceptOneItem, highlightedRowIdx, entryType);
```
robyww, line 73: Same with determineLoadType, just pull out what you need in the function

src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -49,31 +49,39 @@ const tableWarning = 'Only loading the image(s), ignoring any selected table(s).
...
-            sendRegionRequest(fileCacheKey, currentReport);
+            sendRegionRequest(fileCacheKey, report);
             return true;
 
         case LOAD_MOC:
             const mocMeta= {[MetaConst.PREFERRED_HIPS]: getAppHiPSForMoc()};
             if (request.mocOp==='table') mocMeta[MetaConst.IGNORE_MOC]='true';
             //loadToUI = true if request.mocOp==='table', else loadToUI=false
-            sendTableRequest(tableIndices, fileCacheKey, Boolean(request.tablesAsSpectrum==='spectrum'), currentReport, Boolean(request.mocOp==='table'), mocMeta);
+            sendTableRequest(tableIndices, fileCacheKey, Boolean(request.tablesAsSpectrum==='spectrum'), report, Boolean(request.mocOp==='table'), mocMeta);
```
robyww, line 84: I did not catch this in the last PR, you don't need to wrap a `===` with Boolean, it is redundant. The result of a `===` is a boolean. You should fix that though out the swtich.

src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -49,31 +49,39 @@ const tableWarning = 'Only loading the image(s), ignoring any selected table(s).
 
 export function resultSuccess(request) {
 
-    const {currentReport, currentDetailsModel, summaryModel, groupKey, acceptList, uniqueTypes}= request.additionalParams ?? {};
+    const {report, detailsModel, summaryModel, groupKey, acceptList, uniqueTypes,
```
robyww, line 52: I like `report` better than `currentReport` and `detailsModel` better than `currentDetailsModel`

robyww (changes requested):
I have a few code easily fixed comments, but  I could not test it, because I could not make it come up in single upload mode.

kpuriIpac:
> I have a few code easily fixed comments, but I could not test it, because I could not make it come up in single upload mode.

thanks @robyww, making the changes. Sorry I missed the single upload thing on firefly-dev - works locally for me with the same code, but not on the test build for some reason. Looking into it, will re-build after making the code changes and test again.

robyww (approved):

PR #1322, opened 2023-02-17
FIREFLY-1146: Fix target field not updating when searching at selected region in image viewer

jaladh-singhal:
Fixes [FIREFLY-1146](https://jira.ipac.caltech.edu/browse/FIREFLY-1146)

Since FIREFLY-250 was fixed, target is stored in its own field group. So updated the group key where target is set through intiArgs (by search actions)

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1146-image-search-selection-fix/firefly/

when you: show an image -> select an area -> click on binoculars dropdown -> Search FITS/HiPS at region center, it should bring up ImageSearchPanel with target set to the coordinates of the center of my selection (as shown in the dropdown in bold)


robyww (approved):
thanks

PR #1323, opened 2023-02-17
Firefly-1140: Add side bar that list the whole registry

robyww:
#### [Firefly-1140](https://jira.ipac.caltech.edu/browse/FIREFLY-1140): Add side bar with whole registry

 - Add side bar using table
 - Any data set can be select

#### Testing
  - These are URL's that can be used to launch IrsaViewer from atlas type web pages.
  - Try different URLs:
      - https://firefly-1140-atlas-sidebar.irsakudev.ipac.caltech.edu/irsaviewer/?api=dl&id=iras_issa
      - https://firefly-1140-atlas-sidebar.irsakudev.ipac.caltech.edu/irsaviewer/?api=dl&id=wise_z0mgs
      - https://firefly-1140-atlas-sidebar.irsakudev.ipac.caltech.edu/irsaviewer/?api=dl&id=herschel_phpdp&showChooser
      - https://firefly-1140-atlas-sidebar.irsakudev.ipac.caltech.edu/irsaviewer/?api=dl&id=msx&showChooser=true
 - look at the URL doc
      - https://firefly-1140-atlas-sidebar.irsakudev.ipac.caltech.edu/irsaviewer/?api=dl

### DIFF
--- a/src/firefly/js/api/webApiCommands/DatalinkUICommands.js
+++ b/src/firefly/js/api/webApiCommands/DatalinkUICommands.js
@@ -19,9 +19,16 @@ const idExamples= [
         }
     },
     {
-        desc:'load to UI more than one id',
+        desc:'load to UI using id',
         params:{
-            id: ['iras_issa','iras_iris']
+            id: 'wise_z0mgs',
+        }
+    },
+    {
+        desc:'load to UI using id and show the data set chooser',
+        params:{
+            id: 'dss',
+            showChooser: 'true'
         }
     },
 ];
@@ -57,8 +64,9 @@ function loadDLUITable(cmd,params, includeIdType) {
     }
 }
 
-async function getServerAndLoadTablesById(cmd,params) {
+async function getServerAndLoadTablesById(cmd,inParams) {
     const urlRootAry= await getJsonProperty('inventory.serverURLAry');
+    const params= {showChooser:false, ...inParams};
     const initArgs={urlApi:params};
     const urlRoot= urlRootAry[0];
     const makeUrl= (id) =>  urlRoot +'?'+ new URLSearchParams({collection:id}).toString();
@@ -80,9 +88,11 @@ export function getDatalinkUICommands(includeIdType) {
     const execute= (cmd,params) => loadDLUITable(cmd,params,includeIdType);
     const overview= [ 'Load Service Descriptor UI' ];
 
-    const urlOnlyParameters= { url: {desc:'url to datalink UI file'},
+    const urlOnlyParameters= {
+        url: {desc:'url to datalink UI file'},
         [ReservedParams.POSITION.name]: ['coordinates of the search',...ReservedParams.POSITION.desc],
         [ReservedParams.SR.name]: ['radius of search  (optional)',...ReservedParams.SR.desc],
+        showChooser: {desc:'show the dataset chooser'},
         execute: 'true or false - if true execute the search'
     };
     const bothParameters= {...urlOnlyParameters, id : {desc:'ID to predefined data link UI service'}};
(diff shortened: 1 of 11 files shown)


lrebull:
my $0.02 of reactions, which may or may not be what you are looking for:
1. Oh, cool! exciting to see this as complete as it is!! :) 
2. You might not want to know how long I looked for the 'search' button and literally did not find it. (Hint: a really long time. Like, I'd drafted a mail to you asking when the tool was going to be searchable.)
3. In so many of our other tools, the 'back' browser arrow doesn't work. It seems weird to me that now it does, and that is the only way we're expected to go back to the main search screen. It seems like either (a) the search from the search screen should spawn a new tab or window, OR (b) there should be a 'back' button available from the search results screen.
4. We (my team) need to figure out for you the right words and the right link for the documentation in the lower right of the search screen. I don't think it will be "front page" but it might be pointing to the same place you have pointed (the overview.html pages). 
5. Coverage in HiPS map isn't working, by which I mean that the MOC cells that it is overplotting on the HiPS image aren't accurately reflecting the coverage of the data set in question. (Then, see next item, which is an error.) (Or have I misunderstood what it is doing? I expected that the MOC coverage overplotted on the background corresponds to the selected data set.)
6. I can easily request data that is not included in the region covered by the data set, and then it doesn't ever come back, even with an error. Eep, no, it did just come back, but evidence suggests it is timing out rather than erroring out, because it took more than 5 minutes.
7. Default column widths for search results seem strange to me. example: s_region consistently way too narrow; facility_keywords, instrument_keywords, and target_keywords always empty and way, way, way too wide.
8. I tried IRIS and got a data table back at the bottom but "Error-Search for Data product failed" in the image window. (a) weird capitalization on that; (b) why did it fail?
9. I need to get you updated information for that table of data sets. Angela has an email into me; I'll work on that next.

robyww:
@lrebull I made a new build, I clean up the search button layout. It is more consistent with the other panels now. I also made the data set chooser to hide and show.

lrebull:
I am not sure what changed with respect to the search button layout -- for me, it is still in the lower left, so I still would have missed it since I think that's where it was yesterday..? I think?

I like the hide/show data list, but I still can't get it back after doing my first search without using the browser back button, which seems weird.

What is the MOC that is overlaid on the HiPS image? Because it doesn't seem to correspond to the coverage of the selected data set.

Now, when I request data that don't exist, at least it comes back quickly with 'no data found' as opposed to the timeout I was getting yesterday.

robyww:
@lrebull 
> I am not sure what changed with respect to the search button layout -- for me, it is still in the lower left, so I still would have missed it since I think that's where it was yesterday..? I think?

The search button is in the lower left for most everything, but did clean up the layout some, it needed some padding. It looks like the image, tap, catalog, etc

>What is the MOC that is overlaid on the HiPS image? Because it doesn't seem to correspond to the coverage of the selected data set.

I have not done anything with the MOCs yet beyond put the up, the all sky MOCs don't really make sense anyway so I think we will remove them. We I get some better MOC test data I plan to add and remove them as the data set changes.

> I like the hide/show data list, but I still can't get it back after doing my first search without using the browser back button, which seems weird.

How we handle browser back button behavior has not changed with this work.

lrebull:
noting here so that i don't forget that we need to remember to call that column something other than "coverage" because that means someting else even in our context. program type? or something else. need better words.

lrebull:
Worth also taking to heart the suggestion of making sure that the results screen shows what data set you searched on.

robyww:
@lrebull I plan to make a separate ticket for all these loose ends and any bugs or changes on the UI side

PR #1325, opened 2023-03-06
IRSA-5077: Broken 'info' and 'Column Def' links under catalog listing

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-5077

`info` and `ColumnDef` links are broken in IRSA ops.
- Also fixed broken 'Table Selection' links under `name` column
This happens on first load as well as when `Column Def` does not exists

Test URL: https://irsa-5077-info-link.irsakudev.ipac.caltech.edu/irsaviewer/
This is built with `ops` env and uses IRSA ops backend.  It should appear as if it's in ops.

To test when there's no `ColumnDef`, select project = `IRAS`.


ejoliet (approved):
Both issues were tested and are fixed.
Thank you!!

PR #1326, opened 2023-03-09
Firefly-1152: Clean up "Click to Search"

robyww:
#### [Firefly-1152](https://jira.ipac.caltech.edu/browse/FIREFLY-1152): Clean up "Click to Search"

 - FIX: click to search on tap IrsaViewer: it does not set the correct tab
 - Change ellipse selection to a circle, then make it the default (per CCB approval)
 - Reorder menu for polygon or cone so they are at the top, have a message saying inscribed circle
 - Change Search Refinement Tool layer to not always go away with the dialog
 - Leave search refinement tool layer up
 - While layer is up show the click-to-action dropdown
 - Make search dropdown stay visible, make a way to dismiss the layer.
 - When layer up and Search refinment tool not up then the user cannot control it, just dismiss
 - In search refinement tool: put message up for additional searches becoming available when switch between code and polygon

Testing:

 - https://firefly-1152-click.irsakudev.ipac.caltech.edu/irsaviewer
 - Test according to the ticket: [Firefly-1152](https://jira.ipac.caltech.edu/browse/FIREFLY-1152)


lrebull:
- yes, click to search goes properly into TAP catalog search 
- yes, circle region is first.
- yes, click to draw circle region now goes from the target out (as opposed to 'corners' of the circle region)
- click to draw region, red text is "end select"; once you click 'refine search' then the red text changes to "end search marker" and doesn't go away when the refine selection region window goes away. i think that is as designed? (Is that "Change Search Refinement Tool layer to not always go away with the dialog"?)
- in search refinement dialog, when cone is selected, it says "Polygon searches are also available" (that's good). When polygon is selected, it says "Cone and searches are also available" (that's bad). It should be at least "Cone searches are also available".  That bullet in the description says something about additional searches being available when switching between cone and polygon. I see in the pulldown from the binoculars that it changes to tell you explicitly "polygon actions" and "cone actions" -- is that what this is referring to? or is this something else?


the more i ponder, the less sure i am what any of these mean:
- Leave search refinement tool layer up
- While layer is up show the click-to-action dropdown
- Make search dropdown stay visible, make a way to dismiss the layer.
- When layer up and Search refinment tool not up then the user cannot control it, just dismiss

robyww:
@lrebull The ticket was mostly my note taking from out meeting so it is not that clear. However, your interpretation is correct.

> and doesn't go away when the refine selection region window goes away. i think that is as designed?

Yes, that is by design, however we can change the text

> should be at least "Cone searches are also available"

I will fix that.

>  I see in the pulldown from the binoculars that it changes to tell you explicitly "polygon actions" and "cone actions" -- is that what this is referring to? or is this something else?

yet that is what it is referring to

I tried to clean up the help text layout, I am happy to modify it .

lrebull:
Screen Shot 2023-03-10 at 1 37 36 PM
still says "cone and searches"...

PR #1330, opened 2023-03-14
FIREFLY-1153: Derived Columns Updates

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1153
Implemented all 1-10.

Additionally, implemented Future Upgrade: 1, 2, and 4.
- edit and delete derived columns
- meta-computed-column: filtered, selected, row number
- table save:  as displayed or original

Test URL: https://fireflydev.ipac.caltech.edu/firefly-1153-derived-column-updates/firefly

Special cases:
- If you rename or delete a column that's used by sorting, sort order will reset.  The same applies to filters.  
- When a table is filtered and `Row Number` preset is selected, the rows excluded by the filter will be assigned `null` instead of a row number.

Newly added items requiring CCB feedbacks:
- Mode: Custom/Preset  (need help with wording)
- Under Preset
   - Filtered Rows / Selected Rows / Row Number  (need help with wording)
- 'Delete column' is added to the 'Edit' popup.  This is to save space in the TableOptions column name column.
- Table Save:
  - need help with wording on all of the items added. i.e. radio button labels, messages, etc.


vandesai1:
1. "Enter expression" & "Use preset function"
2. Change menu choices to "Set filtered rows to ‘true’ and the rest to ‘false’", "Set selected rows to ‘true’ and the rest to ‘false’", and "Number rows in current sort order"
3. We like this choice. One issue is that every time I delete a column, "Table Options" scrolls back to the top of the table. I think it should do what I said but then keep me in the same place in the table so I can see that the column disappeared.
4a. “Save original table” --> “Save table as originally retrieved”
and change the explanatory note to “The table will be saved in the form originally retrieved into this tool, without filtering, sorting, or any additional columns.“
4b. 
“Save table as displayed” stays as-is BUT change the explanatory note to “The table will be saved in its current state, including its sorting order and derived columns, but excluding rows not accepted by any filters applied, as well as any hidden columns.”

gpdf:
For the record (I know the developers are already aware of this):

* We requested that the number-rows function produce numbers starting from 1, not 0, to match the way the table component itself talks about row counts (in its header), and to mitigate the risk of users conflating "null" and 0 when downloading and further processing a table with such a column.

loitly:
Updated PR with all of the requested changes.  I was also able to resolve the performance issue as well by creating indexes.
Test URL updated with the latest changes.
https://fireflydev.ipac.caltech.edu/firefly-1153-derived-column-updates/firefly

Let me know if you have any other request, otherwise I will merge tomorrow.  @lrebull @gpdf @vandesai1

vandesai1:
1. Mode: Custom/Preset (need help with wording): "Enter expression" & "Use preset function"

- Looks good!

2. Under Preset - Filtered Rows / Selected Rows / Row Number (need help with wording): Change menu choices to "Set filtered rows to ‘true’ and the rest to ‘false’", "Set selected rows to ‘true’ and the rest to ‘false’", and "Number rows in current sort order"

- Looks good!

3. 'Delete column' is added to the 'Edit' popup. This is to save space in the TableOptions column name column: We like this choice. One issue is that every time I delete a column, "Table Options" scrolls back to the top of the table. I think it should do what I said but then keep me in the same place in the table so I can see that the column disappeared.

- Looks good!

4. Table Save: need help with wording on all of the items added. i.e. radio button labels, messages, etc.
4a. “Save original table” --> “Save table as originally retrieved”
and change the explanatory note to “The table will be saved in the form originally retrieved into this tool, without filtering, sorting, or any additional columns.“

- Looks good!

4b. “Save table as displayed” stays as-is BUT change the explanatory note to “The table will be saved in its current state, including its sorting order and derived columns, but excluding rows not accepted by any filters applied, as well as any hidden columns.”

- Looks good!

5. We requested that the number-rows function produce numbers starting from 1, not 0, to match the way the table component itself talks about row counts (in its header), and to mitigate the risk of users conflating "null" and 0 when downloading and further processing a table with such a column.

- Looks good!

6. Newly-added derived columns should be analyzed for whether a category-column filter menu should be offered to the user.

- It works for types long and char, but not double. I think that's ok.

PR #1335, opened 2023-03-23
FIREFLY-1141: column chooser popup

kpuriIpac:
#### [Firefly-1141](https://jira.ipac.caltech.edu/browse/FIREFLY-1141):
- added the ability to allow user to choose upload columns when uploading a table
- **Updates 3/23**
  - changed UI to add a link to the columns description to launch the column chooser dialog as requested @robyww 
  - changed code to keep ColSelectView simple as discussed with @loitly. if multiSelect is true, send back an array of col names, if it's false, then send back the selected col name as before. 
  - testing link updated 

#### Testing:  
  - https://fireflydev.ipac.caltech.edu/firefly-1141-column-chooser/firefly
  - go the TAP panel, select source as IRSA or CADC 
  - Under constraints, select Spatial Type as "**Multi-Object**"
  - Either upload a tbl file or load an existing table
  - You should see a new button under "**Position Columns**" titled "**Choose Columns**"
  - Click on it to make a selection of columns
  - Click on "Poupulate and edit ADQL" to ensure your selections show up here
  - Click on Search to ensure search works as expected with your column selections..

### DIFF
--- a/src/firefly/js/charts/ui/ColSelectView.jsx
+++ b/src/firefly/js/charts/ui/ColSelectView.jsx
@@ -7,7 +7,7 @@ import {dispatchTableRemove}  from '../../tables/TablesCntlr';
 import {clone} from '../../util/WebUtil.js';
 
 import {TablePanel} from '../../tables/ui/TablePanel.jsx';
-import {getTblById, calcColumnWidths} from '../../tables/TableUtil.js';
+import {getTblById, calcColumnWidths, getSelectedData, getColumnValues} from '../../tables/TableUtil.js';
 import {dispatchShowDialog, dispatchHideDialog, isDialogVisible} from '../../core/ComponentCntlr.js';
 import CompleteButton from '../../ui/CompleteButton.jsx';
 import {quoteNonAlphanumeric}  from '../../util/expr/Variable.js';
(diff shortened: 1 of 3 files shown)


src/firefly/js/charts/ui/ColSelectView.jsx
```diff
@@ -154,6 +156,30 @@ function renderCloseAndHelpButtons(tblId,onColSelected,buttonText,popupId) {
...
+
+    for (const [idx, entry] of data.entries()) {
+        if (selectInfo.selectAll) { //ignore selectInfo.exceptions
+            //go through data - for each idx NOT in selectInfo.exceptions, grab data[currIdx][5] - 5 is the idx in uploadInfo.columns
+            if (!(selectInfo.exceptions).has(idx)) selectedIndices.push(data[idx][5]);
+        }
+        else { //selectInfo.exceptions are the indices we want to select
+            if ((selectInfo.exceptions).has(idx)) selectedIndices.push(data[idx][5]);
+        }
+    }
```
kpuriIpac, line 173: I may need to change this part. There is probably a better way of getting the selected columns info (I did it this way for now to get around the user filtering the columns and then making their selections).

kpuriIpac:
Should the size and style of "**Choose Columns**" button be different? It being the same size as the "**Change...**" button to change the upload table might be a bit confusing. It probably needs to be smaller to be more in line with the "**Position Columns**" right above it? Let me know if you have any thoughts @loitly @robyww 

choose_columns

kpuriIpac:
> Should the size and style of "**Choose Columns**" button be different? It being the same size as the "**Change...**" button to change the upload table might be a bit confusing. It probably needs to be smaller to be more in line with the "**Position Columns**" right above it? Let me know if you have any thoughts @loitly @robyww
> 
> choose_columns

and since we use the magnifying glass for columns anyway, maybe "Choose Columns" could just be plain text instead of a button, with a magnifying glass icon next to it launching the column selector dialog instead...

loitly (changes requested):
Please move upload logic out of ColSelectView.

src/firefly/js/charts/ui/ColSelectView.jsx
loitly: Yes, there is.  `firefly.util.table.getSelectedData`

robyww (changes requested):
Before I look to closely at the code I wanted to get the UI right.
- Removed: the Choose columns button.
- Change: The column descriptions to a link.
```js
<a className='ff-href'onClick={() => thePopup()}}>
    <the-stuff-here-aorund-line-250>
</a>
```

Visual of what I mean:

x

loitly (approved):
Thank you.  Looks good.

robyww (approved):
Looks good!  Make the clean up and merge.

PR #1338, opened 2023-03-27
Firefly-1190: optimize byte data loading based on zoom

robyww:
[Firefly-1190](https://jira.ipac.caltech.edu/browse/FIREFLY-1190): optimize byte data loading based on zoom

 - byte data compression size is based on data size and zoom level
 - less compress data is not loaded until the user zooms up.

difficult to test. just review the code

### DIFF
--- a/src/firefly/js/visualize/rawData/RawDataOps.js
+++ b/src/firefly/js/visualize/rawData/RawDataOps.js
@@ -19,6 +19,8 @@ import {Logger} from 'firefly/util/Logger.js';
 const nextColorChangeParams= new Map();
 const colorChangeDonePromises= new Map();
 const imageIdsRequested= new Map();
+const QUARTER_ZOOM_FACT= .15;
+const HALF_ZOOM_FACT= .15;
 
 
 /**
@@ -211,10 +213,12 @@ function clearLocalStretchData(plot) {
     entry.dataType= CLEARED;
 }
 
+// keep code around
 // function isNoisyImage(plot) {
 //     if (isNaN(plot?.webFitsData?.[Band.NO_BAND.value]?.largeBinPercent)) return false;
 //     return (!isThreeColor(plot) && plot.webFitsData[Band.NO_BAND.value].largeBinPercent>.03);
 // }
+
 
 /**
  * @param {WebPlot} plot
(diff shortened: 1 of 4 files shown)


src/firefly/js/visualize/rawData/RawDataOps.js
```diff
@@ -275,40 +279,48 @@ export async function loadStretchData(pv, plot, dispatcher) {
...
     const dataCompress= getFirstDataCompress(plot,mask);
-    const dataSize= plot.dataWidth*plot.dataHeight;
     const {success:firstSuccess, fatal}= await loadStandardStretchData(workerKey, plot,
                   {dataCompress, backgroundUpdate:false, checkForPlotUpdate:!mask}, maskOptions);
 
     if (plotInvalid()) return;
     if (firstSuccess) {
         dispatcher({ type: ImagePlotCntlr.BYTE_DATA_REFRESH, payload:{plotId, imageOverlayId, plotImageId}});
-        if (dataCompress==='FULL') return;
+        if (dataCompress===FULL) return;
```
loitly, line 288: Looks like regardless of the condition, it will 'return' because there's nothing else to do after it leaves the 'if' block.

loitly (approved):
I agree it's hard to review.  The logic of its execution flow could use a few lines of comments if it's intended to be read by another developer.

src/firefly/js/visualize/rawData/RawDataOps.js
```diff
@@ -19,6 +19,8 @@ import {Logger} from 'firefly/util/Logger.js';
 const nextColorChangeParams= new Map();
 const colorChangeDonePromises= new Map();
 const imageIdsRequested= new Map();
+const QUARTER_ZOOM_FACT= .15;
+const HALF_ZOOM_FACT= .15;
```
loitly, line 23: They are the same value, is that right?

robyww: oops

robyww: it should .42

src/firefly/js/visualize/rawData/RawDataOps.js
robyww: yes, I can remove the line, intellij is tell me the same thing.

PR #1339, opened 2023-03-27
FIREFLY-1061: Fix "Page size" appearing in client-table gears dialogs

jaladh-singhal:
Fixes [FIREFLY-1061](https://jira.ipac.caltech.edu/browse/FIREFLY-1061)

- ~Simply removed the Page size element~ remove page element when it is equal to max rows

## Testing 
https://fireflydev.ipac.caltech.edu/firefly-1061-remove-page-size/firefly/?__action=layout.showDropDown&

Go to TAP, click gear icon in "Output Column Selection and Constraints" table, there shouldn't be any page size in top right and everything else should work as expected.

Do a catalog search, the page size should appear on clicking gear icon

### DIFF
--- a/src/firefly/js/tables/TablesCntlr.js
+++ b/src/firefly/js/tables/TablesCntlr.js
@@ -378,6 +378,8 @@ function fixClientTable(tableModel) {
         tableModel = TblUtil.cloneClientTable(tableModel);
     }
 
+    if (!tableModel?.request?.pageSize) set(tableModel, 'request.pageSize', MAX_ROW);
+
     return tableModel;
 }
 
--- a/src/firefly/js/tables/ui/TablePanelOptions.jsx
+++ b/src/firefly/js/tables/ui/TablePanelOptions.jsx
@@ -28,6 +28,7 @@ import {showAddOrUpdateColumn} from './AddOrUpdateColumn.jsx';
 import {BY_SCROLL} from './BasicTableView.jsx';
 
 import EDIT from 'html/images/16x16_edit_icon.png';
+import {MAX_ROW} from 'firefly/tables/TableRequestUtil';
 
 
 export const TablePanelOptions = React.memo(({tbl_ui_id, tbl_id, onChange, onOptionReset, clearFilter}) => {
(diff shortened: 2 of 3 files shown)


src/firefly/js/tables/ui/TablePanelOptions.jsx
```diff
@@ -163,7 +164,7 @@ function Options({uiState, tbl_id, tbl_ui_id, ctm_tbl_id, onOptionReset, onChang
                 </div>
             </div>
             <div>
-                {showPaging &&
+                {showPaging && pageSize!==MAX_ROW &&
```
robyww, line 167: so `MAX_ROW` might not be enough. it might need to be something like `pageSize>=number of rows in table`
Many client tables are not using MAX_ROW they are using something big like 10000.
@loitly what do you think.

loitly: If there's not many places, I would prefer to change them to MAX_ROW.  Using `pageSize >= total row` is not quite the same.  You can have a table with 80 rows while the default pageSize of 100.  In that case, you would still want them to be able to change page size.

loitly: We can also default pageSize to MAX_ROW when not set for client tables.  I think I may have done this already.  But, if not, it's something we can add.

robyww: ok, maybe @jaladh-singhal can track them down

jaladh-singhal: Ok I'll give it a try

src/firefly/js/tables/TablesCntlr.js
```diff
@@ -378,6 +378,8 @@ function fixClientTable(tableModel) {
         tableModel = TblUtil.cloneClientTable(tableModel);
     }
 
+    if (!tableModel?.request?.pageSize) set(tableModel, 'request.pageSize', MAX_ROW);
```
jaladh-singhal, line 381: @robyww this was additionally suggested by @loitly

robyww (approved):

PR #1342, opened 2023-03-30
Firefly-1192: client side obs core packaging

robyww:
#### Testing
This PR is only client side work, it can't be tested until [Firefly-1193](https://jira.ipac.caltech.edu/browse/FIREFLY-1193) is completed


loitly (approved):
Looks good.

PR #1348, opened 2023-04-05
FIREFLY-1183: Combine spectra cleanup

loitly:
Work ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1183
Design ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1171

Notable changes:
- thumbnail view for charts while in normal grid mode
- add show/hide errors checkbox
- remove side-by-side view

All items from the tickets were implemented except for a couple noted in `Implementation notes:` of the work ticket.

Test: 
https://fireflydev.ipac.caltech.edu/firefly-1183-spectra-cleanup/firefly
https://firefly-1183-spectra-cleanup.irsakudev.ipac.caltech.edu/irsaviewer
Follow design ticket to ensure all items are fixed.

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -444,7 +444,6 @@ div.rootStyle img {
 
 
 .text-ellipsis {
-    display: inline-block;
     white-space: nowrap;
     text-overflow: ellipsis;
     overflow: hidden;
--- a/src/firefly/java/edu/caltech/ipac/table/DataGroupPart.java
+++ b/src/firefly/java/edu/caltech/ipac/table/DataGroupPart.java
@@ -16,7 +16,7 @@ public class DataGroupPart implements HasAccessInfos {
 
     public static final String LOADING_STATUS = "Loading-Status";
 
-    public static enum State {COMPLETED, INPROGRESS, FAILED;
+    public enum State {COMPLETED, INPROGRESS, FAILED;
                             public String toString() {
                                 return StringUtils.pad(20, name());
                             }
@@ -26,7 +26,7 @@ public class DataGroupPart implements HasAccessInfos {
     private int startRow;
     private int rowCount;
     private String hasAccessCName = null;
-    private String errorMsg;
+    private Status status;
 
     public DataGroupPart() {
     }
@@ -69,9 +69,12 @@ public class DataGroupPart implements HasAccessInfos {
         this.startRow = startRow;
     }
 
-    public String getErrorMsg() { return errorMsg; }
+    public String getErrorMsg() { return status != null && status.isError() ? status.message : null; }
 
-    public void setErrorMsg(String errorMsg) { this.errorMsg = errorMsg; }
+    public void setErrorMsg(String errorMsg) { setStatus(new Status(500, errorMsg)); }
+
+    public Status getStatus() { return status; }
+    public void setStatus(Status status) { this.status = status; }
 
 //====================================================================
 //  implements HasAccessInfos
@@ -92,5 +95,7 @@ public class DataGroupPart implements HasAccessInfos {
 //====================================================================
 //
 //====================================================================
-
+    public record Status (int code, String message) {
+        public boolean isError() { return code < 200 || code >= 400; }
+    }
 }
--- a/src/firefly/js/charts/dataTypes/FireflyGenericData.js
+++ b/src/firefly/js/charts/dataTypes/FireflyGenericData.js
@@ -329,7 +329,7 @@ export function addScatterChanges({changes, chartId, traceNum, tablesource, tabl
     }
     const hasXErrors = xErrLow.length > 0 || xErr.length > 0;
     const xErrorType = get({fireflyData}, errorTypeFieldKey(traceNum, 'x'));
-    changes[`data.${traceNum}.error_x.visible`] = hasXErrors && xErrorType !== 'none';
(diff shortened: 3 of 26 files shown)


src/firefly/js/ui/ToolbarButton.jsx
```diff
@@ -12,9 +12,9 @@ import CHECK_BOX from 'html/images/black_check-on_10x10.gif';
 import BrowserInfo, {Platform} from 'firefly/util/BrowserInfo.js';
 
 
-export function makeBadge(cnt) {
+export function makeBadge(cnt, style={}) {
     const cName= `ff-badge ${cnt<10 ? 'badge-1-digit' : 'badge-2-digit'}`;
-    return <div className={cName}>{Math.trunc(cnt)}</div>;
+    return <div style={style} className={cName}>{Math.trunc(cnt)}</div>;
```
robyww, line 17: I did this same modification in my PR. I sure you will notice when you rebase.

robyww (approved):
Look's Good (2 bugs)

- Since you are working on it anyway, would `PlotlyChartArea` be an candidate to convert to a functional component? I looked over it it does not look to bad.
- I guess you already add it... but the pinned chart says "pinned images:"
- I really like the grid vs single mode.
- A bug: I am doing some extraction of any type and click on `pin chart/table` button. The table is pinned but the chart is not.

lrebull:
Sorry for the delay.. 

ok. item 1. units. I’m loading in a spitzer spectrum. On ops it plots “F_nu (Jy)” on the y axis and “nu (microns)” on the x-axis.  When ops combines it, it has the same units. On your test build, it has the same units. That's not right - astronomers use nu for frequency and lambda for wavelength, which is in microns. so it should be "lambda (microns)". Just for added complications, in the ticket, it says something about combining them into an x-axis in Hz, which I cannot reproduce, on ops or your test build. (I'm checking with Vandana.)

regarding item 1, when we met earlier this week, you said that if there were no units, you were assuming units, which is why the spitzer spectra didn't work -- they didn't have units. but i have loaded in a spitzer spectrum to ops and your build, and they both behave the same way, or at least they both do now. i thought they behaved differently in your test build when we talked on zoom earlier this week. 

when we did talk on zoom earlier this week, when you tried to combine the spitzer spectra, it didn't give you a choice of the spitzer spectra with the tickboxes because it knew they had no units, so it didn't give them to you as a choice. now, it gives them to me as a choice, but then won't actually combine them -- see screenshot. It would be better if it told me that it couldn't combine them because it didn't understand the units, and better still if it asked me for help in understanding the units. (Like the time series tool tries to guess at the time column but failing that, gives me an opportunity to tell it the time column. That's just my $0.02, and that may be beyond the scope.) (I'm checking with Vandana about whether or not refusing to combine spectra without units is what she wants.)
Screen Shot 2023-04-07 at 8 56 37 PM


item 2 the many pinned plots with improved margins works great. the little badge counter works great too.

item 3 removing side by side button, still show active chart even when just one - all seems good to me - am double checking with Vandana to be sure removing side by side button is what she wanted.

item 4 moved pin chart button down - looks/works great

item 5 combine and show table down - looks/works great

item 6 full screen preserves mode - works great

item 7 remove side by side mode, counter for pinned plots - looks/works great

item 8 changed pop-up title - looks/works great though I can't currently combine charts (as per above)

item 9 no compatible charts found - i can't trigger this.

item 10 pop up stuff - having trouble seeing what the complaints are in the ticket on ops (maybe partially implemented already there??) but new seems ok even if i can't actually merge plots. 

item 11 order collapsing- can't reproduce this or test it (see screenshot above)

item 12 error easier to turn off - looks/works great (i need to remember to propagate this through the docs)

ok. quitting here. will let you know what i hear from Vandana on Monday. Let me know if you want movies to see what it is doing for me on the first parts above.

loitly:
> ok. item 1. units. I’m loading in a spitzer spectrum. On ops it plots “F_nu (Jy)” on the y axis and “nu (microns)” on the x-axis. 
@lrebull can I get your test files?  I can't reproduce with my files.

lrebull:
it won't let me attach tbl files here! i sent them in email.

lrebull:
OK, latest from Vandana - she agrees we should not be assuming units if the tool cannot figure them out. She also agrees with the idea of prohibiting users from combining spectra if it can't understand the units. She's also good with removing the "side by side" choices for pinned plots.  (And she agrees it should be "lambda (microns)" not "nu (microns)".)

loitly:
> OK, latest from Vandana - she agrees we should not be assuming units if the tool cannot figure them out. She also agrees with the idea of prohibiting users from combining spectra if it can't understand the units. She's also good with removing the "side by side" choices for pinned plots. (And she agrees it should be "lambda (microns)" not "nu (microns)".)

`microns` is not on our list of units for wavelength.  
If the unit is one of `A`, `nm`, `um`, `mm`, `cm`, or `m`, then we will show it as `λ [${unit}]`.  
Otherwise, we will use '𝛎 [${unit}]', or '𝛎 [Hz]' if unit is undefined.  
I understand you want to remove these fallback assumptions.  What should I show as the axis label then?
column_name[$unit]?

lrebull:
Ah, ok.
Yes, column_name[$unit] if the units are undefined sounds appropriate.

loitly:
@robyww @lrebull, test URLs updated with all of your requested changes.  Final check before I merge?

lrebull:
It seems to be working. I am a little thrown by images I load in being counted as "pinned images" but I guess that's right.

PR #1351, opened 2023-04-07
Firefly 1193 ObsCore Packager

kpuriIpac:
#### [Firefly-1193](https://jira.ipac.caltech.edu/browse/FIREFLY-1193):
- ObsCore Packager (similar to Sofia) 
- The naming convention I have used is: 
   - if it's a non-datalink file, get the filename from the access_url 
   -  Prepend the file with the product title template entries, or if that's not available, prepend it with '`${obs_collection}-${instrument_name}-${obs_id}`' 
   - For datalink files, I have created a folder with the name '`${obs_collection}-${instrument_name}-${obs_id}`' (or productTitleTemplate) 
      - and we get filename from access_url for every file in the datalink table that we want to download
- _Todo_: I need to finish implementing the logic to use the product title template (which comes from `getAppOptions().tapObsCore?.productTitleTemplate`) and I am sending it to the server via the search request, just need to finish writing the logic for that.  
   - Once this is done, productTitleTemplate will describe what should prepend the filename (or in the case of datalink, what should be the name of the folder)  
- _Todo_: I haven't accounted for duplicate file names. The user's OS might append (1), (2), etc. to duplicate file names but we should take care of it on our end, right? 
- _Todo_: minor, but I can add a check so that if there's only 1 file that we want to download from a datalink table, don't create a folder for it 
- **_Issue_**: I have set the BaseFileName to "ObsCore_Files", which what you should see in the DownloadOptionPanel as well. But when the file is download, it's downloaded as "ObsCore_Files_part1"...I am not sure where "part1" comes from, looking into it. 

**Updates 4/12**
- PR ready
- addressed feedback below from @loitly and wrote file naming code based on discussions with @robyww and @loitly 
- I've added comments through the code to better explain what different parts do (may remove some if there's too many), let me know if you have any questions 
- The logic that was used for file naming was (the getFileName function): 
   - if a productTitleTemplate is available from the client side, use it
   - else, use obs_title to name the file if it's available 
   - else, create a manual template (based on obs_collection, instrument_name and obs_id) 
- added checks for access_url and access_format/content_type (to get file extension) being null 
- BaseFileName (the zip folder name) is also changed to `tbl_id_Files` 
- test builds and instructions updated below 

**Updates 4/13**
- addressed PR comments, highlights:
  - for both file and folder names (and also the table title - which is the name of the downloaded zip folder), got rid of spaces and non-(alphanumeric,period,hyphen,underscore) characters and replaced the latter with either hyphen or underscore as appropriate (hyphen between column names, and underscore for any weird characters within a column name) 
  - creating folder for datalink table only if it has more than one valid file 
  - other small changes and bug fixes 
- _**Potential issue**_:  for some entries, I was getting this error msg in the Data Product viewer: 
    - `No displayable data available for this row: Request failed with status 500: https://fireflydev.ipac.caltech.edu/firefly-1193-package-obscore/firefly/CmdSrv/sync?cmd=upload`
    - to reproduce this error, search for TAP -> CADC -> m1, and under obs_id, search for 2827296
    - also, if you try and download this datalink entry, it succeeds but creates a folder even though there's only one file in the folder. The reason is that the datalink table still has more than 1 valid files (2 `previews` and 1 `this` specifically), but there was an error encountered in packaging these 3 files (see the README file)
    - I am not sure if this is an expected error or not, but wanted to highlight it. 
- both test builds below are updated with @robyww's  TAP UI changes as well

**Updates 4/14**
  - removed checks for uniqueness for file names, as @robyww's changes take care of that now 
  - only added checks for datalink folder name's uniqueness by calling FileUtil's `getUniqueFileNameForGroup` function
  - for non-datalink files, using a temp file to call URLDownload and using this temp file in the packager (instead of using the URL twice) 
     - for datalink files, we still need the access_url when creating the final FileInfo object to be packaged, but this only happens once per datalink entry anyway  

#### Testing:  
**Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1193-package-obscore/firefly
**Irsaviewer** **(DCE)**: https://firefly-1193-package-obscore.irsakudev.ipac.caltech.edu/irsaviewer/dce
- Search for something that returns at least some datalink entries. 
    - On firefly, from the TAP panel, select CADC as TAP service, search for m1 (set row limit to 5 or 10 for faster results).
    -  On DCE, you may click on the MSX collection, click on a point (or, use this: 2h44m14.43s +57d57m24.8s Equ J2000) and search
- Click on the "Prepare Download" button and attempt to download single or multiple files. For datalink tables with more than one valid file that we want to download, a sub-folder will be created for this datalink table. 
- attempt to download non-datalink files as well (jpeg, fits,etc) directly 
- **Testing uniqueness when it comes to naming files**
   - to test that file names are always unique, download a datalink table that has at least 2 "**preview**" entries (in the **semantics** column)
      - for this, in your datakink folder, you should get file names, for instance, that look like this: `obsCollection-instrumentName-obsID-preview.jpeg` and `obsCollection-instrumentName-obsID-preview-1.jpeg `


src/firefly/java/edu/caltech/ipac/table/ObsCorePackager.java
```diff
@@ -0,0 +1,110 @@
...
+            File f = new File(fileName);
+            String extName = f.getName(); //doFolders ? fileName : f.getName();
+            extName = prepend_file_name + "_" + extName;
+
+            FileInfo fileInfo = new FileInfo(access_url, extName, 0);
+            fileInfos.add(fileInfo);
+        }
+
+        HttpServiceInput authCreds = HttpServiceInput.createWithCredential();
+        fileInfos.forEach(fileInfo -> fileInfo.setRequestInfo(authCreds));
```
loitly, line 74: In most cases, `requestUrl` is needed to determine when credential is needed when `createWithCredential(requestUrl)` is called.  Try this...
```
    fileInfos.forEach(fileInfo -> fileInfo.setRequestInfo(HttpServiceInput.createWithCredential(fileInfo.getInternalFilename())));
```

src/firefly/java/edu/caltech/ipac/table/ObsCorePackager.java
```diff
@@ -0,0 +1,110 @@
...
+
+        return Arrays.asList(new FileGroup(fileInfos, null, 0, "ObsCore Download Files"));
+    }
+
+    public static List<FileInfo> parseDatalink(URL url, String prepend_file_name) {
+        List<FileInfo> fileInfos = new ArrayList<>();;
+        try {
+            File tmpFile = File.createTempFile("tmpFile", ".xml");
+            URLDownload.getDataToFile(url, tmpFile);
+            DataGroup[] groups = VoTableReader.voToDataGroups(tmpFile.getAbsolutePath(), false);
```
loitly, line 84: you can directly call `VoTableReader.voToDataGroups` with the `url` no need to create temp file.

loitly (changes requested):
Send to background does not work.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -0,0 +1,209 @@
...
+                    String accessUrl = String.valueOf(dg.getData("access_url", i));
+                    String sem = String.valueOf(dg.getData("semantics", i));
+                    String content_type = String.valueOf(dg.getData("content_type", i));
+
+                    if (accessUrl == null || sem == null) {
+                        //if only semantic (sem) is null, accessUrl may still be available, but we won't know which accessUrls ones to pick
+                        continue;
+                    }
+
+                    if (sem.equalsIgnoreCase("#this") || sem.equalsIgnoreCase("#thumbnail" ) ||
```
robyww, line 131: because of the standard and the way semantics work it  would be better to
```java
sem.toLowerCase().endsWith("#this") 
```
you might want to make that a function to make the code more readable, such as 
```java
    private static boolean testSem(String s, String val) {
       return (s!=null && s.toLowerCase().endsWith(val));  
    }
```

it is allowable to have a semantic that is something like  `https://my.edu/myCustomSemanitc#this`

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -51,15 +52,34 @@ function setupObsCorePackaging(tbl_id) {
...
+            <DownloadButton>
+                <DownloadOptionPanel {...{
+                    updateSearchRequest: updateSearchRequest,
+                    showZipStructure: false, //flattened files, except for datalink
+                    dlParams: {
+                            FileGroupProcessor:'ObsCorePackager',
+                            dlCutout: 'orig',
+                            TitlePrefix: 'ObsCore',
+                            help_id:'table.obsCorePackage',
+                            BaseFileName:`${tbl_id}_Files`}}}/>
```
robyww, line 82: this should not be the `tbl_id` 
```js
getTblById(tbl_id)?.title ?? 'unknown'
```

robyww (changes requested):
I think you are close...

- Clean up the code a little with my suggestions
- Bugs:
   - The folder should be the table name not the id
   - In the data like I think you are generating too many folders and misnaming the folders.
Screenshot 2023-04-13 at 12 43 26 PM

kpuriIpac:
> I think you are close...
> 
> * Clean up the code a little with my suggestions
> * Bugs:
>   
>   * The folder should be the table name not the id
>   * In the data like I think you are generating too many folders and misnaming the folders.
>     Screenshot 2023-04-13 at 12 43 26 PM



> I think you are close...
> 
> * Clean up the code a little with my suggestions
> * Bugs:
>   
>   * The folder should be the table name not the id
>   * In the data like I think you are generating too many folders and misnaming the folders.
>     Screenshot 2023-04-13 at 12 43 26 PM

Thank you! I'll work on fixing the issues, and I will use the table name instead of id for the folders. 
Can you help me re-create this datalink search @robyww? (from the screenshot) I want to try and look more closely as to why it's creating so many folders, and this seems like a good example to work with (because I didn't run into this issue with the tables that I was testing with).

loitly (approved):
I am deferring to Trey for reviewing.  It's better this way.
I do have a comment about file naming convention.
I think naming them #this-xxx and #preview-xxx, is unconventional. 
1.  I think they should be suffix instead of prefix
2. #this indicate the primary data.  no suffix is needed.
for example
```
abc.fits
abc-cutout.fits
abc-preview.png
``` 
Also, in order for the files to be used as input/output as well as to be valid on all common file systems, it's best to stick with the conventional `Use only alphabet characters, numbers, hyphens, underscores, and periods`. 
Just my 2 cents.

robyww:
@kpuriIpac the dataline search - TAP - CADC - target:m1 - then I checked and few and packaged.

robyww:
@kpuriIpac the dataline search - TAP - CADC - target:m1 - then I checked and few and packaged.

robyww:
@kpuriIpac don't include the `#` in the semantic in the filename.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -0,0 +1,209 @@
...
+            extName += "." + extension;
+        }
+        uniqueFileNames.add(extName);
+        return extName;
+    }
+
+    public static List<FileInfo> parseDatalink(URL url, String prepend_file_name) {
+        List<FileInfo> fileInfos = new ArrayList<>();;
+        try {
+            DataGroup[] groups = VoTableReader.voToDataGroups(url.toString(), false);
```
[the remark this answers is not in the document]
kpuriIpac: Thanks, using tmpFile now but I don't think contentType will work in this scope. The url is still the datalink url at this point, so contentType will look like `application/x-votable+xml;content=datalink`. But what I want is the contentType from the accessUrl when I loop through the DataGroup, which looks like `image/jpeg` And that was still a bug in my code though, I was previously calling getExtFromUrl with **url** as opposed to with **accessUrl**. Fixed that.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -0,0 +1,209 @@
...
+import java.io.IOException;
+import java.util.*;
+import java.net.URL;
+
+
+@SearchProcessorImpl(id = "ObsCorePackager")
+public class ObsCorePackager extends FileGroupsProcessor {
+    private static final Logger.LoggerImpl LOGGER = Logger.getLogger();
+
+    private static List<String> acceptableContentTypes = Arrays.asList("png","jpg","jpeg","bmp","fits","tsv", "csv", "tbl",
```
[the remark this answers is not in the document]
kpuriIpac: @loitly yes, this should have been named acceptableExtensionTypes. The idea is that I am looking at the content type (which could look like `application/fits`,` application/x-fits`, `image/jpeg`, etc) and checking if any of the **acceptableExtensionTypes** are present in the content type, and if they are, then use that **acceptableExtensionType** as the file extension (and this step in itself is fallback, only if we don't get extension type from access_format column (or content_type column in datalink table)). 

Does that logic sound okay?

lrebull:
woohoo! the basic functionality is there. i don't think it is sending me email, however -- but that may be a larger problem than just this.  I need to go check the guidelines for filenames ...

lrebull:
filenames look good! they follow the guidelines. :)

lrebull:
nevermind, emails made it, just reallllllyyyyy   slllllloooooooowwwww

kpuriIpac:
> nevermind, emails made it, just reallllllyyyyy slllllloooooooowwwww

yes that's be something we may need to look at later..

robyww (approved):
Great Job! Merge it!

PR #1357, opened 2023-04-18
Firefly-1198:whole table search fix for TAP services not supporting uploads 

kpuriIpac:
#### [Firefly-1198](https://jira.ipac.caltech.edu/browse/FIREFLY-1198)

- the  first bug/issue: if you did a whole table search with a TAP service selected that does not support upload, you would still see the table you just did a whole table search with on the TAP panel, leading to all sorts of issues. 
- This fixes that bug, by checking if this TAP service supports uploads or not, if it does not - don't show the table on the TAP panel for this service
  - instead, display a warning message. 
- **Updates 4/19** 
  - updated the warning message to be displayed after you do a click to action (whole table) search and switch to a service such as NED that does not support uploads
  - fixed the Position Column (from the uploaded table) bugs, see testing instructions below 

- **Updates 4/21** 
  - addressed PR comments, small fixes and using serviceLabel in the warning message now to show TAP service 
  - fixed a bug where if you load a table, and then switch over to a service like NED that doesn't support uploads, then its ADQL query would look incorrect. Reason is that we were adding a prefix to the ADQL query (which we normally do for other uploads, but in this case only the where clause of the ADQL would get the prefix and not the FROM clause - making the query look something like this: 
     - `FROM NEDTAP.objdir WHERE CONTAINS(POINT('ICRS', N.ra, N.dec)`
  - so I passed down **canUpload** (from capabilities to **makeSpatialConstraints** and only set prefix if canUpload is true as well 

Testing: https://fireflydev.ipac.caltech.edu/firefly-1198-whole-table-search-bugs/firefly
- select NED as TAP service 
- search m1 with a small radius 
- do a whole table search on the resulting table
- you should see the warning message next to "Spatial"
- if you switch back to a service that supports uploads (IRSA, CADC, etc.), you should see the table that you just did a whole table search on here
   - and if you switch back to NED, you should see the warning message again. 
- **Updates 4/19**: Instructions to test for position columns bug fixes
  - search for 2 tables (I suggest NED as TAP service and search for m1, and CADC as TAP service and search for m1 - but it can really be any 2 tables)
     - do a click to action whole table search on the first table. Take a note of the position columns. 
     - then, do a click to action whole table search on the second table, the position columns (from the uploaded table) should match the ones from this table and not the first table (that was the bug). 
     - also, in the input field of the position columns, attempt to type in the "**Lon Column**" and "**Lat Column**" values. As you start typing (say "ra"), pick one of the value from the suggestions and make sure that works as expected.
        - this was one of the other bugs with position columns. The column validator still had the columns from the previous table, so when you did a whole table search and manually typed in values in the input field, it wouldn't work as expected (because the column validator would check against the columns of the first table) - so I had to make sure to update the call to `getColValidator` before calling setUploadInfo with the uploadInfo object generated from the whole table search.

### DIFF
--- a/src/firefly/js/ui/tap/SpatialSearch.jsx
+++ b/src/firefly/js/ui/tap/SpatialSearch.jsx
@@ -30,6 +30,7 @@ import {
     getAsEntryForTableName, getColumnAttribute, getTapServices, makeUploadSchema, maybeQuote, tapHelpId
 } from './TapUtil.js';
 import {showColSelectPopup} from 'firefly/charts/ui/ColSelectView';
+import InputFieldLabel from 'firefly/ui/InputFieldLabel';
 
 const CenterLonColumns = 'centerLonColumns';
 const CenterLatColumns = 'centerLatColumns';
@@ -68,7 +69,7 @@ const fldListAry= [ServerParams.USER_TARGET_WORLD_PT,SpatialRegOp,SPATIAL_TYPE,
             SpatialMethod,RadiusSize, PolygonCorners,CenterLonColumns,CenterLatColumns,
     UploadCenterLonColumns, UploadCenterLatColumns, cornerCalcType];
 
-export function SpatialSearch({cols, serviceUrl, columnsModel, tableName, initArgs={}, obsCoreEnabled:requestObsCore, capabilities}) {
+export function SpatialSearch({cols, serviceUrl, serviceLabel, columnsModel, tableName, initArgs={}, obsCoreEnabled:requestObsCore, capabilities}) {
     const {searchParams={}}= initArgs ?? {};
     const obsCoreEnabled= requestObsCore && canSupportAtLeastOneObsCoreOption(capabilities);
     const panelTitle = !obsCoreEnabled ? Spatial : 'Location';
@@ -78,7 +79,7 @@ export function SpatialSearch({cols, serviceUrl, columnsModel, tableName, initAr
     const {canUpload=false}= capabilities ?? {};
 
     const {setConstraintFragment}= useContext(ConstraintContext);
-    const {setVal,getVal,makeFldObj,groupKey}= useContext(FieldGroupCtx);
+    const {setVal,getVal,makeFldObj}= useContext(FieldGroupCtx);
     const [constraintResult, setConstraintResult] = useState({});
     const [getUploadInfo, setUploadInfo]= useFieldGroupValue('uploadInfo');
     const [posOpenMsg, setPosOpenMsg]= useState(TAB_COLUMNS_MSG);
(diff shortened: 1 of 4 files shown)


src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -169,14 +178,20 @@ export function SpatialSearch({cols, serviceUrl, columnsModel, tableName, initAr
...
-                                message={constraintResult?.simpleError??''}
+                                message={constraintResult?.simpleError ?? ''}
                                 initialStateOpen={true} initialStateChecked={true}>
             <div style={{marginTop: 5}}>
                 <ForceFieldGroupValid forceValid={!checkHeaderCtl.isPanelActive()}>
 
+                    {!canUpload && searchParams.uploadInfo &&
+                        <div>
+                            {<InputFieldLabel label={'Spatial Type:'} tooltip={'tooltip'} labelWidth={SpatialLabelSpatial}/>}
+                            {warningMsg('Warning: Upload is not supported from Single Shape services')}
```
robyww, line 191: Let's improve this message like this:

1. pass `serviceLabel` to spacial search, it could be empty, but most the time not
2. message `Single Shape selected: $(serviceLabel||'This service'} does not support upload`

kpuriIpac:
> Mostly looking good. A few ideas for cleanup.
> 
> * one issue: I would like this to also be an improvement it the user just selection an upload table in the normal way in IRSA the switches to NED. It should show the same message. It does not now.

Addressed this issue and other small changes. Showing warning message when you switch to a service that doesn't support upload. 

There was another small issue, when you switch back over to IRSA/CADC (upload supported service) from a service like NED, it would switch over to Single Shape again (even though you were on Multi Object right before switching to NED) - the fix for this is in my latest commit - really simple, just checking if `canUpload` and `uploadInfo` are true and setting `SPATIAL_TYPE` to `MULTI` in that case...but makes it work now when you switch between services

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -78,7 +79,7 @@ export function SpatialSearch({cols, serviceUrl, columnsModel, tableName, initAr
     const {canUpload=false}= capabilities ?? {};
 
     const {setConstraintFragment}= useContext(ConstraintContext);
-    const {setVal,getVal,makeFldObj,groupKey}= useContext(FieldGroupCtx);
+    const {setVal,getVal,setFld,makeFldObj,groupKey}= useContext(FieldGroupCtx);
```
robyww, line 82: Looks like you decided not to use `setFld` you should remove it. I don't think `groupKey` is used anymore here either.

robyww (approved):
Looks good, just fix the one thing and merge.

src/firefly/js/ui/tap/SpatialSearch.jsx
kpuriIpac: Yes, setFld didn't allow me to send a value (to set the initial lon/lat values) so I ended up using setVal. Will remove setFld and groupKey and merge.

robyww: What you are doing is find. with `setFld` you just include include the object `{value:v, other stuff}`

PR #1358, opened 2023-04-18
FIREFLY-712: Table upload not accepting compliant IPAC tables

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-712
- IPAC table requires column type to be lower case.  Convert types to lower case when reading in.

Test: https://fireflydev.ipac.caltech.edu/firefly-712-ipac-type/firefly
Upload the attached sample table file from the ticket.  All column types appeared in lower case and recognized as it should.


kpuriIpac (approved):
Looks good! Can confirm this is working now. Tested it on the test build, and also on ops to recreate the bug.

PR #1359, opened 2023-04-18
FIREFLY-1222 Bug fix in the readout of pixels with value 0.

tgoldina:
This is one line bug fix for incorrect `NaN` values in place of zeros in pixel value display.

Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1222
Test URL: https://fireflydev.ipac.caltech.edu/firefly-1222-zero-val/firefly

To test, upload the FITS image attached to the ticket. Note, that zero values for both float and integer values are now displayed correctly. (They are displayed as `NaN` in https://fireflydev.ipac.caltech.edu/firefly)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/visualize/ClientFitsHeader.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/visualize/ClientFitsHeader.java
@@ -81,7 +81,7 @@ public class ClientFitsHeader implements Serializable, Iterable<String> {
     public double getCDelt2() { return getDoubleHeader(CDELT2); }
     public double getBScale() { return getDoubleHeader(BSCALE); }
     public double getBZero() { return getDoubleHeader(BZERO); }
-    public double getBlankValue() { return getDoubleHeader(BLANK_VALUE); }
+    public double getBlankValue() { return getDoubleHeader(BLANK_VALUE, Double.NaN); }
     public long getDataOffset() { return getLongHeader(DATA_OFFSET); }
 
     public int getIntHeader(String key) { return getIntHeader(key,0); }
--- a/src/firefly/java/edu/caltech/ipac/firefly/visualize/ClientFitsHeader.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/visualize/ClientFitsHeader.java
@@ -75,13 +75,18 @@ public class ClientFitsHeader implements Serializable, Iterable<String> {
     }
 
     public int getPlaneNumber() { return getIntHeader(PLANE_NUMBER); }
-    public int getBixpix() { return getIntHeader(BITPIX); }
+    public int getBitpix() { return getIntHeader(BITPIX); }
     public int getNaxis1() { return getIntHeader(NAXIS1); }
     public int getNaxis2() { return getIntHeader(NAXIS2); }
     public double getCDelt2() { return getDoubleHeader(CDELT2); }
     public double getBScale() { return getDoubleHeader(BSCALE); }
     public double getBZero() { return getDoubleHeader(BZERO); }
-    public double getBlankValue() { return getDoubleHeader(BLANK_VALUE, Double.NaN); }
+
+    public double getBlankValue() {
+        // blank value is only applicable to integer values (BITPIX > 0)
+        return this.getBitpix() > 0 ? getDoubleHeader(BLANK_VALUE, Double.NaN) : Double.NaN;
+    }
+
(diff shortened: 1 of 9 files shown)


robyww (approved):
Thanks!

src/firefly/java/edu/caltech/ipac/firefly/visualize/ClientFitsHeader.java
```diff
@@ -81,7 +81,7 @@ public Iterator<String> iterator() {
     public double getCDelt2() { return getDoubleHeader(CDELT2); }
     public double getBScale() { return getDoubleHeader(BSCALE); }
     public double getBZero() { return getDoubleHeader(BZERO); }
-    public double getBlankValue() { return getDoubleHeader(BLANK_VALUE); }
+    public double getBlankValue() { return getDoubleHeader(BLANK_VALUE, Double.NaN); }
```
gpdf, line 84: I see why this works - essentially we're eliminating the default for `BLANK` - but for the record there's another problem here.   `BSCALE` and `BZERO` are genuinely floating-point parameters (this is explicit in the standard; section 4.4.2.5) but the value of `BLANK` is specified to be an integer:

[FITS 4.0.0](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf) section 4.4.2.5:

> **BLANK keyword.** (...) The value field shall contain an integer that specifies the value that is used within the integer array to represent pixels that have an undefined physical value.

Treating `BLANK` as a double inside Firefly has the advantage that you don't have to maintain a separate flag for "`BLANK` was not specified", but it means that if `BLANK` were ever set to a value that can't be represented exactly in a double, this code would fail.  I *think* this could only happen for 64-bit integer images (`BITPIX=64`), so perhaps this is a safe risk to take -- but formally it is an error.

gpdf (commented):
Not an actual objection, but a concern.

gpdf:
Thinking about it a bit more... I think the standard is clear that `BLANK` should only be respected for integer images.  The mechanism shouldn't be used at all for floating-point images.  From what @tgoldina has said it sounds like Firefly *does* apply `BLANK` to floating-point images, so that should be fixed.

robyww:
@tgoldina if you want what to modify the code so it only respects blank on internet BITPIX, I am fine with that. I think you will need to check other parts of the code for this as well. I think there is another spot in FitsReadUtil

tgoldina:
Thank you for feedback. Respecting BLANK on integer-only BITPIX should not be a problem. 

Supporting int64 values that can not be represented as double is bigger issue:  pixel values in general are passed as double precision in Firefly, see [PixelValue.readValue](https://github.com/Caltech-IPAC/firefly/blob/dev/src/firefly/java/edu/caltech/ipac/visualize/plot/PixelValue.java#L74)

tgoldina:
I have addressed the issue that BLANK should only apply to int value arrays. Updated the test FITS in the ticket to include extensions with various BITPIX and versions without and with BLANK keyword.

I did not change the representation of BLANK value, it is still represented as double as all pixel values.
Large BLANK values that can not be represented precisely with 64-bit float numbers will not be handled correctly.

src/firefly/java/edu/caltech/ipac/visualize/plot/plotdata/FitsReadUtil.java
gpdf, line 1: I think line 501 might be another long-standing bug.  I don't understand why it's not `getNaxis3(h) >= 2`, or equivalently `getNaxis3(h) > 1`.

gpdf (approved):
Other than the unrelated existing possible-bug relating to naxis3/4, looks good to me.

I'll add the remaining unfinished issue regarding the handling of 64-bit BLANK values to FIREFLY-1223.

gpdf:
Updated the comments about the PTF pixel-value kludge.

robyww (changes requested):
The code all looks good.  However, to be clean we need to pass `bitpix` to `ImageStretch.getFluxStandard()` and (regrettably) `ImageStretch.getFluxPalomar()`. They should check `bitpix>0`

tgoldina:
> to be clean we need to pass bitpix to ImageStretch.getFluxStandard() and (regrettably) ImageStretch.getFluxPalomar(). They should check bitpix>0.

I am not sure why we need to do it. The check happens when we get blank value in `FitsReadUtil.getBlankValue`. The blank_value will be Double.NaN if bitpix>0.

robyww (approved):

PR #1361, opened 2023-04-21
Firefly-1212: filter upload tbl columns down to user selected cols

kpuriIpac:
#### [Firefly-1212](https://jira.ipac.caltech.edu/browse/FIREFLY-1212)
- for a table upload, filtered down user selected columns in the upload table 
  - @robyww I wasn't sure how to use the "INCL_COLUMNS" attribute of a TableServerRequest properly for this, but managed to get it working by editing the DataGroup object, let me know if this approach looks okay. 
  - **Updates 4/24**
     - updated code in AsyncTapQuery per @loitly's suggestions 
     - if you upload table/load table, or do a click to action whole table search, the selected columns by default will only be the same as position columns 
        - and if user changes position columns, any of these changes will be reflected by being selected in the columns (and therefore, also in the ADQL) 
     - confirmed table filtering works
     - confirmed **dg.removeDataDefinition** works as expected 
     - the small bug with selecting some columns but them still not being selected when you click "OK" is now fixed 
     - checked with backend team, IRSA does expect ra and dec in lowercase. So we can discuss how to (or if) resolve that if we receive ra/dec values that are capitalized. Either here or in the new ticket to fix the VO Tables <description> tag place: https://jira.ipac.caltech.edu/browse/FIREFLY-1228 

Testing: https://fireflydev.ipac.caltech.edu/firefly-1212-adql/firefly
- select CADC as the TAP service, search for m1
- from the resulting table, do a click to action whole table search (or load the table directly from the TAP panel using the multi object upload dialog)
- search again using this table *without* editing the column selection
  - you will probably get an error that reads something like "Column [dataproduct_type] is ambiguous." 
- now go back to the TAP panel and edit the column selection. Unselect all columns, then select only a few (say, "s_ra" and "s_dec")
   - now if you search again, it should not fail and you should get back a table (as the upload table only includes the columns you selected - and no bad columns (in this case, the **dataproduct_type** column - and possibly some other columns as well)
   - before this fix, if the table included a bad column such as **dataproduct_type** and you did a search even after unselecting this column, your search would still fail (since we upload the whole table anyway). this should now be fixed. 

**Updates for Testing 4/24**
- test build is updated
- change position columns (whatever you change it too, should be added as selected in the columns) 
- try selecting table from Loaded Tables dialog (from multi-object) and from click to action's whole table search. Also try uploading a table from the multi-object dialog. In all 3 cases, ensure that the columns selected by default are only the position columns

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
@@ -7,6 +7,7 @@ import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.server.network.HttpServiceInput;
 import edu.caltech.ipac.table.DataGroup;
+import edu.caltech.ipac.table.DataType;
 import edu.caltech.ipac.table.TableUtil;
 import edu.caltech.ipac.table.io.VoTableWriter;
 import edu.caltech.ipac.util.AppProperties;
@@ -14,6 +15,7 @@ import edu.caltech.ipac.util.StringUtils;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.*;
 
 import static edu.caltech.ipac.table.TableUtil.Format.VO_TABLE;
 
@@ -34,6 +36,7 @@ public class AsyncTapQuery extends UwsJobProcessor {
         String queryStr = request.getParam("QUERY");
         String lang = request.getParam("LANG");
         String maxrecStr = request.getParam("MAXREC");
+        String serverFile = request.getParam("UPLOAD");
         int maxrec = -1; // maxrec is not set
         if (!StringUtils.isEmpty(maxrecStr)) {
             maxrec = Integer.parseInt(maxrecStr);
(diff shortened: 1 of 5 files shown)


src/firefly/js/ui/tap/TapSearchRootPanel.jsx
```diff
@@ -361,6 +366,7 @@ function onTapSearchSubmit(request,serviceUrl,tapBrowserState) {
         const params = {serviceUrl, QUERY: adqlClean};
         if (isUpload) {
             params.UPLOAD= serverFile;
+            params.UPLOAD_COLUMNS= userColumns;
```
robyww, line 369: don't set this if `isADQL` is `true`.

robyww (approved):
This looks good to me. 

 - Check with @loitly to see if there is a function for you purging out the columns. 
 - Fix the one issue I highlighted.
 - after that merge.

src/firefly/java/edu/caltech/ipac/firefly/server/query/AsyncTapQuery.java
```diff
@@ -55,6 +58,19 @@ public HttpServiceInput createInput(TableServerRequest request) throws DataAcces
...
+                    for (DataType entry : data) {
+                        colNames.add(entry.getKeyName());
+                    }
+                    String uploadCols = request.getParam("UPLOAD_COLUMNS");
+                    String[] cols = uploadCols.split(",");
+                    for (String key : colNames) {
+                        boolean contains = Arrays.stream(cols).anyMatch(key::equals);
+                        if (!contains) dg.removeDataDefinition(key);
+                    }
+                }
```
loitly, line 73: This makes sense.  There's a few minor things to check..
- UPLOAD_COLUMNS is empty string
- colNames contains column with commas in the name
- check the 'saved' file to make sure `dg.removeDataDefinition` worked.

code sample ( rebase `dev` to use `TableUtil.splitCols`)
```java
String uploadCols = request.getParam("UPLOAD_COLUMNS");
if (!StringUtils.isEmpty(uploadCols)) {                                     //filter dg columns down to the ones the user has selected
    List<String> cols = Arrays.stream(TableUtil.splitCols(uploadCols))      // use this to allow commas in double-quotes
                    .map(cname -> cname.trim().replaceAll("^\"|\"$", ""))  // remove double-quotes if exists
                    .toList();

    Arrays.asList(dg.getDataDefinitions()).forEach(c -> {
        if (!cols.contains(c.getKeyName())) dg.removeDataDefinition(c.getKeyName());
    });
}
```

loitly (approved):
Changes look fine.  Double check to make sure `dg.removeDataDefinition` worked as expected.

robyww (approved):
I think this is all good.

loitly:
Looks good.

PR #1363, opened 2023-04-24
IRSA-3146 update for SHA Precovery data view.

wmiipac:
see PR:https://github.com/IPAC-SW/irsa-ife/pull/256 for details

### DIFF
--- a/src/firefly/js/metaConvert/ShaRequestList.js
+++ b/src/firefly/js/metaConvert/ShaRequestList.js
@@ -24,17 +24,22 @@ const rangeValues= RangeValues.makeRV({which:SIGMA, lowerValue:-2, upperValue:10
  * @param table
  * @param row
  * @param includeSingle
+ * @param includeStandard
  * @return {{}}
  */
-export function makeShaPlotRequest(table, row, includeSingle) {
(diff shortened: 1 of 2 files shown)


src/firefly/js/metaConvert/ShaRequestList.js
```diff
@@ -44,25 +44,68 @@ export function makeShaPlotRequest(table, row, includeSingle) {
     /*
      const url = https://irsa.ipac.caltech.edu/data/SPITZER/SHA/archive/proc/MIPS003600/r5572864/ch2/bcd/SPITZER_M2_5572864_0007_0061_9_bcd.fits
      */
-    let serverinfo = 'https://irsa.ipac.caltech.edu/data/SPITZER/';
+    const serverinfo = 'https://irsa.ipac.caltech.edu/data/SPITZER/';
+    const mosimgserver = 'https://sha.ipac.caltech.edu/applications/Spitzer/SHA/servlet/ProductDownload?DATASET=level1';
```
robyww, line 48: I don't this should be hard coded. In a future iteration it probably needs to be a property.

robyww (approved):
This is fine, we should think about how to get the URL's in a property before we go out. However, since SHA is so static maybe it is not necessary.

PR #1366, opened 2023-04-26
Firefly-1229: Clean up and fix bug related to Rubin

robyww:
#### [Firefly-1229](https://jira.ipac.caltech.edu/browse/FIREFLY-1229): Testing and Fixes needed for SUIT deployment

- Clean up service descriptor input after DCE work 
- use embedded spacial selection
- Fixed:  issue with embedded selection not working with service descriptions
- Fixed: TAP Upload: size no longer show if upload is from a table (size not known)
- Fixed: TAP Upload: fixed crash in editing columns name when not column exist
- Fixed: SpectrumMetaInspector did not insert the utype when found in FITS
- Fixed: CoverageWatcher crash when looking for the center in ADQL an not there
- Fixed: Change UwsJobProcess rebuild credentials on first redirect
- Fixed: service def panel new show the default HiPS used for coverage and sizes fov on s_region
- Improved: You can now scroll coverage around without if affecting the select point or shape
-  Fixed: service def panel gives correct hips and fov size
- Fixed: check for undefined in ThumbnailView
- Fixed: bad table service always leaves services visible
- coverage settable to show hips only
- obscore table download settable
- Release notes updates

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1229-servdev-panel/firefly
- Rubin: SUIT: data-int (https://data-int.lsst.cloud/portal/app/)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
@@ -25,6 +25,7 @@ import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.text.ParseException;
 import java.time.Instant;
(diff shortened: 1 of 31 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -161,12 +162,29 @@ HttpServiceInput createSingleResultInput(String jobUrl) {
...
             File outFile = File.createTempFile(filename, ".vot", QueryUtil.getTempDir(request));
-            HttpServices.Status status = HttpServices.getData(input, outFile);
+
+            HttpServices.Status status= HttpServices.getData(input, (method -> {
+                    if(HttpServices.isOk(method)) {
+                        try {
+                            return HttpServices.defaultHandler(outFile).handleResponse(method);
+                        } catch (FileNotFoundException e) {
+                            return new HttpServices.Status(400,"Bad Request: Output file not found");
+                        }
```
loitly, line 177: You can ignore exception if you want since file was created above.

loitly (approved):
Code for TAP result's redirect looks good.  Thanks for the fix.
I randomly test the application and did not find any errors.
However, when I did a NED TAP search, follow by 'Use table as an upload to TAP search', I was temporarily confused since nothing happened.  Of course, I then realized NED does not support upload.  Do you have plan to handle this better in the future or, it's acceptable behavior?

robyww:
> However, when I did a NED TAP search, follow by 'Use table as an upload to TAP search', I was temporarily confused since nothing happened. Of course, I then realized NED does not support upload. Do you have plan to handle this better in the future or, it's acceptable behavior?

Yes, that was working in dev I might have broken it, I will look

PR #1367, opened 2023-04-26
IRSA-5175: Fix the right-side of the tri-view to accommodate both Charts and Coverage

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-5175
```
When IRSAViewer is in DCE mode and Coverage appears on the right-hand side of the tri-view, 
it creates a tab-in-tab look which is wrong.  We need to fix it so that Coverage and Charts can coexist in that area.
```
Test: https://fireflydev.ipac.caltech.edu/irsa-5175-triview-top-right/firefly
The feature allowing Coverage to be moved to Chart area is now enabled for Firefly, not just DCE.
To test, search or load a table.  Then, toggle `Coverage: Left/Right/Coverage w/Charts`.  Pin Chart should work as it did before as well.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -93,7 +93,7 @@ export function getMinScatterGLRows() {
 }
 
 /**
- * @returns {boolean} return true if PinnedChartPanel is used.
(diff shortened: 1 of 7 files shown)


robyww (approved):
Looks good! thanks for getting this in.

PR #1368, opened 2023-04-26
FIREFLY-987 Identify tables by EXTNAME, EXTVERSION, EXTLEVEL 

tgoldina:
XXXX-TAB algorithm for computing spectral coordinates requires that the combination of EXTNAME, EXTVERSION, and EXTLEVEL is used to identify the extensions with lookup tables. This PR makes sure that the values of these 3 headers are passed with the WAVELENGTH_TABLE related data to the client.

Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-987
Test deployment: https://fireflydev.ipac.caltech.edu/firefly-987-extid/firefly

To test, check that the displayed wavelength is not NaN for the test files from https://jira.ipac.caltech.edu/browse/FIREFLY-989

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
@@ -3,17 +3,11 @@
  */
 
 package edu.caltech.ipac.firefly.server.visualize.fitseval;
-/**
- * User: roby
- * Date: 7/13/18
- * Time: 10:39 AM
- */
 
 
 import edu.caltech.ipac.firefly.data.RelatedData;
 import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.visualize.WebPlotRequest;
-import edu.caltech.ipac.util.ComparisonUtil;
 import edu.caltech.ipac.visualize.plot.plotdata.FitsRead;
 import edu.caltech.ipac.visualize.plot.plotdata.FitsReadUtil;
 import nom.tam.fits.BasicHDU;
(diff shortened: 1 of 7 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
```diff
@@ -30,167 +24,143 @@
...
      * WCS table lookup is undefined. The specified FITS BINTABLE
      * must contain only one row.
-     * No units conversions are to
-     * @return
+     *
+     * @param HDUs FITS HDUs
+     * @param tabExtId Extension ID of the table lookup extension
+     * @return matching tab extension null if it can not be found
      */
-    private static boolean isValidWCSForTab(BasicHDU<?>[] HDUs){
```
robyww, line 82: so we don't need to validate at all?

robyww (approved):
I think it is fine. I don't quite understand all the decisions in SpectralCubeEval.java, but I think I originally took that from other code. So maybe it was not right in the first place.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/fitseval/SpectralCubeEval.java
tgoldina: > so we don't need to validate at all?

`isValidWCSForTab` was checking that there are no extensions with binary tables that have the same EXTNAME, EXTVER, and EXTLEVEL. (For more than one matching extension, table lookup is undefined.) 

In the updated code, I check for this condition in `findTabExtension`: when looking for matching extension we need to do the same loop and WCS matching, no need to do it twice.

I'll update Javadoc to make it clear and fix the wrong brace placement in line 93.

PR #1369, opened 2023-04-27
Firefly-1237: fixed

robyww:
#### [Firefly-1237](https://jira.ipac.caltech.edu/browse/FIREFLY-1237)
 
 - _Note_- There was a race condition so I used `setTimeout`, in the future it would be worth looking at a bigger solution.

#### Testing
  - https://firefly-1237-polygon.irsakudev.ipac.caltech.edu/irsaviewer

### DIFF
--- a/src/firefly/js/ui/CatalogSearchMethodType.jsx
+++ b/src/firefly/js/ui/CatalogSearchMethodType.jsx
@@ -280,7 +280,7 @@ function SizeArea({groupKey, searchType, imageCornerCalc}) {
             if (!cornerCalcV) {
                 if (sel) setVal('imageCornerCalc', 'area-selection');
             }
-            setVal('polygoncoords', calcCornerString(pv, cornerCalcV));
+            setTimeout( () => setVal('polygoncoords', calcCornerString(pv, cornerCalcV)), 4);
         }
     };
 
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/CatalogSearchMethodType.jsx
```diff
@@ -280,7 +280,7 @@ function SizeArea({groupKey, searchType, imageCornerCalc}) {
             if (!cornerCalcV) {
                 if (sel) setVal('imageCornerCalc', 'area-selection');
             }
-            setVal('polygoncoords', calcCornerString(pv, cornerCalcV));
+            setTimeout( () => setVal('polygoncoords', calcCornerString(pv, cornerCalcV)), 4);
```
loitly, line 283: Curious... if only a 4ms delay was needed, shouldn't a no delay work as well?

loitly (approved):
Code change looks fine.

src/firefly/js/ui/CatalogSearchMethodType.jsx
robyww: probably

lrebull:
i confirm it works!

PR #1371, opened 2023-04-28
FIREFLY-1156: better SQL error message for added column failure

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1156

Test: https://fireflydev.ipac.caltech.edu/firefly-1156-sql-errors/firefly
Load table, then try to make 'Add Column' fail to see error message.  One way is to reference a bad column name in the expression field.  Or, create a type mismatch by picking a data type that's inconsistent with the results of the expression.

You can also look at tests to see several different type of messages.
https://github.com/Caltech-IPAC/firefly/blob/FIREFLY-1156_sql_errors/src/firefly/test/edu/caltech/ipac/table/EmbeddedDbUtilTest.java#L239
https://github.com/Caltech-IPAC/firefly/blob/FIREFLY-1156_sql_errors/src/firefly/test/edu/caltech/ipac/table/EmbeddedDbUtilTest.java#L270

@lrebull It takes a lot of work to construct a precise error message.  And even then, it may not cover all cases.  Are these messages good enough?   If not, can you suggest what I should show?

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/EmbeddedDbUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/EmbeddedDbUtil.java
@@ -23,6 +23,8 @@ import edu.caltech.ipac.table.*;
 import edu.caltech.ipac.util.AppProperties;
 import edu.caltech.ipac.util.StringUtils;
 import org.json.simple.JSONObject;
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.jdbc.BadSqlGrammarException;
 import org.springframework.jdbc.core.BatchPreparedStatementSetter;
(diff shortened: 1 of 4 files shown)


lrebull:
I think you have managed to capture the most likely errors that people will make, and these error messages are enough to point people in the right direction. I tried several of the typos I most frequently make, and it caught them all. I think this will help 95-99% of the situations. Anything else is likely to be pretty pathological! :)

robyww:
@lrebull note that this PR is not in the upcoming release. It you think it needs to be let us know quickly.

robyww (approved):
I like that you are now identifying the error where it is happening instead of giving a general one in the UI.

lrebull:
I don't feel strongly about whether or not it should go in this release. It's up to you guys as to whether or not you want to include it.

PR #1372, opened 2023-04-28
Firefly 1238:  clear polygon

robyww:
#### [FIREFLY-1200](https://jira.ipac.caltech.edu/browse/FIREFLY-1200): Fixed: MOCs added after additional searches
- finishes the work from Firefly-1200. I missed a case

#### Testing
- https://firefly-1238-clear-polygon.irsakudev.ipac.caltech.edu/irsaviewer/dce
- test according to the tickets

### DIFF
--- a/src/firefly/js/visualize/task/PlotHipsTask.js
+++ b/src/firefly/js/visualize/task/PlotHipsTask.js
@@ -304,13 +304,14 @@ async function makeHiPSPlot(rawAction, dispatcher) {
 
 
 
-export function createHiPSMocLayerFromPreloadedTable({tbl_id,title, fitsPath, mocUrl, plotId, visible=false, color, mocGroupDefColorId} ) {
+export function createHiPSMocLayerFromPreloadedTable({tbl_id,title, fitsPath, mocUrl, plotId, visible=false,
+                                                         color, mocGroupDefColorId, attachAllPlot=false} ) {
     const table= getTblById(tbl_id);
     if (!table) return;
     const uniqColName= table.tableData.columns[0].name;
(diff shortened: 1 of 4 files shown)


lrebull:
I agree the MOCs are now fixed! Joyce is looking at the polygon stuff. :)

joycek-ipac:
I've tested the polygon search box and although it still auto-populates with some content, erasing it (and no longer appearing to re-populate) is now possible. If that sort of result is sufficient, then I think that bug may be fixed as well.

src/firefly/js/visualize/task/PlotHipsTask.js
```diff
@@ -304,13 +304,14 @@ async function makeHiPSPlot(rawAction, dispatcher) {
 
 
 
-export function createHiPSMocLayerFromPreloadedTable({tbl_id,title, fitsPath, mocUrl, plotId, visible=false, color, mocGroupDefColorId} ) {
+export function createHiPSMocLayerFromPreloadedTable({tbl_id,title, fitsPath, mocUrl, plotId, visible=false,
+                                                         color, mocGroupDefColorId, attachAllPlot=false} ) {
     const table= getTblById(tbl_id);
```
loitly, line 309: `attachAllPlot` is added to createHiPSMocLayerFromPreloadedTable, but, I don't see where it's used in your PR.  This is normal?

loitly (approved):
Looks good.

src/firefly/js/visualize/task/PlotHipsTask.js
robyww: It is only called from one place and I made the default false. If used again (I think I probably will) there is a case which a it could pass `true` I made it a parameter instead of hardcoding it on line 314

PR #1374, opened 2023-05-01
FIREFLY-1240: flipped y axis doesn't persist in chart thumbnail view

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1240
Test:  https://firefly-1240-pinchart-flipy.irsakudev.ipac.caltech.edu/irsaviewer/

Also fix  #10 from Bugs found in 2023.1 RC to ticket: 
https://confluence.ipac.caltech.edu/display/IRSA/Bugs+found+in+2023.1+RC+%3A+IV%2C+DCE+testing+April+2023
> When you use the preset function to add a column, it appears in the newly created column as true and false. But when you go to try to filter on that newly created column from the filter pulldown at the top of the new column, it says true and <NULL>, not true and false. If you filter on <NULL>, it does what you expect it to do, e.g., just gives you the 'false' values


robyww (approved):
Looks good.

lrebull:
confirm, these both fixed in this build!

PR #1379, opened 2023-05-05
IRSA-3356: Added validator in download dialog for SHA

jaladh-singhal:
Completes an outstanding task in https://github.com/IPAC-SW/irsa-ife/pull/252

In SHA, options for data files (checkbox group) can't be empty. And unlike field level validation, this requires validation on submit. So added prop validateOnSubmit in DownloadDialog and added a validator function in SHA.

## Testing
1. SHA: https://irsa-3356-download-validator.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA/ - see instructions at https://github.com/IPAC-SW/irsa-ife/pull/257

2. Other apps: https://irsa-3356-download-validator.irsakudev.ipac.caltech.edu/applications/sofia/ 
    Make  a search, then select a row in "FORCAST" table and click prepare download, uncheck file type options and submit. Because there wasn't any validator passed in props, it should submit download options form as it used to.

### DIFF
--- a/src/firefly/js/ui/DownloadDialog.jsx
+++ b/src/firefly/js/ui/DownloadDialog.jsx
@@ -141,10 +141,10 @@ const preTitleCss = {
 let dlTitleIdx = 0;
 const newBgKey = () => 'DownloadOptionPanel-' + Date.now();
 
-export function DownloadOptionPanel (props) {
-    const {groupKey, cutoutSize, help_id, children, style, title, dlParams, updateSearchRequest=null, updateDownloadRequest=null} = props;
-    const { cancelText='Cancel', showZipStructure=true, showEmailNotify=true, showFileLocation=true, showTitle=true } = props;
-
+export function DownloadOptionPanel ({groupKey, cutoutSize, help_id, children, style, title, dlParams,
+                                         updateSearchRequest, updateDownloadRequest, validateOnSubmit,
+                                         cancelText='Cancel', showZipStructure=true, showEmailNotify=true,
+                                         showFileLocation=true, showTitle=true, ...props}) {
     const {tbl_id:p_tbl_id, checkSelectedRow} = React.useContext(OptionsContext);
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -142,7 +142,8 @@ let dlTitleIdx = 0;
 const newBgKey = () => 'DownloadOptionPanel-' + Date.now();
 
 export function DownloadOptionPanel (props) {
-    const {groupKey, cutoutSize, help_id, children, style, title, dlParams, updateSearchRequest=null, updateDownloadRequest=null} = props;
+    const {groupKey, cutoutSize, help_id, children, style, title, dlParams, updateSearchRequest=null,
```
robyww, line 145: use `undefined` and not `null` we are trying to use `null` less.

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -161,6 +162,12 @@ export function DownloadOptionPanel (props) {
             return showInfoPopup('You have not chosen any data to download', 'No Data Selected');
         }
 
+        const {valid, message} = validateOnSubmit?.(formInputs) ?? {valid : true, message: ''};
```
robyww, line 165: You might want a default message. Also when valid it true you don’t need `message`. So…
```js

const {valid, message=‘Invalid Input’} = validateOnSubmit?.(formInputs) ?? {valid : true};
```

robyww (approved):
Looks good, just make the cleanup changes.

wmiipac (approved):
looks good.

src/firefly/js/ui/DownloadDialog.jsx
jaladh-singhal: OK, I've added default message in if block instead.

PR #1380, opened 2023-05-08
FIREFLY-1245: handle undefined pageSize

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1245
```js
var wrapperReq = firefly.util.table.makeTblRequest( 
  'IpacTableFromSource', 
  'NED SED', 
  {searchRequest: tblReq}, 
  {filters: '"NED Units" = \'Jy\''}, 
  {pageSize: 50} 
);
```

Normally, Firefly is able to handle undefined pageSize.  However, when it's used in this way(wrapper request with bad pageSize), it defaults to -1 which break paging and cause unexpected exceptions.

Test: https://firefly-1245-handle-undefined-pagesize.irsakudev.ipac.caltech.edu/
Follow instructions in ticket to test.  Or,

TLDR:
`Search for Source` -> m101 -> Search
switch table tab to `NED NED`, then select a different row.

### DIFF
--- a/src/firefly/js/tables/TableUtil.js
+++ b/src/firefly/js/tables/TableUtil.js
@@ -761,7 +761,7 @@ export function processRequest(tableModel, tableRequest, hlRowIdx) {
     let {data, columns} = nTable.tableData;
 
     nTable.request = tableRequest;
-    pageSize = pageSize || data.length || MAX_ROW;
+    pageSize = pageSize > 0 ? pageSize : data.length || MAX_ROW;
 
     if (filters) {
         filterTable(nTable, filters);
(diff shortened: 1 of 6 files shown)


robyww:
@loitly Question: You example shows a 5 parameter `makeTblRequest` which only takes 4 parameters. This is the problem you are fixing? So `pageSize` was never picked up?

robyww (approved):
Looks good.

src/firefly/js/tables/TableUtil.js
```diff
@@ -1403,6 +1403,12 @@ export function getTableState(tbl_id) {
     return TBL_STATE.OK;
 }
 
+export function fixPageSize(pageSize) {
```
dencoded, line 1406: I think would be good to have a couple of tests for `fixPageSize` to make sure it is doing what expected  (especially providing that we relay on this function in more than one places

src/firefly/js/tables/TableUtil.js
```diff
@@ -761,7 +761,7 @@ export function processRequest(tableModel, tableRequest, hlRowIdx) {
     let {data, columns} = nTable.tableData;
 
     nTable.request = tableRequest;
-    pageSize = pageSize || data.length || MAX_ROW;
+    pageSize = pageSize > 0 ? pageSize : data.length || MAX_ROW;
```
dencoded, line 764: I think would be good to have a couple of tests for pagination in TableUtil-test.js like we already do there for sorting and filtering.

dencoded (approved):
Overall looks good to me.
I suggested to add a couple of simple tests so this pagination problem won't hammer us again in the future, if it makes sense. I leave it to you.

src/firefly/js/tables/TableUtil.js
loitly: test added.

src/firefly/js/tables/TableUtil.js
loitly: test added.

PR #1382, opened 2023-05-09
FIREFLY-1242: Preserve filters when 3 color default selection is made

jaladh-singhal:
Fixes [FIREFLY-1242](https://jira.ipac.caltech.edu/browse/FIREFLY-1242)

- When an image with default colors is selected, and filters are present in that color, sets those same filters in other colors' filter field groups too.
- Made sure that this filter field value gets read in initial value

## Testing
https://firefly-1242-3color-default-filters.irsakudev.ipac.caltech.edu/irsaviewer/

Select "3 color composite" -> red channel -> apply mission filter: wise, project filter: all-sky -> select W4 in "WISE ALLWISE Atlas" -> switch to green channel, filters should be preserved and W2 should be automatically selected; switch to blue channel, filters should be preserved and W1 should be selected.

Similarly you can try filtering 2MASS and then test on "2MASS all sky" (where we have default colors available).

### DIFF
--- a/src/firefly/js/ui/ImageSelect.jsx
+++ b/src/firefly/js/ui/ImageSelect.jsx
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import React, {useEffect, useState} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 
 import PropTypes from 'prop-types';
 import {uniqBy, get, countBy, remove, sortBy, isNil} from 'lodash';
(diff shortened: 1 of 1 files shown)


lrebull:
i think this is ok. i made a movie where i sorted myself out mid-movie and this is definitely better than it was, but it did something i didn't expect which confused me. putting it here just for reference. But i think this is working.

https://github.com/Caltech-IPAC/firefly/assets/23038600/cce0db9a-678d-41a8-8411-05eaa83cc0ec

src/firefly/js/ui/ImageSelect.jsx
```diff
@@ -291,7 +291,7 @@ function FilterPanelView({imageMasterData}) {
  * @param {object[]} p.dataList
  * @param {number}   p.maxShown
  */
-function FilterSelect ({type, dataList, maxShown=6}) {
+function FilterSelect ({type, dataList, maxShown=6, groupKey}) {
```
robyww, line 294: I think you are working too hard passing `groupKey` you can just use context.
```js
const {groupKey}= useContext(FieldGroupCtx);
```

You can look at `FieldGroup.jsx`, `makeCtx()` (line 16) to see what is in the FieldGroupCtx

robyww (approved):
Looks good, you are just working too hard.

Make the changes I suggested so that you don't have to pass `groupKey` around. The code will be much cleaner. 

After that -  merge.

PR #1383, opened 2023-05-09
FIREFLY-1069 Multi-dimensional WAVE-TAB algorithm

tgoldina:
https://jira.ipac.caltech.edu/browse/FIREFLY-1069
Extended TAB algorithm to support non-separable spectral coordinates that are changing along both image axes.

Test image is attached to the ticket. Examples of other cubes can be found at [this confluence page](https://confluence.ipac.caltech.edu/x/Mg16IQ).

Reference
[Representations of spectral coordinates in FITS](https://www.aanda.org/articles/aa/pdf/2006/05/aa3818-05.pdf), Section 6.1.1

Other changes in this PR:
- Added `convertToArraySize` convenience method in TableUtil, which folds flattened 1-d array to multi-dimensional according to column header. The same code is used to display array cell value in a table.
- Updated interface to spectral coordinate calculation with the intent to support readout display for multiple spectral coordinates per WCS in future
- Test case for multi-d TAB algorithm
- Fixed NaN wavelength caused by the bug in CD default value.
- Fixed LOG algorithm readout. 

The last two fixed the spectral coordinate display for cube planes in https://data.sdss.org/sas/dr16/manga/spectro/redux/v2_4_3/7957/stack/manga-7957-1902-LOGCUBE.fits.gz
The LOGCUBE data cube has logarithmic wavelength sampling from log10(lambda/Angstroms)=3.5589 to 4.0151 (NWAVE=4563 spectral elements)
Verified that the readout matches the wavelengths in the WAVE extension.


src/firefly/js/visualize/projection/WavelengthHeaderParser.js
```diff
@@ -18,26 +18,186 @@ import {isDefined} from '../../util/WebUtil.js';
...
+            failWarnings.push(`CD3_i${altWcs} is not defined: ${ctype} value is constant`);
+        }
+
+        //check if any value is not defined, use default
+        pc_3j = applyDefaultValues(pc_3j, mijMatrixKeyRoot, which, N);
+
+        const {algorithm, coordType} = getAlgorithmAndType(ctype);
+
+        const canDoPlaneCalc = nAxis === 3 && pc_3j?.length === 3 && pc_3j[0] === 0 && pc_3j[1] === 0 &&
+            allGoodValues(crpix, crval, cdelt && nAxis === 3);
```
tgoldina, line 67: In this context pc_3j is the PC/CD transformation matrix row that corresponds to the spectral coordinate.

We assume that the spectral coordinate value should be displayed per plane  (not per pixel) when there are 3 image coordinates and the spectral coordinate has no dependency on the first two image coordinates.

Is it correct? Should I also check that the value corresponding to the 3-rd image axis is the only non-zero?

src/firefly/js/visualize/projection/Wavelength.js
```diff
@@ -24,22 +26,102 @@ export const VRAD = 'VRAD';
...
+    ['VRAD']: {
+        type: 'Radio velocity',
+        symbol: 'V',
+        defaultUnits: 'm/s',
+    },
+    ['WAVE']: {
+        type: 'Vacuum wavelength',
+        symbol: String.fromCharCode(0x03BB), // λ
+        defaultUnits: 'm',
+    },
```
tgoldina, line 95: Should we add a list of aliases for each item here? It would contain a list of coordinate types that should be treated as the known spectral type even though their CTYPE is not quite conforming to the standard.

src/firefly/js/visualize/projection/WavelengthHeaderParser.js
```diff
@@ -157,234 +291,86 @@ function applyDefaultValues(inArr, keyRoot, which, N) {
...
-    } else if (ctype3.startsWith('LAMBDA')) {
-        wlType = WAVE;
-        algorithm = LINEAR;
-    } else if (ctype3.startsWith('VRAD')) {
-        wlType = VRAD;
-        algorithm = PLANE;
+    // verify that coordinate type is valid
+    const validCoordTypes = Object.keys(spectralCoordTypes);
+    if (!validCoordTypes.includes(coordType)) {
+        return {algorithm, coordType: undefined};
```
tgoldina, line 371: We might need to deal with non-conforming coordinate types here.

src/firefly/js/tables/TableUtil.js
```diff
@@ -511,17 +534,7 @@ export function formatValue(col, val) {
...
-        if (col.type === 'char' && aryDim.length > 0) {
-           aryDim.shift();    // remove first dimension because char array is presented as string
-        }
-
-        if (aryDim.length > 1) {
-            for(let i = 0; i < aryDim.length-1; i++) {
-                val = chunk(val, aryDim[i]);
-            }
-        }
+        val = convertToArraySize(col, val);
```
tgoldina, line 537: @loitly Could you please review the changes to `TableUtil.js`? 

`convertToArraySize` converts one-dimensional as-stored array in an array column to multi-dimensional as-declared array.

src/firefly/js/tables/TableUtil.js
```diff
@@ -486,6 +486,29 @@ export function getRowValues(tableModel, rowIdx) {
...
+
+    const aryDim = (col.arraySize || '').split('x');
+
+    if (col.type === 'char' && aryDim.length > 0) {
+        aryDim.shift();    // remove first dimension because char array is presented as string
+    }
+
+    if (aryDim.length > 1) {
+        for(let i = 0; i < aryDim.length-1; i++) {
+            val = chunk(val, parseInt(aryDim[i]));
```
loitly, line 505: `parseInt` is quirky, meaning you may get unexpected results, e.g. `1x2` -> 1, `1.23` -> 1.
If you're ok with that, then it's fine.  Otherwise, we have a WebUtil.strictParseInt you can use instead.

loitly (approved):
Looks fine to me.

robyww (approved):

PR #1389, opened 2023-05-30
FIREFLY-1253: TAP Query mishandled redirects originating from an "/error" status

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1253

This is a little harder to test because we don't have authenticated TAP services that redirect its `/error` endpoint.  The code is fixed to handle all generic `GET` endpoints.  A test is also written to ensure redirecting is handled properly.

Test: https://fireflydev.ipac.caltech.edu/firefly-1253-tap-query-mishandle-redirects/firefly

Please test for regression bugs.


robyww (approved):
It looks good. I was a little surprise you decided to make such significant changes, but I am fine with it. It looks like you just decided to generalize what was happening in `UwsJobProcessor`

PR #1390, opened 2023-05-30
FIREFLY-1232: Remove StarTable from FITS Table Reader 

kpuriIpac:
#### [Firefly-1232](https://jira.ipac.caltech.edu/browse/FIREFLY-1232)
- The goal of this ticket is to get rid of StarTable as the intermediary step when converting a table from a FITS file to a DataGroup one. We can use hdu/TableHDU directly to create a DataGroup object. 
- The most important new methods are `readFitsTable` (to convert table to a DataGroup object), `addRowToDG`. and `convertHDUToDataType`. <s>I have sill kept calls to `getStarTable` and `convertStarTableToDataGroup` for testing purposes. I can remove that code when I convert this into a PR.</s> StarTable code is now all commented out (will be removed in PR). 
- The test build below is ready and stable and should yield expected results (same as on ops when you upload a FITs file and select and load table entries)
- Some of these changes are based on the `FitsStarTable` class from the uk.ac.starlink.fits as that's what we're trying to replace/get rid of 
- This is done. <s>TODO item (commented as such in the code, a second "TODO" item regarding use of ternary operator is less important so won't mention it here): 
   - at the end of `readFitsTable`, I need to add params info (key value pairs like **XTENSION: BINTABLE**, etc.) to the DataGroup TableMeta.</s>
- **Question**: The "strategy" variable is being passed around but is never really used anywhere in the StarTable to DataGroup conversion. So is it okay to get rid of it? 
- **_Note_**: I noticed a few unused functions @robyww. Are these okay to remove? 
   - `getAttributeValue, fillDataGroup, minValue`

**Updates 6/5 & 6/6**: 
 - some improvements & cleanup based on PR feedback from @robyww 

**Updates 6/8:** 
- Commented out all uses of starlink (can remove these completely later).
- Using **AbstractTableData** class directly instead of BinaryTable when getting "getData()" from TableHDU (this partially fixes ASCIITable and any other non-BinaryTabes not working)
  - there was a small bug in getValAsObject's switch statement. The last case needed to be "class java.lang.String" instead of "java.lang.String". I was never hitting this case before with my fits files, but hit it with the new AsciiTable file. Also, there was no "default" in this switch statement to throw an exception and log an error, so added that as well. 
- finally, wrote our own "**formatClass**" function. Need some feedback on the commented out part of the function from you @robyww. Maybe the commented out part easier to read but a bit longer than what I have now? 
   - checking if the class name contains "[" and then doing further checks does look a bit ugly, but it covers all cases for a given class type, so I thought it was best to combine them all into just one if/else block. 
- test build is updated 

**Updates 6/12:**
- bit more cleanup based on feedback, test build updated  

**Updates 6/13**
- removed old commented out code
- converted from draft PR to regular PR

Testing: 
- https://fireflydev.ipac.caltech.edu/firefly-1232-star-tablehdu/firefly
- I have tested with a few different FITs files
- Upload a fits file from the upload panel. Select any number of table entries and click "Load". 
- The results should be as expected (you may compare the results from ops/dev to ensure everything looks okay). 
- Try and test with different table entries from a few different FITs files if possible.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
@@ -9,36 +9,19 @@ import edu.caltech.ipac.table.DataGroup;
 import edu.caltech.ipac.table.DataObject;
 import edu.caltech.ipac.table.DataType;
 import edu.caltech.ipac.util.StringUtils;
-import edu.caltech.ipac.visualize.plot.ImagePt;
 import edu.caltech.ipac.visualize.plot.plotdata.FitsReadUtil;
-import nom.tam.fits.BasicHDU;
-import nom.tam.fits.Fits;
-import nom.tam.fits.FitsException;
-import nom.tam.fits.FitsFactory;
-import nom.tam.fits.Header;
-import nom.tam.fits.HeaderCard;
-import nom.tam.fits.ImageData;
-import nom.tam.fits.ImageHDU;
-import nom.tam.fits.TableHDU;
-import nom.tam.fits.UndefinedHDU;
+import nom.tam.fits.*;
 import nom.tam.image.compression.hdu.CompressedImageHDU;
 import nom.tam.util.ArrayFuncs;
-import uk.ac.starlink.fits.FitsStarTable;
-import uk.ac.starlink.table.ColumnInfo;
-import uk.ac.starlink.table.DefaultValueInfo;
-import uk.ac.starlink.table.DescribedValue;
-import uk.ac.starlink.table.StarTable;
+import nom.tam.util.Cursor;
 
 import java.io.IOException;
 import java.lang.reflect.Array;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+
+import static edu.caltech.ipac.util.StringUtils.isEmpty;
 
 /**
 * Convert an FITS file or FITS binary table(s) to list of DataGroup.
@@ -49,66 +32,30 @@ public final class FITSTableReader
     private static final Pattern TDISP = Pattern.compile("(A|I|B|O|Z|F|E|EN|ES|G|D)(\\d+)?(?:\\.(\\d+))?.*");
     private static final Pattern EXPONENTIAL = Pattern.compile("E|EN|ES|D");                                    // Table 20 from https://fits.gsfc.nasa.gov/standard30/fits_standard30aa.pdf
 
-
     public static boolean debug = true;
 
     /**
      * Strategies to handle FITS data with column-repeat-count greater than 1(TFORMn = rT):
      *  where n = column index, T = data type, r = repeat count
-     *
+
      * DEFAULT: Set DataType to 'ary'.  Store the full array in DataGroup as an object.
      * When DataGroup is written out into IPAC format, it should only describe the data
      * as type = 'char' and value as type[length].  This is the default strategy if not given.
-     *
+
      * TOP_MOST: Ignore the repeat count portion of the TFORMn Keyword
      * returning only the first datum of the field, even if repeat count is more than 1.
      * This should produce exactly one DataGroup per table.
-     *
-     *
-     *
+
      * ( experiemental.. not sure how it would be use )
      * FULLY_FLATTEN: Generates one DataGroup row for each value of an HDU field.
      * Because each field may have different repeat count (dimension), insert blank
      * when no data is available. This should produce exactly one DataGroup per table.
      * No attribute data added to DataGroup with "FULLY_FLATTEN". Should pass headerCols = null.
-     *
+
      * EXPAND_BEST_FIT: Expands each HDU row into one DataGroup. Fields with lesser count (dimension) will be filled with blanks.
      * EXPAND_REPEAT: Expands each HDU row into one DataGroup. Fields with lesser dimension will be filled with previous values.
      */
     public static final String DEFAULT = "DEFAULT";
-    public static final String TOP_MOST = "TOP_MOST";
-    public static final String FULLY_FLATTEN = "FULLY_FLATTEN";
-    public static final String EXPAND_BEST_FIT = "EXPAND_BEST_FIT";
-    public static final String EXPAND_REPEAT = "EXPAND_REPEAT";
-
-    private static final int maxNumAttributeValues = 80;
-
-    public static void main(String[] args)
-    {
-        String fits_filename = "/Users/ejoliet/devspace/branch/dev/firefly/test-array.fits";//args[0];
-
-        Fits fits= null;
-        try {
-            fits = new Fits(fits_filename);
-            BasicHDU<?>[] hdus = FitsReadUtil.readHDUs(fits);
-            StarTable table = getStarTable(hdus[1], fits_filename, 1);
-            DataGroup data = FITSTableReader.convertFitsToDataGroup(fits_filename, null, FITSTableReader.DEFAULT, 1);
-
-            int columnCount = table.getColumnCount();
-            for (int a= 0; a<columnCount;a++) {
-                ColumnInfo columnInfo = table.getColumnInfo(a);
-                int[] shape = columnInfo.getShape();
-                System.out.println(columnInfo.getName() + " shape length " + shape.length);
-            }
-        } catch (FitsException | IOException | NullPointerException e) {
-            e.printStackTrace();
-        }
-        finally {
-            FitsReadUtil.closeFits(fits);
-        }
-        System.exit(0);
-    }
-
 
     public static DataGroup convertFitsToDataGroup(String fits_filename, Map<String,String> metaInfo,
                                                    String strategy, int table_idx) throws FitsException, IOException {
@@ -118,9 +65,9 @@ public final class FITSTableReader
     /**
      * Convert a table from a FITS file to DataGroup based on table index
      * @param fits_filename the file name
-     * @param dataCols: The names of the columns which will be copied to the data section of the DataGroups.
+     * @param dataCols The names of the columns which will be copied to the data section of the DataGroups.
      *                If dataCols = null, get all the columns into the data group.
-     * @param headerCols: The names of the columns which will be copied to the header section of the DataGroups.
+     * @param headerCols The names of the columns which will be copied to the header section of the DataGroups.
      *                If headerCols = null, get none of the columns into the header of the data group.
      * @param strategy The strategy used to deal with the repeat count  of the data in the given dataCols columns.
      * @param table_idx table index, i.e. HDU number in FITS
@@ -139,7 +86,7 @@ public final class FITSTableReader
             // disable long string for HeaderCard creation while collecting table with table_idx from StarTableFactory to work around
             // the exception error sent from nom.tam.fits.
             FitsFactory.useThreadLocalSettings(true);
-            DataGroup dg;
+            DataGroup result;
             fits = new Fits(fits_filename);
             BasicHDU<?>[] hdus = FitsReadUtil.readHDUs(fits);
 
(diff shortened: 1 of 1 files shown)


src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -295,6 +286,154 @@ private static void logTableReadError(String fitsFilename, int tableIdx, String
...
+            }
+            return base.getClass();
+        }
+    }
+
+    private static DataGroup readFitsTable(BasicHDU<?> hdu,
+                                          String fitsFilename,
+                                          String[] inclCols,
+                                          String[] inclHeaders,
+                                          String strategy) throws IOException, FitsException {
```
loitly, line 329: In this new function, `inclCols` and `inclHeaders` are no longer implemented.  `metaInfo` is removed completely.  And, test is failing.
```
edu.caltech.ipac.firefly.server.util.tables.DataGroupReadTest > fitsTable FAILED
    java.lang.ClassCastException at DataGroupReadTest.java:93
```

kpuriIpac: Yes, inclHeaders will be used for one of my TODOs...(assigning DataGroup meta at the end of this function). and I also just noticed test failing, thanks! Will look into it.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -395,6 +535,114 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
         return dataGroup;
     }
 
+
+    //This function is based on the packageValue function from the FitsStarTable class in the uk.ac.starlink.fits package
+    private static Object getValAsObject(Object base, int icol, boolean[] hasBlank, long[] blanks, boolean[] isScaled,
+                                         double[] scales, double[] zeros) {
+        if (base == null) {
```
loitly, line 542: This function plus `getClassType` is pretty much the bulk of FitsStarTable.java.  I am not sure how this is an improvement unless we decided to abandon starlink completely.  But, if this is the decision, then I suggest writing more unit tests to ensure the commonly encountered tables are read in correctly.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -150,23 +139,25 @@ public static DataGroup convertFitsToDataGroup(String fits_filename,
...
                 FitsFactory.setLongStringsEnabled(true);
-                dg=  (table!=null) ?
-                        convertStarTableToDataGroup(table, dataCols, headerCols, strategy) :
-                        getFitsImageAsTable(hdu,metaInfo);
+                dg=  (table!=null) ? convertStarTableToDataGroup(table, dataCols, headerCols, strategy) : getFitsImageAsTable(hdu,metaInfo);
             }
             String dataTypeHint= metaInfo !=null ? metaInfo.getOrDefault(MetaConst.DATA_TYPE_HINT,"").toLowerCase() : "";
             SpectrumMetaInspector.searchForSpectrum(dg,hdus[table_idx], dataTypeHint.equals("spectrum"));
-            return dg;
+            SpectrumMetaInspector.searchForSpectrum(result,hdus[table_idx], dataTypeHint.equals("spectrum"));
```
loitly, line 159: I am pretty sure one of these two lines are redundant.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -150,23 +139,25 @@ public static DataGroup convertFitsToDataGroup(String fits_filename,
...
-                    dg= convertStarTableToDataGroup(table, dataCols, headerCols, strategy);
+                    //StarTable table = getStarTable(hdus[1], fits_filename, table_idx); // this is really a hack, maybe it should not be here.
+                    result = readFitsTable(hdu, fits_filename, dataCols, headerCols, strategy);
+                    //dg= convertStarTableToDataGroup(table, dataCols, headerCols, strategy);
                 }
             }
             else { // if >0 then try to read it as a FITS table first.
                 FitsFactory.setLongStringsEnabled(false);
                 StarTable table = getStarTable(hdu, fits_filename, table_idx);
+                result = readFitsTable(hdu, fits_filename, dataCols, headerCols, strategy);
```
loitly, line 153: Forgot to comment out `StarTable table = getStarTable(hdu, fits_filename, table_idx);`?

kpuriIpac: Yes I've kept the StarTable stuff only for the draft PR..but I'll remove the code (including the `getStarTable` and `converStarTableToDataGroup` functions and their dependent functions as well) when I turn this into a PR (probably in next push to the branch)

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -393,9 +556,117 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
...
+                long blank = blanks[icol];
+                boolean scaled = isScaled[icol];
+                double scale = scales[icol];
+                double zero = zeros[icol];
+                if (cls == Byte.TYPE) {
+                    byte val = ((byte[])(base))[0];
+                    /*
+                    TODO: Attempt to use ternary operator here instead
+                     */
+                    if (hasblank && val == (byte)((int)blank))
```
robyww, line 562: you don't need to cast blank to do comparisons. Also since this is so repeated you might want to put it on one line.
```java
if (hasblank && val == blank) return null;
```

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -393,9 +556,117 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
...
+                        return null;
+                    if (scaled)
+                        return Double.valueOf((double)val * scale + zero);
+                    else
+                        return Byte.valueOf(val);
+                }
```
robyww, line 568: You might want to try a switch. Combine with the other suggestions. This would make the whole thing look less verbose and be easier to read.
```java
switch (cls.toString()) {
                    case "byte" -> {
                        byte val = ((byte[])(elem))[0];
                        if (hasblank && val == blank) return null;
                        return scaled ? val * scale + zero : val;
                    }
                    case "short" -> {
                        short val = ((short[])(elem))[0];
                        if (hasblank && val == blank) return null;
                        return scaled ? val * scale + zero : val;
                    }
                }
```

robyww: no in this case, string is `"string java.lang.String"`

robyww (commented):
Here are a few comments. I still have more to review, but I am out of time.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -393,9 +556,117 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
...
+                    byte val = ((byte[])(base))[0];
+                    /*
+                    TODO: Attempt to use ternary operator here instead
+                     */
+                    if (hasblank && val == (byte)((int)blank))
+                        return null;
+                    if (scaled)
+                        return Double.valueOf((double)val * scale + zero);
+                    else
+                        return Byte.valueOf(val);
```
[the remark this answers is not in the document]
kpuriIpac: @robyww I made all of the other changes, but I tested this and think using the ternary operator here wouldn't work (because I am trying to give it two different return types). For the following line: `return scaled ? val * scale + zero : val;` - even is **scaled** is **false** and **val** is, say, **byte** or **int** or **short**, **val** will get cast to **double** implicitly if we use a ternary operator here (because `val * scale + zero` is **double**). So I had to leave the if/else as is. 

But removing the explicit class conversion and letting it autobox works as you suggested.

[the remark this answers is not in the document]
robyww: That really surprises me. Have you tried calling the function in the debugger to confirm?

You also might try
```java
return scaled ? (double)(val * scale + zero) : (int)val;
```

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
robyww: More comments. I think blanks only applies to `byte`, `int`, `long` you don't need it for `double` and `float`

robyww: ok, better idea.

write a record class:
```java
    record EvalVal(long blank, boolean scaled, boolean hasblank, double scale, double zero) {
        public Number evalValue(Number val) {
            if (hasblank && val.doubleValue() == blank) return null;
            if (!scaled) return val;
            return val.doubleValue() * scale + zero;
        }
    }
```

in `getValAsObject` (line 582)
```java
EvalVal evaluator= new EvalVal(blanks[icol], isScaled[icol], hasBlank[icol], scales[icol], zeros[icol]);
```

every switch (float example):
```java
 return evaluator.evalValue(((float[])(elem))[0]);
```


```

Comment (kpuriIpac) [reply src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java]: Thanks, great suggestion. Code looks a lot simpler now for `getValAsObject`.

File: src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java [hunk 8]
```diff
@@ -393,9 +540,94 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
         }
         dataGroup.trimToSize();
         return dataGroup;
+    }*/
+
+    static class EvalVal {
```
robyww, line 545: You can make this a `record`. It will take half the space.

kpuriIpac: Ah yes, you mentioned it above too, my bad. Changed to a record and pushed code into existing commit.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -393,9 +540,86 @@ private static DataGroup convertStarTableToDataGroup(StarTable table,
...
+                }
+            }
+
+            if (cls != null && cls.isArray()) {
+                return ArrayFuncs.flatten(elem);
+            } else {
+                return elem instanceof String && isEmpty((String)elem) ? null : elem;
+            }
+        }
+    }
```
robyww, line 603: `getValAsObj` is not clear. After a *lot* study. I realized it is about 4 four cases of `elem`
- `elem` is null
- `elem` is not an array
- `elem` is an array of 1
- `elem` is an array > 1

Here is my propose re-write:
```java
    private static Object getValAsObject(Object elem, int icol, boolean[] hasBlank, long[] blanks, boolean[] isScaled,
                                         double[] scales, double[] zeros) throws FitsException {
        if (elem == null) {
            return null;
        }
        else if (!elem.getClass().isArray()) {
            return elem instanceof String && isEmpty((String)elem) ? null : elem;
        }
        else if (Array.getLength(elem) == 1) {
            String cls = elem.getClass().getComponentType().toString();
            EvalVal evaluator = new EvalVal(blanks[icol], isScaled[icol], hasBlank[icol], scales[icol], zeros[icol]);
            return switch (cls) {
                case "byte" -> evaluator.evalValue(((byte[])(elem))[0]);
                case "short" -> evaluator.evalValue(((short[])(elem))[0]);
                case "int" -> evaluator.evalValue(((int[])(elem))[0]);
                case "long" -> evaluator.evalValue(((long[])(elem))[0]);
                case "float" -> evaluator.evalValue(((float[])(elem))[0]);
                case "double" -> evaluator.evalValue(((double[])(elem))[0]);
                case "boolean" -> Boolean.valueOf(((boolean[])(elem))[0]);
                case "class java.lang.String" -> isEmpty(((String[])(elem))[0]) ? null : ((String[])(elem))[0];
                default -> throw new FitsException( "Unrecognized class type in FITS table file entry: " + cls);
            };
        }
        else {
            return ArrayFuncs.flatten(elem);
        }
    }
```

robyww (commented):
more comments. I am still looking for more fits to test.

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -410,9 +634,133 @@ private static void addRowToDataGroup(StarTable table, DataGroup dataGroup, Link
         } catch (IOException e) {
             logger.error("Unable to read StarTable row:" + rowIdx + "   msg:" + e.getMessage());
         }
+    }*/
+
+    public static Class formatClass(Class c) throws FitsException {
+        String cname = c.getName();
+        if (cname == null) {
```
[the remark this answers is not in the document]
kpuriIpac: @robyww I didn't get a warning from IntelliJ here. And I don't see the `@NotNull` annotation here explicitly, should I add it to the argument?

src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
```diff
@@ -410,9 +634,133 @@ private static void addRowToDataGroup(StarTable table, DataGroup dataGroup, Link
         } catch (IOException e) {
             logger.error("Unable to read StarTable row:" + rowIdx + "   msg:" + e.getMessage());
         }
+    }*/
+
+    public static Class formatClass(Class c) throws FitsException {
```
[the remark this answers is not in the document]
kpuriIpac: didn't get these warnings, will take a look at my settings!

PR #1392, opened 2023-06-06
Firefly-1252: spacial missing

robyww:
#### [Firefly-1252](https://jira.ipac.caltech.edu/browse/FIREFLY-1252): fixed issues with there are not spacial columns

  - changed the processing of tap.additional.services to better reflex the config orider
  - added hide property to tap.additional.services
  - temporal column check

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1252-spacial-missing/firefly
 - test according to the check
 - `tap.additional.services` and temporal column check changes are hard to test outside the Rubin environment

### DIFF
--- a/src/firefly/js/ui/tap/SpatialSearch.jsx
+++ b/src/firefly/js/ui/tap/SpatialSearch.jsx
@@ -127,13 +127,14 @@ export function SpatialSearch({cols, serviceUrl, serviceLabel, columnsModel, tab
         }
     }, [searchParams.radiusInArcSec, searchParams.wp, searchParams.corners, searchParams.uploadInfo]);
 
-
     useEffect(() => {
         const {lon,lat} = formCenterColumns(columnsModel);
         setVal(CenterLonColumns, lon, {validator: getColValidator(cols, true, false), valid: true});
         setVal(CenterLatColumns, lat, {validator: getColValidator(cols, true, false), valid: true});
         const noDefaults= !lon || !lat;
         setVal(posOpenKey, (noDefaults) ? 'open' : 'closed');
+        if (noDefaults) checkHeaderCtl.setPanelActive(false);
+        checkHeaderCtl.setPanelOpen(!noDefaults);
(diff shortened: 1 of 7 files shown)


loitly (approved):
The reported bug is fixed.  Changes look good.  Btw, I added a commit to fix a regression bug when TAP results is not at the first item in the /results list.

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -717,6 +718,9 @@ function makeSpatialConstraints(columnsModel, obsCoreEnabled, fldObj, uploadInfo
         const cenLat= cenLatField?.value;
         errList.checkForError(cenLonField);
         errList.checkForError(cenLatField);
+        if (!cenLon && !cenLat) errList.addError('Lon and Lat columns are not set');
```
gpdf, line 721: I'm not seeing this message displayed on the current data-int deployment; instead, I'm still seeing "Can not be empty.  Please provide value or expression"

src/firefly/js/ui/tap/TapUtil.js
```diff
@@ -287,13 +287,20 @@ export const getAsEntryForTableName= (tableName) => tableName?.[0] ?? 'x';
...
 
-    const modifiedOriginal= tapServices.map( (t) => {
-        const match= additional.find( (a) => a.label===t.label);
-        return match ? {...t,...match} : t;
-    });
-    const trulyAdditional= additional.filter( (a) => !tapServices.find( (t) => t.label===a.label) );
-
-    return [...trulyAdditional,...modifiedOriginal];
+    const mergeAdditional= additional.map( (a) => {
+        if (a.hide) return false;
```
gpdf, line 291: I'm not sure how to use "hide" in the JSON, and it looks like I'll want to in order to provide the cleanest user experience.  Ca you provide an example?

gpdf (commented):
One thing isn't working, on the data-int deployment, as the code above suggests it should: the missing-column-name error message.

PR #1399, opened 2023-06-21
Firefly-1263: optimize fits table read

robyww:
#### [Firefly-1263](https://jira.ipac.caltech.edu/browse/FIREFLY-1263): optimize fits table read

   - Make some simple changes to optimize read very large fits table.

#### Testing

  - https://fireflydev.ipac.caltech.edu/firefly-1263-optimize-fits-table-read/firefly
  - read a large fits table

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/FITSTableReader.java
@@ -340,8 +340,9 @@ public final class FITSTableReader
         DataGroup dataGroup = new DataGroup(fitsFilename, dataTypes);
         // creating DataGroup rows.
         dataGroup.setInitCapacity(nrow);
+        DataType[] dataDefinitions= dataGroup.getDataDefinitions();
         for (int row = 0; row < nrow; row++){
-            addRowToDG(dataGroup, row, hduTable, hasBlank, blanks, isScaled, scales, zeros);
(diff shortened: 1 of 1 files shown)


kpuriIpac:
The code change looks good (should've thought of this!). But on the test build, I get the following error trying to load the big table (from the big 3.5 GB fits file that Gregory linked to: 71 cols x 8633452 rows): 

```
Table Load Error:
Error fetching https://fireflydev.ipac.caltech.edu/firefly-1263-optimize-fits-table-read/firefly/CmdSrv/sync?cmd=tableSearch: code: 502, text:
```

Perhaps the test build needs to have a higher memory limit? (4Gi or 8Gi). 

Trying to test it locally and it doesn't crash, but is stuck in seemingly an endless loop of trying to load, will add breakpoints and try to make it work, and I can report back on what I find. (seems like it runs through all the rows as expected, but doesn't render the table after that for some reason - this might be something on my end)..

robyww:
I think it is memory, but we could give better errors. I was able to load it with a large server (16 GB)

kpuriIpac:
> I think it is memory, but we could give better errors. I was able to load it with a large server (16 GB)

Yes a better error message would help. It appears my local Tomcat is also set to max memory of 16 GB, but this file still doesn't load for me (no error though, so I'm still poking to try and find out why it wouldn't load - it creates the entire **DataGroup** correctly and returns the **result** as well)

robyww:
We can pursuit a better message in a later PR. This work is really just about optimizing the FITS reading part.

PR #1401, opened 2023-06-28
Firefly 1150 object id tap

kpuriIpac:
#### [Firefly-1150](https://jira.ipac.caltech.edu/browse/FIREFLY-1150)
- Added a new Search Helper (ObjectID Search) to the TAP panel under services that support uploads 
- allow user to upload a table and select a column from their uploaded table, and a column from the selected table on the right. This will lead to a new ADQL constraints of the format '`ut.objectID = a.objectID`' (ut = uploaded table, a = first char of the selected table in this case) 
- This is a draft PR because I need to fix one small style issue and at possibly this bug below:
- Bug Description 
  - This is a regression bug. Switching back from **multi-object** to **Single Shape** leads to some incorrect ADQL. For a service, like GAIA (and table is "public.hipparcos"), if under **Spatial** you select **Multi-object** and upload a table. Now if you select **Single Shape** and enter in, say, M31, and click on **Populate and Edit ADQL**. Notice that the **WHERE** clause includes "p.ra" and "p.de", even though the **FROM** clause is just: `FROM public.hipparcos` and not `FROM public.hipparcos as p` (as it would be if there were an uploaded table). 
      - If you try to do a search now, you'd get an error which would include "`Unknown column "p.ra"`" and "`Unknown column "p.de"`"...
      - for some reason, this same behavior does not affect some of the tables under IRSA/CADC. So will need to test this further, but the basic idea of the fix is that if you switch from multi-object to single shape and there's no uploaded table for that search, then "**p.ra**" and "**p.de**" should simply be **ra** and **de** 

**Updates 7/6**: 
   - The Temporal Column and Spatial Type bugs have been fixed
   - For bug description, see @loitly's main comment on this PR below 
       - **Spatial Search bug**
          - multi-object selection was very persistent. This is fixed now by commenting out a use-effect (I don't think there's any need for it anymore, but commented out just so reviewers can take a look at it). Whatever Spatial Type you select now (single shape or multi-object), the selection is respected when you switch services or go to `Edit ADQL` and back to `UI Assisted` mode.  
          - the **_exception_** to this bug is, if you do a whole table search (or "search TAP for row" from the table), then the useEffect that gets triggered (line 103 in `SpatialSearch.jsx`) has ones of its dependencies: _searchParams.uploadInfo_. _searchParams_ comes, as part of _initArgs_, all the way from defaultSearchActions.js (see the `searchWholeTable` function in this file as an example). This useEffect sets `SPATIAL_TYPE` to `MULTI`, which is good the first time a user does a search from a table. But after that, user may switch to single shape and go to `Edit ADQL` or a different service. And if you go back to the previous service or `UI Assisted` mode, multi-object is selected again. I am not really sure what a good fix for this is. 
      - **Temporal Search bug**
        - see my comment below for explanation of this commit 
  -  test build is updated, feel free to try it out!

**Updates 7/10**: 
- there will still some issues with Temporal and Spatial search bugs, as described by @loitly below. These are all fixed now and the test build is updated to reflect these changes. 
- Also includes @robyww's commit to fix `Edit ADQL` and make it sticky (if you select `Edit ADQL`, do a search and click on `TAP` again, it should take you to the `Edit ADQL` mode and not the UI Assisted mode). 

**Updates 7/10 Part 2**:
  - The UI changes to ObjectID Search Helper suggested by @robyww are in. Object ID Search helper is selected when you interact with it (upload a table, or change the column). And the Object ID columns are now open by default as well. 
  - There were a couple of bugs I noticed: The `Uploaded Object ID` field was not validating column entries. And the `Object ID (from the selected table on the right)` field was not updating its columnsModel for the column validator when you switched services/tables correctly. 
  - Fixed these and now the Object ID Search Helper is fully functional and has expected behavior (and bug free, from what I have tested). 

**Updates 7/11**:
- got some of the final PR comments in, involves some code cleanup. Using JSdoc/typdef, and simplifying a for loop to use filter instead, amongst other changes. 

**Testing**: 
- https://fireflydev.ipac.caltech.edu/firefly-1150-objectid-tap/firefly
- for any service that supports uploads (IRSA, CADC, etc.), upload a table using the ObjectID Search helper in the TAP panel and select an ObjecyID column for your uploaded table and for the selected table on the right. For testing purposes, this could be any column you want. 
  - Click on **Populate and edit ADQL**. Make sure the ADQL looks correct. 
  - Ideally, select a service and table that is not too big. This is an issue i've been running into for testing. Because of the `WHERE ut.objectID = a.objectID` type of query, I think every row of the selected table on the right is being checked, which makes sense. So even if you add a row limit, if the table is big enough, the search will be slow. I tried the **CADC's ivoa.ObsCore** table (which has  166,523 rows) and so far that gave me the fastest results... 
- Seelct IRSA as the TAP service. Upload a table under the **Spatial** search helper, and also the **ObjectID Search** helper (make sure both of these are selected as well). Click on **Populate and edit ADQL**. This should give you an error, because we don't support multiple uploads at this time. 
- switch to a service like NED, make sure the **ObjectID Search** helper doesn't show up here

### DIFF
--- a/src/firefly/js/ui/UploadTableChooser.js
+++ b/src/firefly/js/ui/UploadTableChooser.js
@@ -63,7 +63,7 @@ function getFitsColumnInfo(data) {
 }
 
 
-function uploadSubmit(request, setUploadInfo)  {
+function uploadSubmit(request,setUploadInfo,defaultColsEnabled)  {
     if (!request) return false;
     const {additionalParams = {}, fileUpload: serverFile} = request;
     const {detailsModel, report, summaryModel} = additionalParams;
@@ -81,14 +81,14 @@ function uploadSubmit(request, setUploadInfo)  {
     const columns= report.fileFormat===Format.FITS ?
         getFitsColumnInfo(data) :
         data.map(([name,type,u,d]) => ({name, type, units: u?u:'', description: d?d:'', use:true}));
-    const columnsSelected = getSelectedColumns(columns);
+    const columnsSelected = getSelectedColumns(columns,defaultColsEnabled);
     const uploadInfo = {serverFile, fileName, columns:columnsSelected, totalRows, fileSize, tableSource: UPLOAD_TBL_SOURCE};
     setUploadInfo(uploadInfo);
     dispatchHideDialog(dialogId);
     return false;
 }
 
-function existingTableSubmit(request,setUploadInfo) {
+function existingTableSubmit(request,setUploadInfo,defaultColsEnabled) {
     if (!request) return false;
     const tbl = getTblById('existing-table-list-ui');
     const idx = tbl.highlightedRow;
@@ -97,7 +97,7 @@ function existingTableSubmit(request,setUploadInfo) {
     const tableRequest = activeTbl.request;
     const columnData = activeTbl.columns;
     const columns = columnData.map((col) => col.visibility === 'hide' || col.visibility === 'hidden'? ({...col, use:false}) :  ({...col, use:true})); //filter out hidden cols
-    const columnsSelected = getSelectedColumns(columns);
+    const columnsSelected = getSelectedColumns(columns,defaultColsEnabled);
 
     const params ={
         [ServerParams.COMMAND]: ServerParams.TABLE_SAVE,
@@ -127,9 +127,37 @@ function existingTableSubmit(request,setUploadInfo) {
     return false;
 }
 
-function getSelectedColumns(columns) {
-    const {lonCol='', latCol=''} = findTableCenterColumns({tableData:{columns}}) ?? {}; //centerCols
-    const columnsSelected = columns.map((col) => col.name === lonCol || col.name === latCol? ({...col, use:true}) :  ({...col, use:false})); //select position cols only
+/*
+    if showUploadTableChooser is called with a defaultColsEnabled object, use the colTypes and colCount to set the selected columns
+    for this uploaded table
+ */
+function defaultColumnsSelector(columns,colTypes,colCount) {
+    let cols = [];
+    if  (columns == null) return null;
+    let count = 0;
+    for (let i=0; i< columns?.length; i++) {
+        if (count < colCount && colTypes.includes(columns[i].type)) {
+            cols.push(columns[i]);
+            count ++;
+        }
+    }
+    cols = cols.map((col) => col.name.replace(/^"(.*)"$/, '$1'));
+    const defautltCols = columns?.map((col) => (
+        {...col, use:cols.includes((col.name))}));
+
+    return defautltCols;
+}
+
+function getSelectedColumns(columns,defaultColsEnabled) {
+    let columnsSelected;
+    if (defaultColsEnabled != null) { //default cols enabled
+        const colTypes = defaultColsEnabled.colTypes;//['int', 'long', 'string'];
+        columnsSelected = defaultColumnsSelector(columns,colTypes,defaultColsEnabled.colCount);
+    }
+    else {
+        const {lonCol='', latCol=''} = findTableCenterColumns({tableData:{columns}}) ?? {}; //centerCols
+        columnsSelected = columns.map((col) => col.name === lonCol || col.name === latCol? ({...col, use:true}) :  ({...col, use:false})); //select position cols only
+    }
     return columnsSelected;
 }
 
@@ -187,18 +215,18 @@ const LoadedTables= (props) => {
         </div>);
 };
 
-
-export function showUploadTableChooser(setUploadInfo) {
+//showUploadTableChooser may be called with defaultColsEnabledObj, which will replace the regular default selection of lon and lat cols (usually ra,dec)
+export function showUploadTableChooser(setUploadInfo,groupKey= 'table-chooser',defaultColsEnabledObj=null) {
     DialogRootContainer.defineDialog(dialogId,
         <PopupPanel title={'Upload'} layoutPosition={LayoutType.TOP_EDGE_CENTER}>
-            <TapUploadPanel {...{setUploadInfo}}/>
+            <TapUploadPanel {...{setUploadInfo,groupKey,defaultColsEnabledObj}}/>
         </PopupPanel>
     );
     dispatchShowDialog(dialogId);
 }
 
-const TapUploadPanel= ({setUploadInfo,groupKey= 'table-chooser'}) => {
-   return (
+const TapUploadPanel= ({setUploadInfo,groupKey= 'table-chooser',defaultColsEnabledObj}) => {
+    return (
     <FieldGroup groupKey={groupKey}>
         <div style={{ resize: 'both', overflow: 'hidden', zIndex: 1, paddingTop:8, minWidth: 600, minHeight: 500, display: 'flex' }}>
             <FieldGroupTabs initialState={{value: 'upload'}} fieldKey='upload-type-tabs' groupKey={groupKey}
@@ -208,14 +236,14 @@ const TapUploadPanel= ({setUploadInfo,groupKey= 'table-chooser'}) => {
                         style:{height: '100%'},
                         acceptOneItem:true, acceptList:[TABLES], keepState:true, groupKey:groupKey+'-fileUpload',
                         onCancel:() => dispatchHideDialog(dialogId),
-                        onSubmit:(request) => uploadSubmit(request,setUploadInfo),
+                        onSubmit:(request) => uploadSubmit(request,setUploadInfo,defaultColsEnabledObj),
                     }}/>
                 </Tab>
                 <Tab name='Loaded Tables' id='tableLoad' style={{fontSize:'larger'}}>
                         <LoadedTables {...{
                             style:{height: '100%', width:'100%'}, keepState: true, groupKey:groupKey+'-tableLoad',
                             onCancel:() => dispatchHideDialog(dialogId),
-                            onSubmit:(request) => existingTableSubmit(request, setUploadInfo)
+                            onSubmit:(request) => existingTableSubmit(request,setUploadInfo,defaultColsEnabledObj)
                         }}/>
                 </Tab>
             </FieldGroupTabs>
--- a/src/firefly/js/ui/UploadTableChooser.js
+++ b/src/firefly/js/ui/UploadTableChooser.js
@@ -62,7 +62,13 @@ function getFitsColumnInfo(data) {
         });
 }
 
-
+/**
+ * handle submit for an uploaded table
+ * @param request
+ * @param setUploadInfo
+ * @param {DefaultColsEnabled} defaultColsEnabled
+ * @returns {boolean}
+ */
 function uploadSubmit(request,setUploadInfo,defaultColsEnabled)  {
     if (!request) return false;
     const {additionalParams = {}, fileUpload: serverFile} = request;
@@ -81,13 +87,20 @@ function uploadSubmit(request,setUploadInfo,defaultColsEnabled)  {
     const columns= report.fileFormat===Format.FITS ?
         getFitsColumnInfo(data) :
         data.map(([name,type,u,d]) => ({name, type, units: u?u:'', description: d?d:'', use:true}));
-    const columnsSelected = getSelectedColumns(columns,defaultColsEnabled);
+    const columnsSelected = applyDefColumnSelection(columns,defaultColsEnabled);
     const uploadInfo = {serverFile, fileName, columns:columnsSelected, totalRows, fileSize, tableSource: UPLOAD_TBL_SOURCE};
     setUploadInfo(uploadInfo);
     dispatchHideDialog(dialogId);
     return false;
 }
 
+/**
+ * handle submit for an existing table
+ * @param request
+ * @param setUploadInfo
+ * @param {DefaultColsEnabled} defaultColsEnabled
+ * @returns {boolean}
+ */
 function existingTableSubmit(request,setUploadInfo,defaultColsEnabled) {
     if (!request) return false;
     const tbl = getTblById('existing-table-list-ui');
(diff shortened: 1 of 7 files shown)


src/firefly/js/ui/tap/TapSearchRootPanel.jsx
```diff
@@ -424,6 +424,17 @@ function getAdqlQuery(tapBrowserState, showErrors= true) {
...
+    if (isUpload) { //check for more than one upload file (in Spatial and in ObjectID col) - should this be a utility function in constraints.js?
+        const { constraintFragments } = tapBrowserState;
+        const entries = [...constraintFragments.values()];
+        const matchingEntries = entries.filter((c) => Boolean(c.uploadFile && c.TAP_UPLOAD && c.adqlConstraint));
+        if (matchingEntries.length > 1) {
+            if (showErrors) showInfoPopup('More than one upload is not supported at this moment.', 'Error');
+            return;
+        }
+    }
+
```
kpuriIpac, line 437: Does it make sense to have this here? or should it be a utility function in constraints.js? (similar to isTapUpload on line 426 - or I could even combine the 2 and return an object with 2 entries: one would indicate true/false for isTapUpload, and the other would include the number of matching entries (so more than 1 upload would be flagged as false)

src/firefly/js/ui/tap/TapSearchRootPanel.jsx
```diff
@@ -424,6 +424,17 @@ function getAdqlQuery(tapBrowserState, showErrors= true) {
     const tableName = maybeQuote(tapBrowserState?.tableName, true);
     if (!tableName) return;
     const isUpload= isTapUpload(tapBrowserState);
+
+    if (isUpload) { //check for more than one upload file (in Spatial and in ObjectID col) - should this be a utility function in constraints.js?
+        const { constraintFragments } = tapBrowserState;
```
kpuriIpac, line 429: Does it make sense to have this here? or should it be a utility function in constraints.js? (similar to isTapUpload on line 426 - or I could even combine the 2 and return an object with 2 entries: one would indicate true/false for isTapUpload, and the other would include the number of matching entries (so more than 1 upload would be flagged as false)

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -81,14 +81,14 @@ function uploadSubmit(request, setUploadInfo)  {
     const columns= report.fileFormat===Format.FITS ?
         getFitsColumnInfo(data) :
         data.map(([name,type,u,d]) => ({name, type, units: u?u:'', description: d?d:'', use:true}));
-    const columnsSelected = getSelectedColumns(columns);
+    const columnsSelected = getSelectedColumns(columns,defaultColsEnabled);
```
loitly, line 84: Assuming you take the suggested function approach and called parameter `defaultColsSelector`, then this line can be...
```
const columnsSelected = defaultColsSelector?.(columns) || applyDefColumnSelection(columns);
```

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -127,9 +127,37 @@ function existingTableSubmit(request,setUploadInfo) {
...
+    let cols = [];
+    if  (columns == null) return null;
+    let count = 0;
+    for (let i=0; i< columns?.length; i++) {
+        if (count < colCount && colTypes.includes(columns[i].type)) {
+            cols.push(columns[i]);
+            count ++;
+        }
+    }
+    cols = cols.map((col) => col.name.replace(/^"(.*)"$/, '$1'));
```
loitly, line 144: Not 100% what you want here, I am assuming you want the first 2 column with the given `colTypes`.  There are existing util function in TableUtil.js you can use.  `getColsByType` and `stripColumnNameQuotes`.  These functions includes logic that you missed, like hidden column and column names with double-quotes in them.

loitly (commented):
I am only adding a few suggested comments.  These are from browsing through the code.  Use them at your discretion.

src/firefly/js/ui/tap/TemportalSearch.jsx
```diff
@@ -129,8 +130,10 @@ export function TemporalSearch({cols, columnsModel}) {
     useEffect(() => {
         const timeCol= findTimeColumn(columnsModel) ?? '';
         const errMsg= 'Temporal searches require identifying a table column containing a time in MJD.  Please provide a column name.';
-        setVal(TemporalColumns, timeCol, {validator: getColValidator(cols, true, false, errMsg), valid: true});
-        checkHeaderCtl.setPanelOpen(Boolean(timeCol));
+        //if (timeCol != '') setVal(TemporalColumns, timeCol, {validator: getColValidator(cols, true, false, errMsg), valid: true});
+        //setVal(TemporalColumns, timeCol, {validator: getColValidator(cols, true, false, errMsg), valid: true});
+        if (Boolean(timeCol)) checkHeaderCtl.setPanelOpen(true);
+        //checkHeaderCtl.setPanelOpen(Boolean(timeCol));
```
kpuriIpac, line 136: It's my understanding that we were setting the temporal panel to open only if the columnsModel had a time column (as returned by the **findTimeColumn** function) and also setting the temporal column value to this timeCol value returned. That's definitely good UX when timeCol exists but it leads to a few issues when it does not. 

This resulted in the panel automatically closing if you open it, leave the TAP panel (say you go to Edit ADQL) and come back to UI Assisted mode and the Temporal search panel is closed again. That's fixed by the if statement on line 135 here. 

The second issue here is that we were trying to setVal of the **TemporalColumns**  to **timeCol**, which is why even if you selected a temporal column value, and navigated away (to **Edit ADQL**) and back to **UI Assisted**, you would lose your selected column because **timeCol** may be empty.

A potential solution for this is on line 133, we check if a **timeCol** value exists, and only then set the temporal column's value to it. But an issue I find with this is that if the user selects a _different_ temporal column value, and they navigate away from the page and back to it, their selected value will be replaced by the default **timeCol** value returned by **findTimeColumn**. I suggest just leaving this blank and letting the user pick the column, even if one may exist in the columnsModel?

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -127,9 +127,37 @@ function existingTableSubmit(request,setUploadInfo) {
...
-function getSelectedColumns(columns) {
-    const {lonCol='', latCol=''} = findTableCenterColumns({tableData:{columns}}) ?? {}; //centerCols
-    const columnsSelected = columns.map((col) => col.name === lonCol || col.name === latCol? ({...col, use:true}) :  ({...col, use:false})); //select position cols only
+/*
+    if showUploadTableChooser is called with a defaultColsEnabled object, use the colTypes and colCount to set the selected columns
+    for this uploaded table
+ */
+function defaultColumnsSelector(columns,colTypes,colCount) {
+    let cols = [];
+    if  (columns == null) return null;
```
robyww, line 136:
```js
if (!columns?.length) return;
```

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -127,9 +127,37 @@ function existingTableSubmit(request,setUploadInfo) {
...
+    let count = 0;
+    for (let i=0; i< columns?.length; i++) {
+        if (count < colCount && colTypes.includes(columns[i].type)) {
+            cols.push(columns[i]);
+            count ++;
+        }
+    }
```
robyww, line 143: I am not sure but this looks like a `filter` or a `filter` and a `map`. Using though will make she you are doing clearer.

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -92,13 +92,13 @@ export function SpatialSearch({cols, serviceUrl, serviceLabel, columnsModel, tab
...
 
-    useEffect(() => {
+    /*useEffect(() => {
         if (canUpload) {
             if (uploadInfo) setVal(SPATIAL_TYPE,MULTI);
             return;
         }
         setVal(SPATIAL_TYPE,SINGLE);
-    }, [serviceUrl,canUpload]);
+    }, [serviceUrl,canUpload]);*/
```
robyww, line 101: What is this commented out and what behavior does it change?

src/firefly/js/ui/tap/TapSearchRootPanel.jsx
robyww: Yes this looks good. I think you should improve the error message but the logic is good.

src/firefly/js/ui/tap/TemportalSearch.jsx
robyww: I am not quite following. I need to understand the problem. We should talk about the issue. 
> I suggest just leaving the temporal column value blank and letting the user pick the column, even if one may exist in the columnsModel?

Rubin specifically asked us to set the column if it exist.

robyww (changes requested):
@kpuriIpac - this is only a partial review I have not had time to look at Object ID search panel yet. I just wanted to give you some feedback to get you started. I will come back to it after some meetings today.

src/firefly/js/ui/tap/SpatialSearch.jsx
kpuriIpac: Let's say you have IRSA selected as the TAP service. If you switch from `Single Shape` to `Multi-object` once, and then switch back to `Single Shape`. Now if you switch to, say, CADC, this useEffect gets triggered (because of `serviceUrl` changing). And since an uploadInfo object exists, `Multi-object` will be selected here by default instead of `Single Shape` (which was the user's last selection). 

Commenting out the useEffect gets rid of this behavior, while not really affecting anything else.  

Actually, it does cause an issue switching from a `Multi-object` selection to a service that doesn't support uploads like NED. I think this is the final useEffect we'll need here: 

```
useEffect(() => {
        if (!canUpload) setVal(SPATIAL_TYPE,SINGLE);
    }, [serviceUrl,canUpload]);
```

I'll push this change in my next commit so you can take a look.

src/firefly/js/ui/tap/TemportalSearch.jsx
kpuriIpac: Yes we can talk about this, I can share my screen as well when we do.

src/firefly/js/ui/tap/ObjectIDSearch.jsx
```diff
@@ -0,0 +1,241 @@
...
+
+function defaultColsEnabledObj() {
+    return {
+        colTypes: ['int','long','string'],
+        colCount: 3
+    };
+}
+
+function UploadTableSelectorObjectID({uploadInfo, setUploadInfo}) {
+    const {groupKey,setVal}= useContext(FieldGroupCtx); //keeping this for now as setVal may be used if we want to set default ObjectID cols
```
robyww, line 91: not used

robyww (commented):
`ObjectIDSearch.jsx` looks good.

src/firefly/js/ui/tap/SpatialSearch.jsx
robyww: Thanks for helping me understand, the change makes sense.

src/firefly/js/ui/tap/TableSearchMethods.jsx
```diff
@@ -7,6 +7,7 @@ import {ExposureDurationSearch} from './ObsCoreExposureDuration.jsx';
 import {SpatialSearch} from './SpatialSearch.jsx';
 import {TemporalSearch} from './TemportalSearch.jsx';
```
ejoliet, line 8: picky: misspelled component, should be "TemporalSearch"

robyww (commented):
UI comments
- I choose a table and ObjectID search is not enabled. It should be at that point
- Same if I choose a column from the table
- Once they interact with the helper it should be enable
- ObjectID should be Object ID. Probably the scientist will have opinion about the naming.
- The table Object id should default to open since the user will alway have to set it. In fact I don't think it should even be in a disclosure pane but keep it that way for now.
- They both should be open by default.

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -93,11 +93,11 @@ export function SpatialSearch({cols, serviceUrl, serviceLabel, columnsModel, tab
     const updatePanelStatus= makePanelStatusUpdater(checkHeaderCtl.isPanelActive(), Spatial);
 
     useEffect(() => {
-        if (canUpload) {
+        /*if (canUpload) {
             if (uploadInfo) setVal(SPATIAL_TYPE,MULTI);
             return;
-        }
-        setVal(SPATIAL_TYPE,SINGLE);
+        }*/
```
robyww, line 99: remove the commented out lines

robyww (approved):
The UI looks good. I think you are 98% done. 

You still did not address my comments related to defaulting to `null` (we mostly prefer to use `undefined`.) and checking for `null`. Also you have not addressed the JSDoc issues. The  JSDocs should be put in. it is not obvious what `defaultColsEnabled` is. That needs a `typedef`.
 
After you get that done then you can merge.

src/firefly/js/ui/tap/ObjectIDSearch.jsx
```diff
@@ -48,6 +48,10 @@ export function ObjectIDSearch({cols, initArgs={}, capabilities, tableName,colum
         }
     );
 
+    useEffect(() => {
+        if (uploadInfo?.columns) checkHeaderCtl.setPanelActive(true);
+    }, [uploadInfo]);
```
kpuriIpac, line 53: here's the latest change @robyww to select the Object ID Search helper on a table upload. And the useFieldGroupWatch code is right above it.

gpdf:
Initial comments:

* I think maybe the "Object ID Search" search assistant should work more like the "Spatial" one, in that it has a single-ID and a multi-ID capability, and the single-ID capability is present all the time.
* The "main table" (not upload table) ID column should be pre-populated, if possible, with a column name that has UCD `meta.id;meta.main`, if there is one, or if there is only one column with UCD meta.id (without the `meta.main`).
 * If there are multiple columns with `meta.id` as part of their UCD, the column browser should at first come up in a state where they are the only ones shown, i.e., with a filter in place -- which the user can remove if they wish.

One of the things we would do on the server side at IRSA -- and certainly in Rubin, where this is already mostly true -- is make sure that pretty much every table for which this makes sense does in fact have a `meta.id;meta.main` column.

gpdf:
Fun demonstration:

https://github.com/Caltech-IPAC/firefly/assets/13261203/2c7c142f-485d-40c1-b826-2b99ddaebb0b

gpdf (commented):
I've already commented above on trying to achieve more parallel construction with the spatial search helper, and on trying to do some heuristic guessing of IDs.

One other thing that I think is important in this search helper -- so that users don't confuse searching by "ID" with searching by name (as they can do in the spatial-search box): would it be possible, and not too ugly, to include text in the search helper that says, right on the screen:

"Performs an exact match on the ID(s) provided, not a spatial search in the neighborhood of the designated objects."

PR #1402, opened 2023-06-30
FIREFLY 1073: Support for 'POINT' type s_region and prevent coverage from failing

kpuriIpac:
#### [Firefly-1073](https://jira.ipac.caltech.edu/browse/FIREFLY-1073)
- This ticket addresses a few small issues (also see related ticket https://jira.ipac.caltech.edu/browse/IRSA-4866) 
   - Treating **POINT** type s_regions the same as **POSITION** type s_regions fixes this specific issue, and the coverage map loads as expected (see changes in ObsCoreSRegionParser's `parseObsCoreRegion` function).  
   - If s_region wasn't parsed correctly in **ObsCoreSRegionParser.js** (the switch statement on line 132 leads to the default case), then the coverage would fail and the app would crash. Reason for this is that when s_region exists but is parsed as invalid, then the **isCoverageFail** function of **CoveraeViewer.jsx** gets a **covState** value that is a string ('_No coverage available_') instead of an array, which is why `covState.find` fails. Adding a check fixed this. 
   -  Even after the above fix, if you have a s_region value that isn't parsed (let's say if the format of s_region is changed in the future), we should still be able to load coverage map if ra,dec exist. For this, we need to drop (or ignore) s_region if it was parsed as an invalid value. Hence, in CoverageWatcher.js, alongside the check for `isTableWithRegion(table)`, I also added a check for `getRegionAryFromTable(options, table)` returning an empty array (this function calls `parseObsCoreRegion`. If it returns an invalid s_region value,  `getRegionAryFromTable` returns an empty array).  So even if s_region exists but regionAry is empty, we'll not assign **CoverageType** as **REGION**. Which make sure coverage map still loads correctly even if s_region exists but isn't parsed as valid. 
        - I tested this by using the table in the IRSA ticket above, and removing my "**regionShape.point.key**" switch statement in `parseObsCoreRegion` function (while keeping all other changes), and the coverage map still loaded

**Updates 7/5**:
  - fixed a bug which allows isCoverageFail function to go back to its previous state
  - added a console.log note for the user/developer when s_region exists but isn't parsed as valid (possibly because we don't support its format). This will let the developer know we have not displayed s_region on the coverage map. 
     - @ejoliet or anyone else: if you would like to test this locally: Pull my branch, comment out line 134 in ObsCoreSRegionParser.js (`case regionShape.point.key`) and upload the file from the IRSA-4866 ticket above. Then click on coverage map. You should see a message in your console notifying you we didn't pick s_region for the coverage map. 

Testing: 
- https://fireflydev.ipac.caltech.edu/firefly-1073-sregion/firefly
- use the table in the IRSA-4866 ticket above, and load it from the Upload tab. 
- Click on Coverage and make sure the coverage map loads.

### DIFF
--- a/src/firefly/js/visualize/saga/CoverageWatcher.js
+++ b/src/firefly/js/visualize/saga/CoverageWatcher.js
@@ -45,6 +45,7 @@ import {computeCentralPtRadiusAverage} from '../VisUtil.js';
 import {BLANK_HIPS_URL, isBlankHiPSURL, isHiPS} from '../WebPlot';
 import {WebPlotRequest} from '../WebPlotRequest.js';
 import {getSearchTarget} from './CatalogWatcher';
+import {memorizeLastCall} from 'firefly/util/WebUtil';
 
 /**
  * @typedef {Object} CoverageType
@@ -334,8 +335,8 @@ function updateCoverage(tbl_id, viewerId, preparedTables, options, tblCatIdMap, 
                     if (allRowsTable?.tableData?.data?.length > 0) {
                         preparedTables[tbl_id] = allRowsTable;
                         const isRegion = isTableWithRegion(allRowsTable);
-
-                        tblCatIdMap[tbl_id] = (isRegion) ? [centerId(tbl_id), regionId(tbl_id)] : [tbl_id, searchTargetId(tbl_id)];
+                        const regionAry = getRegionAryFromTable(table);
+                        tblCatIdMap[tbl_id] = (isRegion && regionAry.length > 0) ? [centerId(tbl_id), regionId(tbl_id)] : [tbl_id, searchTargetId(tbl_id)];
                         updateCoverageWithData(viewerId, table, options, tbl_id, allRowsTable, preparedTables,
                             isTableUsingRadians(table), tblCatIdMap, preferredHipsSourceURL, preferredHipsSource360URL);
                     }
@@ -509,7 +510,7 @@ function computeSize(options, preparedTables, usesRadians) {
                     ptAry= getBoxAryFromTable(options,t, usesRadians);
                     break;
                 case CoverageType.REGION:
-                    ptAry = getRegionAryFromTable(options, t);
+                    ptAry = getRegionAryFromTable(t);
                     break;
 
             }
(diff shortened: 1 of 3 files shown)


src/firefly/js/visualize/saga/CoverageWatcher.js
```diff
@@ -334,8 +334,8 @@ function updateCoverage(tbl_id, viewerId, preparedTables, options, tblCatIdMap,
                     if (allRowsTable?.tableData?.data?.length > 0) {
                         preparedTables[tbl_id] = allRowsTable;
                         const isRegion = isTableWithRegion(allRowsTable);
-
-                        tblCatIdMap[tbl_id] = (isRegion) ? [centerId(tbl_id), regionId(tbl_id)] : [tbl_id, searchTargetId(tbl_id)];
+                        const regionAry = getRegionAryFromTable(options, table);
```
robyww, line 337: I am concerned this this fairly expensive call is being call to often. It should be wrapped with `memorizeLastCall` (from WebUtil.js)

robyww (approved):
The code looks good. Please do one thing before merge.

use `memorizeLastCall` for `getRegionAryFromTable` - see example in PlotViewUtil.js, line 56.

Also getRegionAryFromTable has an unused parameter `options`, just remove it.

When you have done that then merge.

src/firefly/js/visualize/saga/CoverageWatcher.js
```diff
@@ -769,17 +770,16 @@ function getBoxAryFromTable(options,table, usesRadians){
         .filter( (row) => row.every( (v) => v));
 }
 
-
-function getRegionAryFromTable(options, table) {
+const getRegionAryFromTable = memorizeLastCall( (table) => {
```
kpuriIpac, line 773: @robyww ran into a bug after my own changes. Tried uploading a file without a s_region column. And the coverage map would not load, because on line 338 when I call `const regionAry = getRegionAryFromTable(table);`, in **getRegionAryFromTable** function, we try to access rCol.regionIdx but rCol is null (because` rCol = findTableRegionColumn(table);` and there is no region column!)...

so added a null check for `rCol` here and that fixes this issue. 

Also wrapped memorizeLastCall around **getRegionAryFromTable** as you suggested. Do these 2 changes look good? I'll go ahead and merge if no more issues..

src/firefly/js/visualize/saga/CoverageWatcher.js
```diff
@@ -769,17 +770,16 @@ function getBoxAryFromTable(options,table, usesRadians){
...
     const rCol = findTableRegionColumn(table);
-
+    if (!rCol) return [];
```
kpuriIpac, line 775: @robyww ran into a bug after my own changes. Tried uploading a file without a s_region column. And the coverage map would not load, because on line 338 when I call `const regionAry = getRegionAryFromTable(table);`, in **getRegionAryFromTable** function, we try to access rCol.regionIdx but rCol is null (because` rCol = findTableRegionColumn(table);` and there is no region column!)...

so added a null check for `rCol` here and that fixes this issue. 

Also wrapped memorizeLastCall around **getRegionAryFromTable** as you suggested. Do these 2 changes look good? I'll go ahead and merge if no more issues..

robyww: looks good!

PR #1404, opened 2023-07-05
FIREFLY-1256: Create a property sheet (or table row details) component to use in viewer and API

jaladh-singhal:
Implements [FIREFLY-1256](https://jira.ipac.caltech.edu/browse/FIREFLY-1256)

- [x] Create a PropertySheet component that can be showed in tab in TriViewPanel
- [x] Make it accessible through toolbar too
- [x] Configure default table view options as per ticket
- [x] Add an app-level property to control if it shows in toolbar or as a tab or none
- [x] Gregory's comments (see comment below)
- [x] Backport it in IFE apps that have their own (redundant) implementations of it
- [ ] [Stretch goal] Add a custom view of property sheet instead of default table view

### Viewer
https://firefly-1256-prop-sheet.irsakudev.ipac.caltech.edu/irsaviewer/

1. Catalogs -> IRSA -> search m81 and then search m8. Now check that clicking different rows and switching table tabs updates table in details tab.
2. Catalogs -> TAP -> CADC TAP service -> details tab should show UCD and utype column because they exist in the main table's columns.

### API
https://fireflydev.ipac.caltech.edu/firefly-1256-prop-sheet/firefly/test/tests-main.html

Check IRSA gator table: in toolbar, click on the button left to table options, it should show property sheet as a floating dialog. Selecting rows in the table should update table in this dialog.

### IFE apps
Check https://github.com/IPAC-SW/irsa-ife/pull/277


src/firefly/js/tables/ui/TablePanel.jsx
```diff
@@ -151,6 +154,11 @@ export function TablePanel(props) {
                                  title={TT_OPTIONS}
                                  onClick={showOptionsDialog}
                                  className='PanelToolbar__button options'/> }
+                            {true && //TODO: add flag
+                                <div style={{marginLeft: '4px'}}
+                                     title={TT_PROPERTY_SHEET}
+                                     onClick={showPropertySheetDialog}
+                                     className='PanelToolbar__button propSheet'/> }
```
loitly, line 161: Move it before `Options`.  `Options` should be the far left most button.

src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
```diff
@@ -104,10 +105,10 @@ function RightSide({expanded, closeable, showXyPlots, showMeta, showFits, dataPr
         } else {
             return makeActiveChartTab({viewerId, activeLabel, chartExpandedMode, closeable});
         }
-    };
+    }
 
     const style= {height: '100%'};
-    const defaultSelected= coverageRight ? 'coverage' : showXyPlots ? 'activeCharts' : 'fits';
+    const defaultSelected= true ? 'rowDetails' : coverageRight ? 'coverage' : showXyPlots ? 'activeCharts' : 'fits'; //TODO: remove after testing
     const key= `${showXyPlots&&'xyplot'}-${cov&&'cov'}-${meta&&'meta'}-${fits&&'fits'}`;
```
loitly, line 112: I don't see where it says 'rowDetails' should be the default selected tab.

loitly (commented):
I know it's just a draft.  These are comments I thought of when I finally got a chance to read the tickets.  Also, please include the links to the tickets next time.

src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
jaladh-singhal: This was for debugging

jaladh-singhal: See "TODO: remove after testing"

src/firefly/js/tables/ui/PropertySheet.jsx
```diff
@@ -0,0 +1,50 @@
...
+import {TablePanel} from 'firefly/tables/ui/TablePanel';
+import React, {useEffect} from 'react';
+import {useStoreConnector} from 'firefly/ui/SimpleComponent';
+import {dispatchTableAddLocal} from 'firefly/tables/TablesCntlr';
+
+export function PropertySheet({tbl_group='main', tbl_id, children, fetcher}){
+    const {activeTblId, tableModel} = useStoreConnector(() => {
+        const tblId = tbl_id || getActiveTableId(tbl_group);
+        return {activeTblId: tblId, tableModel: getTblById(tblId)};
+    });
```
loitly, line 11: We talked about using children as a `renderer` and that it should default to <PropertySheetTable {...{tbl_id, highlightedRow}}/>.

In the most basic case: (RowDetails of the active table)
```jsx
  <PropertySheet/>
```

In FIREFLY-603, it mentioned templating. the idea is this.
```jsx
  <PropertySheet>
    <CustomView/>
  </PropertySheet>

function CustomView(p) {
return <div> Hello, ${p.firstName} ${p.lastName}</div>;
}
```
In this scenario, p is the return value of the default `fetcher` function where it transform the highlightedRow into an object of cname/value pairs.  This object is passed in as properties to the renderer.

Then, there's the more advanced use-case where fetcher is used to query external data.
In this case, fetcher is most likely async.  `PropertySheet` should call `fetcher` with `tbl_id`, and `highlightedRow` so it can gather additional table meta.  It should also be smart enough to render 'Loading.. ' status until `fetcher` returns.

For this PR(FIREFLY-602), it only need to implement the default renderer(PropertySheetTable).  But, it should be aware of how this component will eventually grow into.

gpdf:
**Thank you very much!**  This will be very useful in combination with wide tables.

I'm going to comment, right now, only on the actual details tab itself, and on one point about the UI for controlling where it appears on tabs.

So far I have tested today's state of 
https://firefly-1256-prop-sheet.irsakudev.ipac.caltech.edu/irsaviewer
only with a simple (non-ObsCore) catalog search result.  
I will be very interested to see how this feature interacts with the Data Product Viewer environment, though.

### Details tab / property sheet appearance

1. The *Value* and *Units* columns should be right next to each other, because we want to have the user have an experience as much like seeing "431.6 nm" as possible.
2. The UCDs and, if available, UTypes, of the columns should be available in the property sheet.  There are use cases there like "I want to see only the fluxes" where a user would want to be able to filter on UCDs.
  # Putting the above together, I'd suggest the order Name, Value, Units, Description, UCD, UType
4. Just as we've said in many other places, please turn the column-filter header on by default.
5. There seems to be some sort of issue with both integer and floating-point zeros being displayed as blanks; example:

Main table:

image

Details tab for the selected row:

image

### High-level controls

The replacement of "Coverage: Left | Right" with "Coverage Left | Coverage w/Charts" is pretty user-challenging, I think.  It seems like the functionality of "Right" might have been lost?  "Coverage w/Charts" seems to set up a bi-view.

I think we'll have to polish this, but stick with something like this for now, for the Details tab in 2023.2, but for future releases I think we need to rethink how we let users rearrange tabs to over more flexibility.  For instance, I can easily envision wanting to have Charts across from Details, so that I can click on a point in a chart and see the details for that item.

gpdf:
One more point:

There is an existing user-confounding aspect of our column-properties dialog that also appears in the Details tab:

If the user brings up the "gears" dialog for the Details tab, they see:

image

The checkboxes for "Units" and "Data Type" at the top can be very easily confused with the "Units" and "Types" entries in the table below.

I think the "Units" and "Data Type" displays are uninformative for the Details tab and could be suppressed: none of the *columns* in the details tab will ever have units, and the "Data Type" is currently always string ("char").

gpdf:
Another point that just occurred to me: the "checkbox" column should be enabled in the table viewer in the details tab, because, for very wide tables, it's an attractive use case for someone to limit the display to specifically selected columns of the query result (i.e., rows in the details table).

An example of where these would be useful: looking at a catalog overlay on an image, or looking at a scatterplot (e.g., a color-magnitude diagram) a user wants to click on specific sources from the catalog in that display, and then see details on the selected source in the details tab, without having to scroll up and down to find the things they care about.

Please turn on the checkboxes.

jaladh-singhal:
Thanks for your comments @gpdf. For my and reviewers' reference, I'm converting them as TODOs:

- [ ] Columns should be ordered as Name, Value, Units, Description, UCD, UType 
  - [ ] ❓ Type should be renamed as UType or is it a different thing?
  - [ ]  Extract UCD as a column too
- [x] turn the column-filter header on by default
- [ ] Suppress "Units" and "Data Type" in the table options for the row details table
- [ ] fix issue with both integer and floating-point zeros being displayed as blanks
   - it's likely happening because type is being casted to char so 0 is empty string
- [x] turn on the checkboxes
  - ❓ this and above point made me realize a problem we are probably overlooking: In wide table format, each value was of different type (of the type of column it was present in) but converting that to the property sheet/row details, we end up with one column "Values" that has mixed data types (char, float, int, etc.). Right now, we are converting all of them to char, so mathematical filters (like >, =) won't work. We need to think how filtering will work here?


Regarding High-level controls:
> The replacement of "Coverage: Left | Right" with "Coverage Left | Coverage w/Charts" is pretty user-challenging, I think. It seems like the functionality of "Right" might have been lost? "Coverage w/Charts" seems to set up a bi-view. 

This PR doesn't introduce this, it exists similarly in ops. 

> I think we'll have to polish this, but stick with something like this for now, for the Details tab in 2023.2

❓ I agree about the confusion and scope of improvement here. I hope we are tracking it somewhere on jira for future cycles?

@robyww we need to discuss points marked "❓"

gpdf:
> we end up with one column "Values" that has mixed data types

That is an excellent point.  I wonder if we should just disable the filter for this column.  Alternatively, we can just document that its pure-text search treating the values as strings.  There are actually some use cases for that.

gpdf:
> Type should be renamed as UType or is it a different thing
No, "UType" is a different thing in the IVOA data model.  I don't know whether it's actually available to you in the internal data model for tables at the moment or not.  @loitly ?

loitly:
> No, "UType" is a different thing in the IVOA data model. I don't know whether it's actually available to you in the internal data model for tables at the moment or not. @loitly ?

Yes.  There's UType in Firefly table column.

src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
```diff
@@ -93,6 +95,7 @@ function RightSide({expanded, closeable, showXyPlots, showMeta, showFits, dataPr
     const cov= imagesWithCharts || coverageRight;
     const meta= imagesWithCharts && showMeta;
     const fits= imagesWithCharts && showFits;
+    const showPropertySheet = getAppOptions()?.table?.propertySheet === 'integrated';
```
robyww, line 98: I think you should make `'integrated'` and `'popup'` constants in table util.

src/firefly/js/tables/ui/TablePanel.jsx
```diff
@@ -212,6 +221,18 @@ function showTableInfoDialog(tbl_id)  {
...
 
+function showTablePropSheetDialog(tbl_id) {
+    const content = (
+        // TODO: use its own class
+        <div className='TablePanelOptionsWrapper'>
+            <PropertySheet tbl_id={tbl_id}>
+                <PropertySheetTable detailsTblId='rowDetailsTbl'/>
+            </PropertySheet>
+        </div>
+    );
```
robyww, line 232: This looks good.

robyww (commented):
Looks good so far.

-  I like the icon though it you want ahead and make it smaller you could clean up the `i` a little so it would show up better.
- design looks good.
- works in my testing.
- I think you just need to add constants.

src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
jaladh-singhal: Yes, that's better

robyww (approved):

src/firefly/js/tables/ui/PropertySheet.jsx
jaladh-singhal: > We talked about using children as a renderer and that it should default to <PropertySheetTable {...{tbl_id, highlightedRow}}/>.
>
> In the most basic case: (RowDetails of the active table): \<PropertySheet/\>

This would have been problematic because we have `detailsTblId` and `tblOptions` which are only required for table view (child) not for PropertySheet (parent which can be used with custom views as well). @robyww suggested we can create a new component combining these two for providing simple usage. Check `PropertySheetAsTable` component

> In this scenario, p is the return value of the default fetcher function where it transform the highlightedRow into an object of cname/value pairs. This object is passed in as properties to the renderer.

This is not simply name/value pairs because we also have metadata attached to columns (like unit, type, UCD, etc.) that we need to display. That's why I made parent component pass highlightedRow and tbl_id to default/table view so that it can extract information and create a table as needed. I've also added documentation for more clarity. Let me know if I missing something (I'll make changes in next PR)?

> Then, there's the more advanced use-case where fetcher is used to query external data.
In this case, fetcher is most likely async. PropertySheet should call fetcher with tbl_id, and highlightedRow so it can gather additional table meta. It should also be smart enough to render 'Loading.. ' status until fetcher returns.

Yes that's a good point. For now, I've put TODO about using fetcher, and merging this PR because it needs to go in the release this Monday. 

In next PR, I'll work on implementing a custom view using fetcher - which will make it clear for us how to evolve the existing design of these components.

loitly: > problematic because we have detailsTblId and tblOptions

This is not a problem.  These props can be defaulted by `PropertySheetTable`, or it can be passed into from PropertySheet.  It's common for wrapper component to take the props it care about, and pass along the other props to its children without inspection.  An example of such usage is UI library wrapping input field.  e.g.  
```
{tbl_group='main', tbl_id, children, fetcher, ...passAlongProps}
```

> In this scenario, p is the return value of the default fetcher function where it transform the highlightedRow into an object of cname/value pairs. This object is passed in as properties to the renderer.

When `PropertySheet` decides to have a `renderer` and a `fetcher`, then it needs to define what that behavior should be.  I am suggesting that the behavior is.... renderer get passed in as props, `{tbl_id, highlightedRow}` and the returned value of the fetcher call.   The above comment is for when fetcher is not provided.  It would return `{tbl_id, highlightedRow}` plus the values of the highlighted row as key/value pairs.  This would satisfy the 'templating' use-case  mentioned in the ticket.  
What you are doing with `RowDetailsTable` is fine and it does not break the above contract.  It requires `tbl_id, highlightedRow` which will be given.

Because I mistakenly reviewed the PR when it was still in draft, I had to read the ticket to understand what was requested.  Based on my understanding at the time, these are my suggested design elements.  But, you are free to implement it as you see fit.  You probably have more in depth understanding of the issues and how to resolve it.

jaladh-singhal: > It's common for wrapper component to take the props it care about, and pass along the other props to its children without inspection. An example of such usage is UI library wrapping input field. e.g.

Ok this makes sense, esp the example. Although I think we have to add some sort of exception handling in this component to enforce that if children wasn't passed, then `detailsTblId` has to be passed because it's required prop for the default renderer component.

> The above comment is for when fetcher is not provided. It would return {tbl_id, highlightedRow} plus the values of the highlighted row as key/value pairs. This would satisfy the 'templating' use-case mentioned in the ticket.

Ohk I see what you mean. I gave it more thought and I think we can use a data structure that can contain metadata too besides just key-value pairs, something like this:
```
{key: {
  value: valueOfKey,
  metadata: {
    type: typeOfKey,
    units: unitsOfKey,
    ...
  },
 key2: ...
}
```

> Based on my understanding at the time, these are my suggested design elements. But, you are free to implement it as you see fit. You probably have more in depth understanding of the issues and how to resolve it.

Thanks for mentioning that. And I appreciate your input to come up with a better design.

PR #1405, opened 2023-07-05
FIREFLY-1270: Table column selection stopped working

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1270
Regression bug.  Column selection(show/hide) stopped working.  See ticket's video for steps to reproduce.

Test: https://firefly-1270-column-selection.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA
- Verified that SHA's Abstract search still works.  
- Column selection(show/hide) still works as before.

### DIFF
--- a/src/firefly/js/tables/ui/BasicTableView.jsx
+++ b/src/firefly/js/tables/ui/BasicTableView.jsx
@@ -8,7 +8,7 @@ import {Column, Table} from 'fixed-data-table-2';
 import {wrapResizer} from '../../ui/SizeMeConfig.js';
 import {get, set, isEmpty, isUndefined, omitBy, pick} from 'lodash';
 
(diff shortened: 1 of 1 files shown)


jaladh-singhal (approved):
Sorry, I didn't realize that I introduced this bug while adding single column condition. Tested: Abstract search as well as column options' show/hide work as expected.

From what I can gather, `hide` is for temporary visibility (controlled by options panel, etc. - from client side) and `hidden` is for permanent visibility (what is returned from server side). Is that right? Where can I find spec for it?

loitly:
> From what I can gather, `hide` is for temporary visibility (controlled by options panel, etc. - from client side) and `hidden` is for permanent visibility (what is returned from server side). Is that right? Where can I find spec for it?
It's only documented here: https://fireflydev.ipac.caltech.edu/firefly/docs/js/global.html#TableColumn__anchor

You're mostly right.  The initial values came from the server.  These are used when you click reset.  They are then copied to UI state where the user can turn set show/hide on each column.  Because we don't show columns with visibility=hidden to the user, they cannot see it.  However, you can programmatically set visibility to anything you want.

PR #1406, opened 2023-07-10
FIREFLY-1257: UI for identification and entry of JOINs in ADQL queries

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1257

Add `JOINs available` to table with join(s) defined.  Read [Design Ticket](https://jira.ipac.caltech.edu/browse/FIREFLY-861) for details.  
`(jump to table)` is left out.  The `rc-tree` library we are currently using does not support that natively.  Even if possible, it would require a lot of work to get it working correctly.  @gpdf , is this a must have feature?

Also,
- fixed warning messages after react update.

Test: https://fireflydev.ipac.caltech.edu/firefly-1257-tap-join-helper/firefly/
TAP Searches -> Edit ADQL
Then, expand tables to see newly added `JOINs available` node if they are defined.

Here are some services with defined joins: MAST, CADC, 

**Update: Schema Browser filtering capability**
- added **Filter:** input box
This is used to filter the visible tree elements(nodes) based on partial match of the input string.

The algorithm is:
If a parent node matches the input string, then all child nodes will be visible.
If a child node is visible, then its ancestor nodes are visible as well.

The logic is to allow a user to lookup a schema and see all of the tables in the schema.
Lookup a table and see all columns in the table.  Of course, you can filter on the columns too, once it's visible.

Use-case:  CADC, query join data from observation and plane tables. 
TAP -> Edit ADQL -> Service=CADC
Schema Browser -> (+) caom2
-> Filter=obs -> (+) caom2.Observation -> (+) JOINs available -> (+) caom2.plane
-> click on observation table, then caom2-p-o join to complete the join statement
-> Filter=plane -> (+) caom2.Plane -> Filter=metric, to select metrics info from the plane table.


src/firefly/js/ui/tap/TapUtil.js
```diff
@@ -255,6 +255,35 @@ async function doLoadTapColumns(serviceUrl, schemaName, tableName) {
     }
 }
 
+
+export const loadTapKeys = memoize(async (serviceUrl) => {
```
robyww, line 259: I don't really like `memoize`, it has the side effect that the cache can be replace for all usages. That always seemed fragile to me. Also, it only uses the first parameter as the cache key (which does not matter in this case). I wrote a `WebUtil` alternative that I use in several places: `memorizeLastCall`. You can try it if you want to otherwise this is fine.

robyww (approved):
Looks good.

gpdf:
Comment/question: It's not entirely obvious how the schema filter is supposed to work.  I deduced logic like this from what I observed:

- If any node in the tree (leaf node or intermediate node) matches the filter, then that node is shown and every node in the tree "above" it (i.e., closer to the root of the tree) is also shown.
- If an intermediate node matches the filter, then, in addition to the above, every node "below" it in the tree is also shown.

However, it also seems to depend on whether a sub-branch of the tree has actually ever been opened or not: if I start with the tree fully collapsed, some items in the tree seem to be missed.  If I first open -- and then immediately re-close -- subtrees, then searches in them work.

I understand, from an implementation perspective, why this might be the case, but it's not at all an obvious UX.  It's not easy for a user to predict what will happen without taking a sort of software-developer perspective.

It's already a useful feature, but I think people will be surprised by the results.

gpdf:
Strange form-state-updating bug:

If I switch back and forth between the "UI assisted" and "ADQL" screens of the TAP UI, switching back to the "UI assisted" screen is erroneously selecting a multi-object search AND immediately bringing up the corresponding "upload" dialog.

Screen recording: 

https://github.com/Caltech-IPAC/firefly/assets/13261203/20e5b577-5f63-4840-8a65-8653a2a335a9

gpdf (changes requested):
Nice work -- the identification of the existence of JOIN keys, and the ability to drop them into ADQL queries with appropriate syntax, are working as desired.

I was able to navigate CADC's links between CAOM2 tables and construct a query.

A small suggestion: include a leading and trailing space around the inserted text.

I would like to suggest a significant change to the UX, however.  _To be clear, you absolutely did implement what I requested in the design ticket, so this is a suggestion based on what it felt like to actually use it, and it's quite possible you might have preferred doing it this way in the first place._

In most cases, there will only be one way to JOIN from one table to another, so this is the use case we should make easiest. 

Instead of the current nested scheme, where you have to open up the folder for the target table, and then you have to click on what might be a cryptically-named key, let's have the user experience it as clicking on the target table directly.  We still have to allow for there being two ways to get from table A to table B (even though this will be a small fraction of cases), so I've included my full suggestion in an attached image.

Screenshot 2023-07-12 at 6 36 06 PM

----

(EDIT: I see you already pointed out in the description above that the work below can't easily be done for now -- and I'd also forgotten that I had this in the original design ticket.  As I said, though, we can save this for later.  Let's just be sure to make a ticket.)

> It also might be nice for the user to be able to quickly get to the schema for the target table, because the very next thing they'll probably want to do after dropping in the `INNER JOIN` clause is to drop in some columns from that table for the SELECT.
> 
> Screenshot 2023-07-12 at 6 38 02 PM
> 
> (I would really use the "'i'-in-a-circle" icon that we use elsewhere, instead of the emoji ℹ️ above.)
> 
> The idea is that clicking on the (i) would jump the schema browser to the target table, opened up.
> 
> But this last feature could be saved for a future improvement.

loitly:
> Strange form-state-updating bug:
> 
> If I switch back and forth between the "UI assisted" and "ADQL" screens of the TAP UI, switching back to the "UI assisted" screen is erroneously selecting a multi-object search AND immediately bringing up the corresponding "upload" dialog.
> 
> Screen recording:
> 
>  Screen.Recording.2023-07-12.at.6.03.52.PM.mov

I will ignore this because it's not part of the UI I am currently working on.  I've reported similar bugs in @kpuriIpac PR as well.  Maybe this should be a ticket?

loitly:
@gpdf I do like the collapsed look better.  I'll do that.
As for the unexpected behavior in filtering and `jump to`, the main reason is due to 'lazy loading'.  We don't load the 'children' until it's expanded.  After the UI changes, I will evaluate whether it's better to fetch the full schema definition.  If it's not a performance concern, then I will proceed to implement the 'jump to' feature.

robyww:
> After the UI changes, I will evaluate whether it's better to fetch the full schema definition. 

I have concerns about the performance of that. Do you mean every column of every table? In IRSA that would be a lot.

gpdf:
Btw, if the "tree viewer" widget you are using doesn't allow nodes to have a custom icon, you could allow the document icon to display, for each key, but then have the first character in the label be the Unicode character for the chain-link, 🔗 , i.e., U+1F517.

gpdf:
Perhaps this is still a work in progress, but I looked at the build and I see

image

I would really like to see the target **table** name first, before the link name, like in my mockup above.  I'm pretty convinced that'll be easier for users to understand.

loitly:
@gpdf @robyww 

**Summary of Changes:**

- Display all joins under 'JOINs available', without grouping them by table name as mentioned in the ticket.
- Prepend the table name to the label of each join node.
- Add join icon as suggested.
- Instead of including a leading and trailing space around the inserted text, I have decided to include only a leading space. 
  -This logic is now applied when inserting column names as well.  This is already done for the upload_table.
-bThe 'Jump To Table' feature will NOT be implemented.  It is too hard with the current library.

**Additional Changes:**

- Load full schemas and tables for a given service URL.  (NOT columns.  Columms are loaded upon request)
  - This improvement makes Table Schema Filtering work much better than before.
- Apply default sorting when schema_index or table_index are not available. The sorting is in ascending order with lowercase entries placed at the top.

With the improvement, `Edit ADQL` is a lot more responsive and user-friendly.  Performance is not noticeable for `UI assisted`, although from network monitoring, it should also be a bit faster because we made a couple less server calls.

There is a side-effect because of the applied sorting mentioned above.  We are no longer displaying 'natural order' of the schema table.  This in my opinion is an improvement, since we now have a defined ordering scheme for schemas and tables.  They can still provide ordering by providing schema_index and table_index as before.

Test URL updated.  Let me know what you think.

robyww (approved):
It all looks good to me.

kpuriIpac:
> > Strange form-state-updating bug:
> > If I switch back and forth between the "UI assisted" and "ADQL" screens of the TAP UI, switching back to the "UI assisted" screen is erroneously selecting a multi-object search AND immediately bringing up the corresponding "upload" dialog.
> > Screen recording:
> > Screen.Recording.2023-07-12.at.6.03.52.PM.mov
> 
> I will ignore this because it's not part of the UI I am currently working on. I've reported similar bugs in @kpuriIpac PR as well. Maybe this should be a ticket?

BTW, this should be fixed now from my PR on the Object ID work: https://github.com/Caltech-IPAC/firefly/pull/1401 @loitly @gpdf. Let me know if you still run into this (I've tested and seems to work well).

PR #1409, opened 2023-07-18
FIREFLY-1285: SPHEREx photometry table sample fails to load

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1285
- complete TODO item in DataType that maps the remaining VoTable types to Firefly
- remove useShortDesc in DataType.
- add round trip support for VoTable to IPAC to VoTable when VoTable contains types not supported by IPAC table.

From ticket:

> Due to incomplete mappings, it was mapping the missing types to String which causes the Type mismatch exception.  I fixed that as well as cleaning up some old code.  I also added round-trip support for VoTable -> IPAC table -> VoTable even though IPAC table do not support these types.  To do this, I added the type mappings info into the IPAC Table's meta

Test: https://fireflydev.ipac.caltech.edu/firefly-1285-spherex-table-load-fail/firefly
See ticket for link to test table.

-> Upload test table.  It should load fine.  Then, Save as IPAC table
-> Upload IPAC table, content should still look good.  Now, Save as VOTABLE
-> Uploaded saved VOTable.  Table should still look good.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
@@ -8,14 +8,14 @@ import edu.caltech.ipac.firefly.server.util.StopWatch;
 import edu.caltech.ipac.table.*;
 
 import java.io.*;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.stream.Collectors;
 
-import static edu.caltech.ipac.table.IpacTableUtil.createMetaFromColumns;
+import static edu.caltech.ipac.table.DataType.typeToDesc;
+import static edu.caltech.ipac.table.IpacTableUtil.*;
 import static edu.caltech.ipac.table.TableMeta.*;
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
```diff
@@ -94,7 +94,7 @@ private static void save(PrintWriter out, DataGroup dataGroup) throws IOExceptio
                 IpacTableUtil.writeAttributes(out, atts, false);
 
                 dt.setDataType(String.class);
-                dt.setTypeDesc(DataType.CHAR);
+                dt.setTypeDesc("char");
```
robyww, line 97: Why not a constant?

robyww (approved):
Looks good. It is always good when we can more fully support these standards. Thanks for getting to it for fast

src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
loitly: In the past, Firefly mimic data type of IPAC table.  That's why they are used interchangeably.  I just wanted to be clear that this class is for IPAC table format.  Its type reflect IPAC table's.  I can declare a const here, but it seem a bit redundant since it won't use anywhere else.

PR #1411, opened 2023-07-25
Firefly-1291: obscore data products result should show search target

robyww:
#### Testing
- https://firefly-1291-search-target.irsakudev.ipac.caltech.edu/irsaviewer/dce
- do a DCE search, the overlay position should be the search target

### DIFF
--- a/src/firefly/js/metaConvert/ObsCoreConverter.js
+++ b/src/firefly/js/metaConvert/ObsCoreConverter.js
@@ -4,6 +4,7 @@ import {sprintf} from '../externalSource/sprintf.js';
 import {tokenSub} from '../util/WebUtil.js';
 import {PlotAttribute} from '../visualize/PlotAttribute.js';
 import {RangeValues} from '../visualize/RangeValues.js';
+import {getSearchTarget} from '../visualize/saga/CatalogWatcher.js';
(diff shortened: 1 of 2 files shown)


lrebull:
I'm running into this bug while testing https://jira.ipac.caltech.edu/browse/FIREFLY-1196

lrebull:
(i think, anyway. still testing)

lrebull:
If I start a clean session and only ask for one target over many data sets, it works great. I run into trouble if I search on more than one different target.

wmiipac:
I did position searches for few object with AKARI, so there are  few tabs in the table panel.  when navigating through the tabs, target is overlayed on images correctly.  I closed few  table tabs, the target positions are still on the coverage map. Not sure if this is desired, may be confusing to the user.

robyww:
@wmiipac that is the way it has always worked. I don't know if it is desired or not but it is outside the scope of this ticket. If it is a problem we can work on it for another ticket.

robyww:
@lrebull I searched more than one target and it seemed to work for me. I will need more details if there is an issue.

wmiipac (approved):
Works as expected. Thanks for a quick fix!

lrebull:
Huh. i went to make a movie, and naturally I couldn't make it fail. I don't know what was going on before.

PR #1414, opened 2023-07-29
Firefly-1279: TAP Object ID Search Cleanup

kpuriIpac:
#### [Firefly-1279](https://jira.ipac.caltech.edu/browse/FIREFLY-1279)
- for the main table (selected table on the right), the Object ID column is populated with  `meta.id;meta.main` if available, or with the first `meta.id` if `meta.main` is not available. If both are not available, the column remains empty. 
   - _Note_: if a valid column name is found (say a UCD with `meta.id;meta.main` exists), then the `Object ID Search` helper is auto-selected by default. Is this behavior okay? **_Done_**. 
- Added the helpful text (from the ticket) right under the `Object ID Search` helper title (side note: we may also need to update the onlinehelp link next to the title for the Object ID, as this is a new feature so users may want to click on it to learn more about it). 

_Pending tasks_ (as time permits before release): 
- filtering the columns with all that have UCD set to meta.id (if there are multiple), so that user may first choose Object ID from these (and they should be able to remove this filter as well). 
- handling the selection of Object ID search when the user has chosen a file and unselected it. **_Done_**. 

**Updates 7/31:**
- Object ID search being selected when there's a file uploaded is now fixed based on PR comments below @robyww 
- addressed other change as well, Object ID search is now not auto selected (can test by switching over to CADC where the Object ID column is pre populated). 
- test build is updated with these 2 changes 

**Updates 7/31 Part 2: _Filtering_**
- With @loitly's help, was able to get filtering working. Had to use a workaround that doesn't feel super clean to ensure that the filtering is applied even when you change TAP service/tables or even re-open the column selector for the table that you're on multiple times. See my comment about this change below. 
- To test, you'll need to open a table that has more than one `meta.id` entries. I suggest switching over to CADC and choosing `Table Collection: ivoa` and table `ivoa.ObsCore`. 

**Testing**: 
- https://fireflydev.ipac.caltech.edu/firefly-1279-objectid-cleanup/firefly
- go to the TAP UI, make sure you see the new text under the Object ID Search title 
- Switch TAP service from IRSA to CADC. For the default table collection `ivoa` and the default table `ivoa.ObsCore`, the Object ID field should be populated with `obs_collection`. 
    - You can verify that this is correct by clicking on the column selector. There is no UCD with `meta.id;meta.main`, but the first UCD with `meta.id` in it is, indeed, the column selected here: `obs_collection`

### DIFF
--- a/src/firefly/js/ui/tap/ObjectIDSearch.jsx
+++ b/src/firefly/js/ui/tap/ObjectIDSearch.jsx
@@ -3,7 +3,10 @@ import {FieldGroupCtx} from 'firefly/ui/FieldGroup';
 import {ConstraintContext} from 'firefly/ui/tap/Constraints';
 import {useFieldGroupRerender, useFieldGroupValue, useFieldGroupWatch} from 'firefly/ui/SimpleComponent';
 import {
-    getPanelPrefix, makeCollapsibleCheckHeader, makeFieldErrorList, makePanelStatusUpdater
+    getPanelPrefix,
+    makeCollapsibleCheckHeader,
+    makeFieldErrorList,
+    makePanelStatusUpdater
 } from 'firefly/ui/tap/TableSearchHelpers';
 import {getAsEntryForTableName, makeUploadSchema, tapHelpId} from 'firefly/ui/tap/TapUtil';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


robyww:
UI Comment: Object ID Search should not have the checkbox on even if it finds the column if there is not table upload.

To see the problem try CADC/ivoa/Obscore.

src/firefly/js/ui/tap/ObjectIDSearch.jsx
```diff
@@ -50,11 +56,20 @@ export function ObjectIDSearch({cols, initArgs={}, capabilities, tableName,colum
         }
     );
 
+    //clickingSelectCols is toggled each time user clicks on the 'magnifying glass' to select cols, rendering this useEffect
+    useEffect(() => {
+        onTableLoaded(defaultTblId).then((tbl) => {
+            filterTable(cols,tbl);
+        });
+    }, [clickingSelectCols]);
+
```
kpuriIpac: This doesn't feel like the cleanest way of doing this. But I needed a way to render this useEffect when the user clicks on the 'magnifying glass' to open the col selector popup. This is part of the `ColumnFld` component (in ColumnOrExpression.jsx) - so I needed to pass `clickingSelectCols` and its setter to this component (see line 265 of ObjectIDSearch.jsx). 

Now if you switch tables, or reopen the same column selector multiple times, it should retain the original filtering based on `meta.id` UCD entries (if multiple `meta.id` entries were not found, then no filtering is applied). 

Thoughts, @loitly @robyww?

src/firefly/js/ui/tap/ObjectIDSearch.jsx
```diff
@@ -175,14 +224,14 @@ function UploadTableSelectorObjectID({uploadInfo, setUploadInfo}) {
...
-                objectKey:UploadObjectIDColumn }} />
+                objectKey:UploadObjectIDColumn}} />
             }
         </div>
     );
 }
 
 function ObjectIDCol({objectCol, style={},cols, objectKey, openKey, labelStyle,
-                           headerTitle, headerPostTitle = '', openPreMessage='', headerStyle}) {
+                           headerTitle, headerPostTitle = '', openPreMessage='', headerStyle,tbl_id=null,clickingSelectCols=null,setClickingSelectCols=null}) {
```
loitly, line 234: It's a lot cleaner if you just move the useState/useEffect to this component.  Passing these value/function as props to its children is messy especially when it's not needed.

loitly (commented):
I am only commenting of the code changes.  I don't understand the ticket enough to review the PR.

robyww (approved):
Looks good.

PR #1420, opened 2023-08-15
Firefly-1298: Update obscore to support related grids of images

robyww:
[Firefly-1298](https://jira.ipac.caltech.edu/browse/FIREFLY-1298):Update obscore, service descriptors to support more  features

  - Support obscore and datalink grid view
  - Full support for service descriptors and all views
  - Support for `#this-spectrum` and `#this-image-grid`
  - Major refactoring of the whole DataProducts handling
      - hopefully the whole system is more maintainable 
      - files renamed to make more sense
      - some new files created
      - functions moved around
      - directories `vo` and `missions` (the missions use the more deprecated style)
  - Added more documentation
  - Fixing results view for non obscore tables with  service descriptors that point to datalinks
  - Support multiple multi product viewers for different types of data
  - Support parameters into service descriptor from other sources
      - this is how we are going to add cutout values injected into the service descriptor
  - Fixed: wrong row and table with datalink

Testing:

  - use Euclid PR for testing.
  - https://github.com/IPAC-SW/irsa-ife/pull/288
  - https://firefly-1298-dl.irsakudev.ipac.caltech.edu/applications/euclid


src/firefly/js/metaConvert/DataLinkProcessor.js
```diff
@@ -50,14 +50,17 @@ const isTooBig= (size) => size>GIG;
...
  * @param {string} [params.baseTitle]
  * @return {DataProductsDisplayType}
  */
 export function processDatalinkTable({sourceTable, row, datalinkTable, activateParams, baseTitle=undefined,
-                                     additionalServiceDescMenuList, doFileAnalysis=true, onlyRelatedGrid = false}) {
-    const dataSource= getObsCoreAccessURL(sourceTable,row);
+                                     additionalServiceDescMenuList, dlUrl, doFileAnalysis=true, onlyRelatedGrid = false}) {
+    let dataSource= getObsCoreAccessURL(sourceTable,row);
+    //for catalog tables, datalink tables are not linked via access_url, so dataSource will be empty from the above call
+    if (dataSource === '') dataSource = dlUrl;
```
kpuriIpac, line 63: @robyww the reason the cutouts and spectra were showing, but the datalink table wasn't coming up is because for our catalog (_MER-catalog-search.xml_) table, the call to `getObsCoreAccessURL` would return an empty string. Because that would attempt to get the access_url links to the datalink tables (which an obscore type table like _MER-tiles-search.xml_ has, but our catalog table doesn't). 

So instead I decided to pass the datalink url directly from `ServDescConverter.js` to the `processDatalinkTable` function, and use it when the dataSource is empty (in the case of non-obscore tables). 

With this fix, the cutouts, spectra and datalink table all show up in the results view (but I'll need to update the test build and push new endpoint changes to kubernetes for this to work for others).

loitly (approved):
I see a lot of good refactoring work done here.  We are building a lot of functionalities into `data product table`(?).  I think it's time we formally write a specification for the format of this table.  I see you have some already.  But, it may not be obvious that's where to look for.  Maybe add these types of documentation under `/docs`.

kpuriIpac (approved):
Great work! Didn't go into great detail looking into all of the code changes, but the documentation looks great. This also makes the Euclid prototype come together and really work well, especially the grid view for cutouts and the support for multiple MultiProductViewerContainer.

PR #1423, opened 2023-08-29
FIREFLY-1282: Add multi-results support for UWS jobs

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1282

Generalize Firefly's UWS support  to accommodate multiple results being returned.
In this case, it should return an ObsCore-like table.

Currently, there isn't an existing UWS service with results like this.  Therefore, I added a mock-tap service that can return tap result at `/results/result`, but multiple results at the `/results` endpoint.  
This mock service has been deployed to https://nightly.irsakudev.ipac.caltech.edu/mock-tap

To test, go to: https://fireflydev.ipac.caltech.edu/firefly-1282-uws-multi-results/firefly
- Upload -> Upload from URL
  - Enter URL:  https://nightly.irsakudev.ipac.caltech.edu/mock-tap/async/45
  - Click Upload and then Load

You can use any integer instead of 45.  The mock service provides two sets of data, one for odd-numbered jobs and the other for even-numbered jobs.

Also, you can use `https://nightly.irsakudev.ipac.caltech.edu/mock-tap` as a TAP service within the TAP Panel.  This approach can be valuable when you need to mock TAP data for testing purposes in the future.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/JobInfo.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/JobInfo.java
@@ -39,7 +39,7 @@ public class JobInfo implements Serializable {
     private int executionDuration = LIFE_SPAN;
     private Instant destruction;
     private Map<String, String> params = new HashMap<>();
-    private List<String> results = new ArrayList<>();
+    private List<Result> results = new ArrayList<>();
     private Error error;
 
     // these are additional info that's not in uws:job defined props
@@ -105,11 +105,11 @@ public class JobInfo implements Serializable {
         this.error = error;
     }
 
-    public List<String> getResults() {
+    public List<Result> getResults() {
         return results;
     }
 
-    public void setResults(List<String> results) {
+    public void setResults(List<Result> results) {
         this.results = new ArrayList<>(results);
     }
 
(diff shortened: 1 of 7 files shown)


src/firefly/java/edu/caltech/ipac/firefly/core/background/JobInfo.java
```diff
@@ -204,6 +204,7 @@ public SrvParam getSrvParams() {
 //====================================================================
 
     public static record Error ( int code, String msg){}
+    public static record Result(String href, String mimeType, String size){};
```
robyww, line 207: Is there any more result data that we could keep?

robyww (approved):
This all looks goods. I think creating a mock service is great! I am seeing more and more needs for stuff like this. Clearly we might need more testing when  we hit a real service but for now this is all we need.

Thanks

src/firefly/java/edu/caltech/ipac/firefly/core/background/JobInfo.java
loitly: The schema is hard to read.  But, after reading it again, I found an attribute I missed; `xlink:type`.  I could add this.  But, I have no idea how it's used.  It's not documented anywhere except in the schema.

PR #1424, opened 2023-08-29
FIREFLY-1118: Update recenter action to allow coordinate system to be a string

jaladh-singhal:
Fixes [FIREFLY-1118](https://jira.ipac.caltech.edu/browse/FIREFLY-1118)

`recenterActionCreator` expects `centrePt` field in payload to be a `WorldPtObj` for `WorldPt` type. This works as long as payload is coming from the UI (because things are in the same JS world) but it will fail if payload comes through `WebSocketListener` i.e. used when calling it from FF python client.

Using `makeWorldPt` also encouraged me to use `CoordSys.parse` which is smart enough to parse any `cSys` string in the payload, giving more flexibility overall.

## Testing 
Make sure this doesn't break existing recenter option in image toolbar

Testing instruction for firefly python client in https://github.com/Caltech-IPAC/firefly_client/pull/53

### DIFF
--- a/src/firefly/js/visualize/task/PlotChangeTask.js
+++ b/src/firefly/js/visualize/task/PlotChangeTask.js
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {isNumber, isArray} from 'lodash';
+import {isNumber, isArray, set} from 'lodash';
 import {logger} from '../../util/Logger.js';
 import ImagePlotCntlr, { IMAGE_PLOT_KEY, dispatchWcsMatch, ActionScope, visRoot} from '../ImagePlotCntlr.js';
 import {
@@ -25,6 +25,7 @@ import {loadStretchData, queueChangeLocalRawDataColor} from '../rawData/RawDataO
 import {dispatchAddTaskCount, dispatchRemoveTaskCount} from '../../core/AppDataCntlr.js';
 import {Band} from '../Band.js';
 import {makeCubeCtxAry, populateFromHeader} from 'firefly/visualize/task/CreateTaskUtil.js';
+import {parseAnyPt} from 'firefly/visualize/Point';
 
 
 //=======================================================================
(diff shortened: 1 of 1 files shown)


src/firefly/js/visualize/task/PlotChangeTask.js
```diff
@@ -64,7 +66,17 @@ const dispatchAndMaybeMatch= (rawAction) => (dispatcher,getState) => {
     };
 
 
-export const recenterActionCreator= (rawAction) => dispatchAndMaybeMatch(rawAction);
+export const recenterActionCreator = (rawAction) => {
+    // `recenter` reducer expects centerPt to be WorldPtObj
+    const {centerPt} = rawAction.payload;
+    if (centerPt?.type === Point.W_PT && (!centerPt?.cSys || isString(centerPt?.cSys))) {
+        // type is WorldPt, and cSys is str or undefined (aka is not CoordinateSys object)
+        rawAction.payload.centerPt = makeWorldPt(centerPt.x, centerPt.y, CoordSys.parse(centerPt.cSys));
```
robyww, line 74: Firefly currently support two formats for work point.
- The world point object
- The serialized version `'ra;dec;coords'` for example `'1.111;2.222;EQ_J2000'`

You are actually proposing the third format a json object with cSys a string and not a constant that we then parse to a world point object. 

I think that is ok but this is not the place for it. It needs to be consistent across firefly.  It would be better for you to call `parseWorldPt` in `Point.js` and enhance it to support this format. In that case `parseWorldPt` should do all the work.

robyww (changes requested):
You on the right track but the work should be in `parseWorldPt`

src/firefly/js/visualize/task/PlotChangeTask.js
jaladh-singhal: I realised the serialsed version can do the same job if I use `parseAnyPt`. This even allows passing not WorlPt points like "SCREEN_PIXEL" etc. Thanks for the heads up!

jaladh-singhal: For testing: check if following works correctly:
- recentre option in image toolbar 
- existing method in notebook `fc.set_pan` with coord=
  - 'j2000'
  - 'image'
  - 'SCREEN_PIXEL'

robyww (approved):

PR #1426, opened 2023-09-12
FIREFLY-1310: Prototype of Drag Drop UI for file uploads 

kpuriIpac:
#### [Firefly-1310](https://jira.ipac.caltech.edu/browse/FIREFLY-1310)
- a simple drag drop UI for file uploads - probably just the first iteration of many, would like to iterate based on UI feedback received 
- We need to consider how big to make the space for drag & drop, and what happens to it after a user uploads a file, since we immediately display the file analysis results 
   - potentially make the entire window draggable/droppable for files? 

**Updates 10/10**
 - The entire window is draggable/droppable for files now 
 - added text that indicates you can choose a file from your OS's file explorer _or_ drag and drop a file
 - You can switch to **_Upload from URL_** and upload files from there too (this should switch the window back to **_Upload from file_** as well). 
    - To swtich the radio buttons back to **_Upload from file_**, I call **fireValueChange** in the **RadioGroupInputField** component (line 37)
    - This leads to some warnings in the console though (regarding updating a component while rendering a different component and updating state during an existing state transition). I am trying to fix it, but open to suggestions. 
  - The way this is handled now: The **FileDropZone** component sets state of some values on a file drop event. A useEffect in **FileUpload** component (in FileUpload.jsx) then handles the onDrop event. 
  - Finally, I discovered 3 existing bugs in our upload workflow during my work. I'll be creating a ticket for that. 
    - Here's that ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1332. I added short videos for each bug, haven't dived into them yet, let me know if you have any feedback. 

**Updates 10/20**
- Made some cleanup and other changes based on @robyww's feedback
  - especially for RadioGroupInputField, got rid of those changes by simply using useFieldGroupValue's setter, much simpler, and cleaner! 
  - also added the big drag and drop text that goes away when you upload a file. Make sure to test this on the TAP panel dialog (multi-object upload) as well as the Hips/MOC upload dialog. 
  

**Testing** (updated as of 10/10): 
- **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1310-drag-drop-file-ui/firefly
- **IrsaViewer**: https://firefly-1310-drag-drop-file-ui.irsakudev.ipac.caltech.edu/irsaviewer
- Go to "Upload" and test the drag and drop feature to upload a file 
- Switch to **Upload from URL** and attempt to drag and drop files here as well 
- From TAP searchers, select **Spatial type**: **Multi Object** and attempt to drag and drop files into this popup upload dialog 
- From Image Search, choose **Image Type**: **View HiPS Images** and search for m1. From the results view, from the **HiPS/MOC** menu dropdown, click on **Add MOC layer**, switch tabs to **Use my MOC** and attempt to drag/drop a MOC fits file here (or any other file - those will just not be accepted). 
    - this one is important to test as the HiPs upload popup dialog calls **FileUploadViewPanel** directly (instead of **FileUpoadDropdown**). For this reason, I had to move my drag and drop component (**FileDropZone**) inside **FileUploadViewPanel**.

### DIFF
--- a/src/firefly/js/ui/FileUpload.jsx
+++ b/src/firefly/js/ui/FileUpload.jsx
@@ -1,4 +1,4 @@
-import React, {memo} from 'react';
+import React, {memo, useEffect} from 'react';
 import PropTypes from 'prop-types';
 import {get, has, isFunction, isNil, isString} from 'lodash';
 
@@ -8,9 +8,8 @@ import {useFieldGroupConnector} from './FieldGroupConnector.jsx';
 import LOADING from 'html/images/gxt/loading.gif';
 import {upload} from '../rpc/CoreServices.js';
 
-
 function FileUploadView({fileType, isLoading, label, valid, wrapperStyle,  message, onChange, value, labelWidth,
-                         innerStyle, isFromURL, onUrlAnalysis, fileNameStyle}) {
+                         innerStyle, isFromURL, onUrlAnalysis, canDragDrop,fileNameStyle}) {
     var style = !isFromURL ? Object.assign({color: 'transparent', border: 'none', background: 'none'}, innerStyle) : (innerStyle || {});
     var fileName = (!isFromURL && value) ?  value.split(/(\\|\/)/g).pop() : 'No file chosen';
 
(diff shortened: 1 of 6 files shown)


src/firefly/js/ui/FileUpload.jsx
```diff
@@ -9,6 +9,40 @@ import LOADING from 'html/images/gxt/loading.gif';
...
+    const handleDragEnter = (e) => {
+        e.preventDefault();
+        setIsDragging(true);
+    };
+
+    const handleDragLeave = () => {
+        setIsDragging(false);
+    };
+
+    const handleDrop = (e) => {
```
loitly, line 25: Consider using `useCallback` when passing functions to another component to avoid unnecessary re-renders.

loitly (approved):
Good effort.  It Looks really nice.  I would move `Choose Files` and the text under it into the box like how you suggested it.  That way it's clear that it's an 'or' operation.

src/firefly/js/ui/FileUpload.jsx
```diff
@@ -89,7 +99,14 @@ FileUploadView.defaultProps = {
 export const FileUpload= memo( (props) => {
     const {viewProps, fireValueChange}=  useFieldGroupConnector(props);
     let modViewProps;
-    if (viewProps.isFromURL) {
+
+    useEffect(() => {
+        if (viewProps.isFileDrop) {
+            handleChange(viewProps.dropEvent, fireValueChange, viewProps.fileType, viewProps.fileAnalysis, viewProps.setFileDrop);
+        }
```
robyww, line 106: is approach is fine. Just don't use isFileDrop use `dropEvent`. I think you can get away with not using get/setFileDrop at all.

src/firefly/js/ui/FileUpload.jsx
```diff
@@ -148,10 +167,14 @@ function doUrlAnalysis(value, fireValueChange, type, fileAnalysis) {
...
-    var file = get(ev, 'target.files.0');
-    var displayValue = get(ev, 'target.value');
-
+function handleChange(ev, fireValueChange, type, fileAnalysis, setFileDrop) {
+    let file = get(ev, 'target.files.0');
+    let displayValue = get(ev, 'target.value'); //file.name;
+    if (ev.type === 'drop') { //drag drop files - instead of picking file from 'Choose File'
+        file = Array.from(ev.dataTransfer.files)[0];
+        displayValue = file.name;
+        setFileDrop(false);
```
robyww, line 176: There is probably a way to not need this line.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -162,16 +165,18 @@ export function FileUploadViewPanel({setSubmitText, acceptList, acceptOneItem})
...
             flexDirection: 'column' }}>
+            <FileDropZone {...{setFileDrop, setDropEvent}} style={{height:'100%', width: '100%'}}>
                 <div className='FileUpload'>
                     <div className='FileUpload__input'>
                         <RadioGroupInputField
                             initialState={{value: uploadMethod[0].value}}
                             fieldKey={uploadOptions}
                             alignment={'horizontal'}
                             options={uploadMethod}
+                            isFileDrop={isFileDrop}
```
robyww, line 176: I think you can change with and use an `useEffect` and `useFieldGroupValue` to change this field and not need extra properties.

robyww (changes requested):
It is all working but needs some more clean up. We can go over them when you are ready.

Summary of suggestions.
- Don't modify RadioGroupInputField, control the value instead with useEffect
- Only pass dropEvent to `FileUpload` and maybe a prop to control the message.
- remove  `const [isFileDrop, setFileDrop] = useState(() => false);` altogether. Control stuff by checking the event.

robyww (approved):
This looks good. I really appreciate you pursuing this idea. It is a major UI improvement.

 I think you can remove `setDropEvent` as a prop in `FileUpload`. Look at my two suggestions and play with that and see if you can get it to work. Either way go ahead and merge after that.

PR #1429, opened 2023-09-14
Firefly-1311: fixed: hips tile should not be smoothed at highest resolution

robyww:
#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1311-hips-smoothing/firefly
- zoom all the way into a hips images. The lowest level should be blocky


lrebull:
yes, i see pixels in the hips, and it is different than ops.

PR #1434, opened 2023-09-22
IRSA-5479: return empty string is app options specifies empty productTitleTemplate string

kpuriIpac:
#### [[IRSA-5479]](https://jira.ipac.caltech.edu/browse/IRSA-5479)
- this change is to return an empty string for obs core title if app options specifies a an empty tapObsCore.productTitleTemplate as discussed (aimed at euclid - but generic and can apply to other apps as well) 

**Testing**: 
- tested it locally and it works ('`Euclid, `' is taken out from the cutouts title), don't have a test build available for this as I'm still working on other irsa-ife changes for euclid. once I push  a commit for that, I can create a test build as well. 
  - actually, pushed the irsa-ife change and created a test build here: https://irsa-5479-cutouts-title.irsakudev.ipac.caltech.edu/applications/euclid/search
  - you should be able to do a region explorer search and see that '`Euclid, `' is now taken out of the title.

### DIFF
--- a/src/firefly/js/metaConvert/vo/ObsCoreConverter.js
+++ b/src/firefly/js/metaConvert/vo/ObsCoreConverter.js
@@ -211,6 +211,7 @@ function getObsCoreRowMetaInfo(table,row) {
 function createObsCoreTitle(table,row) {
  // 1. try a template
     const template= getAppOptions().tapObsCore?.productTitleTemplate;
+    if (!(template??'').trim()) return '';
     const templateColNames= template && getColNameFromTemplate(template);
     const columns= getColumns(table);
     if (templateColNames?.length && columns?.length) {


src/firefly/js/metaConvert/vo/ObsCoreConverter.js
```diff
@@ -211,6 +211,7 @@ function getObsCoreRowMetaInfo(table,row) {
 function createObsCoreTitle(table,row) {
  // 1. try a template
     const template= getAppOptions().tapObsCore?.productTitleTemplate;
+    if (typeof template !== 'undefined' && template.trim() === '') return ''; //if template is an empty string, assume that's what user wants and return empty string as is
```
robyww, line 214: template can be undefined and you to use it.

use: WebUtil - isDefined
```js
isDefined(getAppOptions().tapObsCore?.productTitleTemplate) 
```

robyww (changes requested):
What you are doing does not actually check if the key is defined.

src/firefly/js/metaConvert/vo/ObsCoreConverter.js
robyww: ok, after you slack message, I responded to just use lodash trim(), but after even more thought...

```js
if (!(template??'').trim()) return '';
```
should be fine.

robyww (approved):
Do that line and merge.

PR #1435, opened 2023-09-28
Firefly-1317: Improved readout handling

robyww:
#### [Firefly-1317](https://jira.ipac.caltech.edu/browse/FIREFLY-1317): Improved readout handling

 - Many details and Test FITS is image in [FIREFLY-1223](https://jira.ipac.caltech.edu/browse/FIREFLY-1223)
 - fixed the bug with a large 64 big blank value (BLANK) header
 - show "undefined" for blank values
 - handle HEX display
 - give accurate long readout: reliably display the exact value of a 64-bit integer 
 - revamped the passing of flux data to use JSON

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1317-integer/firefly
 - Use design ticket for testing
 - switch between decimal and hexadecimal readout
 - readout switch to HEX is made by clicking on the `Value:` on the mouse readout much like you do changing coordinate system.


lrebull:
@robyww where do you want us to test this? i don't see a build linked here where we can test it. Is it really on dev already? Eric is going to test this for us.

ebratton01:
@robyww I have tested the feature and it works:

1.  It prompts the user to switch readout values between decimal and hexadecimal once they click on "value:" or "flux:"
2. It displays "undefined" for very large integers (blank values), or "0x7fffffffffffffff (undefined)" when in hexadecimal
3. It correctly converts the integer values from decimal to hexadecimal

But, it does not correctly convert floating point decimal values to hexadecimal. I tested it using Spitzer SEIP catalog searching for M81 and 500 arcsecond cutout size.

Specific coordinate: EQ-J2000: 9h54m48.01s +69d03m16.5s
Pixel decimal flux value is 5137958 MJy/sr and should convert to hexadecimal value 4E6626.
Instead pixel hexdecimal value reads 0x40a46a26 MJy/sr, which converts to 1,084,516,902 MJy/sr.


Screenshot 2023-10-02 at 3 12 50 PM
Screenshot 2023-10-02 at 3 12 35 PM
Screenshot 2023-10-02 at 3 10 56 PM
Screenshot 2023-10-02 at 3 12 06 PM

lrebull:
(just trying to help facilitate in the translation between "scientist" and "developer" -- if it's not meant to translate floating point to hex, then it shouldn't quietly attempt it and present something as if it's succeeded... if it's not meant to translate floating point, then it should fail elegantly, not quietly do it wrong...)

robyww:
@ebratton01 - Maybe we need a little more documentation on the dialog

> But, it does not correctly convert floating point decimal values to hexadecimal. I tested it using Spitzer SEIP catalog searching for M81 and 500 arcsecond cutout size.

It is not trying to convert the hex to the flux, it is trying to give the hex view of the ieee floating point. This was in the design ticket but it is a little difficult to understand.

robyww:
> (just trying to help facilitate in the translation between "scientist" and "developer" -- if it's not meant to translate floating point to hex, then it shouldn't quietly attempt it and present something as if it's succeeded... if it's not meant to translate floating point, then it should fail elegantly, not quietly do it wrong...)

It is not doing anything wrong. It is showing the raw value in the FITs file as a hexadecimal number. It is not the value of the number in the fits file it is the raw representation of it. if you were to dump the FITS file this is what you would see.

gpdf (changes requested):
Tested with files from SOFIA (64-bit float and 64-bit integer), 32-bit float files from WISE, etc., and 16-bit integer files from SDSS.  I wasn't able to quickly scare up a 32-bit integer file.

It looks like it's basically what was intended.  A couple of issues:

1)

When lock-by-click is on, and you change the decimal/hexadecimal mode, the display of the already-selected pixel value doesn't change.  This is true regardless of whether the value display is in "compact" mode (at the bottom of the window) or in "inspector" mode (with a popped-out dialog).  This needs to work, as a natural thing to do if debugging some weird value is to click back and forth between the two.

2)

It looks to me as if the display is incorrect for 16-bit integer images.

It displays, e.g., 0xffffa161 for a pixel in an "SDSS r" image obtained from the usual IRSA Viewer image search.  These images have BITPIX=16.  The value should be displayed with four hex digits.

I've attached a sample image in this format.

[image_SDSS-r.fits.zip](https://github.com/Caltech-IPAC/firefly/files/12811377/image_SDSS-r.fits.zip)

gpdf:
Also just a remark to reinforce the earlier discussion: the point of the "hex" feature is exactly as said above: to show the actual binary value in the FITS file.  So 0x40a46a26, in IEEE 32-bit floating point format, is the number 5.13795757... (showing some excess digits for clarity), which is exactly what the regular decimal floating-point display in Firefly shows.

https://www.h-schmidt.net/FloatConverter/IEEE754.html is a nice conversion tool.

ebratton01:
> Also just a remark to reinforce the earlier discussion: the point of the "hex" feature is exactly as said above: to show the actual binary value in the FITS file. So 0x40a46a26, in IEEE 32-bit floating point format, is the number 5.13795757... (showing some excess digits for clarity), which is exactly what the regular decimal floating-point display in Firefly shows.
> 
> https://www.h-schmidt.net/FloatConverter/IEEE754.html is a nice conversion tool.

I see, thanks I didn't know that. I'll keep this tool in mind.

robyww:
@gpdf The build is updated with your requested changes and bug fixes.

gpdf (changes requested):
I do see that the display of 16-bit integer images is fixed.  Thank you!

However, I'm still having some issues.  I loaded a SOFIA image:

[image_F0894_HA_POL_08018676_HAWEHWPE_CAL_047-054B.fits.zip](https://github.com/Caltech-IPAC/firefly/files/12825830/image_F0894_HA_POL_08018676_HAWEHWPE_CAL_047-054B.fits.zip)

which has both 64-bit float (double) HDUs and one 64-bit integer HDU.

Specifically, I loaded extension 1 and extension 9, which are double and int64, respectively.

When the image viewer was displaying the integer extension, I brought up the new display-mode dialog, and found that in order to get the value display for the integer extension to change, I had to change the setting for floating-point pixel value display.  The radio buttons for the integer display had no effect.

In addition, the dialog goes away as soon as you click on a single radio button, so it's impossible to change both settings to hex in the same action.

Finally, and I'm sorry for not noticing this before, the dialog needs a (?) icon for a future help link.  I realize that there's already a pretty good explanation in the dialog itself, but we're just trying to be consistent.

robyww:
Todo:

- [x] hex monospace
- [x] auto-close on the dialog
- [ ] help link
- [ ] look at last bug

new build: 11/6

PR #1439, opened 2023-10-17
Firefly-1325: Datalink 1.1

robyww:
#### [FIREFLY-1325](https://jira.ipac.caltech.edu/browse/FIREFLY-1325) - Support Datalink 1.1

- support datalink 1.1
- refactor VOAnalyzer and make it a directory
  - `VoAnalyzer.js` is now a `voAnalyzer` directory with:
       - ColumnsModelInfo.js, TableAnalysis.js, VoCoreUtils.js, VoTableRecognizer.js
       - SpectrumDM.js, VoConst.js, VoDataLinkServDef.js, vo-typedefs.jsdoc

- Cleanup of HealpixIndex
- fixes: [Firefly-1323](https://jira.ipac.caltech.edu/browse/FIREFLY-1323)

#### Testing

- do the Euclid searches it should work as before
- https://firefly-1325-datelink11.irsakudev.ipac.caltech.edu/applications/euclid

##### Review - file to focus on.
- `VoDataLinkServDef.js` - added 1.1 support
- a new DLAnalysis object to how the analysis of the semitics. 
- everything else under `voAnalyzer/` is break up of `VoAnalyzer.js`
- Many, many files are just import changes
- `DLGenAnalyzeSearch.js` - now uses same tools ` VoDataLinkServDef.js` to parse datalink file
- The refactoring expose some circular dependency import issues. This is solved my moving some use of constants inside of functions instead of at the module level.


kpuriIpac:
Will continue reviewing tomorrow, but I noticed that the test build doesn't really work. The results are the same as before for region explorer results, the RHS displays: `no support for related grid in datalink file`. 

And for source inspector results, it either says: `No preview available` or `No data to display: Could not retrieve datalink data, TypeError: Cannot read properties of null (reading 'includes')`

kpuriIpac:
> Will continue reviewing tomorrow, but I noticed that the test build doesn't really work. The results are the same as before for region explorer results, the RHS displays: `no support for related grid in datalink file`.
> 
> And for source inspector results, it either says: `No preview available` or `No data to display: Could not retrieve datalink data, TypeError: Cannot read properties of null (reading 'includes')`

I cleared cache and tested this on VPN, so at least that wasn't the issue. But let me know if these still work for you and I can check why it doesn't load for me?

robyww:
@kpuriIpac I think I must have built it against the wrong branch. The about box showed the wrong commit. Anyway I just rebuilt it and it is working.

kpuriIpac (approved):
Looks good. Looked at `VoDataLinkServDef.js` more closely than others, and skimmed through the other changes. Nothing stood out as incorrect. Splitting up the files into `voAnalyzer/ `directory makes sense.

PR #1441, opened 2023-10-26
Firefly 1320: Search Panel routes for Euclid SphereX

kpuriIpac:
#### [Firefly-1320](https://jira.ipac.caltech.edu/browse/FIREFLY-1320)
- irsa-ife PR: https://github.com/IPAC-SW/irsa-ife/pull/294
- React Router's `useNavigate()` may be used only in the context of a <Router> component which is why I had to separate the `SearchPanel` component into a new `SearchPanelRoutes` component (if I tried to use `useNavigate()` in the existing `SearchPanel` component, things would be fine for euclid and spherex apps, but would break for wise and other apps
- SearchPanelRoutes now works with the euclid app, which means we would have had two submit and two cancel buttons on the search page (since SearchPanel internally calls FormPanel, but the euclid search components - RegionExplorer, Source Inspector also call FormPanel)
   - So I had to add a conditional in FormPanel component whose default value is always true to display the Search/Cancel buttons, and only false from SearchPanelRoutes (for Euclid, SphereX). So the rest of the app should be unaffected but there may still be a better way to do this instead of modifying FormPanel directly? 
- **_ToDo_**: if user enters a path like `euclid/search/region-explorer`, then the search page should be displayed with Region Explorer selected as the search. Right now though, going to a route like `euclid/search/region-explorer` yields nothing at all, so this may be a irsa-ife change with react router needed to make this work. 

**Update 10/30**: 
- Got rid of SearchPanelRoutes component based on @loitly's feedback 

**Update 11/3**
- Got rid of making conditional changes *inside* the `FormPanel` component though, to not display the Search/Cancel buttons when it's a route based app. Instead, I was able to just conditionally not render the `FormPanel` for route based apps in SearchPanel.jsx's `SearchForm` component (see irsa-ife PR for more details on why I chose to do this)

**Update 11/10**
- removed `routerStyle`, just using `SearchPanel's` default style 
- after some testing, to make the MOCs for these route based apps take up the entire width available, I had to add the following style `style={{width: '100%', height: '100%'}}` to the parent div of the `<div className='SearchPanel' style={style}>` (~line 45 in SearchPanel.jsx)
  - this and the change in the menu.css for `.SearchPanel` allows the mocs to take up the entire width available. Even though this is a change that shouldn't affect other apps (as width and height are set to 100% instead of hardcoded values), I still built and tested the following apps (apps that use `SearchPanel`) to ensure this change didn't affect their look: ztf, SHA, Sofia, FinderChart and wise. 
- not conditionally rendering `FormPanel` anymore (instead removed the `FormPanel` calls from the individual search components in euclid, so that `SearchPanel` pretty much remains the same as before)
- not using the watcher in `RouterChange` anymore (and changed its name to `RouterSearchItem` instead). Instead, using searchInfo (`search`) available from `SearchItem` directly as @loitly suggested
- general cleanup 

**Testing**: 
- **Euclid**: https://firefly-1320-searchpanel-routes.irsakudev.ipac.caltech.edu/applications/euclid
- select one of 3 different search options, click search. Click on the "Search" tab from the nav bar to go back to the search page. Your previous search should be selected here. Proof of concept for `SearchPanel` (or rather `SearchPanelRoutes`) working with route based app.

### DIFF
--- a/src/firefly/js/ui/SearchPanel.jsx
+++ b/src/firefly/js/ui/SearchPanel.jsx
@@ -11,6 +11,7 @@ import {FormPanel} from './FormPanel.jsx';
 import {useStoreConnector} from './SimpleComponent.jsx';
 import {StatefulTabs, Tab} from './panel/TabPanel.jsx';
 import {makeSearchOnce} from '../util/WebUtil';
+import {useNavigate, useInRouterContext} from 'react-router-dom';
 
 const changeSearchOptionOnce= makeSearchOnce(); // setup options to immediately execute the search the first time
 
@@ -31,7 +32,7 @@ export function SearchPanel({style={}, initArgs={}}) {
     const isSingleSearch = Object.keys(allSearchItems).length === 1;
     if (isSingleSearch) {
         return (
-            <div>
+            <div className='SearchPanel-box'>
                 {title && <h2 style={{textAlign: 'center'}}>{title}</h2>}
                 <SearchForm style={{height: 'auto'}} searchItem={searchItem} initArgs={initArgs}/>
             </div>
@@ -41,7 +42,7 @@ export function SearchPanel({style={}, initArgs={}}) {
     if (flow === 'vertical') {
         const sideBar = <SideBar {...{activeSearch, groups}}/> ;
         return (
-            <div>
+            <div className='SearchPanel-box'>
                 {title && <h2 style={{textAlign: 'center'}}>{title}</h2>}
                 <div className='SearchPanel' style={style}>
                     {sideBar}
@@ -56,7 +57,7 @@ export function SearchPanel({style={}, initArgs={}}) {
     } else {
         const onTabSelect = (index,id,name) => dispatchUpdateAppData(set({}, ['searches', 'activeSearch'], name));
         return (
-            <div>
+            <div className='SearchPanel-box'>
                 {title && <h2 style={{textAlign: 'center'}}>{title}</h2>}
                 <StatefulTabs componentKey={`SearchPanel_${title}`} onTabSelect={onTabSelect} resizable={true} useFlex={true} borderless={true} contentStyle={{backgroundColor: 'transparent'}}>
                     {searchesAsTabs(allSearchItems, initArgs)}
@@ -138,10 +139,21 @@ function SearchGroup({group, activeSearch}) {
     );
 }
 
+function RouterSearchItem({onClick, ttips, clsname, label, search}) {
+    const navigate = useNavigate();
+    const handleClick = () => {
+        onClick();
+        navigate(search.path);
+    };
+    return <div className='SearchPanel__searchItem' onClick={handleClick} title={ttips}><span className={clsname}>{label}</span></div>;
+}
+
 function SearchItem({search, activeSearch}) {
+    const isUsingRouter = useInRouterContext();
     const ttips = search.desc || search.title || search.name;
     const label = search.title || search.name;
     const clsname = search.name ===  activeSearch ? 'selected' : 'normal';
     const onClick = () => dispatchUpdateAppData(set({}, ['searches', 'activeSearch'], search.name));
-    return <div className='SearchPanel__searchItem' onClick={onClick} title={ttips}><span className={clsname}>{label}</span></div>;
+    if (isUsingRouter) return <RouterSearchItem {...{onClick, ttips, clsname, label, search}}/>;
+    else return <div className='SearchPanel__searchItem' onClick={onClick} title={ttips}><span className={clsname}>{label}</span></div>;
 }
--- a/src/firefly/js/ui/FormPanel.jsx
+++ b/src/firefly/js/ui/FormPanel.jsx
@@ -9,7 +9,7 @@ import * as TablesCntlr from '../tables/TablesCntlr.js';
 import {HelpIcon} from './HelpIcon.jsx';
 import {dispatchHideDropDown} from '../core/LayoutCntlr.js';
 import {makeTblRequest} from '../tables/TableRequestUtil.js';
-import {get, isNil} from 'lodash';
+import {isNil} from 'lodash';
 import {dispatchFormCancel, dispatchFormSubmit} from 'firefly/core/AppDataCntlr.js';
 
 function handleFailure() {
@@ -46,8 +46,7 @@ function createSuccessHandler(action, params={}, title, onSubmit) {
 
 export const FormPanel = function (props) {
     const { children, onSuccess, onSubmit, onCancel=dispatchHideDropDown, onError, groupKey, groupsToUse,
-        action, params, title, getDoOnClickFunc,
-        submitText='Search',cancelText='Cancel', help_id, changeMasking,
+        action, params, title, getDoOnClickFunc, submitText='Search',cancelText='Cancel', help_id, changeMasking,
         includeUnmounted=false, extraWidgets=[], extraWidgetsRight=[]} = props;
     let { style, inputStyle, submitBarStyle, buttonStyle} = props;
 


src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -12,12 +12,12 @@ import {useStoreConnector} from './SimpleComponent.jsx';
...
 import {makeSearchOnce} from '../util/WebUtil';
 
-const changeSearchOptionOnce= makeSearchOnce(); // setup options to immediately execute the search the first time
+export const changeSearchOptionOnce= makeSearchOnce(); // setup options to immediately execute the search the first time
 
 export function SearchPanel({style={}, initArgs={}}) {
     const searchPanelState = useStoreConnector(getSearchInfo);
 
-    const {activeSearch, groups, allSearchItems={}, flow='vertical', title} = searchPanelState;
+    const {activeSearch, groups, allSearchItems={}, flow='vertical', title, searchStyle} = searchPanelState;
```
loitly, line 20: Adding style into `searchInfo` seem awkward to me.  There has to be a better way.  There's already a `style` prop that you can use to effect the style of this component.  (Btw, you removed it.  This is normally not a good idea.  It may affect existing code.)
There is also a `SearchPanel` class that you can use at the css level.  Regardless, I don't think adding `searchStyle` as a state is a good idea.

loitly (changes requested):
There's no reason to create a separate component.  I've mentioned to you before that there's a method you can use to check whether a component is rendered in a react router context.  I searched and found this.  It should work.  I tested it in WISE.

```   
 const isUsingRouter = useInRouterContext();
 console.log('isUsingRouter:' + isUsingRouter);
```

Given this, I have to ask if you can redo this without introducing the 2nd component.  Thanks.

kpuriIpac:
> There's no reason to create a separate component. I've mentioned to you before that there's a method you can use to check whether a component is rendered in a react router context. I searched and found this. It should work. I tested it in WISE.
> 
> ```
>  const isUsingRouter = useInRouterContext();
>  console.log('isUsingRouter:' + isUsingRouter);
> ```
> 
> Given this, I have to ask if you can redo this without introducing the 2nd component. Thanks.

Thanks @loitly. I'll use `useInRouterContext()` (I tested too and it works) and update the style use. 

The reason I created a new component was because of the call to `useNavigate()` (which I then use to call `navigate(searchItem.path)`). Apparently, if the context in which the component is called is not react-router, then there's an error (so for example if I use `useNavigate()` in `SearchPanel`, then the wise app won't work). And since `useNavigate` is a react hook, I can't call it conditionally based on `isUsingRouter's` value. I can show you this later, or do you have any suggestions based on this info?

loitly:
> or do you have any suggestions based on this info?

One way is to add the useInRouterContext check here: https://github.com/Caltech-IPAC/firefly/blob/114443900f8203c6d0b9316357fe3d62052319fc/src/firefly/js/ui/SearchPanel.jsx#L141

At line 146, if it's true, then you can render the component that's reactrouter-awared, otherwise, return what it is now.

kpuriIpac:
> > or do you have any suggestions based on this info?
> 
> One way is to add the useInRouterContext check here:
> 
> https://github.com/Caltech-IPAC/firefly/blob/114443900f8203c6d0b9316357fe3d62052319fc/src/firefly/js/ui/SearchPanel.jsx#L141
> 
> At line 146, if it's true, then you can render the component that's reactrouter-awared, otherwise, return what it is now.

Thanks @loitly, this worked great. Got rid of the SearchPanelRoutes component in the latest commit. Updated the test build. Still sending in style (`routerStyle`) for router based apps to the `searchPanelState`, but using defualt `style` prop again for all others. Will try and fix this to use the style prop for router based apps too.

src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -138,10 +147,28 @@ function SearchGroup({group, activeSearch}) {
     );
 }
 
+function RouterChange({onClick, ttips, clsname, label}) {
+    const searchPanelState = useStoreConnector(getSearchInfo);
+    const navigate = useNavigate();
+
+    const {activeSearch, allSearchItems = {}} = searchPanelState;
+    const searchItem = allSearchItems[activeSearch];
```
loitly, line 155: `RouterChange` is not a good name.  It should be something that indicate it's a presentation of a Router's SearchItem.
Being that it's a presenter, why does it need to watch for data changes?
`    const searchPanelState = useStoreConnector(getSearchInfo);
`
All of the needed data should already be available from `SearchItem`.

src/firefly/js/ui/Menu.css
```diff
@@ -67,6 +67,8 @@
     min-width: 800px;
     min-height: 400px;
     margin: -3px;
+    height: 100%;
+    width: 100%;
```
loitly, line 71: Why were these added?  It should not if it were added to fix issues that's only related to router support.

src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -41,7 +42,7 @@ export function SearchPanel({style={}, initArgs={}}) {
     if (flow === 'vertical') {
         const sideBar = <SideBar {...{activeSearch, groups}}/> ;
         return (
-            <div>
+            <div style={{width: '100%', height: '100%'}}>
```
loitly, line 45: Same concern as Menu.css.  These are generic code.  Are you sure it is needed by everyone.

src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -114,7 +115,7 @@ function SearchForm({searchItem, style, initArgs}) {
     );
 }
 
-function SideBar({activeSearch, groups=[]}) {
+export function SideBar({activeSearch, groups=[]}) {
```
loitly, line 118: Who is using these exports?

src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -138,10 +139,21 @@ function SearchGroup({group, activeSearch}) {
     );
 }
 
+function RouterSearchItem({onClick, ttips, clsname, label, search}) {
+    const navigate = useNavigate();
+    const handleClick = () => {
+        onClick();
+        navigate(search.path);
+    };
+    return <div className='SearchPanel__searchItem' onClick={handleClick} title={ttips}><span className={clsname}>{label}</span></div>;
```
loitly, line 148: Perfect.

src/firefly/js/ui/SearchPanel.jsx
kpuriIpac: The `SearchForm` and `SideBar` exports were unused, removed them (I think those were for the `SearchPanelRouter` component I had previously created, but forgot to remove the exports while cleaning that up)

src/firefly/js/ui/SearchPanel.jsx
kpuriIpac: Hey @loitly, so you're right, these are not needed by the other apps. But mainly for router apps (the idea being they end up using the entire screen's width and height for the MOC background). I found that adding these doesn't affect other apps, but nonetheless, it still applies these styles to other apps, which is why I initially had conditional styles for router based apps. If I don't add the style here and in menu.css, then this is what euclid, for example, looks like (similar to Wise/SHA, etc.): 

euclid_ss3

If we don't want to add style in this way or conditionally if it's router based app, do you have a suggestion on how I could approach this?

kpuriIpac: Actually @loitly, I was able to get rid of this change from menu.css at least! 

(based on your suggestion in irsa-ife PR) By adding the following to euclid.css instead: 

```
.SearchPanel {
    height: 100%;
    width: 100%;
}
```

But I think I still need to add `style={{width: '100%', height: '100%'}}` to the div in SearchPanel.jsx, becuase I tinkered with it a lot, and without adding it there, we end up with something like this: 

euclid_ss2

loitly: I see what the problem is.  I would add `className='SearchPanel-Box` to line: 36, 46, and 61.  This allow other to override as needed.  Putting fixed style without the ability to override may present an issue later.

src/firefly/js/ui/SearchPanel.jsx
loitly: I see you also remove Search.jsx.  But, you forgot to remove the classNames that was added to euclid.css.

src/firefly/js/ui/SearchPanel.jsx
kpuriIpac: @loitly just to confirm, you mean I should add this className to the existing div elements right? (which would be lines 35, 45 and 60?) so it would look like, for example, for lines 45-47: 

```
<div className='SearchPanel__box'>
                {title && <h2 style={{textAlign: 'center'}}>{title}</h2>}
                <div className='SearchPanel' style={style}>
```

loitly: No.  Here's an example...
```
            <div className='SearchPanel-Box'>
                {title && <h2 style={{textAlign: 'center'}}>{title}</h2>}
                <div className='SearchPanel' style={isUsingRouter ? routerStyle : style}>
                    {sideBar}
                    {searchItem &&
                    <div className='SearchPanel__form'>
                        <SearchForm searchItem={searchItem}  initArgs={initArgs}/>
                    </div>
                    }
            </div>

```

kpuriIpac: wait, I'm confused, isn't that the same? (the div that `className='SearchPanel-Box'` is being added to?)

kpuriIpac: @loitly can you see the latest commit here? I removed the unused css and added style for `SearchPanel-Box` (that's in the irsa-ife PR) and over here I added the` SearchPanel-Box` className in 3 places. Does that look right?

loitly (approved):
Looks good.

src/firefly/js/ui/SearchPanel.jsx
```diff
@@ -101,20 +103,27 @@ function executeOK(clickFunc,initArgs,searchItem) {
...
+
+    else {
+        return (
+            <FormPanel groupKey={name} style={style}
+                       getDoOnClickFunc={(clickFunc) => executeOK(clickFunc,initArgs,searchItem) }
+                       {...rest}>
+                <Render {...{searchItem, initArgs}} />
+            </FormPanel>
+        );
+    }
```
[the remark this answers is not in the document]
kpuriIpac: @loitly just pinging to bring this up for discussion on Monday. Now that we're handling stuff a little differently (In Euclid, `Search` and `Cancel` buttons need to be _inside_ the EmbeddedPositionSearch Panel for `Search By Position: Region Explorer and Source Position`), so we need to get rid of the `Search` and `Cancel` buttons outside (basically removing the call to `FormPanel` here)...

but, we still need to keep the call to `FormPanel` for `Source Inspector: Search By ID`...So yeah, a little complicated, I'll think of something but just wanted to put this on your radar for discussion on Monday.

PR #1446, opened 2023-11-08
FIREFLY-1318: Add documentation link to DCE result table

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1318
Show link to dataset's documentation page in DCE result table.

Test: https://firefly-1318-dce-doclink-in-result.irsakudev.ipac.caltech.edu/irsaviewer/dce
Test with one table as well as multiple tables.
Should also test to see when there's more than one docrows.  (I don't know of a dataset that do).


robyww:
I thought you were going to make this a button like `Prepare Download`. As it is having a button, then a like then the navigation then the toolbar looks awkward. Buttons on left and toolbar on the right would look more even.

loitly:
> I thought you were going to make this a button like `Prepare Download`. As it is having a button, then a like then the navigation then the toolbar looks awkward. Buttons on left and toolbar on the right would look more even.

Having a button as link is also not a very good look.  But, I'll make the change so you guys can review it on Friday.

robyww (approved):

robyww:
Wait to we show it to the sci-ccb before merging.

PR #1447, opened 2023-11-10
FIREFLY-1297: Add support for null values to all table's data types.

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1297
Fixed: Firefly incorrectly interprets null values as 'false' when a column is specified as a Boolean.
This also expand NULL support to all data types. 

Test: https://firefly-1297-table-support-null-primitives.irsakudev.ipac.caltech.edu/irsaviewer/dce
- Select Facility: MSX
- Search on any covered area with 300 arcsec radius.

In the result table, `environment_photometric` column should be all NULLs instead of `false`.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
+++ b/src/firefly/java/edu/caltech/ipac/table/PrimitiveList.java
@@ -31,10 +31,11 @@ public interface PrimitiveList {
         }
     }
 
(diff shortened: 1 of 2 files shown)


robyww (approved):

lrebull:
this works!

PR #1451, opened 2023-11-21
Firefly-871: Add wavelength to "line extract" and "pick points" tools

robyww:
#### Testing:  [Firefly-871](https://jira.ipac.caltech.edu/browse/FIREFLY-871)
  - https://fireflydev.ipac.caltech.edu/firefly-871-wl/firefly 
  - upload wl.fits
  - extract line or points
  - save to table
  - there should be a column for wavelength
  - [wl.fits.zip](https://github.com/Caltech-IPAC/firefly/files/13430210/wl.fits.zip)


lrebull:
There is a column in the extracted line called 'wavelength' but i don't have a scientifically valid file to test for scientific validity...

PR #1452, opened 2023-11-27
Firefly-1235: Adding region file format to table save when position columns are available

kpuriIpac:
#### [Firefly-1235](https://jira.ipac.caltech.edu/browse/FIREFLY-1235)
- adds option of file format for region file to table save when position columns are available 
- downloads region file if selected 
- @robyww as I mentioned in the standup, not a lot of changes here, I'll still work on some cleanup but if you could take a quick look at the code to make sure it makes sense overall (or point out if something stands out as incorrect)

**Updates 12/1**
- added support for saving files with one column array entry of ra,dec values (assuming entry arrays are always going to be type `double`), test build updated. 

**Updates 11/29**
  - Please see my comment in `TableAnalysis.js` line 34. Is there a table I can use to test and write code for single column entries for center columns? 
  - What should the behavior look like when users filter out the center columns (one or both), @robyww? as of now by adding a null check for either of the columns, I'm still writing and saving a file with just `J2000; color=blue` in it.  Should I throw an error instead?
  - I manually tested a table with no center columns (but edited client side to still show `Region (.reg)` as a file format) and tried to save it, got an error as expected. (although it's a generic error message`Request failed with status 500: http://localhost:8080/firefly/CmdSrv/sync?cmd=tableSave` but the exception on server side logs `Unable to find center columns`). This client side error message we get is from the `catch` statement on line 94-95 in` fetch.js` (I'm trying to figure out how to display the errors from the response object - so that it's more descriptive)...

**Testing**: 
- https://fireflydev.ipac.caltech.edu/firefly-1235-table-region/firefly
- do a TAP search that includes position columns in the resulting table (e.g.: ra,dec)
- try to save this table
- you should see .reg (or region) as one of the File Format options
- download the table as a region file and verify if the resulting file looks ok (should only have position column entries).

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/EmbeddedDbProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/EmbeddedDbProcessor.java
@@ -6,6 +6,7 @@ package edu.caltech.ipac.firefly.server.query;
 import edu.caltech.ipac.table.TableUtil;
 import edu.caltech.ipac.table.io.IpacTableException;
 import edu.caltech.ipac.table.io.IpacTableWriter;
+import edu.caltech.ipac.table.io.RegionTableWriter;
 import edu.caltech.ipac.table.io.DsvTableIO;
 import edu.caltech.ipac.table.io.VoTableWriter;
 import edu.caltech.ipac.firefly.data.FileInfo;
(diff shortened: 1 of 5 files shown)


kpuriIpac:
@robyww I addressed the comments above and pushed a new commit. Updated the test build. Let me know what you think.

robyww (approved):
Looks good.

lrebull:
Woo hoo! this works! It's kind of crazy how happy this makes me.

Is this the right time to ask for some sort of indicator in the pop-up from the "save" icon when invoked from the image to remind users that the region file when saved from there does not, in fact, save the entire catalog, but just saves the decimated catalog? We could do something as static as this:
Screen Shot 2023-11-28 at 9 18 57 AM

lrebull:
i think they're all linked to the ticket in question here...
https://jira.ipac.caltech.edu/browse/FIREFLY-955
https://jira.ipac.caltech.edu/browse/FIREFLY-578

kpuriIpac:
@robyww is there some stuff that can still go in this PR from the CCB ticket, like one of the things I noticed in https://jira.ipac.caltech.edu/browse/FIREFLY-955: 

`The table name (from the table tab) should be saved as a "#" comment at the top of the file.`

Or is this PR good as is, for now?

PR #1456, opened 2023-12-01
Firefly-1350: Add MOC outline style

robyww:
#### [Firefly-1350](https://jira.ipac.caltech.edu/browse/FIREFLY-1350): Add MOC outline style

 - Do outline with MOC
 - put in the right opacity
 - Moc default style
 - moc max depth to 5 with an appOption

Testing:
- https://firefly-1350-moc-outline.irsakudev.ipac.caltech.edu/irsaviewer/dce
- click around on display, all the MOCs should come up as outlines
- choose the drawing layers to change to other modes
- _note:_ sometime moc outlines seem like they are double for certain ones. This is really the best we can do.

### DIFF
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -156,7 +156,9 @@ const defFireflyOptions = {
         useForImageSearch: true,
         hipsSources: 'all',
         defHipsSources: {source: 'irsa', label: 'Featured'},
-        mergedListPriority: 'irsa'
+        mergedListPriority: 'irsa',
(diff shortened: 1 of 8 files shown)


loitly (approved):
Changes look good.

lrebull:
Everything I've tried so far has resulted in double outlines, but I still think this is an improvement over what it was before. See attached. 
Screen Shot 2023-12-01 at 1 59 43 PM

lrebull:
this one is less of an improvement...

Screen Shot 2023-12-01 at 2 00 45 PM

PR #1459, opened 2023-12-07
FIREFLY-1353: Add arrow buttons for image scrolling in ImageTableRowViewer

jaladh-singhal:
Fixes [FIREFLY-1353](https://jira.ipac.caltech.edu/browse/FIREFLY-1353)

- Added react-slick slider to use in ImageTableRowViewer. slick-carousel package is also added for the css needed for react-slick
- Expose layout/container creating function in MultiItemViewerView so that items (images) can be rendered inside slider
- Synchronized changes in slide with change in highlighted row and vice versa (note: this is where logic gets bit complicated because we have to prevent it from looping)
- Created a custom slider arrow button for above purpose and to cancel the event when slider is hitting the left edge or right edge (so that it doesn't slide and active image moves to center)
- Made change in active plot through UI, change the highlighted row 
- Make height of slider (and inner components) take height of its parent container - that's why CSS sheet was added to override default slick styles

## Testing
https://firefly-1353-arrow-btn-imagerowviewer.irsakudev.ipac.caltech.edu/applications/spherex/

Make sure spectrum table is open - it's easier to test that way

- Click arrows, it should change the highlighted row. Note that slider won't really move until the active image comes to centre. Afterwards it will slide the slider along with moving the active image by one.
- Click on any row in table (or point in plot), it should change the slider position and active image. Also try changing the table pages.
-  Click on any image that is not active, it should change the highlighted row ~and move the slider to bring to center.~ preserving the slider position. On subsequent arrow clicks, test that it will move active plot/row or slider to bring it to center.

### DIFF
--- a/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
+++ b/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
@@ -9,6 +9,7 @@ import {isNil, xor} from 'lodash';
 import Slider from 'react-slick';
 import 'slick-carousel/slick/slick.css';
 import 'slick-carousel/slick/slick-theme.css';
+import './ImageTableRowViewer.css';
 
 import {useFieldGroupValue, useStoreConnector} from '../../ui/SimpleComponent.jsx';
 import {getTblById} from '../../tables/TableUtil.js';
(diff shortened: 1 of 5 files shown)


src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -48,8 +54,80 @@ export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutout
...
+        if(table?.highlightedRow !== activePlotRowNum) {
+            //on active plot change, change highlighted row if not same
+            dispatchTableHighlight(tbl_id, activePlotRowNum);
+        }
+    }, [activePlotId]);
+
+    const sliderRef = useRef(null);
+    const slideChangeByArrows = useRef(false);
+
+    const makeImageSlider = (viewerItemIds, makeItemViewer) => {
```
robyww, line 79: move makeImageSlider out of the component. It is just too big and make reading it hard.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -64,7 +142,7 @@ export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutout
         <MultiImageViewer
             {...{viewerId, Toolbar, insideFlex, defaultImageCnt, maxImageCnt, tableId:tbl_id,
                 makeRequestFromRow,defaultCutoutSizeAS,
-                wcsMatchType, activePlotId,
+                wcsMatchType, activePlotId, makeCustomLayout: makeImageSlider,
```
robyww, line 145: if you move `makeImageSlide` out this will need to become something liek
```js
makeCustomLayout: (viewerItemIds, makeItemViewer) => makeImageSlider(viewerItemIds, makeItemViewer,table,viewerId, etc)
```

robyww (commented):
Besides my comment below...

In `Toolbar` add two props.
 - `useCutoutSize` to determine is you show the `RadioGroupInputField` at line 205
 -  `defaultWcsMatchType` which can be 'Target' or 'Standard'
     - look at line 252 and 255
     - on line 186 call the correct function
     - change the text between `WCS Match` and `TargetMatch`

jaladh-singhal:
> useCutoutSize to determine is you show the RadioGroupInputField at line 205

`defaultCutoutSizeAS` is already being used to check whether to show input field at L211 - setting it null in spherex app hides it, I tested it

> defaultWcsMatchType which can be 'Target' or 'Standard'
> look at line 252 and 255
> on line 186 call the correct function
> change the text between WCS Match and TargetMatch

This makes sense but this will create 3 states: WCS Match (existing), no Match (existing), and Target Match (new one to add). I don't think changing just text of checkbox will be enough for user to control it. Do you want me to replace it with a dropdown (or radio button group) instead?

robyww:
There are many state for wcs match. The dropdown handles all of this. The checkbox is just setting the type of short cut that the UI is showing.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
jaladh-singhal: I moved it out in a component of its own

robyww:
@jaladh-singhal 

I found another issue this morning. Already loaded images should not be reloading. I am going to have to look at the code more closely to figure out why.

jaladh-singhal:
> @jaladh-singhal
>
> 
> I found another issue this morning. Already loaded images should not be reloading. I am going to have to look at the code more closely to figure out why.
> 
> There are many errors on the console. I think it is probably related to all the image deletion and creation.

I don't see errors in the console but yes I noticed reloading of image viewer in each slide which shouldn't happen. I think it has to do with position of images changing in slides array with arrow click. 

I'm wondering at this point, it might be worth writing our own image slider - visually, it's just css translate and transform. We can then also implement virtualization that react-slick doesn't support.

robyww (approved):
Works great. Looks read nice. Very smooth.

Great Job!

jaladh-singhal:
https://swiperjs.com/demos#virtual-slides - leaving this here, if need to be used in future instead of react-slick

PR #1462, opened 2023-12-12
Firefly-1366 dialogs

robyww:
#### [FIREFLY-1366](https://jira.ipac.caltech.edu/browse/FIREFLY-1366)

  - converted dialogs to be Joy UI based
  - also implemented: [FIREFLY-1368](https://jira.ipac.caltech.edu/browse/FIREFLY-1368)

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1366-dialogs/firefly
- popup a dialog of any time testing, input, etc


loitly (approved):
Nice.  Finally, more modern looking.  How hard is it to remove the blue 'x' top right button?  We normally have the 'Close' or 'Ok' button already.  Plus, it's currently has a 'Delete Image' tooltip for every dialog.

robyww:
I will fix the tooltip. I think we need to keep the `X`, but we could change the icon to another `X`. I just don't have one now. I think the uses like having an `X` in the dialogs.

PR #1464, opened 2023-12-16
FIREFLY-1356: UI Conversion: convert TableView to Joy UI

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1356

Also:
- Add theme to API mode
- Scale down theme's font-size to our preferred size to avoid customizing every component's size.

What is converted…
- table and table toolbar
- table save dialog
- Add a column dialog
    - Including Expression Creator (search icon in Expression field)
- Table Info dialog
- Table Options dialog
    - both Column Options and Advanced Filter

What is not converted..
- TabPanel
- InputAreaField (Advanced Filter Dialog)
- CollapsiblePanel

Issues need resolving…
- Need to establish what is the default non-primary button should be
    - the current default is  too light, it bleeds into the background.
    - I suggest using ‘solid’ for dialog (ok/cancel) and ‘outlined’ for all others
- RadioGroupInputField label’s uses wrong font-family.
- InputField:  why label is one size down?
    - e.g. if size is ‘md’, label is ‘sm’.
- Standardize on..
    - Variant usage
    - Typography level
    - Color usage

To test: https://fireflydev.ipac.caltech.edu/firefly-1356-joyui-table/firefly
- Load a table, see the appearance of the new table.
- Click on every table's toolbar button to ensure they are also modified

Then, test API mode: https://fireflydev.ipac.caltech.edu/firefly-1356-joyui-table/firefly/test/tests-table.html

### DIFF
--- a/src/firefly/js/ui/ThemeSetup.js
+++ b/src/firefly/js/ui/ThemeSetup.js
@@ -1,61 +1,46 @@
 import {extendTheme} from '@mui/joy';
 
+
 export function getTheme() {
     return extendTheme({
         components: {
             JoyButton: {
                 defaultProps: {
-                    size:'sm' ,
                     variant:'soft' ,
                     color:'neutral'
                 }
             },
             JoyInput: {
-                defaultProps: { size:'sm' }
+                styleOverrides: {
+                    root: {
+                        minHeight: '1.75rem',
+                    },
+                },
             },
             JoyIconButton: {
                 defaultProps: {
-                    size:'sm' ,
                     variant:'plain',
                     color:'neutral',
                 }
             },
-            JoyFormControl: {
-                defaultProps: { size:'sm' }
-            },
             JoyFormLabel: {
-                // styleOverrides: {
-                //     root: ({theme}) => ({lineHeight:'50px'})
+                // defaultProps: {
+                //     sx : {
+                //         '--FormLabel-lineHeight' : 1.1
+                //     }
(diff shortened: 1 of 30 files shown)


src/firefly/js/ui/ThemeSetup.js
```diff
@@ -1,61 +1,46 @@
...
+                // defaultProps: {
+                //     sx : {
+                //         '--FormLabel-lineHeight' : 1.1
+                //     }
                 // }
-                defaultProps: {
-                    size:'sm',
-                    sx : {
-                        '--FormLabel-lineHeight' : 1.1
-                    }
```
robyww, line 34: I might need to revisit this one later. The labels don't seem to sit close enough to the input component. However, we can see how the small font plays our first.

robyww (approved):
This all looks good. I do think you should address the table toolbar background in this PR.

UI issues:
  - table toolbar background should not longer be gray. We should use sheet or card and pick up Joy UI color scheme.
  - We need to settle on a toolbar color. Right now I am leading towards just white (Sheet backgound). That is more like the Interactivism recomendation.
  - close in table options should be on left and showing as the primary button.

Not sure but we should seriously consider:
  - make table header look like joy UI table header.
    - We current use a linear-gradient, do we want to keep doing this? 
    - Joy UI look for the header:

Screenshot 2023-12-18 at 9 12 55 AM

PR #1465, opened 2023-12-19
Firefly-1367: toolbar converted to JoyUI

robyww:
#### [Firefly-1367](https://jira.ipac.caltech.edu/browse/FIREFLY-1367): Toolbar converted to JoyUI

- Convert toolbar, `ToolbarButton`, and dropdowns to JoyUI
- This work is _not_ using the JoyUI dropdowns. I converted ours to use JoyUI components. 
   - JoyUI dropdowns are very different than ours. I am not sure JoyUI dropdowns even support all the features.
   - This is something we will have to revisit at a later time.
- Updated badge to use the JoyUI one throughout the application
- Update CloseButton to be just a JoyUI button with a Icon
- More cleanup from the font size changes
- Also implemented, Extraction Dialog to JoyUI: [Firefly-1369](https://jira.ipac.caltech.edu/browse/FIREFLY-1369)

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1367-toolbar/firefly/


loitly (approved):
I personally don't like the yellow outlined as the secondary button.  I also don't like it used as badge.  I think I've seen badge with primary color.  
I would suggest using a scheme like this: https://mui.com/joy-ui/customization/theme-colors/#changing-the-global-variant-tokens
Or, if there's one given by the UI consulting group, we can use that.  Also, we do not have to stick to only 6 colors.  We can introduce one or 2 more if needed.

robyww:
- The button work was to make sure all the non-primary buttons are using the default. I can change it back to soft/neutral.
- I just tried the solid/primary for the badge. it works.
- I am happen to have more colors but I don't want to do be the one that does that (at least now now).

PR #1466, opened 2023-12-21
Firefly 1374 color dialog and other dialogs

robyww:
#### [Firefly-1374](https://jira.ipac.caltech.edu/browse/FIREFLY-1374)

UI Conversion:
- Color/Stretch dialog
- Color Picker dialog
- Point Shape Picker dialog
- Search Refinement Tool dialog
- some continuing clean up from other changes
- toolbar button icons inverse color for dark mode

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1374-color-dialog/firefly
- Try any of the above dialogs

### DIFF
--- a/src/firefly/js/ui/ColorPicker.jsx
+++ b/src/firefly/js/ui/ColorPicker.jsx
@@ -2,6 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
+import {Stack, Tooltip, Typography} from '@mui/joy';
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import {dispatchShowDialog} from '../core/ComponentCntlr.js';
@@ -16,12 +17,16 @@ import {getRGBA} from 'firefly/util/Color.js';
 const DEF_PRESET_COLORS= ['#D0021B', '#F5A623', '#F8E71C', '#8B572A', '#7ED321', '#417505', '#BD10E0',
     '#9013FE', '#4A90E2', '#50E3C2', '#B8E986', '#000000', '#4A4A4A', '#9B9B9B', '#FFFFFF'];
 
-const dialogTip=
-`Choose a color:
-- graphically
-- as an RGB hex color code (does not support alpha)
-- by entering RGB values (0-255) and an alpha in percent (0-100)
-- from the swatches at the bottom`;
+const dialogTip= (
+    <Typography sx={{'ul li': {listStyleType:'circle'}}}>
+        Choose a color
+        <ul>
+            <li>graphically</li>
+            <li>as an RGB hex color code (does not support alpha)</li>
+            <li>by entering RGB values (0-255) and an alpha in percent (0-100)</li>
+            <li>from the swatches at the bottom</li>
+        </ul>
+    </Typography>);
 
 
 export function showColorPickerDialog(color, callbackOnOKOnly, callbackOnBoth, cb, Id = '',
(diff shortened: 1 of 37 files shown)


src/firefly/js/ui/ColorPicker.jsx
```diff
@@ -16,12 +17,16 @@ import {getRGBA} from 'firefly/util/Color.js';
...
     '#9013FE', '#4A90E2', '#50E3C2', '#B8E986', '#000000', '#4A4A4A', '#9B9B9B', '#FFFFFF'];
 
-const dialogTip=
-`Choose a color:
-- graphically
-- as an RGB hex color code (does not support alpha)
-- by entering RGB values (0-255) and an alpha in percent (0-100)
-- from the swatches at the bottom`;
+const dialogTip= (
+    <Typography sx={{'ul li': {listStyleType:'circle'}}}>
```
loitly, line 21: You can get away with only `li` if you want to since it applies to only this block.

src/firefly/js/ui/ColorPicker.jsx
```diff
@@ -63,25 +68,27 @@ function ColorPickerWrapper ({callback,color,callbackOnOKOnly, callbackOnBoth,
 
 
     return (
-        <div>
-            <div title={dialogTip}>
+        <Tooltip title={dialogTip} placement='right'>
```
loitly, line 71: We need to set policy on Tooltip usage.  Is it always, or can `title` be used as well?

src/firefly/js/ui/DropDownMenu.jsx
```diff
@@ -25,7 +25,7 @@ function placeDropDown(e,x,y, beforeVisible) {
 export function SingleColumnMenu({children}) {
     return (
         <DropDownMenu>
-            <Stack direction='column'>
+            <Stack direction='column' sx={{'.ff-toolbar-button':{width:'100%', justifyContent:'flex-start'}}}>
```
loitly, line 28: `width:'100%'` can be replaced with `width:1`.

src/firefly/js/ui/ToolbarButton.jsx
```diff
@@ -48,13 +48,15 @@ function getShortCutInfo(shortcutKey) {
 export const ToolbarButton = memo((props) => {
     const {
         icon,text='',tip='',badgeCount=0,enabled=true, visible=true,
-        imageStyle={}, shortcutKey='', color='neutral',
-        disableHiding, active, sx,
+        imageStyle={}, shortcutKey='', color='neutral', variant='plain',
+        disableHiding, active, sx:inSX,
```
loitly, line 52: This is a ToolbarButton, you would not expect `sx` to apply to `Tooltip` element.  If given, it should apply to the root.  I think slotProps should be used instead.

src/firefly/js/visualize/ui/ColorDialog.jsx
```diff
@@ -96,35 +97,31 @@ function renderThreeColorView(plot, rFields, gFields, bFields, rgbFields, isHueP
...
-                        style={{padding: '2px 0 7px 10px'}}
-                        onSuccess={(request)=>replot3ColorHuePreserving(request)}
-                        onFail={invalidMessage}
-                        text='Refresh'
-                        dialogId='ColorStretchDialog'
+                        closeOnValid={false} sx={{pt:.25, pb:1, pl:1}} text='Refresh' dialogId='ColorStretchDialog'
+                        onFail={invalidMessage} onSuccess={(request)=>replot3ColorHuePreserving(request)}
                     />
-                    <div style={{ textAlign:'right', padding: '2px 10px'}}>
+                    <Box sx={{ textAlign:'right', py:.25, px:1}}>
```
loitly, line 119: I prefer `1/4` over decimal.  Of course, it's just a matter of preference.  
Also, I used fixed number of pixels in places where spacing should not change the appearance.  I don't know if this is the case here.

loitly (approved):
Changes look good.  We should make toggle buttons look different from normal buttons.  I find it's hard to tell them apart.

src/firefly/js/ui/ToolbarButton.jsx
robyww: now that I think about it there is a way to do what I want with out modifying the passed sx.

src/firefly/js/ui/DropDownMenu.jsx
robyww: interesting. Yes I remember that now.

PR #1468, opened 2023-12-22
FIREFLY-1355: Convert TabPanel to Joy UI

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1355
- added scrolling to tabs.
- convert table content overflow drop-down (...)

Test: https://fireflydev.ipac.caltech.edu/firefly-1355-joyui-tabpanel/firefly/
Test wherever tabs are used.  
Keep in mind these features of TabPanel:
- Tab may be removable
- Tab will resize itself as more tabs are added.  It will also resizes as more spaces are mad available

Added Feature:
- A `min-width=100` is set on the tabs.  When there are too many to fit, it will overflow.  They become scrollable at this point.  To scroll, hover over the tab headers, press the Shift key, and then use the scroll wheel.

### DIFF
--- a/src/firefly/js/tables/ui/TablePanel.jsx
+++ b/src/firefly/js/tables/ui/TablePanel.jsx
@@ -3,8 +3,8 @@
  */
 
 import React, {useEffect} from 'react';
-import {Box, Link, Stack, Typography, Sheet} from '@mui/joy';
-import PropTypes from 'prop-types';
+import {Box, Stack, Typography, Sheet, Chip} from '@mui/joy';
+import PropTypes, {object, shape} from 'prop-types';
 import {defer, truncate, get, set} from 'lodash';
 import {getAppOptions, getSearchActions} from '../../core/AppDataCntlr.js';
 import {ActionsDropDownButton, isTableActionsDropVisible} from '../../ui/ActionsDropDownButton.jsx';
@@ -56,8 +56,7 @@ const TT_PROPERTY_SHEET = 'Show details for the selected row';
 
 export const TBL_CLZ_NAME = 'FF-Table';
 
-export function TablePanel(props) {
-    let {tbl_id, tbl_ui_id, tableModel, sx, ...options} = props;
+export function TablePanel({tbl_id, tbl_ui_id, tableModel, variant='outlined', sx, slotProps, ...options}) {
     tbl_id = tbl_id || tableModel?.tbl_id || uniqueTblId();
     tbl_ui_id = tbl_ui_id || `${tbl_id}-ui`;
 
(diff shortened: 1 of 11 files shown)


src/firefly/js/tables/ui/TablePanel.jsx
```diff
@@ -94,16 +93,16 @@ export function TablePanel(props) {
     if ([TBL_STATE.ERROR,TBL_STATE.LOADING].includes(tstate))  return <NotReady {...{showTitle, tbl_id, title, removable, backgroundable, error}}/>;
 
     return (
-        <Box sx={sx} position='relative' width={1} height={1}>
+        <Sheet {...{variant, sx:{boxSizing: 'border-box', position:'relative', width:1, height:1, ...sx}, ...slotProps?.tablePanel}}>
```
robyww, line 96: I see all the `1`'s instread of `100%`!

src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -311,8 +209,146 @@ FieldGroupTabs.propTypes = {
...
+                        '&::before': {
+                            content: '""',
+                            display: 'block',
+                            position: 'absolute',
+                            height: 2,
+                            bottom: -2,
+                            left: 0,
+                            right: 0,
+                            bgcolor: activeBg,
+                        },
```
robyww, line 289: You might want to add some documentation here. I have no idea what this is doing.

src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -311,8 +209,146 @@ FieldGroupTabs.propTypes = {
...
+                    {startDecorator}
+                </ListItemDecorator>
+            )}
+            {label}
+            {removable &&
+                <Chip variant='soft'
+                      onClick={(e) => {
+                          onTabRemove && onTabRemove(name);
+                          e.stopPropagation && e.stopPropagation();
+                }}>x</Chip>
```
robyww, line 331: I just read out Chip. There is a `ChipDelete`, maybe we should start using that.

robyww (approved):
Look good. That application is really starting to come together!

robyww:
There is a padding of `8px` somewhere in the tab component that probably should not be there. Both the charts and the images are affected by it.

robyww:
One more though about this. To take out the padding for the visualizers means our toolbars need to be white.

loitly:
> One more thought about this. To take out the padding for the visualizers means our toolbars need to be white.

Yes.  The padding on the TabPanel is by design.  This is so that the TabContent appears connected to the active tab.  In table, I switch over to white toolbar when added to Tabs.  We should do the same for other components.

PR #1469, opened 2023-12-23
Firefly-1381: JoyUI: converrt small components and clean up

robyww:
#### [FIREFLY-1381](https://jira.ipac.caltech.edu/browse/FIREFLY-1381): JoyUI: clean up and small component

- finished TAP except for AdvancedADQL.jsx
- finished image search panel
- catalog panel
- revamp mouse MouseReadoutBottomLine.jsx
- remove react-select
- finished catalog panel
- working indicator
- added ChipDelete
- fixed image panel: scrolling and checkboxes not wrapping

#### Testing:

- https://fireflydev.ipac.caltech.edu/firefly-1381/firefly
- look at TAP, image panel, catalog panel


loitly (approved):
It's looking good.  Here are a few comments I have.

Image Panel:  
- The double outlined created by stacking the sections on top of each other makes it looks like it's a bug.  I suggest adding 1/4 of spacing between them.

TAP Panel: 
- The top and bottom box-shadow in `.TapSearch__section` looks awkward now that the background is white.  I suggest using a plain outlined Sheet instead.
- I think adding a separator between the 3 collapsible sections(accordion) would look clearer.  Either spacing or a line like the demo in JoyUI.

Catalog panel:
- Should replace the table List box with a JoyUI equivalent.

robyww:
> I think adding a separator between the 3 collapsible sections(accordion) would look clearer. Either spacing or a line like the demo in JoyUI.

I had already played with a line before but just now I played with just spacing. I think that works so I will do that.

robyww:
> The top and bottom box-shadow in .TapSearch__section looks awkward now that the background is white. I suggest using a plain outlined Sheet instead.

I did not use outline but I cleaned it up. I think it looks better.

robyww:
> Catalog panel: Should replace the table List box with a JoyUI equivalent.

I will defer this to another ticket

PR #1470, opened 2023-12-27
FIREFLY-1289: Change to new static list of image zoom levels

gpdf:
After this change, instead of the present irregular system, for seriously zoomed-out images we will continue to use large steps - now consistent factors of 2.  Otherwise, throughout the zoom range we will use four steps per factor of two,
i.e., the 4th root of 2 or about 1.189 / 19% steps, rounded to three digits after the decimal.

This commit is a candidate for back-porting to 2023.3.


robyww:
I am going to close this PR and create a new one that I can put into `rc-2023.3`

PR #1472, opened 2024-01-04
Firefly-1341: Enhanced EmeddedPositionSearchPanel for upload

kpuriIpac:
**[Firefly-1341](https://jira.ipac.caltech.edu/browse/FIREFLY-1341)**: 
**IRSA-IFE PR for Euclid**: https://github.com/IPAC-SW/irsa-ife/pull/303

- `UploadTableChooser` shows up and works with `EmbeddedPositionSearchPanel` (conditionally, because all components using EmbeddedPositionSearchPanel may not want an upload option, like Source Inspector in Euclid or some DCE tables)
- removed column selector for Euclid by passing in `uploadTable` flag (should change this name) to `UploadTableSelector` 
    - the flag enables showing column selector for an uploaded table on TAP, but not for Euclid (or other apps that call `UploadTableSelector` with this flag set to true) 

**Updates 1/31**:
- doing `dispatchTableSearch` on all table uploads and not going to results view directly. Made this work using initial approach suggested by @robyww. Important files to look at: `LayoutCntlr.js`, `HydraManager.js` and `FireflyViewerManager.js`
  - In LayoutCntlr.js, checking if the tbl_group of the uploaded table is main (as it's a different group for things like DataProduct layer, but we don't want to go to result view for those either) 
  - instead of `tbl?.request?.META_INFO?.UploadTable)` I would like to use @robyww's suggested `getBooleanMetaEntry`, but for some reason even when I call it with the correct table/tbl_id and MetaInfo keyword, it returned nothing (I'll try and fix this - only outstanding issue from my end before review feedback). 
- Added some conditional logic inside SearchPanel's `SearchForm` component based on @loitly's recommendation. This seems to work quite well. (for the `useFormPanel` call, look at `euclid.js` in the IFE PR). 
- @robyww the request object in RegionExplorer now has the `uploadInfo` object. (For some reason just `setUploadInfo` on the `uploadInfo` object did not work directly, so in `UploadTableChooser`, I made use of `register` (FieldGroupCtx) to add `uploadInfo` as part of `additionalParams` of the request object (similar to FileUpoadViewPanel). 
- Finally, an unrelated issue: 
  - **overlays** are still not showing up on Euclid when you upload a table (when I tested the uploads with DCE briefly, the overlays showed up as expected). 

**Test Build (updated 1/31 - some issue with the builds - will rebuild all three 2/1 morning):** 
  - Firefly: https://fireflydev.ipac.caltech.edu/firefly-1341-possearchpanelupload/firefly
  - Euclid: https://firefly-1341-possearchpanelupload.irsakudev.ipac.caltech.edu/applications/euclid
  - DCE: https://firefly-1341-possearchpanelupload.irsakudev.ipac.caltech.edu/irsaviewer/dce
- For **Firefly**, go to the TAP panel
    - upload a table form the Multi-Object popup panel (or multiple tables one after another)
    - `Load table` will now just close the popup but not go to the results view 
    - click on `Search` or `Cancel` and you should see this uploaded table in the results view 
    - click on search, and you should see the uploaded tables in the results view 
- For **Euclid**: 
   - In region explorer, click on **Multi-object** and attempt to upload a table (or multiple tables)
   - You should not see a column chooser here, and the position columns' collapsible panel should also be open by default
   - Click **Search** and you should see these upload tables in the results view 
- For **DCE**:
  - Make sure the search on DCE works as expected, and this work on EmbeddedPositionSearchPanel hasn't affected anything there

### DIFF
--- a/src/firefly/js/core/LayoutCntlr.js
+++ b/src/firefly/js/core/LayoutCntlr.js
@@ -8,15 +8,19 @@ import {get, isEqual, isEmpty, filter, pick, uniqBy} from 'lodash';
 import Enum from 'enum';
 import {flux} from './ReduxFlux';
 import {clone} from '../util/WebUtil.js';
-import {smartMerge, getActiveTableId} from '../tables/TableUtil.js';
+import {
+    smartMerge, getActiveTableId, getTblById, findGroupByTblId,
+} from '../tables/TableUtil.js';
 import {getDropDownNames} from '../ui/Menu.jsx';
-import ImagePlotCntlr, {IMAGE_PLOT_KEY, visRoot} from '../visualize/ImagePlotCntlr.js';
-import {TBL_RESULTS_ADDED, TBL_RESULTS_REMOVE, TABLE_REMOVE, TABLE_SPACE_PATH} from '../tables/TablesCntlr.js';
+import ImagePlotCntlr, {visRoot} from '../visualize/ImagePlotCntlr.js';
+import { TBL_RESULTS_ADDED, TBL_RESULTS_REMOVE, TABLE_REMOVE, TABLE_SPACE_PATH, TBL_RESULTS_ACTIVE, TABLE_LOADED
+} from '../tables/TablesCntlr.js';
 import {CHART_ADD, CHART_REMOVE, CHART_SPACE_PATH} from '../charts/ChartsCntlr.js';
 import {REPLACE_VIEWER_ITEMS} from '../visualize/MultiViewCntlr.js';
 import {REINIT_APP} from './AppDataCntlr.js';
 import {getDefaultChartProps} from '../charts/ChartUtil.js';
 import {getPlotViewAry} from 'firefly/visualize/PlotViewUtil.js';
+import {MetaConst} from 'firefly/data/MetaConst';
 
 export const LAYOUT_PATH = 'layout';
 
@@ -314,12 +318,26 @@ export function dropDownHandler(layoutInfo, action) {
     switch (action.type) {
         case CHART_ADD:
         case TBL_RESULTS_ADDED:
+        case TBL_RESULTS_ACTIVE:
+        case TABLE_LOADED:
+            const tbl_id= action.type === CHART_ADD ? action.payload.groupId : action.payload.tbl_id;
+            if (findGroupByTblId(tbl_id)!=='main' || getTblById(tbl_id)?.request?.META_INFO?.[MetaConst.UPLOAD_TABLE]) {
+                return layoutInfo;
+            }
+            return smartMerge(layoutInfo, {dropDown: {visible: false}});
         case REPLACE_VIEWER_ITEMS :
         case ImagePlotCntlr.PLOT_IMAGE :
             return smartMerge(layoutInfo, {dropDown: {visible: false}});
         case ImagePlotCntlr.PLOT_IMAGE_START :
-            const {useForSearchResults= true}= action.payload.pvOptions;
-            const visible= !useForSearchResults && layoutInfo.dropDown.visible;
+            const VISUALIZED_TABLE_IDS = action.payload.attributes.VISUALIZED_TABLE_IDS;
+            if (VISUALIZED_TABLE_IDS && VISUALIZED_TABLE_IDS.length > 0) {
+                const lastId = VISUALIZED_TABLE_IDS[VISUALIZED_TABLE_IDS.length - 1];
+                // Check if the last entry contains 'Upload_Tbl' - and return layoutInfo as is if it does
+                const containsUploadTbl = lastId.includes('Upload_Tbl');
+                if (containsUploadTbl) return layoutInfo;
+            }
+            const {useForSearchResults= true, useForCoverage}= action.payload.pvOptions;
+            const visible= (!useForCoverage && !useForSearchResults) && layoutInfo.dropDown.visible;
             return smartMerge(layoutInfo, {dropDown: {visible}});
         case CHART_REMOVE:
         case SHOW_DROPDOWN:
--- a/src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
+++ b/src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
@@ -7,12 +7,13 @@ import {filter} from 'lodash';
 import {getAppOptions} from '../../core/AppDataCntlr.js';
 
 import {LO_VIEW, SHOW_DROPDOWN, SET_LAYOUT_MODE, getLayouInfo, dispatchUpdateLayoutInfo, dropDownManager} from '../../core/LayoutCntlr.js';
-import {smartMerge} from '../../tables/TableUtil.js';
+import {getTblById, smartMerge} from '../../tables/TableUtil.js';
 import {TBL_RESULTS_ADDED, TABLE_LOADED, TABLE_REMOVE, TBL_RESULTS_ACTIVE, TBL_RESULTS_REMOVE} from '../../tables/TablesCntlr.js';
 import {CHART_ADD, CHART_REMOVE} from '../../charts/ChartsCntlr.js';
 
 import ImagePlotCntlr from '../../visualize/ImagePlotCntlr.js';
 import {REPLACE_VIEWER_ITEMS} from '../../visualize/MultiViewCntlr.js';
+import {MetaConst} from 'firefly/data/MetaConst';
 
 /**
  * this manager manages what main components get display on the screen.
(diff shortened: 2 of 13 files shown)


src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -64,7 +69,19 @@ function uploadSubmit(request,setUploadInfo,defaultColsEnabled)  {
...
-    const uploadInfo = {serverFile, fileName, columns:columnsSelected, totalRows, fileSize, tableSource: UPLOAD_TBL_SOURCE};
+    const uploadInfo = {serverFile, fileName, columns:columnsSelected, totalRows, fileSize, tableSource: UPLOAD_TBL_SOURCE, tblCount};
+    //Todo: uploadInfo has tblCount ... but make fileName = fileName + tblCount?
+    tblCount++;
+
+    const META_INFO= {
+        [MetaConst.CATALOG_OVERLAY_TYPE]:'TRUE'
+    };
+    const options=  {META_INFO};
+    const tblReq = makeFileRequest(fileName, serverFile, null, options);
```
kpuriIpac, line 80: @robyww I'm adding the `META_INFO` for `CATALOG_OVERLAY_TYPE` here.

src/firefly/js/ui/dynamic/DynComponents.jsx
```diff
@@ -177,6 +177,7 @@ export function PositionAndPolyFieldEmbed({fieldDefAry, plotId, toolbarHelpId, i
                                        otherComponents, WrapperComponent}) {
 
     const polyType = findFieldDefType(fieldDefAry, POLYGON);
+    //const uploadType = findFieldDefType(fieldDefAry, UPLOAD);
     const posType = findFieldDefType(fieldDefAry, POINT) ?? findFieldDefType(fieldDefAry, POSITION);
```
kpuriIpac, line 181: @robyww following up on my slack question, this is what I meant. I don't think we'll get any upload info from the `fieldDefAry`, right? so is it safe to just set `useUpload = true` (see line 204) before the call to `EmbeddedPositionSearchPanel`?  that'll always give an option to upload a table in all of DCE...

src/firefly/js/templates/fireflyviewer/FireflyViewerManager.js
```diff
@@ -89,6 +89,12 @@ function onAnyAction(layoutInfo, action, views) {
         case CHART_ADD:
         case CHART_REMOVE:
         {
+            if (action.type === TBL_RESULTS_ADDED || action.type ===  TABLE_LOADED) {
+                const tbl = getTblById(action.payload.tbl_id);
+                if (tbl?.request?.META_INFO?.UploadTable) { //tbl?.tableMeta?.UploadTable
+                    return layoutInfo;
+                }
+            }
```
robyww, line 97: This is going to have to be in two or three of the templates, particularly hydra template

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -316,6 +317,14 @@ export function dropDownHandler(layoutInfo, action) {
         case TBL_RESULTS_ADDED:
         case REPLACE_VIEWER_ITEMS :
         case ImagePlotCntlr.PLOT_IMAGE :
+            if (action.type === TBL_RESULTS_ADDED || action.type ===  TABLE_LOADED
+                || action.type === TBL_RESULTS_ACTIVE || action.type === CHART_ADD) {
+                //TODO: Review: is it okay to get tblId from gropupId when action.type is CHART_ADD?
+                const tbl = action.type === CHART_ADD ? getTblById(action.payload.groupId) : getTblById(action.payload.tbl_id);
+                if (tbl?.request?.META_INFO?.UploadTable) {
+                    return layoutInfo;
+                }
```
robyww, line 326: this code is not quite right. I think you should clean it up. You have an `if` inside a `case` which mostly should not happen. it should be another case. Also, some of the if `action.type === TBL_RESULTS_ACTIVE` and `action.type ===  TABLE_LOADED` can never be true because they are not part of the case.

src/firefly/js/ui/UploadTableSelector.jsx
```diff
@@ -0,0 +1,172 @@
+import PropTypes from 'prop-types';
+import {FieldGroupCtx} from 'firefly/ui/FieldGroup';
```
robyww, line 2: remember this file will have to be ported to JoyUI based on what was already done with `UploadTableSelector`

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -316,6 +317,14 @@ export function dropDownHandler(layoutInfo, action) {
         case TBL_RESULTS_ADDED:
         case REPLACE_VIEWER_ITEMS :
         case ImagePlotCntlr.PLOT_IMAGE :
+            if (action.type === TBL_RESULTS_ADDED || action.type ===  TABLE_LOADED
+                || action.type === TBL_RESULTS_ACTIVE || action.type === CHART_ADD) {
+                //TODO: Review: is it okay to get tblId from gropupId when action.type is CHART_ADD?
+                const tbl = action.type === CHART_ADD ? getTblById(action.payload.groupId) : getTblById(action.payload.tbl_id);
+                if (tbl?.request?.META_INFO?.UploadTable) {
```
robyww, line 324: use TableUtil.js, `getBooleanMetaEntry`

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -1,16 +1,25 @@
...
+    CONE_AREA_OPTIONS,
+    CONE_AREA_OPTIONS_UPLOAD,
+    CONE_CHOICE_KEY,
+    POLY_CHOICE_KEY,
+    UPLOAD_CHOICE_KEY
+} from '../../visualize/ui/CommonUIKeys.js';
 import {HiPSTargetView} from '../../visualize/ui/TargetHiPSPanel.jsx';
 import {RadioGroupInputField} from '../RadioGroupInputField.jsx';
-import {useFieldGroupValue} from '../SimpleComponent.jsx';
+import {useFieldGroupValue, useFieldGroupWatch} from '../SimpleComponent.jsx';
```
robyww, line 13: `useFieldGroupWatch` is not used

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -70,28 +82,44 @@ export function EmbeddedPositionSearchPanel({
                                                 insetSpacial=true,
                                                 otherComponents= undefined,
                                                 usePosition= true,
+                                                useUpload = false,
                                                 usePolygon= true,
+                                                uploadTblReqs,
```
robyww, line 87: is `uploadTblReqs` this still necessary since the tables are already loaded?

robyww (commented):
This is the main comment. The other stuff is clean up

I don’t think you need `const [uploadTblReqs, setUploadTblReqs] = useState([]);` at all. Not passed to EmbeddedPositionSearchPanel.

The request will have an uploadInfo I think that is all you need.  you will need to modify upload into to include tbl_id

src/firefly/js/ui/UploadTableSelector.jsx
kpuriIpac: got it. this will be the main work then I suppose in the rebase after I merge this PR?

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -314,10 +316,28 @@ export function dropDownHandler(layoutInfo, action) {
...
         case ImagePlotCntlr.PLOT_IMAGE_START :
+            const VISUALIZED_TABLE_IDS = action.payload.attributes.VISUALIZED_TABLE_IDS;
+            if (VISUALIZED_TABLE_IDS && VISUALIZED_TABLE_IDS.length > 0) {
+                const lastId = VISUALIZED_TABLE_IDS[VISUALIZED_TABLE_IDS.length - 1];
+                // Check if the last entry contains 'Upload_Tbl' - and return layoutInfo as is if it does
+                const containsUploadTbl = lastId.includes('Upload_Tbl');
+                if (containsUploadTbl) return layoutInfo;
+            }
             const {useForSearchResults= true}= action.payload.pvOptions;
             const visible= !useForSearchResults && layoutInfo.dropDown.visible;
```
robyww, line 342: This will not work in all cases but, these two lines are a slight modification of the existing code and fix an existing problem.
```js
const {useForSearchResults= true, useForCoverage}= action.payload.pvOptions;
const visible= (!useForCoverage && !useForSearchResults) && layoutInfo.dropDown.visible;
```

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -314,10 +316,28 @@ export function dropDownHandler(layoutInfo, action) {
...
+        case TABLE_LOADED:
+            const tbl = action.type === CHART_ADD ? getTblById(action.payload.groupId) : getTblById(action.payload.tbl_id);
+            if (action.payload.options?.tbl_group !== 'main') {
+                return layoutInfo; //do not load results view if tbl group is not main
+            }
+            //Todo: const isUploadTbl= getBooleanMetaEntry(tbl, MetaConst.UPLOAD_TABLE, false);
+            if (tbl?.request?.META_INFO?.UploadTable) {
+                return layoutInfo;
+            }
+            return smartMerge(layoutInfo, {dropDown: {visible: false}});
```
robyww, line 329: I think this is cleaner
```js
const tbl_id= action.type === CHART_ADD ? action.payload.groupId : action.payload.tbl_id;
if (findGroupByTblId(tbl_id)!=='main' || getTblById(tbl_id)?.request?.META_INFO?.[MetaConst.UPLOAD_TABLE]) {
     return layoutInfo;
}
```

robyww (approved):
OK, do this clean up and I think it is ready to merge.

PR #1473, opened 2024-01-08
FIREFLY-1379: UI Conversion: Table cleanup

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1379

- support dark mode
I can add a way to switch between dark/light in the 'core changes' ticket.  But, I can't find it.

- use Joy 'warning' for highlighted row
What was used for images is too dark.  There's not enough contrast for black text.  I am using `warning-200` instead.  

- fix tooltip formatting
Because tooltip prop is expecting a string, I am adding a `whiteSpace:'pre'` into `InputFieldView.jsx` so that it will respect line breaks.  Let me know what you think.

- fix table not shrinking

Test: https://fireflydev.ipac.caltech.edu/firefly-1379-table-cleanup/firefly


robyww:
The table shrinking issue is there there.
Two way to see it
- load a FITS, go to fits header dialog, stretch it out then shrink it donw.
- go to tap, make browser larger then smaller and you will see it with the column constraints.

PR #1475, opened 2024-01-09
FIREFLY-1387: UI Conversion: Additional core functionalities

loitly:
- Provide a way for application to override Firefly's default theme.
- Package google font as part of Firefly to avoid external linking of google font.
- Use device's colorMode as default.  Provide option to set colorMode as well as document how to enable mode switching in the future.
Read comments in FireflyRoot.jsx for details.  Please provide feedbacks if you see issues.

- Use CssBaseline
I enabled `ScopedCssBaseline`.  See https://mui.com/joy-ui/react-css-baseline/ for more details.

This is a good idea going forward.  However, because it changes box-sizing model, it may breaks places where we use offset to fix alignment where we should have set box-sizing to 'border-box'.  I saw and fixed a couple.  I don't know how many more places this may affect.  I am ok reverting this change.  But, if there's not many places, we should keep this.

Test: https://fireflydev.ipac.caltech.edu/firefly-1387-colormode/firefly/

Everything should work the same.  You will noticed a slight smoothing of the text applied by CssBaseline.
You can also test switching between dark/light using your device's mode.  
```
    // Firefly is setup to take the device's color mode
    // To set device mode on your mac; AppleMenu -> System Settings -> Appearance
```

Also, be aware when you use `sx` to apply styles using selectors.
```
<Sheet sx = {{'& .PagingBar__button' ...   // selects all elements with this class from THIS node down.

// without the '&', it is global to Sheet
<Sheet sx = {{'.PagingBar__button' ...   // selects all elements with this class from Sheet down.
```

I've already seen a few leaked styles.

### DIFF
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -4,7 +4,6 @@
  */
 
 import 'isomorphic-fetch';
-import '@fontsource/inter';
 import React from 'react';
 import {createRoot} from 'react-dom/client';
 import {set, defer, once} from 'lodash';
@@ -131,6 +130,10 @@ const defFireflyOptions = {
         {id: 'vocat'},
         {id: 'nedcat'}
     ],
+    theme: {
+        customized: undefined,          // a function that returns a customized theme
+        colorMode: undefined,           // can be 'dark' or 'light'.  When not specified(default), it will use device's settings.
+    },
     MenuItemKeys: {},
     imageTabs: undefined,
     irsaCatalogFilter: undefined,
--- a/src/firefly/js/tables/ui/BasicTableView.jsx
+++ b/src/firefly/js/tables/ui/BasicTableView.jsx
@@ -4,7 +4,6 @@
 
 import React, {useCallback, useEffect} from 'react';
 import {Box, Typography} from '@mui/joy';
-import { useColorScheme } from '@mui/joy/styles';
 import PropTypes from 'prop-types';
 import {Column, Table} from 'fixed-data-table-2';
 import {wrapResizer} from '../../ui/SizeMeConfig.js';
@@ -19,6 +18,7 @@ import {CellWrapper, HeaderCell, makeDefaultRenderer, SelectableCell, Selectable
 import {useStoreConnector} from '../../ui/SimpleComponent.jsx';
 import {dispatchTableUiUpdate, TBL_UI_UPDATE} from '../TablesCntlr.js';
 import {Logger} from '../../util/Logger.js';
+import { useColorMode } from '../../ui/FireflyRoot.jsx';
 
 import 'fixed-data-table-2/dist/fixed-data-table.css';
 import './TablePanel.css';
@@ -32,7 +32,7 @@ export const BY_SCROLL = 'byScroll';
 
 const BasicTableViewInternal = React.memo((props) => {
 
-    const { mode } = useColorScheme();
+    const { activeMode } = useColorMode();
 
     const {width, height} = props.size;
     const {columns, data, hlRowIdx, renderers, bgColor, selectInfoCls, callbacks, rowHeight, rowHeightGetter, showHeader=true,
@@ -149,7 +149,7 @@ const BasicTableViewInternal = React.memo((props) => {
         else return null;
     };
 
-    const hlColor = mode === 'dark' ? 'background.surface' : undefined;
+    const hlColor = activeMode === 'dark' ? 'background.surface' : undefined;
 
     return (
         <Box tabIndex='-1' onKeyDown={onKeyDown}
--- a/src/firefly/js/tables/ui/TablePanel.jsx
+++ b/src/firefly/js/tables/ui/TablePanel.jsx
@@ -37,10 +37,16 @@ import {AddColumnBtn} from './AddOrUpdateColumn.jsx';
 
 import FILTER from 'html/images/icons-2014/24x24_Filter.png';
 import OUTLINE_EXPAND from 'html/images/icons-2014/24x24_ExpandArrowsWhiteOutline.png';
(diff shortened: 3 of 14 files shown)


src/firefly/js/Firefly.js
```diff
@@ -131,6 +131,10 @@ const defFireflyOptions = {
         {id: 'vocat'},
         {id: 'nedcat'}
     ],
+    theme: {
+        customized: undefined,          // a function that returns a customized theme
+        colorMode: undefined,           // dark/light mode.  Default to device's mode.
```
robyww, line 136: You should add a couple of commented out lines for the next few weeks. That way it is quick to change mode.

```js
        // colorMode: () => 'dark',           // dark mode
        // colorMode: () => 'light',          // light mode
```

src/firefly/js/ui/FireflyRoot.jsx
```diff
@@ -1,31 +1,68 @@
...
 
+
+/**
+ * A wrapper around useColorScheme to return the current active mode, which should be either light or dark
+ * @return {object}
+ */
+export function useColorMode() {
+    const { mode, systemMode, ...rest } = useColorScheme();
+    const activeMode = mode === 'system' ? systemMode : mode;
+    return {activeMode, mode, systemMode, ...rest};
```
robyww, line 66: I feel like this is incomplete. I don't think system mode works (I tested with my mac). Even it is did it does not have the event listeners to change it when system mode is set to auto or changes.

some googling show we can do like this.
```
 window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', () => /*something*/)
```
I think it is worth looking into `window.matchMedia` then `addEventListener`

src/firefly/js/ui/FireflyRoot.jsx
```diff
@@ -1,31 +1,68 @@
...
-
-
+import React, {useEffect} from 'react';
+import {ScopedCssBaseline, extendTheme, CssVarsProvider, useColorScheme, GlobalStyles} from '@mui/joy';
 
+import {defaultTheme} from './ThemeSetup.js';
+import {getAppOptions} from '../core/AppDataCntlr.js';
+import {logger} from '../util/Logger.js';
 
+import '@fontsource/inter/200.css'; // Lighter
```
robyww, line 8: So we don't have to have this in package.json? it just pulls it in?

robyww (changes requested):
The structure all looks good. However I see two issues.
 - This fonts are not quite right. They are a little lighter. It is not clear to me when this is. Maybe it is missing an import?
       *UPDATE*- I mentioned this below, this come from `CssBaseline`, so we should just keep the new look.
 - All the dark/light infrastructure does not seem as fully connected as it needs to be. Ultimately we should have three modes `dark`, `light` and `auto`. `auto` should take the system settings and change with the system settings.

src/firefly/js/tables/ui/TablePanel.jsx
```diff
@@ -295,10 +295,10 @@ function ToolBar({tbl_id, tbl_ui_id, connector, tblState, slotProps}) {
 
     if (!showToolbar) return null;
 
-    const sx = mode === 'dark' ?                                            // temporary solution for dark-mode; need real icons
+    const sx = activeMode === 'dark' ?
                 ({'& .PagingBar__button, .PanelToolbar__button': {
-                    backgroundColor: 'text.primary'
-                }}) : undefined;
+                    filter : 'invert(1)'}
+                }) : undefined;
```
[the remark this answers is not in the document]
loitly: Existing code prior to ToolbarButton. Let me try to convert.

src/firefly/js/tables/ui/BasicTableView.jsx
```diff
@@ -149,7 +149,7 @@ const BasicTableViewInternal = React.memo((props) => {
         else return null;
     };
 
-    const hlColor = mode === 'dark' ? 'background.surface' : undefined;
+    const hlColor = activeMode === 'dark' ? 'background.surface' : undefined;
```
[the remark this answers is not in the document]
loitly: Under light mode, the text color is black while surface is white.  When switched over to dark, the text color is white and the surface is black.  But, for highlighted row, we do not switch the row color.  So, there's not enough contrast to read white text over light beige background.  That's why I switch to black color in dark mode.

robyww:
Seems like the font thing come from `CssBaseline` so I think it is fine. So we should just go with what you have.

src/firefly/js/Firefly.js
loitly: colorMode is a string 'dark' or 'light'.  I will clarify in the existing comment.

src/firefly/js/ui/FireflyRoot.jsx
loitly: This works fine for me.  I can demonstrate it for you.  The culprit is `setMode` used in existing code(fixed in my PR).  I tried to explained this in the comments.  `setMode` uses browser local storage.  Once set, it will ALWAYS take that value regardless of what your device setting is.

This is the intended design for our color mode.
```
    const colorMode = getAppOptions().theme?.colorMode?.();
    // if colorMode is set; must use this mode
    // To add mode switching; make sure colorMode is not set.
    // Then, add a component(Button) to trigger 'setMode'.
    // setMode store the value in localStorage.  It will use this mode
    // until that value is cleared or kill/restart browser.
```

src/firefly/js/ui/FireflyRoot.jsx
loitly: After looking into this, I realized that you already included the npm package `@fontsource/inter`.  All you need is to import the different weight css files to use it.

robyww: So are you going to change package.json? or is it fine?

robyww (approved):

PR #1480, opened 2024-01-18
Firefly 1392: Convert BackgroundMonitor and a few other components to JoyUI

kpuriIpac:
**[Firefly-1392](https://jira.ipac.caltech.edu/browse/FIREFLY-1392)**: 

- Changed the "Hide" back to a button as suggested, and make the 'X' next to the results smaller 
- If `ProgressBar.jsx` is not being used anywhere else (I don't think it's being used in ife either), then it may be safe to remove it
- Background Monitor conversion complete, also converted `Prepare Download` and `Data Help` buttons when you do a DCE search (and touched a few other relevant files like WorkspaceSelectPane.jsx as well). 
  - _Note_: I don't think the background monitor email notification is working. (I think this is a known issue? so is it okay to ignore as scope of this ticket?)  

**Updates 1/25:** 
- Addressed the comments and feedback @robyww. This is done or mostly done I think, added a couple of comments for you in the code. 
- Got rid of border altogether for the results. I think it looks clean, but let me know if you think we should still use a `Divider` (I think it may make it a bit messy personally? but I can give it a shot or we can revisit after feedback from scientists later). 
- **Testing**: Test firefly as well as DCE (`Prepare Download` and `Data Help` buttons), and also the background monitor when you send stuff to it from `Prepare Download `dialog. 
- Updated the ui-notes as well. 

**Testing**:  
   - Firefly: https://fireflydev.ipac.caltech.edu/firefly-1392-ui-bg-monitor/firefly
   - IRSAviewer (DCE): https://firefly-1392-ui-bg-monitor.irsakudev.ipac.caltech.edu/irsaviewer/dce
      - Provide any feedback on the UI for `Background Monitor`, `prepare download` dialog and the Background Monitor launched from the `Prepare Download` dialog (has some different text/layout).
- Do a big search on firefly and click `send to background` to test out the new `LinearProgress` bar by JoyUI (and abort the job once too, make sure it looks ok) 
-  Clicking on `Enable email notification` checkbox moves the elements up (existing behavior), tried to keep the 3 elements (hide, checkbox and help icon) linear / on the same level 
- For all of the results, got rid of the colored background, but still kept the border (can remove this)
- For DCE: 
  - Do a search, then notice the `Prepare Download` and `Data Help` buttons on the table (top left). These have been individually converted + added some space between the two.  
  - Click on `Prepare Download` after selecting a few rows. Make sure this dialog looks okay too. Click on save to `Workspace` here to test out the warning message (converted this to a Typography warning, but still kept the border around the message - should I get rid of that?)
     - if `showWarnings` is true (in `DownloadDialog.jsx`), you get another warning (~line 212), I converted this to a Typography and tested locally as well (no borders around this warning, so for consistency I can get rid of it from the workspace warning as well). 
   - Finally, click `Prepare Download `again and send to background, then take a look at the background monitor here as well to ensure it looks okay. 
   - One thing to note is that the new Linear Progress bar looks quite faint here on DCE as the progressDesc text is initially empty for a few seconds depending on the search empty (versus for a regular table search sent to background it reads "UWS Job Submitted..." instantly, making it more noticeable.


robyww:
> The ChipDelete (X) next to the jobs looks too big, tried to make it smaller but that didn't work. Will get back to this, but let me know if there are any suggestions.

```js
 <ChipDelete {...{ onClick: () => doTheDelete(), sx:{'--Chip-deleteSize': '1.9em'} }}/>
```

src/firefly/js/core/background/BackgroundMonitor.jsx
```diff
@@ -166,25 +177,27 @@ function JobHeader({jobInfo}) {
...
+        <Stack alignItems={'center'} flexDirection={'row'}>
+            <Stack direction='row' alignItems='center' sx={{ overflow: 'hidden' }} title={label}>
+                <Typography level='title-md' alignItems={'center'} sx={{fontWeight: 'bold', ml: 0.375, whiteSpace: 'nowrap', overflow: 'hidden',
+                    textOverflow: 'ellipsis'}}>{label}
+                </Typography>
                 <img src={INFO} onClick={showInfo} className='JobInfo__items--link'/>
-            </div>
-            <div style={{display: 'inline-flex', alignItems: 'center', paddingLeft: 5, flexShrink: 0}}>
+            </Stack>
+            <Stack direction='row' alignItems='center' sx={{ pl: 1, flexShrink: 0 }}>
```
robyww, line 189: add `spacing={1}`

robyww (commented):
UI comments

- The hide button (if it is there) should be a primary button not  a chip
- I agree the ChipDelete should be smaller
- There should not be a background color on the results
-

robyww (changes requested):
I made a bunch a little comment that fall into following categories.
- We want to use border as little as possible. If we have to there is a technique we can talk about.
- In general I have found using `Divider` is a better option
- we don't want to set color directly we want to use JoyUI color such as 'primary', 'neutral' and 'warning' etc
- sx mostly is not necessary for `Box` and `Stack` (maybe when you are passing a class based style, I am not sure)
- I think ToolbarButton is a better option for those table buttons you are changing.
- we are never using bold and italic directly. If there use should click use `Link` or `Chip`
- Some of this stuff I think @loitly and I have been doing intuitively. You are welcome to update [ui-notes.md](https://github.com/Caltech-IPAC/firefly/blob/UI-conversion/docs/ui-notes.md) with anything that was not clear

kpuriIpac:
> I made a bunch a little comment that fall into following categories.
> 
> * We want to use border as little as possible. If we have to there is a technique we can talk about.
> * In general I have found using `Divider` is a better option
> * we don't want to set color directly we want to use JoyUI color such as 'primary', 'neutral' and 'warning' etc
> * sx mostly is not necessary for `Box` and `Stack` (maybe when you are passing a class based style, I am not sure)
> * I think ToolbarButton is a better option for those table buttons you are changing.
> * we are never using bold and italic directly. If there use should click use `Link` or `Chip`
> * Some of this stuff I think @loitly and I have been doing intuitively. You are welcome to update [ui-notes.md](https://github.com/Caltech-IPAC/firefly/blob/UI-conversion/docs/ui-notes.md) with anything that was not clear

Thanks @robyww. Working on these and yes I'll add some of this to [ui-notes.md] as well.

src/firefly/js/core/background/BackgroundMonitor.jsx
```diff
@@ -58,8 +60,8 @@ export function showBackgroundMonitor(show=true) {
 
     return (
         <Stack justifyContent='space-between' spacing={1} alignItems='left' minHeight={200}
-               minWidth={330} maxWidth={590} sx={{resize: 'both', overflow: 'auto'}}>
-            <Box sx={{ flexGrow: 1, overflow: 'auto' }}>
+                    minWidth={330} maxWidth={590} sx={{resize: 'both', overflow: 'auto'}}>
```
kpuriIpac: @robyww I made the other changes. I left this sx here with this Stack because merging it as <Stack {...{ _properties here_ }}> makes the reisze not work for the Background monitor popup (I am not sure why though).

src/firefly/js/core/background/BackgroundMonitor.jsx
```diff
@@ -264,23 +256,19 @@ function PackageItem({SINGLE, jobId, index}) {
 
 
     const action = dlreq?.WS_DEST_PATH
-        ? (<Typography justifyContent={'flex-end'} width={175} height={20} sx={{color:'dimgray', mx:'0.5', my:'auto', display: 'inline-flex',
-            whiteSpace: 'nowrap',
-        }}>
+        ? (<Link>
```
kpuriIpac: and @robyww in my cleanup I noticed there was never an actual link for `Download to Workspace` (so if `dlreq?.WS_DEST_PATH` exists, then you just see the text for `Download to Workspace`). Left this as is for now.

robyww (approved):
Looks good. Go ahead and merge.

PR #1443, opened 2023-11-01
FIREFLY-1304: Implement spectral redshift correction

jaladh-singhal:
Implements [FIREFLY-1304](https://jira.ipac.caltech.edu/browse/FIREFLY-1304)

On a high-level:
- Made spectral frame data in the table available to the chart data as `fireflyData`
- Added UI for spectral frame and redshift input in Spectrum options
  -  Also updated it based on early feedback from scientists
- Made changes in spectral frame UI inputs trigger changes in X ColumnOrExpression and X Label - aka redshift correction
  - Also apply those changes to other traces of the chart
- [x] Combine expressions for unit conversion and redshift correction, also fix label combination
- [x] Make it work for CUSTOM or other ref pos too
- [x] Chart should be initialized based on initial state of spectral frame options

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1304-redshift-correction/firefly/

Created test files for each of the following 3 case available at https://github.com/jaladh-singhal/z-correction-data/

1. TOPOCENTER - Download [Spitzer_IRS_SL_PBCD-topocenter.vot](https://github.com/jaladh-singhal/z-correction-data/blob/main/Spitzer_IRS_SL_PBCD-topocenter.vot) and upload it to firefly. Open chart options, change spectral frame to rest frame and then in resulting options, choose user-specifed one. Enter 0.5, and apply, check that x axis data and label changed in the chart for all traces.

   Also note that the rest frame options have values that are present in table metadata (retrievable via clicking info button in table). And changing value in dropdown (rest/observed) preserves selection in rest frame radio group options as well as input field.

   Stress test it for other edge cases that can occur due to selection in other fields in spectrum options.

2. CUSTOM - Download [Spitzer_IRS_SL_PBCD-custom.vot](https://github.com/jaladh-singhal/z-correction-data/blob/main/Spitzer_IRS_SL_PBCD-custom.vot) and upload it to firefly. Open chart options, spectral frame should have a custom redshift option selected by default. Change options and play around like in TOPOCENTER case. Also note that formula has multiplication factor in this case.

3. Other - Download [Spitzer_IRS_SL_PBCD-barycenter.vot](https://github.com/jaladh-singhal/z-correction-data/blob/main/Spitzer_IRS_SL_PBCD-barycenter.vot) and upload it to firefly. Open chart options, spectral frame shouldn't be editable as per design and x-axis label in the chart should show the spectral frame.

PR #1485, opened 2024-01-26
FIREFLY-1386: Convert chart options panel to joy UI

jaladh-singhal:
Fixes [FIREFLY-1386](https://jira.ipac.caltech.edu/browse/FIREFLY-1386)

- [x] Converted:
  - [x] ScatterOptions 
  - [x] BasicOptions
  - [x] SpectrumOptions
  - [x] FireflyHistogramOptions, HistogramOptions (converted to functional component too)
  - [x] HeatmapOptions
  - [x] Errors
  - [x] NewTracePanel
  - [x] ChartSelectPanel (the topmost component)
  - [x] ❓ PlotlyHistogramOptions - it's not being used anywhere, do we need it?
- [x] Fixed some issues with input components:
  - ColumnOrExpression, SuggestBoxInputField - exposed slotProps and sx
  - CheckboxGroupInputField - FormControl was being applied only to label, but it needs to be applied to both label and input group for proper spacing & layout when used in horizontal context
  - ListBoxInputField - fixed event handler throwing errors due to ev being null in some cases
- [x] Cleanup `fieldProps` in all Options as they are no longer being used
- [ ] Use type=switch for checkbox in Error and heatmap colorscale (simply turning it on in props is throwing errors)
- [ ] Fix why Colorscale list input field shows "default" on first render of chart options in firefly test-charts heatmap example
- [ ] Make search icon in ColumnOrExpression, more integrated like in Target Panel
- [ ] Redesign form layout to show input fields with group relationship more effectively (esp in Spectrum Options' spectral and flux axis, Histogram Options' algorithm) and adjust group based spacing which isn't perfect yet
- [ ] ❓ make input field widths same - "add column" panel in table does that, but it's tricky to do so in char options because of more complex design
- [ ] make it possible to set a React node as RadioGroupInputField label - currently, we can only set string-only label and we are using layout-workarounds for usecases when radio input needs a text field which are difficult to edit/maintain.

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1386-chart-options/firefly/

Try chart options in all the examples present at /test/tests-chart.html
- mostly focus on 4 different types: scatter, heatmap, histogram, spectrum
- try modifying and adding new chart
- spectrum example there isn't complex enough - refer to #1443 for testing instructions

Since there are so many permutations, it might be also helpful to compare these examples with old options panel at https://fireflydev.ipac.caltech.edu/ui-conversion/firefly/test/tests-chart.html


src/firefly/js/ui/FormPanel.jsx
```diff
@@ -80,7 +81,7 @@ export const FormPanel = function (props) {
     }, []);
 
     return (
-        <Sheet className='ff-FormPanel' style={style}>
+        <Sheet className='ff-FormPanel' style={style} sx={sx}>
```
robyww, line 84: glad you added `sx` here

src/firefly/js/ui/SuggestBoxInputField.jsx
```diff
@@ -260,7 +260,7 @@ export class SuggestBoxInputFieldView extends PureComponent {
     render() {
 
         const {displayValue, valid, message, highlightedIdx, isOpen, inputWidth, suggestions, mouseTrigger } = this.state;
-        const {label, tooltip, inline, renderSuggestion, wrapperStyle, placeholder, slotProps={},
+        const {label, tooltip, inline, renderSuggestion, wrapperStyle, placeholder, slotProps={}, sx,
```
robyww, line 263: at some point we should remove wrapperStyle. I have been taking it out of other components.

robyww (approved):
@jaladh-singhal This PR is looking pretty good. It is big but I tried to look over each file. I think you might need to break out the remaining task into a second ticket. I am going to go ahead and approve. What don't you do any small things you can do quickly and make a second ticket for the rest of the todos.

robyww:
> PlotlyHistogramOptions - it's not being used anywhere, do we need it?
remove it.

PR #1486, opened 2024-01-26
FIREFLY-1394: Convert Chart and related components to JoyUI.

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1394

Convert ChartPanel and related components to JoyUI.  Most noticeable is the Toolbar.
Also made Plotly takes JoyUI surface color when drawing charts.  This part is good for dark mode, too.

https://fireflydev.ipac.caltech.edu/firefly-1394-chart-ui-conversion/firefly/
Test all features of chart except for any popup.  
- Load one or more charts
- Expanded mode
- Multi-chart grid mode as well as single mode
- Pin chart


src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -25,6 +26,12 @@ function adjustLayout(layout={}) {
     set(layout, 'yaxis.ticksuffix', hasOppositeY ? '  ' : '');
     set(layout, 'margin.b', hasOppositeX ? MIN_MARGIN_PX: X_TICKLBL_PX);
     set(layout, 'margin.t', hasOppositeX ? X_TICKLBL_PX: MIN_MARGIN_PX + (hasTitle ? TITLE_PX: 0));
+
+    // make background same as app's background color
+    const bgSurface = getTheme()?.palette?.background?.surface?.split(',')[1].slice(0,-1);     // plotly will only take a color string
+    if (!layout?.paper_bgcolor) set(layout, 'paper_bgcolor', bgSurface);
+    if (!layout?.plot_bgcolor)  set(layout, 'plot_bgcolor', bgSurface);
```
robyww, line 33: You could eventually start checking for dark mode and change it. I did that with hips.

robyww (approved):
Looks fine.

In my PR I changed a couple of icons in the table (settings, filter). So we are going going to have some conflict. in `TablePane.jsx` and probably `Buttons.jsx`

Looks like the only toolbar buttons that are not supporting dark move now are the three chart moves (zoom, pan, select). We can work that out eventually.

src/firefly/js/charts/ui/PlotlyChartArea.jsx
loitly: `background.surface` supports both mode.  but, we need to reload browser.  i can add useColorMode so that it watches for mode change in my next PR.

PR #1488, opened 2024-02-02
FIREFLY-1398: Export additional functions to JS API

loitly:
- Added all exported functions and constants from TableUtil and TableRequestUtil
- Added all exported functions and constants from ChartUtil and selected functions from ChartCntlr

Test: https://fireflydev.ipac.caltech.edu/firefly-1398-api-more-util-func/firefly/test/tests-chart.html
Bring up Chrome DevTools, then inspect `firefly.util.table` and `firefly.util.chart`.


robyww (approved):
Looks good. I think you should move the PR to rc-2023.3

PR #1491, opened 2024-02-05
FIREFLY-1402: more chart related UI conversion changes

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1402

Fixed so far:

Table UI:
Apply full borders to table cells, set the font size to 12px, reduce the row height to 20px(back to pre-Joy's height), and unset font smoothing.

13.5: 
VO TAP search, look at "Output Column Selection and Constraints" part of the screen. Scroll left and right. The part of the "Name" column that is not occupied by text is transparent! you can see the contents of the rest of the table as you scroll left/right.

18:
when in dark mode, the text color used for links is unreadable
(found in: iv image results, when many images loaded, list of images pop up)

Test: https://fireflydev.ipac.caltech.edu/firefly-1402-more-chart-conversion/firefly/

### DIFF
--- a/src/firefly/js/tables/ui/BasicTableView.jsx
+++ b/src/firefly/js/tables/ui/BasicTableView.jsx
@@ -31,7 +31,7 @@ export const BY_SCROLL = 'byScroll';
 const BasicTableViewInternal = React.memo((props) => {
 
     const {width, height} = props.size;
-    const {columns, data, hlRowIdx, renderers, bgColor, selectInfoCls, callbacks, rowHeight, rowHeightGetter, showHeader=true,
+    const {columns, data, hlRowIdx, renderers, selectInfoCls, callbacks, rowHeight, rowHeightGetter, showHeader=true,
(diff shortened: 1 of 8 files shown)


robyww (approved):

lrebull:
- I think the fonts look much better!
- 13.5 is fixed!
- 18 is better, I think.
- I think the table looks better!
- 41 - i noticed this appears to be fixed though not listed here .. ah, it's in the referenced ticket. chasing the listed items there too
- 43 - fixed here, not noted here or in referenced ticket
- **21 is still a thing**
- **36, 38, 40 is still not done**
- 39 is fixed
- 42 seems to still have a slight bold, and italics - not sure if this is what you wanted to stay or not. i think it's ok, but then i'm used to it from ops..?
- **45, 49, 50 all still a thing**
- **47 is still a thing and behaves weirdly on my laptop as well as my desktop**

I think that's everything, or did I miss something?
confluence page updated accordingly.

PR #1496, opened 2024-02-09
FIREFLY-1408: JoyUI reported bugs

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1408

Only the listed bug reports were fixed.  Will rollover the unfinished work to another ticket.

Bug report fixed: https://confluence.ipac.caltech.edu/display/IRSA/Testing+Feb+2024+-+Joy+UI%2C+mostly
21, 36, 38, 39, 40, 41, 42, 45, 47, 49, 50

Notes on specific item:
39: Not exactly sure what this is asking for.  I added a border to Chart’s toolbar.  Is that what you want?
41: IRSA TAP returns columns with a `width` attribute.  Firefly respects this value.  However, this may result in some columns being rendered too narrow for their content.
45: Implemented to retain the previous behavior but with the JoyUI look.
47: Fixed all except for the exact position of the tooltip.  This is generic code.  May need to rethink.

Test: https://fireflydev.ipac.caltech.edu/firefly-1408-chart-ui-conversion/firefly/

#### Check When confirmed
 
- [ ] 21
- [x] 36
- [x] 38
- [x] 39
- [x] 40
- [x] 41
- [x] 42
- [x] 47
- [x] 49
- [x] 50

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -524,16 +524,3 @@ div.rootStyle img {
     flex-direction: column;
     flex-grow: 1;
 }
-
-.required-msg {
-    font-style: italic;
-    margin: 10px;
-}
-.required-msg::before {
-    content: 'Required fields are marked';
-}
-.required::after, .required-msg::after {
-    color: red;
-    margin-left: 3px;
-    content: '*';
-}
--- a/src/firefly/js/charts/ui/ChartPanel.jsx
+++ b/src/firefly/js/charts/ui/ChartPanel.jsx
@@ -1,6 +1,6 @@
 import React, {useEffect} from 'react';
 import PropTypes from 'prop-types';
-import {Box, Stack} from '@mui/joy';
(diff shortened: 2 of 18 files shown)


robyww (approved):
Code looks good. Wait for @lrebull 's feedback.

lrebull:
21 - i now see that there is whitespace that is maybe part of the gear icon? so the line in the table just stops, rather than running into the gear. if this is ok with you, then we can leave it here. (it might be a result of my old OS...)
Screen Shot 2024-02-09 at 11 06 20 AM

lrebull:
everything else is ok!

loitly:
> 21 - i now see that there is whitespace that is maybe part of the gear icon? so the line in the table just stops, rather than running into the gear. if this is ok with you, then we can leave it here. (it might be a result of my old OS...) 

I adjusted the header after doing 21.  That's why it did not fit perfectly into the header.  I'll fix it in my next ticket.

PR #1497, opened 2024-02-13
FIREFLY-1412: Finish Charts conversion left over from FIREFLY-1408

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1412

Converted the listed files to JoyUI.

One significant piece is the addition of `AutoCompleteInput` to replace `SuggestBoxInputField`.  `SuggestBoxInputField` is still in used by other component.  The behavior may not be exactly the same.  But, using JoyUI's Autocomplete will allow us to add more features moving forward.

https://fireflydev.ipac.caltech.edu/firefly-1412-more-chart-ui/firefly/
Things to test:
- [ ] ColumnOrExpression:  Chart options: x/y/error/color map/ size map
- [ ] ColSelectView: popup to select column.  Added double-click to select then close dialog
- [ ] Chart save dialog
- [ ] ChartSelectDropdown:  from top menu button
- [ ] Combine chart popup
- [ ] Chart filter popup
- [ ] Fixed Table Advance Filter
- [ ] Other minor cleanup

### DIFF
--- a/src/firefly/js/charts/ui/ChartPanel.jsx
+++ b/src/firefly/js/charts/ui/ChartPanel.jsx
@@ -89,7 +89,7 @@ export const ChartToolbar = (props={}) => {
     // logic for PinnedChartContainer toolbar added here
     if (allowPinnedCharts()) {
         return (
-            <Stack id='chart-toolbar' direction='row' alignItems='center' spacing={1}>
+            <Stack id='chart-toolbar' direction='row' alignItems='center'>
                 <CombineChart {...{viewerId, tbl_group}} />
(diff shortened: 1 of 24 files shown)


robyww (approved):
The `AutoComplete` is very nice! My only concern that the don't size by pixel units. Rubin has talked about skinning with a different font.

loitly (commented):
Added reviewer's suggestions.  Test instance updated.
@lrebull I agree with the hyperactive tooltip.  But, this is the default.  I tone down the tooltip for column expression only.  @robyww will decide whether the default should be tweaked.

PR #1499, opened 2024-02-16
FIREFLY-1370: Implement Cascade-style plots for spectra

jaladh-singhal:
Fixes [FIREFLY-1370](https://jira.ipac.caltech.edu/browse/FIREFLY-1370)

Adds cascading option to combine chart dialog which makes sure that plots don't overlap when combined.

Note: I changed two icon in chart panel's pinned mode (feel free to suggest better icons):
- Show tables - because the arrow was pointing in wrong direction in Bi-view (I believe it was created with Tri-view in mind)
- Combine chart - because it was looking more like swapping than combining

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1370-cascade-plots/firefly

- Upload spectra files linked in the ticket (which have same wavelength range), pin charts and then combine -> Apply cascading with default padding -> check combined charts don't overlap.
- Decrease padding to make cascading of spectra tighter.

This achieves what design ticket is asking for, but more testing is needed to find out the edge cases where it will fail.
- It is known that it will fail if charts with different spectral axis units are combined - that's being tracked separately at FIREFLY-1414


lrebull:
IANAS (I Am Not A Spectroscopist) but it does seem to be working as specified. 
I think the workflow for the merging of spectra is a little counter-intuitive, but that's a larger problem than just this implementation.
I am not sure about the icons, either, but I don't have any better suggestions.
The explanation of what you're doing for the padding and cascading is very clear and so users should be able to understand what it's doing and what it's asking for.

src/firefly/js/charts/ui/CombineChart.jsx
```diff
@@ -178,9 +182,33 @@ const CombineChartDialog = ({onComplete}) => {
...
 
+
+const CascadePlots = ({doCascading, setDoCascading}) => {
+    const paddingValidator = (val) => Validate.floatRange(-1,1,0,'Cascade Padding', val, false);
+    return (
+        <Stack spacing={1}>
+            <FormControl orientation={'horizontal'} sx={{'.MuiSwitch-root': {margin: 0}}}>
+                <FormLabel>Apply Cascading: </FormLabel>
+                <Switch checked={doCascading} onChange={(e)=>setDoCascading(e.target.checked)}/>
+            </FormControl>
```
loitly, line 193: If a SwitchInputField isn't already present, we should add one.  It'll remove the need to handle state and populate props outside of `fields`.

loitly (approved):
Great job on this nice, clean solution. However, while I was able to test successfully using your sample tables, it didn't work with another set of tables I have. I'll send them to you. I am guessing it might be related to the double negative (--) appended to the formula.

src/firefly/js/charts/ui/CombineChart.jsx
robyww: You can use `<CheckBoxInputField type='switch/>`

PR #1501, opened 2024-02-20
FIREFLY-1415: Table options default size is too small

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1415

- [x] Changed popup size to 65% of viewport for table options and table info(i-icon).
- [ ] Table info: fixed layout issue; grouped collapsible panels for cleaner look.
- [ ] Replaced table cell ‘more’ (…) popup with JoyUI DropDown.  Beside it being a better solution, it also fixes the issue where the old popup appeared cropped at the edges.
- [ ] Converted column dropdown filter(down arrow button) to JoyUI Dropdown and revamp its buttons.

Test: https://fireflydev.ipac.caltech.edu/firefly-1415-tableoptions/firefly

### DIFF
--- a/src/firefly/js/tables/ui/TableInfo.jsx
+++ b/src/firefly/js/tables/ui/TableInfo.jsx
@@ -5,7 +5,7 @@ import {Typography, Box, Link, Stack} from '@mui/joy';
 import {isEmpty} from 'lodash';
 import {getTblById, hasAuxData} from '../TableUtil.js';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
-import {CollapsiblePanel} from '../../ui/panel/CollapsiblePanel.jsx';
+import {CollapsibleGroup, CollapsibleItem} from '../../ui/panel/CollapsiblePanel.jsx';
 import {ContentEllipsis} from './TableRenderer.js';
 import {ObjectTree} from './ObjectTree.jsx';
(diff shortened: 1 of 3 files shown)


lrebull:
I'm not 100% sure what the changes are (from the user perspective) for the 2nd, 3rd, and 4th tickboxes are above - i have checked the build and it looks ok, and seems to behave ok. Some of the stuff seems to behave the same way it did on ops, which isn't entirely consistent with the wording above, so I'm just going to assume it's ok...?

loitly:
Item 4 is the most obvious among the other 3.  No more 'x' icon; 'filter' is now replaced with a big 'Apply' button on the bottom.  'clear' link is now replaced with a small(chip) clear button on top to the checkboxes.

The flow is to do all of the selections including 'clear', then apply at the end.  Doing it interactively may pose a performance issue for BIG table.

image

loitly:
The other 2 items, as long as it's still working properly, then that's good enough.

robyww:
This might be outside the scope a little but I don't think the link here are Joy UI

Screenshot 2024-02-20 at 4 21 54 PM

robyww (approved):
Looks good. I think it looks better.

I put a separate note about TableMeta links

lrebull:
ahhh ok. i also get the filters expanding up on my screen, not down, which kind of hindered my understanding a bit.

loitly:
> I put a separate note about TableMeta links

No worries.  I fixed it.

PR #1503, opened 2024-02-21
Firefly-1418: UI cleanup

robyww:
#### [Firefly-1418](https://jira.ipac.caltech.edu/browse/FIREFLY-1418): UI cleanup

 - dropdown should go up if no room down
 - tap
      -  general tap: show services by default
      - clean up layout for different modes
 - menu: small tweeks to make it look better
 - AppConfigDrawer will close when you click the app icon

#### Testing

 - https://firefly-1418-ui-cleanup.irsakudev.ipac.caltech.edu/irsaviewer
 - other tap stuff is mostly better spacing
 - do tap search, make table area small, click on microscope
 - close draw by clicking on app icon
 - There should be a little line under the results tab when it is not selected (like the other tabs)
 - the side bar result item is better aligned

### DIFF
--- a/src/firefly/js/ui/DropDownMenu.jsx
+++ b/src/firefly/js/ui/DropDownMenu.jsx
@@ -4,7 +4,7 @@ import {isFunction} from 'lodash';
 import React, {useContext, useEffect, useRef, useState} from 'react';
 import {Box, Button, Card, Stack, Tooltip} from '@mui/joy';
 import PropTypes from 'prop-types';
-import {DropDownDirCTX} from './DropDownDirContext.js';
+import {DropDownDirCtx} from './DropDownDirContext.js';
 
 const DEFAULT_DROPDOWN_DIR = 'right';
 export const DROP_DOWN_WRAPPER_CLASSNAME= 'ff-dropdown-menu';
@@ -17,8 +17,13 @@ function placeDropDown(e,x,y, beforeVisible) {
     if (left<5) left= 5;
     e.style.left= left +'px';
     e.style.top= (pos.y + 10)+'px';
-    if (isFunction(beforeVisible)) beforeVisible(e);
+    let isAbove= false;
+    if (isFunction(beforeVisible)) {
+        const {above}= beforeVisible(e) ?? {above:false};
+        isAbove= above;
+    }
     e.style.visibility='visible';
+    return isAbove;
 }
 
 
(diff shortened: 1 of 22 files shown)


src/firefly/js/ui/DropDownMenu.jsx
```diff
@@ -4,7 +4,7 @@ import {isFunction} from 'lodash';
 import React, {useContext, useEffect, useRef, useState} from 'react';
 import {Box, Button, Card, Stack, Tooltip} from '@mui/joy';
 import PropTypes from 'prop-types';
-import {DropDownDirCTX} from './DropDownDirContext.js';
+import {DropDownDirCtx} from './DropDownDirContext.js';
```
loitly, line 7: Are you aware that JoyUI has a Dropdown component that will handle all of these things for you?  I've converted table's dropdown to it.

src/firefly/js/ui/UploadTableSelector.jsx
```diff
@@ -127,8 +127,8 @@ export function CenterColumns({lonCol,latCol, sx, cols, lonKey, latKey, openKey,
              <Typography component='div'>
                  <Stack {...{direction:'row', alignItems:'baseline', spacing:1}}>
                      <span>{headerTitle}</span>
-                     <Typography color='warning'>{(lonCol || latCol) ? `${lonCol || 'unset'}, ${latCol || 'unset'}` : 'unset'}</Typography>
-                     <Typography level='body-sm'>{`${headerPostTitle}`}</Typography>
+                     <Typography level={'body-md'} sx={{fontWeight:'normal'}}>{(lonCol || latCol) ? `${lonCol || 'unset'}, ${latCol || 'unset'}` : 'unset'}</Typography>
+                     <Typography level='body-sm' sx={{fontWeight:'normal'}}>{`${headerPostTitle}`}</Typography>
```
loitly, line 131: Why did you have to adjust fontWeight directly?  Typography level is suppose to do that.  Having a finite number of level is so that we will not have too many variation of font-size/weight.  Similar concept to variant and color.

loitly (approved):
Tested and it worked.  I have a few comments on the way it looked.  But, they are just my opinions.

- TAP
    - Spacing between Service and project is too much compare to the other fields
    - I don’t like using on/off switch as a toggle for URL vs TAP list.  
    - The 4 toggle buttons on the right does not look good.
    - Why is the (?) for IRSA Tables so far away compared to the other (?)
    - IRSA Tables section does not shrink well.  (resize horizontally)
- Menu
    - Some spacing between menu item and ‘Hide Tab’ would be nice.
        - Here is where I think the ‘on/off’ switch would be appropriate

src/firefly/js/ui/UploadTableSelector.jsx
robyww: yes, I know, but the `Typography` is inside a button which is forcing it to bold. So to actually get `body-md` it had to force the `fontWeight`. I was a little surprised by that.

src/firefly/js/ui/DropDownMenu.jsx
robyww: In a PR a few weeks ago I documented how the JoyUI dropdown would be very difficult to convert to. We would have to rewrite a lot of code. I experimented with it a lot. Also I don't think is does cascades, but I never got that far.

robyww:
> Here is where I think the ‘on/off’ switch would be appropriate

I don't agree that on/off is the right approach for the hide tab, but can get user feedback. I think any of those toggle approaches are confusing though they might look a little better. The big question is "What am I turning off?"

robyww:
All of this is sort of out of scope, but I will address some of it

> Spacing between Service and project is too much compare to the other fields

fixing
> I don’t like using on/off switch as a toggle for URL vs TAP list

I sort of like it but I don't have that strong of opinion. Maybe I can find a better place for it. But ultimately we should wait for user feedback. 
> The 4 toggle buttons on the right does not look good.

Not sure what to do here.

>Why is the (?) for IRSA Tables so far away compared to the other (?)

So that the (?) does not jump around when you switch services. I am open to sugestions.

> IRSA Tables section does not shrink well. (resize horizontally)

I will work on it.

PR #1508, opened 2024-02-29
FIREFLY-1416: UI Conversion: Update landing page to have a polished look

jaladh-singhal:
Fixes [FIREFLY-1416](https://jira.ipac.caltech.edu/browse/FIREFLY-1416)

- [x] Make landing page as per [figma design](https://www.figma.com/file/YD5gkxtxfHhXiGWe7aSEBC/FF-Navigation?type=design&node-id=52%3A700&mode=design&t=PBDhUCQNbcf4gjsr-1)
- [x] Make landing page - empty results and app branding customisable
   - [x] refactor to slots design
   - [x] try building it with SUIT and adjust
- [x] Create a better way to show arrows pointing banner tabs and background tab
   - [x] Make such component using snackbar - AppHint (currently it's specific to landing page pointing banner and is not generalized enough because positioning it on app level like a tooltip is tricky)   
   - [x] Make it appear only first-time user is loading app
- [x] Polish tabs and banner
   - [x] Make the tab border extend to entire app to make it look everything below banner look like a tab panel
   - [x] Make dividers in banner more like in figma design
   - [x] Try ~shadow~, tab hover fix, make font shrink but not tab height on resize, and other tweaks
- [x] Polish side-menu/drawer

## Testing
Firefly: https://fireflydev.ipac.caltech.edu/firefly-1416-polish-landing/firefly/

IV: https://firefly-1416-polish-landing.irsakudev.ipac.caltech.edu/irsaviewer/

SUIT: https://suit-ui-conversion.irsakudev.ipac.caltech.edu/suit/

- Check landing page and notice hint about using tabs
- Do a long running search (I tried obscore searches) and send it to background to see background monitor hint
- Switch between dev and this version of apps to see minor (but important) differences in banner 

### New
- Check that both hints only appear first time app is loaded (note: you might need to clear local storage if you are doing repeated testing)
- Check drawer is following the design we finalized in slack
- Try resizing window, the tabs will shrink in width but not in height. Also hover of tabs is now same as Results.
- Check `LandingPage` component (renamed from DefaultLandingPage) being customized by IV [in this PR](https://github.com/IPAC-SW/irsa-ife/pull/310) and by SUIT [in this commit](https://github.com/lsst/suit/commit/8af10c20f79e0caaf697ce8d921f332680ca3bb8). 

## Early feedback needed on:
- Is landing page component customizable enough or does it require restructuring?
- Does AppHints look in appearance and functionality? (note: I have chosen solid primary to make them distinguishable from alerts and I think I have seen similar looking hints on some other Material based app on internet that I can't recall)
- The Empty results section doesn't occupy full height (unlike Figma design) - do we want that? Also, do we want to reduce spacing between inner elements (aka make the layout tighter) on smaller window sizes?
- Can I use browser local storage for storing one-time-show flag of app hints (similar to how theme is stored and remembered)?
- Tab hover background is bluish for Results tab and greyish for all other tabs - is it intentional? If not, I'd like to make it bluish for all other tabs because visually, it fits well with the soft primary background.


robyww (commented):
This is looking really good.  I like the hints though I am not sure we want primary/solid.
The landing page looks good. I like the improvement on the menu tabs.

To keep preferences, use in `AppDataCntrl.js`:
```js
dispatchAddPreference()
getPreference()
```

I am fine with you fixing the tap hover.

robyww:
To test the landing page flexibility. check out and build `https://github.com/lsst/suit.git`.

I copied the old landing page. Make a new version based on the branding. The branch is `suit-ui-conversion`

loitly (commented):
Changes look good.  
AppHints: I was unexpectedly surprised by the primary tooltip pointer.  But, I agree that was needed to draw my attention since we are not blurring the page.
Tab hover: I agree.
Empty results:  If you want to make it responsive, you can.  But, it looks fine as is.

LandingPage Customizable:
What you have will probably work fine.  But, if you want to truly make it customizable, you can adopt JoyUI `slots/slotProps` concept which is much more robust.  You clearly have 3 slots, not counting root.  You can define 
```js  
slots {
  top: defaults to DefaultAppBranding
  bottom: defaults to EmptyResults
  hints: defaults to AppHint        
}
slotProps {
  top
  bottom
  hints
}
```
With the simple suggestion above, you no longer need `emptyResultsProps` and can easily customize the default slots, switch out the slot's component, or even swap between top/bottom.

This comment is only for you to consider.  No action expected.

src/firefly/js/templates/fireflyviewer/DefaultLandingPage.jsx
```diff
@@ -21,6 +22,9 @@ export function DefaultLandingPage() {
 
     return (
         <Sheet className='ff-ResultsPanel-StandardView' sx={{width: 1, height: 1}}>
+            {showLandingHelp && <AppHint hintText={'Choose a search from the tabs menu'} sx={{left: 200 + (bannerMiddleStyle?.marginLeft ?? 0)}}/>}
+            {haveBgJobs && <AppHint hintText={'Load job results from background monitor'} sx={{right: 8}} tipPlacement='end'/>}
```
loitly, line 26: I would not rely on `bannerMiddleStyle` for positioning.  It was a bad choice to added it and may be removed in the future.  I would use a reasonable offset instead.  And, if you were doing slotProps, this could be a prop each app can set.

jaladh-singhal: Ideally, I wanted to make AppHint hookable to a particular component (how joy ui tooltip works) but that will require JS position calculation magic because AppHint needs to be created at higher level in DOM (direct descendent of body) and has to watch position of banner tabs, which changes in each app.

I didn't realise I can expose the offset prop and let each app set it. Thanks! Also slots design sounds really good - I'll refactor it accordingly.

src/firefly/js/ui/AppPropertiesCtx.jsx
```diff
@@ -6,6 +6,5 @@ export const AppPropertiesCtx = React.createContext( {
     landingPage: undefined,
     footer: undefined,
     showUserInfo: false,
-    showLandingHelp: true,
```
jaladh-singhal, line 9: Removed it because now same can be achieved by `<LandingPage slots={{tabsMenuHint: false or null}}/>`

src/firefly/js/templates/fireflyviewer/LandingPage.jsx
```diff
@@ -0,0 +1,220 @@
...
+                     {...bgMonitorHintProps}
+                     sx={{right: 8, ...bgMonitorHintProps?.sx}}/>
+        ),
+        topSection=(
+            <DefaultAppBranding appTitle={appTitle} {...topSectionProps}/>
+        ),
+        bottomSection=(
+            <EmptyResults {...bottomSectionProps}/>
+        )
+    } = slots;
```
loitly, line 42: JoyUI implements `slots/slotProps` differently.  `slots` are of elementType, not Node.
Maybe, something like this.
```js
const Top = slots?.top || DefaultAppBranding;
const topProps = slots?.top ? {something} : defaultAppBrandingProps;

return (
    <Top {...topProps} {...slotProps?.top}/>
);

```
After reviewing your code, this may work better for us, since we may not need the extra flexibility.  But, I am a little concern about using `slots/slotProps`, yet implementing it differently.  If you would like to keep this, I suggest changing slots to `sections` or something other then `slots`.

jaladh-singhal: Just to make sure I understand it right: We still can't use elementType as string  (e.g. 'div') - how JoyUI does it. But we can mimic same behavior by passing React component aka function (`DefaultAppBranding`) instead of the Node (`<DefaultAppBrnading {...someProps}/>`)?

src/firefly/js/ui/Menu.jsx
```diff
@@ -96,11 +109,12 @@ export function Menu() {
     return (
         <Stack direction='row' justifyContent={'space-between'}>
             <MenuTabBarJoyTabBased {...{menuTabItems,size,selected,bCntAdd,windowWidth,dropDown}}/>
-            <Stack direction='row'>
-                {showBgMonitor && <Divider orientation='vertical' sx={{mx:1}} />}
+
+            <Stack direction='row' divider={<Divider orientation='vertical' sx={{mx:0.5, my: 1}}/>}>
+                <React.Fragment/>
```
robyww, line 114: I did not realize you could put a divider in a stack. That's nice!

src/firefly/js/ui/Menu.jsx
```diff
@@ -129,21 +143,18 @@ export function MenuItemButton({menuItem, icon, size='lg', clickHandler, isWorki
 
 
 
-function tabDivider(activeBg,size) {
+function tabDivider(size, placeAtEnd=true) {
     return {
         '&[aria-selected="false"]': {            // add pipes after non-selected tabs
-            '&::after': {
+            [`&::${placeAtEnd ? 'after' : 'before'}`]: {
```
robyww, line 149: good idea

robyww (approved):
Looks good. Good job!

loitly (approved):
Code changes look good.

PR #1517, opened 2024-03-16
FIREFLY-1427: Provide option for HydraViewer apps to use Menu instead of search panel.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1427
More changes here: 
https://github.com/IPAC-SW/irsa-ife/pull/311
https://github.com/lsst/suit/pull/51

Provide option for HydraViewer apps to use Menu instead of search panel.
- To enable, call dispatchLoadSearches with renderAsMenuItems=true.
- Add 'primary: true|false’ to SearchItem; this is then passed to MenuItem.

**Also:**
- Fix TabPanel's inconsistent usage of onTabSelect.

- Add slots/slotProps to the application-level component, e.g., FireflyLayout. The three slots are: drawer, banner, and dropdown. 
  - Maybe we should move landingPage into a slot as well, but not in this PR.

- Add meaningful slots to Banner. This makes it cleaner and more robust.

- Moved and cleaned up many app-level props, e.g., banner*Style, enableVersionDialog, showTitleOnBanner, drawerComponent.

- Fix appTitle so that it is now only a string. Replace TimeSeries usage with slotProps.banner.title, where it can be any renderable type.

- Made a small change to LandingPage after a discussion with Jaladh.

- Changes to HydraViewer to support many irsa-ui apps:

  - Apply banner/icon styling to support the new JoyUI banner.
  - Apply a fix to our non-theme-aware icons.
  - Customize EmptyResultSection.

**irsa-ui apps:**

Apply a generic fix to all of its appIcons and use a special fix for FinderChart.
Using WISE as an example, customize EmptyResults and render searches as menu items.

**Utility added:**

`setIf`: Deep set a value but only if the predicate returns true. Defaults to `isUndefined` to set only if it has not been set.
`<Slot/>`: A wrapper/renderer component that interprets/enforces our usage of slotProps.

Test for Firefly and Slate: 
https://fireflydev.ipac.caltech.edu/firefly-1427-search-panel-sidebar/firefly/

For SUIT:
https://firefly-1427-search-panel-sidebar.irsakudev.ipac.caltech.edu/suit/

For irsaviewer, finderchart, sofia, ztf, SHA, and WISE, go here then follow links:
https://firefly-1427-search-panel-sidebar.irsakudev.ipac.caltech.edu/

Because many core files were touched, we need regression testing.  
For irsa-ui apps, check to see that landing page is added and icons appeared okay in light and dark mode.
WISE is the only one where SearchPanel were replaced with sidebar MenuI items.  Use it to see how the flow is like if converted.

### DIFF
--- a/src/firefly/js/templates/hydra/HydraViewer.jsx
+++ b/src/firefly/js/templates/hydra/HydraViewer.jsx
@@ -4,7 +4,9 @@
 
 
 import React, {useEffect} from 'react';
-import PropTypes from 'prop-types';
+import PropTypes, {element, object, shape, string} from 'prop-types';
+import {Typography} from '@mui/joy';
+import {cloneDeep} from 'lodash/lang.js';
 
 import {flux} from '../../core/ReduxFlux.js';
 import {dispatchSetMenu, dispatchOnAppReady, dispatchNotifyRemoteAppReady, getSearchInfo} from '../../core/AppDataCntlr.js';
@@ -21,13 +23,17 @@ import {dispatchSetLayoutMode, LO_MODE} from '../../core/LayoutCntlr';
 import {getExpandedChartProps} from '../../charts/ChartsCntlr.js';
 import {DEFAULT_PLOT2D_VIEWER_ID} from '../../visualize/MultiViewCntlr.js';
 import App from 'firefly/ui/App.jsx';
-import {useStoreConnector} from 'firefly/ui/SimpleComponent.jsx';
+import {Slot, useStoreConnector} from 'firefly/ui/SimpleComponent.jsx';
+import {searchClickHandler, SearchPanel} from 'firefly/ui/SearchPanel.jsx';
+import {LandingPage} from 'firefly/templates/fireflyviewer/LandingPage.jsx';
+import {Stacker} from 'firefly/ui/Stacker.jsx';
+import {setIf as setIfUndefined} from 'firefly/util/WebUtil.js';
 
 
-/**
+/*
  * This is a viewer.
  */
-export function HydraViewer({menu, ...props}) {
+export function HydraViewer({menu, slotProps, ...props}) {
 
 
     useEffect(() => {
@@ -42,9 +48,23 @@ export function HydraViewer({menu, ...props}) {
         });
     }, []);
 
+    const mSlotProps = cloneDeep(slotProps || {});
+
+    // defaultView is used when the requested view does not match any of the predefined views in dropdown.
+    // with Hydra's SearchPanel, component(view) are defined by a search's renderer, i.g. renderStandardView
+    // this is a convenience way to direct all undefined requests to SearchPanel.
+    setIfUndefined(mSlotProps,'dropdown.defaultView', <SearchPanel/>);
+
+    // adjust banner for appIcon
+    setIfUndefined(mSlotProps,'banner.slotProps.icon.style.marginTop', -43);
+    setIfUndefined(mSlotProps,'banner.slotProps.tabs.pl', '120px');
+
+    setIfUndefined(mSlotProps,'landing.icon', props?.appIcon);
+    setIfUndefined(mSlotProps,'landing.title', props?.appTitle);
+
     return (
-        <App {...props}>
-            <ResultSection/>
+        <App slotProps={mSlotProps} {...props}>
+            <ResultSection slotProps={mSlotProps}/>
         </App>
     );
 }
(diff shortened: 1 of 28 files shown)


kpuriIpac:
Haven't looked at the code very closely yet, but tried Finderchart and it works great. I think in my new ticket I can incorporate your changes in to finish up the landing page (mostly just the text for `Welcome to FinderChart`). It should be similar and straightforward for other irsa-ui apps alos. 

The logo inversion (filter brightness) works great for dark/light mode. Tried it on Wise as well as FinderChart!

kpuriIpac:
This is unrelated to the work in this ticket, but one weird "bug" I found (which generally affects all our landing pages) I found was: 

Since we can now upload files on the results/landing page directly by drag and drop, if you accidentally select and drag the logo on the landing page, it'll open the Catalog Upload tab, which I thought was amusing. See video below. I say "bug" in quotes because this is technically a feature. 

https://github.com/Caltech-IPAC/firefly/assets/110492183/5b521052-798b-462e-8e41-8847a22dc860

robyww (changes requested):
Good job getting the searches into the AppDrawer. Overall the `slotProp` approach seems to accomplish the goals of flexibility. However, I do have some concerns about the implementation details.

- `slotProps` are very deep.
- I think we need to remove `App`, it makes `slotProps` confusing since all it does is wrap FireflyLayout. If you don't do that then the right thing to do is add yet another layer to `slotProps`. For RoutedApp maybe you could maybe use a slot to replace MainPanel. 
- If you don't want to do any of that then `App` needs to have the same propTypes as `FireflyLayout.propTypes`. I have gotten lost in all the app props pass through several times.
- See my comments about `setIf` for default props. I was hoping to get away from string path approaches.
- I think we need to pass a `img` instead of a string for app icon. It creates a few problems but solves more.
- I am concerned about the overuse of `Slot`. I think it should only be used where we know it is necessary. It adds complexity and implies some flexibility that may not be there. I have noticed the same with JoyUI. JoyUI seems to have many, many untested cases with replacing the component.


UI Issues

- The AppHint is not displaying the first time on a empty landing page. I think this might be because the landing page is not the default but it should be fixed.

loitly:
> UI Issues
> * The AppHint is not displaying the first time on a empty landing page.
This is intentionally turned off.  It seem awkward to have it appeared while switching tabs.  But, I have no problem turning it back on.

The other items, we need to talk.  They were all brought in to solve existing problems.  If I remove them, then all you will get is SearchItems as MenuItems.

robyww (approved):

src/firefly/js/templates/hydra/HydraViewer.jsx
```diff
@@ -135,3 +165,46 @@ function showExpandedView ({expanded}) {
...
+        });
+    }
+    return menuItems;
+}
+
+function HydraLanding({icon, title, desc, slotProps, ...props} ) {
+
+    const Greetings = () => (
+        <Stacker startDecorator={icon} direction='column' alignItems='start'>
+            <Typography level='h4'>{title || 'Welcome message here'}</Typography>
```
jaladh-singhal, line 191: If we add appTitle this can use "Welcome to ${appTitle}" as fallback (how Default Landing Page does it)

src/firefly/js/ui/SimpleComponent.jsx
```diff
@@ -254,3 +256,32 @@ export function useDebugCycle({id, render=true, mount=true}) {
...
+    customMsg: 'I am Joe'
+ }
+
+ <Slot component={EmptyResults} {...defProps} slotProps={passedInSlotProps}/>
+ This will return:
+   <MyResults text='hello world' customMsg='I am Joe'/>
+ */
+export function Slot({component, slotProps={}, ...props}) {
+    const {component:Component=component, ...pProps} = slotProps;
+    props = smartMerge(props, pProps);
```
jaladh-singhal, line 285: This will ensure that deep merging of sx I was doing in LandingPage, right? (visually, things are looking same)

jaladh-singhal (approved):
I have only skimmed code except few files - it looks great and makes things cleaner. Thanks for documentation of new functions added.

Visually, I found two things:
- on resizing the width of ife apps (HydraViewer), the tabs size doesn't shrink like it does in Firefly, IRSAViewer (FireflyViewer) - also bg monitor text doesn't convert into an icon.
- the grey ife apps' icons look better than white but I don't think it looks perfect esp with light blue and greenish-blue backgrounds behind it. We can maybe do some css image masking to make it look better?

src/firefly/js/templates/hydra/HydraViewer.jsx
```diff
@@ -135,3 +165,46 @@ function showExpandedView ({expanded}) {
...
+            menuItems.push({label: k, action:k, primary: !!v?.primary, clickHandler: searchClickHandler});
+        });
+    }
+    return menuItems;
+}
+
+function HydraLanding({icon, title, desc, slotProps, ...props} ) {
+
+    const Greetings = () => (
+        <Stacker startDecorator={icon} direction='column' alignItems='start'>
```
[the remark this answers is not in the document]
loitly: Agree.  I added the skeleton code to demonstrate where to customize the landing page.  It is my intention to have the work done in another ticket.  Hence, the `message here` comment.

src/firefly/js/ui/SimpleComponent.jsx
loitly: `smartMerge` will merge plain object and array.  But, having that logic here allow it to change without having to change it everywhere.

PR #1518, opened 2024-03-21
Firefly-1441: clean up icons for pngs

robyww:
#### [Firefly-1441](https://jira.ipac.caltech.edu/browse/FIREFLY-1441): clean up icons for pngs

   - zoom icons for images and charts
   - copy to clipboard icons
   - info icons in images selections and column adding
   - join icon in advanced tap
   - dropdown indicator now svg
   - removed commented out code
   - charts now use <ChipDelete/>
   - Not all icons changed: Some icons there are no good solutions for

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1441-icons/firefly
 - look at zoom, charts delete, copy clipboard, image dropdown


loitly (approved):
Changes look fine.  However, I noticed the new icons has the default JoyUI behavior where it will 'darken' when mouse over.  I remembered you applied an override that makes all icons always appears 'darken'.  You should do the same to these new icons.

PR #1519, opened 2024-03-21
FIREFLY-1401: UI Conversion: Improve chart options panel

jaladh-singhal:
Fixes [FIREFLY-1401](https://jira.ipac.caltech.edu/browse/FIREFLY-1401) (mostly)

- Removed `type=switch` from CheckboxGroupInputField because it was buggy and switch is meant to be used as singleton (not in group)
- Created SwitchInputField to have a field group connected Joy-UI switch with tooltip and labels and used it in Error and other places using CheckboxGroupInputField type=switch
- Fixed wrong usage of switch in TAP panel and Edit ADQL
- Fixed bug in heatmap options colorscale field
- Used CollapsiblePanelGroup to group Chart and Trace options's collapsibles in different chart modes
- Converted ImageTableRowViewer to joyUI
- Fixed rubin chart#3  bug

## Testing
Firefly: https://fireflydev.ipac.caltech.edu/firefly-1401-improve-chat-opts/firefly/

SphereX: https://firefly-1401-improve-chat-opts.irsakudev.ipac.caltech.edu/applications/spherex/

- Check switch(es) in:
    - Add Chart -> X and Y Error fields
    - TAP Panel -> CADC -> "Enter my URL" switch with helper text below the input box, "Use Image search" switch 
    - Edit ADQL -> insert fully qualified column name with helper text below label
- Go to [firefly chart tests](https://fireflydev.ipac.caltech.edu/firefly-1401-improve-chat-opts/firefly/test/tests-chart.html) -> #7heatmap -> chart options -> colorscale field should show "Greens" (instead of no option selected "Default" - which was the bug)
- Open Add Chart or Chart Options -> Change plot type to Scatter, Heatmap, Histogram -> in all 3 of them, the "Trace Options" and "Chart Options" accordions should be grouped
- In spherex -> make a search (leads to forced photometry) -> select "spectra preview" tab -> the Image Slider at the bottom should look converted and work as before. Additional: also test expanded mode - close button should look better than before.
- Rubin bug as explained in the link in ticket should no longer be there - open chart options, select a column from the suggestion list, the tooltip should disappear, also hover over tooltip will now won't keep it in focus which was making it interfere with mouse movement over other fields

### DIFF
--- a/src/firefly/js/charts/ui/ColumnOrExpression.jsx
+++ b/src/firefly/js/charts/ui/ColumnOrExpression.jsx
@@ -3,6 +3,7 @@
  */
 import React, {useContext} from 'react';
 import PropTypes, {object, shape} from 'prop-types';
+import {defaultsDeep} from 'lodash';
 import {Expression} from '../../util/expr/Expression.js';
 import {dispatchValueChange} from '../../fieldGroup/FieldGroupCntlr.js';
 import {getFieldVal} from '../../fieldGroup/FieldGroupUtils.js';
@@ -133,8 +134,12 @@ export function ColumnFld({cols, groupKey, fieldKey, initValue, label, tooltip='
                            }}
                            disableClearable={true}
                            endDecorator={!readOnly && helper ? helper : undefined}
-                           {...{required,readOnly,placeholder,slotProps,sx}}
-                            slotProps={{tooltip: {enterNextDelay:1500}}}
+                           {...{required,readOnly,placeholder,sx}}
+                           slotProps={defaultsDeep(slotProps,
+                               {tooltip: {
+                                   disableInteractive: true, //to hide tooltip when hovering over it, to prevent its long size blocking other fields
+                               }}
+                           )}
 
         />
     );
--- a/src/firefly/js/charts/ui/options/HeatmapOptions.jsx
+++ b/src/firefly/js/charts/ui/options/HeatmapOptions.jsx
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, {useContext, useEffect} from 'react';
 import {get} from 'lodash';
 
 import {getChartData} from '../../ChartsCntlr.js';
-import {FieldGroup} from '../../../ui/FieldGroup.jsx';
+import {FieldGroup, FieldGroupCtx} from '../../../ui/FieldGroup.jsx';
 
 import {toBoolean} from '../../../util/WebUtil.js';
 import {intValidator} from '../../../util/Validate.js';
@@ -15,7 +15,10 @@ import {getColValStats} from '../../TableStatsCntlr.js';
 import {ColumnOrExpression} from '../ColumnOrExpression.jsx';
 import {ALL_COLORSCALE_NAMES, PlotlyCS} from '../../Colorscale.js';
 import {getChartProps} from '../../ChartUtil.js';
-import {FieldGroupCollapsible} from '../../../ui/panel/CollapsiblePanel.jsx';
+import {
+    CollapsibleGroup,
+    FieldGroupCollapsibleItem
+} from '../../../ui/panel/CollapsiblePanel.jsx';
 import {Stack, Typography} from '@mui/joy';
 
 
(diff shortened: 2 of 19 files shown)


src/firefly/js/charts/ui/options/HeatmapOptions.jsx
```diff
@@ -116,13 +126,22 @@ export function fieldReducer({chartId, activeTrace}) {
...
+export function TableSourcesOptions({tablesource={}, activeTrace, groupKey, chartId, orientation='horizontal'}) {
     // _tables.  is prefixed the fieldKey.  it will be replaced with 'tables::val' on submitChanges.
     const tbl_id = get(tablesource, 'tbl_id');
     const colValStats = getColValStats(tbl_id);
     const xyProps = (xOrY) => ({fldPath:`_tables.data.${activeTrace}.${xOrY}`, label: `${xOrY.toUpperCase()}:`,
         name: xOrY.toUpperCase(), nullAllowed: false, colValStats, groupKey, slotProps: {control: {orientation}}});
 
+    //on the first render, update store with the color scale value reduced from chart data
+    //otherwise FieldGroupConnector is replacing it with stale state (1st option in ListBox)
+    useEffect(()=>{
```
jaladh-singhal, line 138: This is the workaround I was talking about. Setting debugger at https://github.com/Caltech-IPAC/firefly/blob/0352891e4ff4be9b48b2c1d3bccec6230237f6a7/src/firefly/js/ui/FieldGroupConnector.jsx#L194  and https://github.com/Caltech-IPAC/firefly/blob/0352891e4ff4be9b48b2c1d3bccec6230237f6a7/src/firefly/js/ui/FieldGroupConnector.jsx#L204, you'll notice that `newInitState` has correct value for colorscale field (that came by reducing chart data) but for some reason, at line 204 the value fetched from store is stale ("Default" for undefined - which is calculated initially before reducer was called) and the following logic replaces this stale value with correct value.

robyww: I think chart still uses a reducerFunc since it was written early on. This sometimes creates issues. We are trying to get away from this approach altogether. Therefore your solution is fine. Instead of using the dispatch, use the hook.
```js
    const {setVal,getVal} = useContext(FieldGroupCtx);
```

_then_

```js
   colorscaleName && setVal(`fireflyData.${activeTrace}.colorscale`, colorscaleName);
```

robyww (approved):
This looks good.

I like the tap `Select TAP Service` UI. I think you found the right combination.

I have a few clean up comments. Do those and I think it is ready to merge.

src/firefly/js/ui/SwitchInputField.jsx
```diff
@@ -0,0 +1,78 @@
...
+    const {valid, message} = validator?.(value) || {valid: true};
+
+    fireValueChange({ value, message, valid });
+    onChange?.(value);
+}
+
+
+const convertToBoolean = (value) => {
+    if (isBoolean(value)) return value;
+    else if (isString(value)) {
```
loitly, line 45: `toBoolean` in WebUtils may be doing the same.

src/firefly/js/ui/tap/AdvancedADQL.jsx
```diff
@@ -287,24 +286,22 @@ export function AdvancedADQL({adqlKey, defAdqlKey, serviceUrl, capabilities, sty
...
-                                            }} />
-                                        </Tooltip>
+                                    <Switch {...{ size:'md', sx:{alignSelf:'flex-start'},
+                                        endDecorator: (
+                                            <Stack>
+                                                <Typography>
+                                                    Insert fully-qualified column names (recommended for table joins)
+                                                </Typography>
+                                                <FormHelperText>
+                                                    When selecting a column from the Schema browser use the full qualified name
```
loitly, line 296: I like the use of FormHelperText but we should use it consistently throughout.  I don't particularly like how it looks as is.  But, we can always apply a change in theme to have it effect everywhere in the app.

loitly (approved):
Looks good!

src/firefly/js/charts/ui/options/HeatmapOptions.jsx
jaladh-singhal: Ok thanks - I updated it!

src/firefly/js/charts/ui/ColumnOrExpression.jsx
```diff
@@ -133,8 +134,12 @@ export function ColumnFld({cols, groupKey, fieldKey, initValue, label, tooltip='
                            }}
                            disableClearable={true}
                            endDecorator={!readOnly && helper ? helper : undefined}
-                           {...{required,readOnly,placeholder,slotProps,sx}}
-                            slotProps={{tooltip: {enterNextDelay:1500}}}
+                           {...{required,readOnly,placeholder,sx}}
+                           slotProps={defaultsDeep(slotProps,
+                               {tooltip: {
+                                   disableInteractive: true, //to hide tooltip when hovering over it, to prevent its long size blocking other fields
```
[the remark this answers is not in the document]
jaladh-singhal: Yeah I thought of it - if copy/pasting is needed we can maybe put the text as a form helper text

PR #1520, opened 2024-03-21
Firefly-1439: fix regression issues

robyww:
#### [Firefly-1439](https://jira.ipac.caltech.edu/browse/FIREFLY-1439): fix regression issues 

  - fixed: triview not checking store initLoadCompleted in api loads
  - fixed: `TableUtil.js` not checking for bigint before calling isNan
  - fixed: test had wrong naif id 2000002 instead of 20000002

#### Testing
   - https://fireflydev.ipac.caltech.edu/firefly-1439-regression/firefly/test/tests-viewer.html
      - load the hips or a table
   - https://fireflydev.ipac.caltech.edu/firefly-1439-regression/firefly/test/test-coverage.html?test=MOST+Orbital+path+of+Pallas
      - should load

### DIFF
--- a/src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
+++ b/src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
@@ -25,7 +25,7 @@ import {
 } from '../../visualize/ui/TriViewImageSection.jsx';
 import {ResultsPanel} from './ResultsPanel.jsx';
 
-const stateKeys= ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'images', 'coverageSide'];
+const stateKeys= ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'images', 'coverageSide', 'initLoadCompleted'];
 const LEFT= 'LEFT';
 const RIGHT= 'RIGHT';
 const triViewKey= 'images | tables | xyplots';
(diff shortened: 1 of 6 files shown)


src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
```diff
@@ -25,7 +25,7 @@ import {
 } from '../../visualize/ui/TriViewImageSection.jsx';
 import {ResultsPanel} from './ResultsPanel.jsx';
 
-const stateKeys= ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'images', 'coverageSide'];
+const stateKeys= ['title', 'mode', 'showTables', 'showImages', 'showXyPlots', 'images', 'coverageSide', 'initLoadCompleted'];
```
loitly, line 28: What is 'initLoadCompleted'?  It is not the same as onAppReady?

loitly (commented):
Changes look fine.  However, I can't get any of the `3 - Remote API: Tri-view: Send MOC Viewer Test` tests to load.

src/firefly/js/templates/fireflyviewer/TriViewPanel.jsx
robyww: `initLoadCompleted` is a boolean that is set in LayoutInfo after the first table or image or plot has loaded. It been around for three or four years.

The problem was that TriViewPanel should have gotten it from `getLayoutInfo()` with the other stuff instead of relying on a stale property. I am not sure how it worked before. 

As I think about it, maybe it would be better for TriViewPanel just to maintained a static boolean. I will think about it.

robyww:
> However, I can't get any of the 3 - Remote API: Tri-view: Send MOC Viewer Test tests to load.

The MOS's load fine for me on the build. I did notice the that table loading for going very slow on the k8s build. Maybe not enough memory. However, it does work and it quit fast on the local build.

loitly:
> The MOS's load fine for me on the build. I did notice the that table loading for going very slow on the k8s build. Maybe not enough memory. However, it does work and it quit fast on the local build.

You are right.  It's loading fine now.  Maybe backend service was overloaded when I was testing it yesterday.

loitly (approved):

PR #1521, opened 2024-03-22
Firefly-1443: improve tab resizing

robyww:
#### Testing
 - wise: https://firefly-1442-tabs.irsakudev.ipac.caltech.edu/applications/wise
 - firefly: https://fireflydev.ipac.caltech.edu/firefly-1442-tabs/firefly
 - add tabs and resize the application. They should shrink and grow


loitly (commented):
It's an improvement.  The font now resizes more accurately to the screen size, and the transition of `Background Monitor` to an icon is working.  However, it still does not address how it renders poorly when it's too wide.  See attachment.
image

One quick and dirty way to address this is to set overflow:hidden.  I am not suggesting it.    But, leaving it like this also looks bad.

robyww:
I have not even tried that. I did not know. I was really trying to work with the tabs so that is really out of range of the task. That said, I will take a look at it on Monday.

src/firefly/js/ui/Menu.jsx
```diff
@@ -94,26 +106,48 @@ export function Menu() {
...
+    return (
+        <Stack direction='row' justifyContent={'space-between'}
+               ref={(c) => {
+                   if (c) {
+                       setTabBarElement(c);
+                   }
+               }}>
+            <MenuTabBarJoyTabBased {...{menuTabItems,size,selected,dropDown, setElement}}/>
+
+            <Stack direction='row' divider={<Divider orientation='vertical' sx={{mr:0.5, mt: 1.25, mb:3/4}}/>}>
```
jaladh-singhal, line 147: Thanks for fixing this - top and bottom margin now is perfect.

[Not a big deal] I'd keep mx:0.5 except for 1st divider which needs mr: 0.5 only - because now hover background of icon button looks misaligned.

jaladh-singhal (approved):
It works as intended but I agree with Loi's comment - we need to find a more robust solution other than shrinking fontsize because it has limits.

Another thing I noticed on wise is: help icon shrinks but BG monitor icon doesn't which looks bit awkward. I think we can keep size intact for icon buttons, because help icon gets really small to notice.

src/firefly/js/ui/Menu.jsx
robyww: Thanks for catching that.  I changed it because it did not look good when the two dividers got close together. I wanted them to appear to merge but it was offset and did not look right. I will work with it more.

loitly:
Looks better.  Thanks.

PR #1526, opened 2024-03-28
Firefly-1450: Multi fixes and object id feature

robyww:
#### Firefly-1450: Multi fixes and object id feature 

 - [Firefly-1450](https://jira.ipac.caltech.edu/browse/FIREFLY-1450): object id search using select...in
 - from: https://confluence.ipac.caltech.edu/pages/viewpage.action?spaceKey=IRSA&title=Testing+April+2024
 - fix: 3
 - fix: 4
 - fix: 11
 - fix: 12- tap: clicking twice on the tables menu for it to work
 - fix: 13
 - fix: 20 - 1 & 2
 - fix: 21

@kpuriIpac :
 -  the object Id feature is in `ObjectIDSearch.jsx` and small changes in `ColumnConstrainsPanel.jsx`, `Constraints.jsx`, `TapSearchRootPanel.jsx`
 - the rest is just label changes and small bug fixes

#### Testing

  - https://firefly-1450-object-id.irsakudev.ipac.caltech.edu/irsaviewer
  - test select...in query
  - test fixes

#### Confirm

- [ ] using object id upload
- [x] 3
- [x] 4
- [x] 11
- [x] 12
- [x] 13
- [x] 20
- [x] 21

### DIFF
--- a/src/firefly/js/tables/ui/TablePanel.jsx
+++ b/src/firefly/js/tables/ui/TablePanel.jsx
@@ -333,7 +333,7 @@ function ToolBar({tbl_id, tbl_ui_id, connector, tblState, slotProps}) {
                 { expandable && !expandedMode &&
                 <ExpandButton tip={TT_EXPAND} onClick={expandTable}/>
                 }
-                { help_id && <div style={{marginTop:-10}}> <HelpIcon helpId={help_id} /> </div>}
+                { help_id && <HelpIcon helpId={help_id} />}
             </Stack>
         </Sheet>
     );
--- a/src/firefly/js/ui/CheckboxGroupInputField.jsx
+++ b/src/firefly/js/ui/CheckboxGroupInputField.jsx
@@ -1,5 +1,5 @@
 import React, {memo} from 'react';
-import PropTypes, {object, shape} from 'prop-types';
+import {element, bool, object, shape, oneOfType, string, func, arrayOf} from 'prop-types';
 import {Checkbox, FormControl, FormLabel, Stack, Tooltip} from '@mui/joy';
 import {useFieldGroupConnector} from './FieldGroupConnector.jsx';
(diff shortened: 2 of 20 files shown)


lrebull:
- 3 is done, with the caveat that we don't have all the remaining actions/wording yet for the rest of the list
- 12 - the "2 clicks to pick from the first menu" part is done, but i am still getting "this page is slowing down firefox. to speed up your browser, stop this page" when i try to use VizieR for any TAP searches. To reproduce : your build, General TAP, VizieR (CDS) TAP service, J_AA from the Table Collection menu, and then try to do anything at all with the "tables" menu.  Is this an "us" problem or a "them" problem?
- 13 is not fixed at all for me (do you want a movie?)
- 20 - Gator table help marker is fixed, as is "fraome", and as you noted above overplot's "ok" is not. 
- 21 seems to work but i am having a terrible time getting your build to load **any** images at the moment - see screenshot
Screenshot 2024-03-29 at 9 48 08 AM

kpuriIpac (approved):
Code looks good. Tested the feature with some breakpoints too (both using `Select IN` explicitly for searches that support uploads, and for searches like NED) and it works as expected. 

I have a small UI suggestion. With this new addition, for searches that don't support upload (like NED), should we mention that these searches uses `Select IN` by default? (it may be obvious to users already - but since searches that support uploads now show this option, I thought it might be a good touch to acknowledge this for searches that don't).

If it makes sense, then this message could possible go under the existing Object ID info message (~line 117): 

```
 <Typography level='body-xs'>Performs an exact match on the ID(s) provided, not a spatial search in the neighborhood of the designated objects.</Typography>
 {!canUpload && <Typography level='body-xs'>This search uses "Select IN" style SQL as this service does not support uploads.</Typography>}
```

robyww:
@kpuriIpac I was wondering about that, I will add it.

PR #1527, opened 2024-03-29
IRSA-5895 this change is needed for the search button stay at bottom …

wmiipac:
Add flexGrow to the FormPanel, this will put ife app `search ` button at the bottom of the search panel.

### DIFF
--- a/src/firefly/js/ui/FormPanel.jsx
+++ b/src/firefly/js/ui/FormPanel.jsx
@@ -79,7 +79,7 @@ export const FormPanel = function (props) {
     );
 
     return (
-        <Stack component={Sheet} className='ff-FormPanel' spacing={1} p={1} height={1} sx={{...style, ...sx}}>
+        <Stack component={Sheet} className='ff-FormPanel' flexGrow={1} spacing={1} p={1} height={1} sx={{...style, ...sx}}>
             <Stack flexGrow={1} sx={{...inputStyle, ...slotProps?.input}}>
                 {children}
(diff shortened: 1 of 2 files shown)


src/firefly/js/ui/FormPanel.jsx
```diff
@@ -79,7 +79,7 @@ export const FormPanel = function (props) {
     );
 
     return (
-        <Stack component={Sheet} className='ff-FormPanel' spacing={1} p={1} height={1} sx={{...style, ...sx}}>
+        <Stack component={Sheet} className='ff-FormPanel' flexGrow={1} spacing={1} p={1} height={1} sx={{...style, ...sx}}>
```
loitly, line 82: You should not add a `flexGrow` here unless you have confirmed that it did not break anywhere else.  This component it used in many places.  Instead, you can add `flexGrow` to `sx`.

wmiipac: > You should not add a `flexGrow` here unless you have confirmed that it did not break anywhere else. This component it used in many places. Instead, you can add `flexGrow` to `sx`.

@loitly I made change as you requested, please take a look 

> You should not add a `flexGrow` here unless you have confirmed that it did not break anywhere else. This component it used in many places. Instead, you can add `flexGrow` to `sx`.

@loitly I reverted change to FormPanel and added change to SearchPanel where is uses FormPanel component with  flexGrow added

loitly (approved):

PR #1528, opened 2024-04-02
FIREFLY-1452: fix bugs in UploadTableChooser popup

kpuriIpac:
**[FIREFLY-1452](https://jira.ipac.caltech.edu/browse/FIREFLY-1452)**: 
- fixed the bugs (`Loaded Tables `width increasing) and styling issue in UploadTableChooser (`Load` button is now at the bottom by default - set height of `Box` equal to the `minHeight` we have for resizing - that was the issue, but took some time to make it work as expected). 
- ~~got rid of the Cancel buttons. Do we want them here?t~~   

**Testing**: 
- https://fireflydev.ipac.caltech.edu/firefly-1452-upload-popup/firefly
- Test the upload table chooser popup by clicking on `Multi-Object` under Spatial Search on TAP. 
- After uploading a table, click on the Loaded Tables tab and make sure it looks okay (by default, without resizing the popup).

### DIFF
--- a/src/firefly/js/ui/UploadTableChooser.js
+++ b/src/firefly/js/ui/UploadTableChooser.js
@@ -1,4 +1,4 @@
-import {Stack} from '@mui/joy';
+import {Stack, Typography, Box} from '@mui/joy';
 import React from 'react';
 
 import {dispatchHideDialog, dispatchShowDialog} from '../core/ComponentCntlr.js';
@@ -17,10 +17,10 @@ import {TablePanel} from 'firefly/tables/ui/TablePanel';
 import {FormPanel} from 'firefly/ui/FormPanel';
 import {ServerParams} from 'firefly/data/ServerParams';
 import {doJsonRequest} from 'firefly/core/JsonUtils';
-import {dispatchHideDropDown} from 'firefly/core/LayoutCntlr';
 import {dispatchTableSearch} from 'firefly/tables/TablesCntlr';
 import {MetaConst} from 'firefly/data/MetaConst';
 import {makeFileRequest} from 'firefly/api/ApiUtilTable';
+import {dispatchHideDropDown} from 'firefly/core/LayoutCntlr';
 
 const dialogId = 'Upload-spatial-table';
 const UPLOAD_TBL_SOURCE= 'UPLOAD_TBL_SOURCE';
@@ -190,10 +190,9 @@ function applyDefColumnSelection(columns,defaultColsEnabled) {
 
 const NoTables = () => {
     return (
-        <div style={{margin: '40px 10px 10px 10px', fontSize: 'large',
-            padding: '5px 0 5px 0', textAlign: 'center', width: '100%'}}>
+        <Typography level={'title-lg'} color={'neutral'} textAlign='center' width={1} height={1} pt={2} mt={2}>
             {'No tables available to load.'}
-        </div>
+        </Typography>
     );
 };
 
@@ -214,7 +213,7 @@ const LoadedTables= (props) => {
     if (data.length === 0) {
         return <NoTables/>;
     }
-    const columns = [{name: 'Table Title', width: 45}, {name: 'No. of Cols', width: 15}, {name: 'No. of Rows', width:15},
+    const columns = [{name: 'Table Title', width: 40}, {name: 'No. of Cols', width: 15}, {name: 'No. of Rows', width:15},
         {name: 'tblId', visibility: 'hidden'}];
     const highlightedRow = 0; //default selection
     const tbl_id = 'existing-table-list';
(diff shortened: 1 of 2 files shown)


src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -268,28 +264,30 @@ export function showUploadTableChooser(setUploadInfo,groupKey= 'table-chooser',d
...
-                        <LoadedTables {...{
-                            style:{height: '100%', width:'100%'}, keepState: true, groupKey:groupKey+'-tableLoad',
-                            onCancel:() => dispatchHideDialog(dialogId),
-                            onSubmit:(request) => existingTableSubmit(request,setUploadInfo,defaultColsEnabledObj)
-                        }}/>
-                </Tab>
-            </FieldGroupTabs>
-        </Stack>
-    </FieldGroup> );};
+        <Box sx={{height:500, display: 'flex', flexDirection: 'column', resize: 'both', overflow: 'hidden', minHeight: 500, minWidth: 600}}>
```
robyww, line 267: You should make this a stack

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -268,28 +264,30 @@ export function showUploadTableChooser(setUploadInfo,groupKey= 'table-chooser',d
...
+            <FieldGroup groupKey={groupKey} sx={{ flexGrow: 1 }}>
+                <Stack sx={{zIndex: 1, pt: 1, height: '100%'}}>
```
robyww, line 269: don't need sx with stack

robyww (approved):
Looks Good. Do the following and merge

- The PR should be against `rc-2024.1` not `dev` since it is going in the release.
- Keep the cancel button. It should be there for dialogs. Removing is only for the tabs.
- a couple of clean up comments.

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -268,28 +265,30 @@ export function showUploadTableChooser(setUploadInfo,groupKey= 'table-chooser',d
...
-                            style:{height: '100%', width:'100%'}, keepState: true, groupKey:groupKey+'-tableLoad',
-                            onCancel:() => dispatchHideDialog(dialogId),
-                            onSubmit:(request) => existingTableSubmit(request,setUploadInfo,defaultColsEnabledObj)
-                        }}/>
-                </Tab>
-            </FieldGroupTabs>
-        </Stack>
-    </FieldGroup> );};
+        <Stack height='35rem' sx={{resize:'both', overflowY:'auto',minHeight:'35rem', minWidth:'40rem'}}>
+            <FieldGroup groupKey={groupKey} sx={{ flexGrow: 1 }}>
```
kpuriIpac, line 269: @robyww can you take a look at the `overflyY:auto` I added here? 
Reason: after you upload a table, if you make the panel bigger, and then make it smaller again, you lose the `Load` and `Cancel` buttons (probably needs more cleanup later to just fix this problem - but for now adding `overflowY:auto` ensures user will still have access to those buttons by scrolling).

src/firefly/js/ui/UploadTableChooser.js
kpuriIpac: Done. made the changes and re-added cancel buttons and made the PR against `rc-2024.1`.

kpuriIpac:
Made some changes to `FileUploadViewPanel`. This affects the `Upload Dialog` and also the `Upload Tab`. Test both to make sure the table shrinks in both cases upon resize (increase the size and then decrease it to test this). Since Loaded Tables calls its own `TablePanel` (it doesn't use `FileUploadViewPanel`), I was able to set `sx={{position: 'absolute', inset:0}}` for that one directly. @loitly @robyww

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -214,32 +213,31 @@ const LoadedTables= (props) => {
...
         {name: 'tblId', visibility: 'hidden'}];
     const highlightedRow = 0; //default selection
     const tbl_id = 'existing-table-list';
 
     const tableModel = {tbl_id, tableData: {columns, data}, highlightedRow, totalRows: data.length};
 
     return (
-        <div style={{margin: '10px', position: 'relative', width: '100%', display: 'flex', alignItems: 'stretch', flexDirection: 'column'}}>
-            <div style={{color: 'gray', fontSize: 'larger', lineHeight: '1.3em'}}>
+        <Box sx={{ width: 1, height: 1, display: 'flex', flexDirection: 'column'}}>
```
loitly, line 224: `display: 'flex', flexDirection: 'column'` is a Stack.  Should use that instead.
```jsx
<Stack width={1} height={1}>
```

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -622,18 +622,22 @@ function SingleDataSet({type, desc, detailsModel, report, UNKNOWN_FORMAT, curren
     const showDetails= supported && detailsModel;
     const isUWSJobFile = isUWS(report);
     const jobUrl = report.parts[0].url;
-
     return (
-        <Stack {...{direction:'row', flex:'1 1 auto', justifyContent: showDetails || isUWSJobFile?'start':'center'}}>
-            <Box sx={{pt:4, pr:2, ml: isUWSJobFile?4:0}}>
+        <Stack {...{direction:'row', flex:'1 1 auto', justifyContent: showDetails || isUWSJobFile?'start':'center', position:'relative'}}>
+            <Box sx={{pt:4, pr:2, ml: isUWSJobFile?4:0, flexShrink:0}}>
```
loitly, line 627: Same here.  Should use Stack's spacing to layout children instead of them adjusting padding and margin.

loitly (approved):
UI looks fine.  Cleanup code if you want to.

PR #1530, opened 2024-04-02
IRSA-5895 ZTF JoyUI conversion

wmiipac:
Set File location option in download option panel to horizontal. 
To test in the dev build:
https://irsa-5895-ztf.irsakudev.ipac.caltech.edu/applications/ztf

Do a position search and try `Prepare Download` in the result page.


loitly (approved):
Looks good.

PR #1533, opened 2024-04-03
FIREFLY-1449: Fix 2 pushpin in DCE and other chart bugs

jaladh-singhal:
Fixes [FIREFLY-1449](https://jira.ipac.caltech.edu/browse/FIREFLY-1449)

- Made PinChart to not show up when viewer is Data Product Viewer 
   - because current pinning functionality is designed to handle only charts that have their data tables in store (DP viewer's chart data gets replaced each time row is changed). To pin chart shown in DP Viewer, first pin its table and then from active charts, pin it.
   - this was also a bug in ops and the two pin buttons should have never existed until chart pinning can handle DP viewer - @lrebull agrees to remove it
- Exposed showAddChart as a prop in MultiChartViewer and set it false by default, and true only for ActiveChartsPanel
- Moved x, y mappings from first call of reducer to initial state of components to prevent it setting x, y axes options as undefined.

## Testing 
https://firefly-1449-2pins-chart-bugs.irsakudev.ipac.caltech.edu/irsaviewer/

- Follow the steps in ticket, check that pushpin only appears next to table|spectrum, and not in chart toolbar
- Check toolbars of Data Product Viewer, Pinned Chart Viewer, there shouldn't be "add chart" icon button present there. It should only be present in Active Chart Viewer
- Catalogs -> m81, 3600 arcsec -> active chart -> heatmap options (gear icon) -> chart options expand -> x and y axes title should be prefilled. Also change any field there, apply -> close -> reopen, the entries should persist.

### DIFF
--- a/src/firefly/js/charts/ui/ChartsContainer.jsx
+++ b/src/firefly/js/charts/ui/ChartsContainer.jsx
@@ -148,7 +148,7 @@ export const ActiveChartsPanel = (props) => {
 
         if (!useOnlyChartsInViewer) {
             return watchTblGroup(viewerId, tbl_group, addDefaultChart, useOnlyChartsInViewer);
-        };
+        }
 
     }, [tbl_group, viewerId, chartId, useOnlyChartsInViewer]);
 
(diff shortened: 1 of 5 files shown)


robyww (approved):
Looks good.

- Change the PR to be against `rc-2024.1`
- Wait until @lrebull test it before merging.

lrebull:
Yes, this seems fixed and clear to me!

PR #1534, opened 2024-04-05
IRSA-5960: Added PDF icon for use in FinderChart 

kpuriIpac:
**[IRSA-5960](https://jira.ipac.caltech.edu/browse/IRSA-5960)**: 
- IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/314

**Testing**: 
- Finder Chart: https://irsa-5960-finderchart-final-cleanup.irsakudev.ipac.caltech.edu/applications/finderchart
- do a search and see the toolbar icons, and notice the new PDF icon

### DIFF
--- a/src/firefly/js/templates/hydra/HydraViewer.jsx
+++ b/src/firefly/js/templates/hydra/HydraViewer.jsx
@@ -4,7 +4,7 @@
 
 
 import React, {useEffect} from 'react';
-import PropTypes, {element, object, shape, string} from 'prop-types';
+import PropTypes, {element, node, object, shape, string} from 'prop-types';
 import {Typography} from '@mui/joy';
 import {cloneDeep} from 'lodash/lang.js';
 
@@ -168,7 +168,7 @@ export function HydraLanding({icon, title, desc, slotProps={}, ...props} ) {
 HydraLanding.propTypes = {
     icon: element,
     title: string,
-    desc: string,
+    desc: node,
     ...LandingPage.propTypes,
 };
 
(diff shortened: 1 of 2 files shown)


lrebull:
Finder chart icon in the upper left corner still makes my skin crawl, sorry... 

On the side menu, there is one instance of "Finderchart" that initially I was going to say should be changed to "Finder Chart" but now that I look at what it actually does, that heading should be changed to "IRSA Searches" and probably also have added one of the same IRSA-specific VO TAP searches that IRSA Viewer has in that side menu. (It looks strange to just have one item under that heading anyway.)

The rest of it looks fine on a first pass! I'm going to play around with it and work on the docs; surely I'll find some more stuff that way!

robyww (approved):

kpuriIpac:
> Finder chart icon in the upper left corner still makes my skin crawl, sorry...
> 
> On the side menu, there is one instance of "Finderchart" that initially I was going to say should be changed to "Finder Chart" but now that I look at what it actually does, that heading should be changed to "IRSA Searches" and probably also have added one of the same IRSA-specific VO TAP searches that IRSA Viewer has in that side menu. (It looks strange to just have one item under that heading anyway.)
> 
> The rest of it looks fine on a first pass! I'm going to play around with it and work on the docs; surely I'll find some more stuff that way!

Hi @lrebull, before I merge this in, based on this screenshot below (from IRSA Viewer): 

Screenshot 2024-04-15 at 11 39 09 AM

What searches from here would you like to see under "IRSA Searches" in the Finder Chart sidebar?

robyww (approved):

lrebull:
We definitely need the 'upload' tab.

Under IRSA:
Let's get rid of the IRSA Data Collections (which doesn't make much sense here, and I don't know how you'd manage the results). I am not sure exactly how you'd manage Images, come to think of it. Catalogs and VO TAP (for IRSA) makes sense to me, and a new search would pull new images. How would you manage the output of an images search using the images tab?

Under "External Archive..."
Let's get rid of the CADC ObsCore (which doesn't make much sense here, and can be obtained via a regular multi-archive TAP if someone really wants) and the "Multi-Archive VO Cone Search" (which is rarely used as far as i know and just unnecessarily complicated). That just leaves two things under this header, but I think that's fine.

PR #1539, opened 2024-04-11
IRSA-5918: Cleanup inner layout of EmbeddedPositionSearchPanel

jaladh-singhal:
See https://github.com/IPAC-SW/irsa-ife/pull/319 for ticket details & testing instructions

- Made some cosmetic changes (mostly spacing related) in the EmbeddedPositionSearchPanel to clean up its inner layout


loitly (approved):

robyww (changes requested):
The component needs to be tightly spaced. The image shows a before and after in DCE.

We don't want to space it this much.

Screenshot 2024-04-11 at 3 36 42 PM

jaladh-singhal:
> The component needs to be tightly spaced. The image shows a before and after in DCE.
> 
> We don't want to space it this much.

@robyww I updated this because this will be converted to a collapsible panel when @kpuriIpac finishes his ticket - so height won't be an issue. But I'm fine with reverting it.

> Also it has a center alignment for the target and a left for the radius. You should keep it like before.

In that case I'd make search radius input field full width because the helper text looks awkward extending beyond the input's width.

loitly:
Regarding spacing, the component should expose slotProps, enabling users to define appropriate settings. This is something to keep in mind when it comes time to refactor.

kpuriIpac:
> Regarding spacing, the component should expose slotProps, enabling users to define appropriate settings. This is something to keep in mind when it comes time to refactor.

Yes, `EmbeddedPositionSearchPanel`'s props needs to use our slotProps approach. I'd only changed one to use it in my last ticket (`searchRoot`). I can do the rest in my current ticket, unless @jaladh-singhal takes on it in this one.

robyww:
> @robyww I updated this because this will be converted to a collapsible panel when @kpuriIpac finishes his ticket - so height won't be an issue. But I'm fine with reverting it.

The uses are pretty sensitive to this. I think even with collapsable it should be tight.

> In that case I'd make search radius input field full width because the helper text looks awkward extending beyond the input's width.

I guess we can try that and see how it looks

> Regarding spacing, the component should expose slotProps, enabling users to define appropriate settings

I agree, make the spacing with slotProps. Many of the parameters need this approach anyway. Just default it to a tight layout like before.

robyww (approved):
The slotProps look good and I think the search radius being 100% is fine.  

However it is still not, by default, as compact as it was before. This is something that the science teams called out in the JoyUI conversion. For DCE they want this panel tight. 

Make the current layout as tight as before and then it is fine to merge.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -212,6 +216,8 @@ EmbeddedPositionSearchPanel.propTypes= {
     searchItem: object,
     initArgs: object,
     slotProps: shape({
-        searchRoot: object
+        searchRoot: object,
+        searchInnerLayout: object,
+        searchTypeCone: object
```
loitly, line 221: The current code is not slots friendly. Therefore, I see several issues with this slotProps declaration. However, refactoring it to utilize slotProps is beyond the scope of this PR. I would suggest adding a comment to indicate that a complete refactor is acceptable and not to consider these slots as absolutely required.

PR #1541, opened 2024-04-16
FIREFLY-1459: Create DALI utility functions to support standard parameters like UPLOAD and MAX_REC.

loitly:
- create base class DaliQuery to search DAL services
- update UwsProcessor and AsyncTapQuery to make use of these functions

Mostly new code with some unit testing.  I ran test and it passed.  I built SHEREx for regression testing.

https://firefly-1459-dali-processor.irsakudev.ipac.caltech.edu/applications/spherex/

- Spectrophotometry Tool first so the results tab is populated
- Then, try TAP upload to ensure it's still working


robyww (approved):
This is hard to review, but it looks fine to me.

PR #1540, opened 2024-04-15
FIREFLY-1396: Display overflow indicator when present in VOTable.

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1396

Although ticket said 'TAP', it can be applied to any VOTABLE with a QUERY_STATUS of OVERFLOW.

Test: https://fireflydev.ipac.caltech.edu/firefly-1396-votable-overflow/firefly/demo/table-api.html
Since there aren't any services providing this status, use this test page to mimic its return. 
Include this JSON value in the META_INFO field.
```json
{
"QUERY_STATUS" : "OVERFLOW",
"backgroundable": true
}
```
Or, edit a VOTABLE with this status, then upload.

PR #1542, opened 2024-04-16
FIREFLY-1396: Display TAP overflow indicator when present in VOTable

loitly:
The INFO tags inside the RESOURCE were not populated, which is why the overflowed TAP results were not marked. This was an oversight on my part. Please test again.

https://fireflydev.ipac.caltech.edu/firefly-1396-votable-overflow/firefly/

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/JsonTableUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/table/JsonTableUtil.java
@@ -538,6 +538,7 @@ public class JsonTableUtil {
             applyIfNotEmpty(ri.getDesc(),  v -> json.put("desc", v));
             if (ri.getGroups().size() > 0) json.put("groups", toJsonGroupInfos(ri.getGroups()));
             if (ri.getParams().size() > 0) json.put("params", toJsonParamInfos(ri.getParams()));
+            if (ri.getInfos().size() > 0)  json.put("infos", ri.getInfos());
             return json;
         }).collect(Collectors.toList());
     }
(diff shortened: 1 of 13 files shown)


gpdf (changes requested):
Functionally this looks like it works now for IRSA TAP.

I would still like to see the feature from the FIREFLY-556 mock-up where, in the table-info dialog, the words "Results truncated" appear after the warning triangle.  See https://github.com/Caltech-IPAC/firefly/pull/1540#discussion_r1568031617 for an inline comment on the previous PR at the point where I think this would apply.

I've tested this with a manually uploaded VOTable from the Rubin TAP service as well and verified that the "OK followed by OVERFLOW" style of that VOTable also works.  However, I noticed that the overflow is not reported in the Table Info "(i)" dialog for a manually uploaded table.  That wasn't covered by the original ticket, of course, but it seems odd from a user perspective to have it show up in the top material of the table but not in "table info".

The appearance of the "Table Info" dialog for a manually uploaded table is also odd:

image

It looks blank even though the "Table Metadata" tab is active and has content.  If there's only one tab, shouldn't it always be active?

loitly:
> "Results truncated"

Sorry.  Overlooked.  It's added now.  Please test.

> The appearance of the "Table Info" dialog

I can't reproduce.

gpdf:
Confirmed that "Results truncated" appears now.  I think it would be better if the mouseover applied both to those words and to the warning-triangle icon, if that change is easy to make.

Confirmed that added columns now have their distinctive colors again.

Also cannot reproduce the odd behavior of the Table Info dialog.  I'll keep an eye out for this in the future.

I still think that it would be good for the overflow to be reflected somehow in the Table Info dialog, for uploaded tables with overflows, but since this is a rare situation, I'm not asking for this to be changed on this PR.

gpdf (approved):
Approved; if modifying the mouseover scope as mentioned in my comment above is fast, please do that, otherwise this is good to go.

PR #1544, opened 2024-04-19
FIREFLY-1461: Refactor FormPanel and introduce slots

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1461
- Document the usage of this component
- Perform minor cleanup of CompleteButton
- Fix all usage of FormPanel to comply with new api

More changes here: https://github.com/IPAC-SW/irsa-ife/pull/321

See src/firefly/js/ui/FormPanel.jsx for changes made.  
Because there were lots of changes throughout, please help with regression testing.
Especially Upload, DynamicUI*, and DCE related usages.

Tests:
https://fireflydev.ipac.caltech.edu/firefly-1461-refactor-formpanel/firefly/

https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/
- irsaviewer/dce/timeseries/gator
- [finderchart](https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/applications/finderchart/)
- [sofia](https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/applications/sofia/)
- [SHA](https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA/)
- [WISE](https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/applications/wise/)
- [euclid](https://firefly-1461-refactor-formpanel.irsakudev.ipac.caltech.edu/applications/euclid)

### DIFF
--- a/src/firefly/js/ui/CompleteButton.jsx
+++ b/src/firefly/js/ui/CompleteButton.jsx
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {delay, isFunction, isUndefined} from 'lodash';
+import Arrays, {delay, isFunction, isUndefined} from 'lodash';
 import React, {useContext, useEffect} from 'react';
 import {Box, Button} from '@mui/joy';
 import {object, string, func, bool, oneOfType, arrayOf} from 'prop-types';
@@ -41,17 +41,16 @@ async function collectInputs(registeredComponents, groupKey, includeUnmounted= f
     const valid = await validateFieldGroup(groups, includeUnmounted);
 
     let inputs;
-    if (isFunction(groupKey)) {
-        const groupsToValidate= groupKey();
-        inputs = groupsToValidate.reduce( (obj,gkey) => {
+    if (Arrays.isArray(groups)) {
+        inputs = groups.reduce( (obj,gkey) => {
             obj[gkey]= getFieldGroupResults(gkey,includeUnmounted);
             return obj;
         },{});
     }
     else {
-        inputs = getFieldGroupResults(groupKey,includeUnmounted);
+        inputs = getFieldGroupResults(groups,includeUnmounted);
     }
-    return [addComponentData(addComponentData(inputs,registeredComponents)), valid];
+    return [addComponentData(inputs,registeredComponents), valid];
 }
 
 
@@ -114,8 +113,7 @@ export function CompleteButton ({onFail, onSuccess, groupKey=null, text='OK',
 CompleteButton.propTypes= {
     onFail: func,
     onSuccess: func,
-    // groupsToUse: func,                  // this is confusing and not necessary.  groupKey should be [string,func].
-    groupKey: oneOfType([ string, func ]),  // when func, it returns array of string
+    groupKey: oneOfType([ string, arrayOf(string), func ]),  // when func, it returns array of string
     text: string,
     closeOnValid: bool,
     dialogId: string,
--- a/src/firefly/js/ui/CompleteButton.jsx
+++ b/src/firefly/js/ui/CompleteButton.jsx
@@ -33,59 +33,52 @@ function addComponentData(inR={}, registeredComponents={}) {
 }
 
 
-function validUpdate(valid,onSuccess,onFail,registeredComponents, closeOnValid,groupKey,
-                     dialogId, groupsToUse, includeUnmounted= false) {
-    const funcToCall = valid ? onSuccess : onFail;
-    let continueValid= true;
+async function collectInputs(registeredComponents, groupKey, includeUnmounted= false) {
 
-    if (Array.isArray(groupKey)) {
-        const groupsToValidate= groupsToUse();
-        const requestObj = groupsToValidate.reduce( (obj,groupKey) => {
-            obj[groupKey]= getFieldGroupResults(groupKey,includeUnmounted);
+    if (!groupKey) return [{}, true];   // if no groupKey, then no parameters to collect and therefore pass validation
+
+    const groups = isFunction(groupKey) ? groupKey() : groupKey;
+    const valid = await validateFieldGroup(groups, includeUnmounted);
+
+    let inputs;
+    if (isFunction(groupKey)) {
+        const groupsToValidate= groupKey();
+        inputs = groupsToValidate.reduce( (obj,gkey) => {
+            obj[gkey]= getFieldGroupResults(gkey,includeUnmounted);
             return obj;
         },{});
-        if (funcToCall) funcToCall(addComponentData(requestObj,registeredComponents));
     }
     else {
-        const request = getFieldGroupResults(groupKey,includeUnmounted);
-        if (funcToCall) continueValid= funcToCall(addComponentData(request,registeredComponents));
+        inputs = getFieldGroupResults(groupKey,includeUnmounted);
     }
-
-    const stillValid= valid && (continueValid ?? true);
-
-    if (stillValid && dialogId && closeOnValid) dispatchHideDialog(dialogId);
-}
-
-function onClick(onSuccess,onFail,registeredComponents, closeOnValid,groupKey,
-                 dialogId,groupsToUse, includeUnmounted, changeMasking, requireAllValid) {
-    if (groupKey) {
-        if (changeMasking) changeMasking(true);
-        validateFieldGroup(groupsToUse(), includeUnmounted)
-            .then( (valid)=> {
-                if (changeMasking) changeMasking(false);
-                validUpdate(valid||!requireAllValid,onSuccess,onFail,registeredComponents,closeOnValid,groupKey,dialogId, groupsToUse, includeUnmounted);
-            });
-    }
-    else {
-        if (onSuccess) onSuccess();
-        if (dialogId && closeOnValid) dispatchHideDialog(dialogId);
-    }
+    return [addComponentData(addComponentData(inputs,registeredComponents)), valid];
 }
 
 
+async function onClick(onSuccess, onFail, registeredComponents, closeOnValid, groupKey,
+                       dialogId, includeUnmounted, changeMasking, requireAllValid) {
+
+    changeMasking?.(true);
+    const [inputs, valid] = await collectInputs(registeredComponents, groupKey, includeUnmounted);
+    changeMasking?.(false);
+
+    const passValidation = valid || !requireAllValid;
+    const funcToCall = (passValidation) ? onSuccess : onFail;
+    const continueValid = funcToCall?.(inputs, passValidation) ?? true;
+    if (continueValid && dialogId && closeOnValid) dispatchHideDialog(dialogId);
+}
 
 
 export function CompleteButton ({onFail, onSuccess, groupKey=null, text='OK',
                           color=undefined, variant=undefined, requireAllValid=true,
                           closeOnValid=true, dialogId,includeUnmounted= false, primary=true,
-                          groupsToUse= () => groupKey,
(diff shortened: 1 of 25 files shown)


src/firefly/js/ui/CompleteButton.jsx
```diff
@@ -121,8 +114,8 @@ export function CompleteButton ({onFail, onSuccess, groupKey=null, text='OK',
 CompleteButton.propTypes= {
     onFail: func,
     onSuccess: func,
-    groupsToUse: func,
-    groupKey: oneOfType([ string, arrayOf(string) ]),
+    // groupsToUse: func,                  // this is confusing and not necessary.  groupKey should be [string,func].
+    groupKey: oneOfType([ string, func ]),  // when func, it returns array of string
```
robyww, line 118: I think this should be:
```js
groupKey: oneOfType([ string, arrayOf(string), func ])
```

src/firefly/js/ui/CompleteButton.jsx
```diff
@@ -33,59 +33,52 @@ function addComponentData(inR={}, registeredComponents={}) {
...
-    }
+async function collectInputs(registeredComponents, groupKey, includeUnmounted= false) {
 
-    const stillValid= valid && (continueValid ?? true);
+    if (!groupKey) return [{}, true];   // if no groupKey, then no parameters to collect and therefore pass validation
 
-    if (stillValid && dialogId && closeOnValid) dispatchHideDialog(dialogId);
-}
+    const groups = isFunction(groupKey) ? groupKey() : groupKey;
+    const valid = await validateFieldGroup(groups, includeUnmounted);
```
robyww, line 41: notice groups can be an array or a string `validateFieldGroup` accepts both. groupKey can be an array.

src/firefly/js/ui/CompleteButton.jsx
```diff
@@ -33,59 +33,52 @@ function addComponentData(inR={}, registeredComponents={}) {
...
+            obj[gkey]= getFieldGroupResults(gkey,includeUnmounted);
+            return obj;
+        },{});
     }
     else {
-        if (onSuccess) onSuccess();
-        if (dialogId && closeOnValid) dispatchHideDialog(dialogId);
+        inputs = getFieldGroupResults(groupKey,includeUnmounted);
     }
+    return [addComponentData(addComponentData(inputs,registeredComponents)), valid];
```
robyww, line 54: I think this should be
```js
return [addComponentData(inputs,registeredComponents), valid];
```

src/firefly/js/ui/CompleteButton.jsx
```diff
@@ -121,8 +114,8 @@ export function CompleteButton ({onFail, onSuccess, groupKey=null, text='OK',
 CompleteButton.propTypes= {
     onFail: func,
     onSuccess: func,
-    groupsToUse: func,
-    groupKey: oneOfType([ string, arrayOf(string) ]),
+    // groupsToUse: func,                  // this is confusing and not necessary.  groupKey should be [string,func].
```
robyww, line 117: fine to remove the commented line.

robyww (changes requested):
The FormPanel work is great. Everything is easier to follow. 

A couple of issues:
- Major: `CompleteButton` desperately needed some clean up and it is much easier to follow now. However, I think there are  a couple of bugs that I have noted below.
- Minor: I think we should use a boolean to control if the cancel button shows and not a slot prop.

src/firefly/js/ui/CompleteButton.jsx
```diff
@@ -33,59 +33,52 @@ function addComponentData(inR={}, registeredComponents={}) {
...
-                if (changeMasking) changeMasking(false);
-                validUpdate(valid||!requireAllValid,onSuccess,onFail,registeredComponents,closeOnValid,groupKey,dialogId, groupsToUse, includeUnmounted);
-            });
+    let inputs;
+    if (isFunction(groupKey)) {
+        const groupsToValidate= groupKey();
+        inputs = groupsToValidate.reduce( (obj,gkey) => {
+            obj[gkey]= getFieldGroupResults(gkey,includeUnmounted);
+            return obj;
+        },{});
```
robyww, line 49: I think this should be:
```js
    if (Arrays.isArray(groups)) {
        inputs = groups.reduce( (obj,gkey) => {
            obj[gkey]= getFieldGroupResults(gkey,includeUnmounted);
            return obj;
        },{});
    else {
        inputs = getFieldGroupResults(groupKey,includeUnmounted);
    }
```

robyww (commented):
I missed something

src/firefly/js/ui/CompleteButton.jsx
robyww: `ColorDialog.jsx` line 164 has the following:
```jsx
                  <CompleteButton
                        text='Refresh' dialogId='ColorStretchDialog' includeUnmounted={true}
                        groupKey={['colorDialogTabs',RED_PANEL,GREEN_PANEL,BLUE_PANEL]}
                        closeOnValid={false} sx={{pt:.25, pb:1, pl:1}}
                        onSuccess={replot(usedBands)} onFail={invalidMessage}
                    />
```

src/firefly/js/visualize/ui/IrsaCatalogSearch.jsx
```diff
@@ -39,15 +39,16 @@ export function IrsaCatalogSearch() {
...
+                groupKey={() => ([irsaCatalogGroupKey, gkeySpacial])}
+                onSuccess={onSearchSubmit}
+                onError={onSearchFail}
                 help_id={'catalogs.irsacatalogs'}
-                cancelText=''
-                slotProps={{input: {overflow: 'auto'}}}
-            >
+                slotProps={{
+                    input: {overflow: 'auto'},
+                    cancelBtn: {component:null}
```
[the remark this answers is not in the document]
loitly: In regards to `component=null`, I thought it was a clever way of reusing an existing reserved prop to turning on/off a slot.  We can create our own reserved prop to do that same, if that is important to you.  Just make sure it does not clash with other commonly used props, like `display`.

As for `useCancelButton: false`, I don't think it's a good idea.  This is how we end up with all top-level props to begin with.  If there is standard way of turning on/of a slot, then that's what we should use.

src/firefly/js/ui/CompleteButton.jsx
loitly: If the purpose of `func` is to only return `arrayOf(string)`, why complicate the api?  
`()=>(['abc', 'def'])` is  not hard to write and still very readable.
But, if you insist, I can change it.

src/firefly/js/ui/CompleteButton.jsx
loitly: Yes.  That's why I resolved `groupKey` in case it's a func.

src/firefly/js/ui/CompleteButton.jsx
robyww: I did some checking we never use func as far as I can tell, but we do have an array of string. See below.

src/firefly/js/visualize/ui/IrsaCatalogSearch.jsx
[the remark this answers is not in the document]
robyww: Not everything that JoyUI does like this is via a slot. They do expose common props at the top level and use slots of less used settings.

src/firefly/js/ui/CompleteButton.jsx
loitly: You are right.  I was suppose to go back to change all places that uses an array `groupKey` to a function instead.  But, if you prefer supporting 3 type of groupKey, i'll change it at the completeButton.

src/firefly/js/visualize/ui/IrsaCatalogSearch.jsx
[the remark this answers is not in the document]
loitly: I will revert cancel button changes.  It will keep its original implementation without the ability to customize.  We will expose `onCancel` and `cancelText` at the top level as before.  When `cancelText` is nullish, then no cancel button will be rendered.  This is not how I would have done it.  But, it's inline with what you are asking for.

src/firefly/js/ui/CompleteButton.jsx
loitly: I will change `groupKey` to `oneOfType([ string, arrayOf(string), func ])` as suggested.

src/firefly/js/ui/CompleteButton.jsx
robyww: I actually think we could take out the function. I don’t think we use it. I think it was an idea I had long ago the was never used. 

I hope to get away from the array. Our change in the last couple of years has made array mostly unnecessary (I think). But until we refactor everything  we will not be sure.

loitly: We used 'func' [here](https://github.com/Caltech-IPAC/firefly/blob/653149882416ce72dada04a861ff7ec612c0ab0d/src/firefly/js/visualize/ui/ImageSearchPanelV2.jsx#L125)
I removed the prop and combine it with `groupKey`.  It makes more sense to me.

src/firefly/js/visualize/ui/ImageSearchPanelV2.jsx
```diff
@@ -119,17 +119,22 @@ function ImageSearchPanel({sx, resizable=true, onSubmit, gridSupport = false, mu
...
-                        onCancel = {onCancel}
-                        cancelText={onCancel?'Cancel':''}
-                        help_id = {'basics.searching'}>
+                        onCancel= {onCancel}
+                        help_id = {'basics.searching'}
+                        slotProps={{
+                            input: {bgcolor:'transparent', p:0, border:0},
+                            searchBar: {flexShrink: 0, px:1, pb:1},
+                            cancelBtn: {
+                                component: onCancel ? undefined : null,
```
jaladh-singhal, line 131: After tabs layout, now that we are using "cancel" button infrequently, will it be worth moving this logic to FormPanel?

This will allow to have cancel button not shown (null) by default, and shows up only if someone passed onCancel

src/firefly/js/visualize/ui/IrsaCatalogSearch.jsx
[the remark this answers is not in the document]
jaladh-singhal: Maybe changing the defaults for cancel btn (see my previous comment) can help with this too while keeping the slots/slotProps design (which should be kept IMO because that's how complete button is)

jaladh-singhal (approved):
Code looks cleaner and it will be easier to modify form panel styles now (which I really found struggling when working with it in ChartSelectPanel)

Not sure if following is in the scope of this PR: in firefly, switching between Images and TAP tab causes search button to jump. If possible, we should set padding/spacing defaults that work for most of search panels using this FormPanel component to enforce consistency in their outer layout.

loitly:
@jaladh-singhal That's a good point.  We don't use 'cancel' to hidedropdown anymore, which means it does nothing.  No reason to have it shown by default.

loitly:
> @jaladh-singhal That's a good point. We don't use 'cancel' to hidedropdown anymore, which means it does nothing. No reason to have it shown by default.

I just checked.  Reversing the use case did not help.  It has the same usage issue @robyww complained about.  Since I've already reverted both cancelText and completeText, I will just leave it as it.  It's not significant.

loitly:
@robyww I am done.  Let me know if there's anything else you would like to change.

robyww (approved):

PR #1545, opened 2024-04-23
FIREFLY-1463:  Add CorsFilter

loitly:
- apply CorsFilter to font files
- moved cors logic to this class

The way I tested it:
- change compose.yml to use port 8000 instead of 8080.
- run docker locally
- change src/firefly/html/test/template.html to load from 'http://localhost:8000/firefly_loader.js'

Then, I load `http://facade-local:8080/firefly/test/tests-table.html`.
`facade-local` is alias to localhost.  I think you can use localhost to achieve the same results.

image

image

### DIFF
--- a/config/web.xml
+++ b/config/web.xml
@@ -44,6 +44,15 @@
         <url-pattern>/sticky/firefly/events</url-pattern>
     </filter-mapping>
 
+    <!--CORS Filter-->
+    <filter>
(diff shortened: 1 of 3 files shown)


jaladh-singhal (approved):
I checked out this branch and tested with Jupyter lab extension - I'm no longer seeing CORS errors for font files in console - thanks!

PR #1548, opened 2024-04-26
Firefly-1451: Embedded Position Search Panel Enhancements

kpuriIpac:
**[FIREFLY-1451](https://jira.ipac.caltech.edu/browse/FIREFLY-1451)**: 

- see ticket details
- made header as small as possible when search panel is open, and when collapsed, display a search summary. 
- reduced opacity of search panel on mouse click anywhere outside the search panel, but hover/mouseover the search panel brings it back to 100% opacity

- corresponding irsa-ife PR: https://github.com/IPAC-SW/irsa-ife/pull/323

**Testing**: 
- Euclid: https://firefly-1451-embed-panel-enhanced.irsakudev.ipac.caltech.edu/applications/euclid
   - Play around with the searches, collapse the panel for cone and multi-object (before and uploading a file - should work in both cases). 
- DCE: https://firefly-1451-embed-panel-enhanced.irsakudev.ipac.caltech.edu/irsaviewer/dce
   - the search panel will be a bit buggy for DCE right now when collapsed, since there's no groupKey being sent to `EmbeddedPositionSearchPanel`
   - but if we wrap the call to `EmbeddedPositionSearchPanel` in a FieldGroup with a groupKey, then this may work as well (or if we decide not to make it collapsible for DCE, then the collapse behavior can be made conditional in `EmbeddedPositionSearchPanel`).

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -1,12 +1,12 @@
-import {Sheet, Stack} from '@mui/joy';
-import React, {useEffect} from 'react';
+import {Sheet, Stack, Typography} from '@mui/joy';
+import React, {useContext, useEffect, useState} from 'react';
 import {oneOfType, oneOf, element, bool, string, number, arrayOf, object, func, shape} from 'prop-types';
 import CoordinateSys from '../../visualize/CoordSys.js';
 import {CONE_AREA_OPTIONS, CONE_AREA_OPTIONS_UPLOAD, CONE_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY
 } from '../../visualize/ui/CommonUIKeys.js';
 import {HiPSTargetView} from '../../visualize/ui/TargetHiPSPanel.jsx';
 import {RadioGroupInputField} from '../RadioGroupInputField.jsx';
-import {useFieldGroupValue} from '../SimpleComponent.jsx';
+import {Slot, useFieldGroupRerender, useFieldGroupValue} from '../SimpleComponent.jsx';
 import {SizeInputFields} from '../SizeInputField.jsx';
 import {DEF_TARGET_PANEL_KEY, TargetPanel} from '../TargetPanel.jsx';
 import {CONE_AREA_KEY} from './DynamicDef.js';
@@ -14,6 +14,12 @@ import {DEF_AREA_EXAMPLE, PolygonField} from './DynComponents.jsx';
 
 import {UploadTableSelector} from 'firefly/ui/UploadTableSelector';
 import {showUploadTableChooser} from 'firefly/ui/UploadTableChooser';
+import {CollapsibleGroup, CollapsibleItem} from 'firefly/ui/panel/CollapsiblePanel';
+import {FormPanel} from 'firefly/ui/FormPanel';
+import {parseWorldPt} from 'firefly/visualize/Point';
+import {formatWorldPtToString} from 'firefly/visualize/ui/WorldPtFormat';
+import {getFieldGroupResults} from 'firefly/fieldGroup/FieldGroupUtils';
+import {FieldGroupCtx} from 'firefly/ui/FieldGroup';
 /**
  * Create search panel with HiPS Viewer with an embedded target/area selection
  * All properties are optional
@@ -46,6 +52,7 @@ import {showUploadTableChooser} from 'firefly/ui/UploadTableChooser';
  * @param {Boolean} [props.useUpload] - true to allow uploads
  * @param {Object} [props.searchItem] - used for URL API (similar to SearchPanel)
  * @param {Object} [props.initArgs] - used for URL API (similar to SearchPanel)
+ * @param {function} [props.doSearch] - used for FormPanel's onSuccess search func from collapsed search panel
  * @return {JSX.Element|boolean}
  * @constructor
  */
@@ -78,13 +85,17 @@ export function EmbeddedPositionSearchPanel({
                                                 usePolygon= true,
                                                 searchItem,
                                                 initArgs,
-                                                slotProps
+                                                slotProps,
+                                                doSearch
                                             }
 ) {
 
     const [getConeAreaOp, setConeAreaOp] = useFieldGroupValue(CONE_AREA_KEY);
     const [getUploadInfo, setUploadInfo]= useFieldGroupValue('uploadInfo');
     const uploadInfo= getUploadInfo() || undefined;
+
+    const [isHovered, setIsHovered] = useState(true);
+    const [isSearchPanel, setIsSearchPanel] = useState(false);
 
     //conditionally show UploadTableChooser only when uploadInfo is empty - TAP like behavior
     useEffect(() => {
@@ -108,7 +119,7 @@ export function EmbeddedPositionSearchPanel({
     };
 
     const internals= (
-        <Stack spacing={.5} sx={{pt: insetSpacial ? 0 : 1}} {...slotProps?.searchInnerLayout}>
+        <Stack spacing={0.5} sx={{pt: insetSpacial ? 0 : 1}} {...slotProps?.searchInnerLayout}>
             {doToggle && <RadioGroupInputField {...{
                 sx:{alignSelf: 'center'},
                 fieldKey: CONE_AREA_KEY, orientation: 'horizontal',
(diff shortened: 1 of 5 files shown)


src/firefly/js/ui/panel/CollapsiblePanel.jsx
```diff
@@ -157,11 +162,13 @@ CollapsibleItemView.propTypes = {
     componentKey: string,
     isOpen: bool,
     header: oneOfType([string, func, element]),
+    expandedHeader: oneOfType([string, func, element]),
```
loitly, line 165: Instead of adding new prop and slotProps, I suggest fixing the `header` prop so that it does what it was designed to do.  `header` should be:

```js
    header: oneOfType([node, func]),      // func is called with isOpen as a parameter
```

robyww (commented):
This is looking good. The only thing I see now is making the header small when the panel is opened.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -269,11 +269,29 @@ export function EmbeddedPositionSearchPanel({
...
+                <CollapsibleGroup
+                    sx={{borderWidth: '0px',
+                        '& .MuiAccordionSummary-root': {
+                            paddingBlockStart: '0.5rem',
+                            paddingBlockEnd: '0.5rem',
+                            minBlockSize: '1rem',
+                            '&.Mui-expanded': {
+                                height: '1rem '
+                            }
+                    }}}>
```
kpuriIpac, line 281: @loitly do you mind reviewing this latest commit? it controls the styling for the header when the search panel is open (the idea was to make it as small as possible, as the scientists don't want the search panel to take up too much space) but still keep it collapsible. 

Feel free to review other stuff too, but if this looks ok I can go ahead and merge this PR.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,32 +164,136 @@ export function EmbeddedPositionSearchPanel({
...
 
+    function SearchSummary({title, searchType, coords, radius, fileName, rows, showCoords, showRadius, showFileInfo, sx}) {
+        return (
+            <Stack sx={sx}>
+                {title && <><Typography color={'warning'} level='title'>{title}</Typography></>}
+                <Typography color={'neutral'} level='body-md'>
+                    <Typography component='span' color={'primary'}>Search Type</Typography>: {searchType}
+                    {showRadius && searchType === 'Cone' && <>
+                        , <Typography component='span' color={'primary'}>Search Radius</Typography>: {radius}
+                    </>}
```
loitly, line 175: The combination of `<></>`, `, ` and nested Typography is hard to look at.  It would make it a lot easier to read if you define a simple function so you don't repeat yourself.
```js
  const keyVal = (k,v) => <>, <Typography component='span' color={'primary'}>{k}</Typography>: {v}</>;
```
I don't think `component='span'` is needed, but I guess you did it for a reason.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,32 +164,136 @@ export function EmbeddedPositionSearchPanel({
...
         <div key='targetGroup'
-             style={{display: 'flex', flexDirection: 'column', alignItems: 'center', alignSelf: 'stretch', height:'100%',
-                 paddingBottom:insetSpacial?0:20, position: 'relative'}}>
-            <HiPSTargetView {...{
+             onMouseDown={() => {
+                      if (!isSearchPanel) setIsHovered(false);
+                      else setIsHovered(true);
+             }}
+            style={{display: 'flex', flexDirection: 'column', alignItems: 'center', alignSelf: 'stretch', height:'100%',
+                         paddingBottom:insetSpacial?0:20, position: 'relative'}}>
```
loitly, line 240: This should be a Stack.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,32 +164,136 @@ export function EmbeddedPositionSearchPanel({
...
+                            px: 1/2,
+                            bottom: '1.5rem',
+                            maxWidth: '50em',
+                            left: 3,
+                            opacity: isHovered ? '100%' : '40%',
+                            ...slotProps?.searchRoot?.sx
+                        })
+                }}>
+                <CollapsibleGroup
+                    sx={{borderWidth: '0px',
```
loitly, line 273: CollapsibleGroup uses `variant='outlined'` by default.  To remove border, you should set a different variant='plain'.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,32 +164,136 @@ export function EmbeddedPositionSearchPanel({
...
+                        '& .MuiAccordionSummary-root': {
+                            paddingBlockStart: '0.5rem',
+                            paddingBlockEnd: '0.5rem',
+                            minBlockSize: '1rem',
+                            '&.Mui-expanded': {
+                                height: '1rem '
+                            }
```
loitly, line 280: I don't think these add any value except for additional padding which, in my opinion, is not needed.  It however makes it really hard for someone else to modify months from now.

kpuriIpac:
> Bug: Clicking around while in collapsed mode did not update the coordinate. It should, right?
> 
> Other feedbacks:
> 
> * too much padding around the outside.  Too much padding between Coordinates and Search button.
> * title `Search Summary` is redundant and a waste of real estate.
> 

Oh yes that's a bug, thanks for pointing it out. the Header component does not rerender unless you do a mouse over (and then it updates the coordinates). 

The padding between Coordinates and Search button is because `coords` is treated as  one long string, and even after setting CSS like `whiteSpace: 'normal'`, it still doesn't get rid of the extra space. Looking into both of these issues.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,32 +164,136 @@ export function EmbeddedPositionSearchPanel({
...
+
+        const defSlotProps = {
+            searchSummary: {
+                title: undefined,
+                searchType, coords, radius, fileName, rows,
+                showCoords: false,
+                showRadius: false,
+                showFileInfo: false,
+                sx:{}
+            },
```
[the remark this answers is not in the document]
kpuriIpac: About this, you're right about the first part. I need not add `searchType, coords, radius, fileName, rows` to searchSummary/defSlotProps as these are not meant to be overridden. So I'll just pass these directly to `SearchSummary`. 

As for the second part, @robyww wanted the Search Summary to be customizable by whoever's calling it (for example: see how I'm calling it for Euclid here: https://github.com/IPAC-SW/irsa-ife/pull/323/files). DCE may want to call it differently, and for instance, may not want to show `radius` even when it's available. In fact, even though `title` does take up extra space, it can optionally be left empty as well. 

That was the idea behind keeping the Search Summary customizable by any component calling `EmbeddedPositionSearchPanel`... But I am open to better ideas (or even a better default). Right now the default is to just show the `Search Type` (`Cone/Polygon/Multi-object`) and nothing else. And I can get rid of the `title` option altogether if it seems like there would be no good use case for it.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
kpuriIpac: Okay so the reasoning behind this one is interesting. From what I can tell Joy/MUI's AccordionSummary assigns the following values: 

``` 
paddingBlockStart: var(--ListItem-paddingY)
paddingBlockEnd: var(--ListItem-paddingY)
minBlockSize: var(--ListItem-minHeight) 
```

which are inherited from the default `JoyAccordionGroup` as: 
``` 
--ListItem-minHeight: 2.25rem;
--ListItem-paddingY: 0.25rem;
--ListItem-paddingX: 0.75rem; 
```

Now I could potentially set these in ThemeSetup.js if we wanted to use these values (the ones I overrode) elsewhere too, but it actually looks good in other places, for example this one from Firefly Image tab (ImageSelect.jsx). 

Screenshot 2024-05-09 at 7 29 14 PM
 as the default. 

Now the same (without me overriding the values) looks like this on Euclid (compare it with the test build): 

Screenshot 2024-05-09 at 7 33 39 PM

which @robyww mentioned makes the header too big when the search panel is open. 

So in order to make it smaller, I added this extra bit of styling. It's possible there may be simpler way to set the height, but I tried a few things and unless I overrode (or unchecked) this default style, my desired height for the header/summary was not respected. 
 
Anyway, if we decide to keep this, I can add a small comment explaining this choice for future developer to understand at first glance.

kpuriIpac:
@loitly I addressed most of your feedback in the latest commit. 

- bug is fixed (by using `useFieldGroupValue` which allows component to re-render when the coords change by interacting with the Hips image)
- this allowed for better formatting of the text too. Especially for cone, I decided to use `formatWorldPtToString` which makes the summary much cleaner
- for `SearchSummary`, I used a small variation of your suggested `keyVal` function, which allows me to account for the last entry (so as to not put a comma after that in the summary text). 
- finally, just need to fix up the style (padding) a bit more when the panel is collapsed and then I think then this will be ready, but let me know if you have any comments on these changes so far.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -164,26 +166,33 @@ export function EmbeddedPositionSearchPanel({
...
-                    }
-                    {showFileInfo && searchType === 'Multi-Object' && fileName && rows &&
-                        <>
-                            , <Typography component='span' color={'primary'}>File Name</Typography>: {fileName}
-                            , <Typography component='span' color={'primary'}>Rows</Typography>: {rows}
-                        </>
-                    }
+                    {keyValuePairs.map((pair, index) =>
+                        keyVal(pair.k, pair.v, index === keyValuePairs.length - 1)
+                    )}
```
loitly, line 195: I think you misunderstood.  Now, it looks worse than before.  What I meant is this,
```js
    { keyVal('Search Type', searchType) }
    { showRadius && searchType === 'Cone' && keyVal('Search Radius', radius) }
    { showCoords && searchType !== 'Multi-Object' && coords && keyVal('Coordinates', coords) }
```
If you want to insert a `,` in between, add a `<Stack divider={', '} direction='row'>` around them.  Of course, your keyVal function have to return an element and not fragment for this to work.
This is only to give you an idea of how it could have been done.  As long as it's working for you, I am ok you leaving it as is.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -204,40 +220,42 @@ export function EmbeddedPositionSearchPanel({
...
                         slotProps={{
-                            searchBar: {px:0, py:0, justifyContent: 'left'},
+                            searchBar: {p:0, justifyContent: 'right'},
                         }}
                         cancelText=''>
-                        <Slot component={SearchSummary} {...defSlotProps.searchSummary} slotProps={slotProps?.searchSummary}/>
+                        <Stack {...{width:'100%', alignItems:'center'}}>
+                            <Slot component={SearchSummary} {...defSlotProps.searchSummary} searchType={searchType}
+                                  coords={coords} radius={radius} fileName={fileName}
+                                  rows={rows} slotProps={slotProps?.searchSummary}/>
```
loitly, line 245: This is exactly the same as when you left them in defSlotProps.  Why move them out here?

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -204,40 +220,42 @@ export function EmbeddedPositionSearchPanel({
...
-             onMouseDown={() => {
-                      if (!isSearchPanel) setIsHovered(false);
-                      else setIsHovered(true);
-             }}
-            style={{display: 'flex', flexDirection: 'column', alignItems: 'center', alignSelf: 'stretch', height:'100%',
-                         paddingBottom:insetSpacial?0:20, position: 'relative'}}>
+        <Stack key='targetGroup' alignItems='center' height='100%' paddingBottom={insetSpacial ? 0 : 20}
+           onMouseDown={() => {
+                  if (!isSearchPanel) setIsHovered(false);
+                  else setIsHovered(true);
```
loitly, line 256: Isn't this `setIsHovered(!isSearchPanel);` ?

loitly (approved):
I confirm coordinates are updating while in collapsed mode now.  Ok to merge.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
kpuriIpac: Yes you're right, I can make it simpler, it would be `setIsHovered(isSearchPanel)` (whatever the value of `isSearchPanel` is)

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
kpuriIpac: Yeah this makes sense and the code is definitely cleaner using this suggested approach, but even after setting `direction='row'` for the divider stack, the formatting still looks kind of weird (I can add a screenshot if you'd like to take a look), so I'll leave this as is.

kpuriIpac:
@robyww (For **_Monday May 13_** - I'll ping you on Slack on Monday): 

- Made some more changes & cleanup after @loitly's feedback 
  - fixed a bug and it looks cleaner overall especially when the panel is collapsed  
- Was ready to merge, but noticed that search wasn't working for DCE when the panel is collapsed, and that's because I was not passing through the `submitSearch` method (that `EmbeddedPositionSearchPanel` requires to use use in its Header's `FormPanel` when collapsed). Fixed this in the latest commit, but please have a look at that. 
  - Also passing some `slotProps` from DCE for `SearchSummary` (`showCoords:true, showRadius:true`)
- [Euclid](https://firefly-1451-embed-panel-enhanced.irsakudev.ipac.caltech.edu/applications/euclid) test build & [DCE](https://firefly-1451-embed-panel-enhanced.irsakudev.ipac.caltech.edu/irsaviewer/dce) test build
- I'll wait for final feedback/okay to merge because of this DCE change.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,34 +166,151 @@ export function EmbeddedPositionSearchPanel({
...
+                        keyVal(pair.k, pair.v, index === keyValuePairs.length - 1)
+                    )}
+                </Typography>
+            </Stack>
+        );
+    }
+
+    const Header = function(isOpen) {
+        const {groupKey} = useContext(FieldGroupCtx);
+        const reqObj = getFieldGroupResults(groupKey,true);
```
kpuriIpac, line 203: @loitly I think I get what you were trying to say earlier - instead of sending 
`searchType, coords, radius, fileName, rows` to `SearchSummary` component, I could just send/use this request object directly in `SearchSummary`? is that right?

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -155,34 +166,151 @@ export function EmbeddedPositionSearchPanel({
...
+        const defSlotProps = {
+            searchSummary: {
+                title: undefined,
+                showCoords: false,
+                showRadius: false,
+                showFileInfo: false,
+                searchType, coords, radius, fileName, rows,
+                sx:{}
+            },
+        };
```
robyww, line 229: #### All this work should be done in `SearchSummary`
- for force a rerender use:
```js
useFieldGroupRerender([targetKey,sizeKey,polygonKey,CONE_AREA_KEY]);

```

#### `SearchSummary` should only take a request

```jsx
 <Slot component={SearchSummary} request={reqObj}/>
```

   - `SearchSummary` is just a default implementation that will probably always be replaced. We may may need to make a helper components for replacing it but that will be in future steps.

   - `SearchSummary`'s job is to take a request object and make the header UI.

robyww (changes requested):

robyww (approved):
`Header` and `SearchSummary` should be declared outside of the `EmbeddedPositionSearchPanel`. Move those out and then it will be fine to merge.

PR #1550, opened 2024-04-30
Firefly-1472: bug fix: menu should adjust with browser set fonts sizes

robyww:
#### Testing
- firefy: https://fireflydev.ipac.caltech.edu/firefly-1472-menu-large-font/firefly
- IrsaViewer: https://firefly-1472-menu-large-font.irsakudev.ipac.caltech.edu/irsaviewer
- using chrome, set preferences -> Appearance -> Font Size: set to `Large` or `Very Large`
- Looks at ops to see the problem: https://irsa.ipac.caltech.edu/irsaviewer/
- Menu bar should grow and shrink correctly

### DIFF
--- a/src/firefly/js/FFEntryPoint.js
+++ b/src/firefly/js/FFEntryPoint.js
@@ -23,7 +23,7 @@ import APP_ICON from 'html/images/fftools-logo-offset-small-42x42.png';
 const defProps = {
     appTitle: 'Firefly',
     initLoadingMessage: window?.firefly?.options?.initLoadingMessage,
-    appIcon: <img src={APP_ICON}/>
+    appIcon: <img src={APP_ICON} style={{width:38}}/>
 };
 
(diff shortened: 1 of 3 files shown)


lrebull:
looks good to me!

jaladh-singhal (approved):
Looks good to me!

Although IRSA icon can use more top margin (20 px instead of 6px) when browser font is set to large (I tried setting it in rem, that didn't help)
image

[Out of scope of this PR] I'm not sure how much of this larger browser fonts setting are we planning to support. You can see non-uniform icon sizes in the toolbars of different viewers in the triview results.

loitly:
Menu does not handle too many tabs well.
image

Also, you can see from the screenshot, TAP panel is not rendering those buttons correctly.
My guess is, there will be more of these issues if we look for it.  For now, the menu alignment is fixed.

robyww:
@loitly I think you found a bug, a useCallback need a dependency so width was calculated wrong in a certain case.

PR #1551, opened 2024-04-30
Firefly-1473: clean up HiPS toolbar

robyww:
#### Testing
- firefly: https://fireflydev.ipac.caltech.edu/firefly-1473-hips-toolbar/firefly
- IV: https://firefly-1473-hips-toolbar.irsakudev.ipac.caltech.edu/irsaviewer
- Load a hips image and look the hips context toolbar
- do a catalog search and look at the hips context toolbar


lrebull:
I see what you are doing and I think it looks good, but I think @gpdf needs to weigh in on whether or not it addresses what concerns he had last week. 
I did not see any difference in these menus between having a catalog or not, FWIW.

robyww:
if you just load a hips that is not a coverage, the you don't the conversion options.

gpdf (changes requested):
I like the general change.   It feels like a more readily understandable UI.

I would swap the positions of the controls so that they are in the order 

"HiPS/MOC"  //  "Galactic/Aitoff"  //  "Center Galactic"

It's not clear why "Center Galactic" shouldn't be available in both "Galactic" modes (spherical and Aitoff).

In the Coverage pane, I like the general direction, but I'm not at all comfortable with the word "Convert" for this -- not least because there is a technical ability in the HiPS infrastructure to extract FITS data from HiPS images.

I'm trying to provide rapid feedback and that's going to be at the cost of me suggesting a better word right now.  But I think it's going to be hard for users to understand without solid _a priori_ knowledge of what it's supposed to do.

"Auto zoom-in to FITS" is roughly the concept, and on the menu that appears when actually looking at FITS coverage the menu item should convey "Auto zoom-out to HiPS".  I'm not claiming that those are the best words, but they are at least explanatory.

I also think we should warn the user of what FITS they're likely to get, because we are *not* implementing the most obvious thing, which would be to zoom in to the FITS that the HiPS came from.

gpdf:
I did give it another few minutes' thought.  How about this:

On HiPS coverage images:
   * (checkable selection) Auto Zoom-in to 2MASS FITS [where the string "2MASS" comes from the Firefly config, the same place where the actual choice is controlled]
   * Switch to 2MASS FITS image [mouseover is "Switch to 2MASS FITS image at current view center; coverage extent will be limited"]

On FITS coverage images:
  * (checkable selection) Auto Zoom-out to HiPS [no name needed, assuming that it remembers whatever HiPS was most recently selected]
  * Switch to All-Sky (HiPS) image

gpdf:
Finally, I think the "HiPS/MOC" menu should be relabeled "HiPS/MOC Selection", with tooltip "Select all-sky images and data collection coverage maps".  If you can make it responsive, so that the existing short name is used when space is limited, that would be cool.

gpdf:
_(Just noticed that there's a typo in the branch name that has caused this PR to be linked to an unrelated Firefly design ticket, FIREFLY-1474.)_

robyww:
@gpdf I am implementing the other changes but `HiPS / MOC Selection` is just too long for a dropdown. 

@gpdf @lrebull In fact the am going to try to shorten the other dropdown that show the coord sys and projection. To something like `Eq J / Spherical`

lrebull:
Yeah, we definitely need to work on wording and talk/think through what order things are in and whether or not they should be consolidated... i don't have any new thoughts in comparison to yesterday where once I understood what was going on my main thought was, "ok, cool, I see where you are going, but we need better words, and I don't know what those are." Not super helpful, I realize ...

lrebull:
ah, i was looking at yesterday's build.

ok, i'm on an allWISE hips image. when i turn on "auto zoom in to fits wise", indeed, zooming in gets me a WISE image. but zooming back out dumps me into a DSS hips image, not the hips image i was in before! same for the "switch to" menu choice.

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
-];
-
-function HipsFitsConvertButton({pv}) {
+function HipsOptionsDropdown({pv}) {
+    const AITOFF_TEXT= 'Aitoff';
+    const SPHER_TEXT= 'Spherical';
     const plot= primePlot(pv);
     if (!plot) return undefined;
+    const {plotId}= plot;
+    const csysText= plot.imageCoordSys===CoordinateSys.GALACTIC ? 'Gal' : 'Eq J';
```
gpdf, line 95: I don't think "Eq J" is going to be very meaningful.  The "J" makes it more difficult to understand, not less.

If you are certain that "Equatorial / Spherical" is not going to work, then "Equ" is IMO a better abbreviation than "Eq J".  But it's still pretty cryptic if you don't already know what the purpose of the menu is.

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
     return (
-        <RadioGroupInputFieldView {...{
-            options: projOptions, value, buttonGroup:true, title:buttonGroupTip,
-            sx: {pl:.5, button:{height:'1.5em'}},
-            onChange: (ev) => dispatchChangeCenterOfProjection({plotId:pv.plotId, fullSky:ev.target.value==='aitoff'})
-        }}/>
+        <DropDownToolbarButton {...{
+            sx:{button:{minWidth: '11rem', px:0}},
+            text,
+            tip:'Change HiPS projection between spherical (180 deg) and Aitoff (360 deg)',
```
gpdf, line 172: This tooltip is outdated, as it only refers to one function of the menu.

For the menu, perhaps "Change image orientation and projection".

Because the equatorial vs galactic choice is first in the dropdown name, it should also be first in the menu.  I would name it just "Orientation" and the mouseover for that can be "Choose the orientation of the HiPS all-sky image".

I'm avoiding "coordinate system" because I've realized that it may conflict in the users' minds with their choice of what coordinate system to _overlay_ on the image.

The "HiPS Projection" part of the menu would now be second, and I would just call it "Projection" in line with the comments about reducing the prominence of the technical term "HiPS".  The mouseover for this subsection header in the menu can then be "Choose the projection for the all-sky HiPS image".

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
+                        text: SPHER_TEXT,
+                        tip: 'All-sky multi-resolution picture with spherical projection, up to 180 degrees',
+                        onClick: () => dispatchChangeCenterOfProjection({plotId: pv.plotId, fullSky: false})
+                    }}/>
+                    <ToolbarButton {...{
+                        hasCheckBox: true,
+                        checkBoxOn: isHiPSAitoff(plot),
+                        key: 'change aitoff',
+                        text: AITOFF_TEXT,
+                        tip: 'All-sky multi-resolution picture with AITOFF projection, up to 360 degrees',
```
gpdf, line 149: "image with Aitoff", please

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
+                        onClick: () => dispatchChangeCenterOfProjection({plotId: pv.plotId, fullSky: true})
+                    }}/>
+                    <DropDownVerticalSeparator useLine={true}/>
+                    <Typography sx={{whiteSpace:'nowrap'}}>HiPS Coordinate System</Typography>
+                    <ToolbarButton text='Galactic' tip='Use Galactic coordinate system' key={'gal'}
+                                   hasCheckBox={true}
+                                   checkBoxOn={plot.imageCoordSys===CoordinateSys.GALACTIC}
+                                   onClick={()=>dispatchChangeHiPS( {plotId,  coordSys: CoordinateSys.GALACTIC})}/>
+                    <ToolbarButton text='Eq J2000' tip='Use Equitorial J2000 coordinate system' key={'eqj'}
```
gpdf, line 158: "Equatorial"

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
+            </>}
+            {isHiPS(plot) &&
+                <>
+                    <Typography>HiPS Projection</Typography>
+                    <ToolbarButton {...{
+                        hasCheckBox: true,
+                        checkBoxOn: !isHiPSAitoff(plot),
+                        key: 'change Hips',
+                        text: SPHER_TEXT,
+                        tip: 'All-sky multi-resolution picture with spherical projection, up to 180 degrees',
```
gpdf, line 141: "image" rather than "picture", please

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
```diff
@@ -94,65 +86,104 @@ function changeAutoConvert(pv, auto) {
...
+function HiPSImageTable({pv}) {
     if (!primePlot(pv)) return null;
     const mi= pv.plotViewCtx.menuItemKeys;
 
     const dropDown= (
-        <SingleColumnMenu>
+        <SingleColumnMenu sx={ (theme) => ({
+            background:ctxToolbarBG(theme, 98),
+        })}>
             <ToolbarButton text={'Change HiPS'} tip={'Choose a different HiPS Survey'}
```
gpdf, line 187: "Choose a different HiPS (all-sky multi-resolution) image"

gpdf (changes requested):
I think given @lrebull's comments we're going to need a demonstration to the CCB-D for final closure.

src/firefly/js/visualize/ui/VisCtxToolbarView.jsx
robyww: I can change it, we probably need @lrebull... but you don't think an astronomer would know what Eq J means?

robyww:
> ok, i'm on an allWISE hips image. when i turn on "auto zoom in to fits wise", indeed, zooming in gets me a WISE image. but zooming back out dumps me into a DSS hips image, not the hips image i was in before! same for the "switch to" menu choice.

@lrebull this is a way it has always worked. I thing it always switches back to the default coverage HiPS. Fixing that (if it is  worth it) would need to be in another ticket

robyww:
build update with all of the feedback

PR #1552, opened 2024-05-01
FIREFLY-1187: Update TAP ObsCore.jsx to support more options

jaladh-singhal:
Fixes [FIREFLY-1187](https://jira.ipac.caltech.edu/browse/FIREFLY-1187)

- For ObsCore Instrument, Collection, and SubType, created a `ObsCoreInputField` that can render as ValidationField or multi-select AutocompleteInput, depending on the options supplied from AppOptions
- Updated ADQL generation code to allow parsing of options from ListBoxInput 
- Added a search processor `ObsCoreMetadataQuery` that fires queries for different obscore columns in parallel to retrieve metadata 
- Added client-side code in TAP root panel to fire the above query for each obscore capable service when selected, and then passes the resulting table metadata table to ObsCore panel. Also caches the results in same way as schema, tables, & columns 
- Enabled multiple mode in Autocomplete. Note: this requires disabling freeSolo for the reason mentioned in code comments

## Testing
For seeing autocomplete input field mode, check suit build (I updated SUIT.js to remove placeholders [this PR](https://github.com/lsst/suit/pull/52)): have to build locally with this branch name for both suit and firefly because of security token not working in jenkins.

DP0.2 Images tab -> the autocomplete for instrument, collection, subtype should have options. Then try instrument: LSSTCam, <NULL>; subtype: lsst.raw -> 
- check ADQL generated for it should have constraints for these fields
- submit and check that in results, the table should only have these values in their corresponding columns (use filters to inspect)

In firefly build: https://fireflydev.ipac.caltech.edu/firefly-1187-obscore-options/firefly/, check that:
- Column input field in chart options isn't broken because of changes in Autocomplete state logic
- Additionally, select TAP -> CADC -> ObsTAP: wait for 5.5 min for ObsCore options to populate, you can go to Networks tab in Devtools to see the query was fired. Once results are retrieved, ADQL should look ok and results should be as expected.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
@@ -1,0 +1,111 @@
+package edu.caltech.ipac.firefly.server.query;
+
+import edu.caltech.ipac.firefly.data.TableServerRequest;
+import edu.caltech.ipac.firefly.server.util.Logger;
+import edu.caltech.ipac.firefly.server.util.QueryUtil;
+import edu.caltech.ipac.firefly.server.util.StopWatch;
+import edu.caltech.ipac.table.DataGroup;
+import edu.caltech.ipac.table.DataObject;
+import edu.caltech.ipac.table.DataType;
+import edu.caltech.ipac.util.StringUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+import java.util.stream.Collectors;
+
+import static edu.caltech.ipac.firefly.server.query.ObsCoreMetadataQuery.*;
+import static edu.caltech.ipac.firefly.server.query.UwsJobProcessor.getTableResult;
+
+
+@SearchProcessorImpl(id = ID, params = {
+        @ParamDoc(name = SVC_URL, desc = "base TAP url endpoint excluding '/sync'"),
+        @ParamDoc(name = OBSCORE_TNAME, desc = "obscore schema.table"),
+        @ParamDoc(name = COLUMNS, desc = "columns to retrieve metadata for"),
+})
+public class ObsCoreMetadataQuery extends EmbeddedDbProcessor {
+    public static final String ID = "ObsCoreMetadataQuery";
+    public static final String SVC_URL = "serviceUrl";
+    public static final String OBSCORE_TNAME = "obscoreTable";
+    public static final String COLUMNS = "columns";
+    public static final String QUERY_FRAGMENT = "/sync?REQUEST=doQuery&LANG=ADQL&QUERY=";
+    public static final Logger.LoggerImpl LOGGER = Logger.getLogger();
+
+
+    public DataGroup fetchDataGroup(TableServerRequest request) throws DataAccessException {
+        List<String> columns = StringUtils.asList(request.getParam(COLUMNS), ",");
+        List<Future<DataGroup>> futures = new ArrayList<>();
+        List<Throwable> exceptions = new ArrayList<>();
+        StopWatch timer = StopWatch.getInstance();
+
+        ExecutorService executor = Executors.newCachedThreadPool();
+        try {
+            timer.start("All queries");
+            // Submit queries for each column in parallel
+            for (String column : columns) {
+                Future<DataGroup> future = executor.submit(() -> {
+                    timer.start(column+" query");
+                    DataGroup columnMetaData = fetchColumnMetaData(request, column);
+                    timer.printLog(column+" query");
+                    return columnMetaData;
+                });
+                futures.add(future);
+            }
+
+            // Retrieve DataGroup from each future one by one
+            List<DataGroup> columnsMetadata = new ArrayList<>();
+            for (Future<DataGroup> future : futures) {
+                try {
+                    columnsMetadata.add(future.get());  // note: blocking main thread until completion of this future, will still lead to completion of the futures running in other threads
+                } catch (InterruptedException | ExecutionException e) {
+                    exceptions.add(e);
+                    columnsMetadata.add(null);
+                }
+            }
+            timer.printLog("All queries");
+
+            // If all futures failed, throw a combined exception; otherwise log exceptions (if occurred)
+            if (exceptions.size() == futures.size()) throw new DataAccessException(mergeErrorMsg(exceptions));
+            else if (exceptions.size() > 0) LOGGER.warn(mergeErrorMsg(exceptions));
+
+            return mergeColumnsMetadata(columns, columnsMetadata);
+        } finally {
+            executor.shutdown();
+        }
+    }
+
+
+    private DataGroup mergeColumnsMetadata(List<String> columns, List<DataGroup> columnsMetadata) {
+        DataGroup table = new DataGroup("ObsCore Metadata", new DataType[]{
+                new DataType("column_name", String.class),
+                new DataType("column_options", String.class)
+        });
+
+        for (int i=0; i<columns.size(); i++) {
+            String colName = columns.get(i);
+            DataGroup colMetadata = columnsMetadata.get(i);
+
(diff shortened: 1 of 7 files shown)


src/firefly/js/ui/tap/TapUtil.js
```diff
@@ -268,6 +269,29 @@ async function doLoadTapColumns(serviceUrl, schemaName, tableName) {
     }
 }
 
+export const supportsObsCoreMetadataLoad = (serviceUrl) => {
+    return true || serviceUrl.toLowerCase().includes('lsst.cloud'); //for testing compatibility with other TAP services
```
jaladh-singhal, line 273: I have left this on in case you want to test CADC - the options will populate in the instrument and collection fields in 5.5 min from the time you selected the service.

Note: MAST ObsCore won't work because it blocks any query running longer than 2min

src/firefly/js/ui/tap/TapViewType.jsx
```diff
@@ -272,6 +273,20 @@ function BasicUI(props) {
...
     };
 
+    const loadObsCoreMeta = (serviceUrl, obsCoreTableModel) => {
+        const [, obsCoreTable] = obsCoreTableModel?.tableData?.data?.[0];
+        if (!obsCoreTable || !supportsObsCoreMetadataLoad(serviceUrl)) {
+            setObsCoreMetadataModel(undefined); //indicates loading attempt wasn't made
+        }
+        else {
+            setObsCoreMetadataModel({loading: true}); //indicates intermediate state until tableModel is available (Promise is resolved/rejected)
+            const columns = ['obs_collection', 'instrument_name', 'facility_name', 'dataproduct_subtype'].join(',');
```
robyww, line 283: Why not put this list in `loadObsCoreMetadata()`? I think it would be better to push it into the utility since these are standard columns.

robyww (approved):
This looks good. Just fix a couple of the small things I noted.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
```diff
@@ -0,0 +1,107 @@
...
+    public static final Logger.LoggerImpl LOGGER = Logger.getLogger();
+
+
+    public DataGroup fetchDataGroup(TableServerRequest request) throws DataAccessException {
+        List<String> columns = StringUtils.asList(request.getParam(COLUMNS), ",");
+        List<Future<DataGroup>> futures = new ArrayList<>();
+        List<Throwable> exceptions = new ArrayList<>();
+
+        // Submit queries for each column in parallel
+        ExecutorService executor = Executors.newCachedThreadPool();
```
loitly, line 40: You don't want to use edu.caltech.ipac.firefly.server.ServerContext#SHORT_TASK_EXEC?  
It should work the same.
But, if you want to create your own, wrap `executor.shutdown();` around a try/catch/finalize` to ensure it is cleaned up properly.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
```diff
@@ -0,0 +1,107 @@
...
+        for (Future<DataGroup> future : futures) {
+            try {
+                columnsMetadata.add(future.get());  // note: blocking main thread until completion of this future, will still lead to completion of the futures running in other threads
+            } catch (InterruptedException | ExecutionException e) {
+                exceptions.add(e);
+                columnsMetadata.add(null);
+            }
+        }
+
+        LOGGER.debug(String.format("All queries returned in  %d msec%n", System.currentTimeMillis() - sTime));
```
loitly, line 63: We have a convenient class(`StopWatch`) that will do this for you.  It is also hooked to logger, but at `TRACE` level.  This allow you to set as many debugging logs as you want during development and not have it show up in normal runtime logs.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
```diff
@@ -0,0 +1,107 @@
...
+        executor.shutdown();
+
+        // If all futures failed, throw a combined exception; otherwise log exceptions (if occurred)
+        if (exceptions.size() == futures.size()) throw new DataAccessException(mergeErrorMsg(exceptions));
+        else if (exceptions.size() > 0) LOGGER.error(mergeErrorMsg(exceptions));
```
loitly, line 68: If this is not an error, let's log it as 'warning'.

src/firefly/js/Firefly.js
```diff
@@ -183,6 +183,7 @@ const defFireflyOptions = {
     },
     tapObsCore: {
         enableObsCoreDownload: true,
+        debug: true, //TODO: remove when merging
```
loitly, line 186: Don't forget to remove.  Also, we have a client logger(Logger.js) what works similar to server-side logger.  This is a good alternative to implement logging.

loitly (approved):
Code looks fine.  Just a few comments.

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -173,7 +173,7 @@ public DataGroup getResult(TableServerRequest request) throws DataAccessExceptio
     }
 
     static String getFilename(String urlStr) {
-        return urlStr.replace("(http:|https:)", "").replace("/", "");
+        return urlStr.replaceAll("(http:|https:|/|\\?|)", "");
```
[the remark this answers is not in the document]
jaladh-singhal: Makes sense but won't it overwrite other files because of the same name?

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCoreMetadataQuery.java
jaladh-singhal: `SHORT_TASK_EXEC` was the reason I was getting null pointer exceptions I shared with you earlier. That's why I created my own executor.

About cleanup are you referring to something like this: https://github.com/Caltech-IPAC/firefly/blob/a6b372576b1558ec51b7e305fc2b864369f773bf/src/firefly/java/edu/caltech/ipac/firefly/server/ServerContext.java#L870-L877 ?

If yes, I don't know what to put in finally.

loitly:
```java
ExecutorService executor = Executors.newCachedThreadPool();
try {
  // your code
} finally {
    executor.shutdown(); 
}
```
This ensure shutdown is called when encountering uncaught exceptions.

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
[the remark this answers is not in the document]
jaladh-singhal: Yes I didn't realize it's just a prefix - I removed it and added "results-".

src/firefly/js/Firefly.js
```diff
@@ -182,7 +182,7 @@ const defFireflyOptions = {
         canCreateExtractionTable: false,
     },
     tapObsCore: {
-        enableObsCoreDownload: true,
+        enableObsCoreDownload: true
```
robyww, line 185: how about `enableObsCoreAttributesQuery`. The name is not quite clear.

robyww (approved):
Looks fine, just change the name.

src/firefly/js/Firefly.js
jaladh-singhal: Oh, I didn't mean to change it (I accidentally removed comma), the prop is in SUIT.js called `tapObsCore.enableMetadataLoad` (L176)

loitly (commented):
I saw your changes.  Looks good.

PR #1560, opened 2024-05-21
Firefly-1478: fixes items: 3,4,6,7

robyww:
#### [Firefly-1478](https://jira.ipac.caltech.edu/browse/FIREFLY-1478): fixes items 3,4,6,7 from ticket

- also - support noirlab's old version of tap
- API mode: bug fix: chart toolbar background now consistent with images and tables (variant: soft)

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1478-3-4-6-7/firefly
- test according to ticket  3,4,6,7
- helpful notes
   - choose IRSA - spitzer - popup the tables
   - choose Gaia - gaiadr3 - popup the tables
- [x]    3 - tables selection button not bold
- [x]    4 - popup icon at far right of tables button
- [x]    6 - schema in tables selection dialog
- [x]    7 - Gaia - tables selection dialog show with columns different

### DIFF
--- a/src/firefly/js/ui/tap/TapUtil.js
+++ b/src/firefly/js/ui/tap/TapUtil.js
@@ -14,6 +14,7 @@ const qFragment = '/sync?REQUEST=doQuery&LANG=ADQL&';
 export const ADQL_LINE_LENGTH = 100;
 export const ADQL_UPLOAD_TABLE_NAME= 'upload_table';
 export const TAP_UPLOAD_SCHEMA= 'TAP_UPLOAD';
+const EMPTY_SCHEMA_NAME= 'EMPTY_SCHEMA_NAME';
 
 
 // cache objects - they only grow, but I don't think they will ever get too big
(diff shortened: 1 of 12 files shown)


src/firefly/js/ui/tap/TapUtil.js
```diff
@@ -188,9 +202,25 @@ async function doLoadTapTables(serviceUrl, schemaName) {
         }
         const schemaNameIdx = getColumnIdx(tableModel, 'schema_name');
         const tableNameIdx = getColumnIdx(tableModel, 'table_name');
+        const descIdx = getColumnIdx(tableModel, 'table_desc');
+        const doSwap= descIdx<tableNameIdx;
 
         const rTable = omit(tableModel, 'tableData.data');
         rTable.tableData.data = tableModel.tableData.data.filter((row) => row[schemaNameIdx] === schemaName);
+        if (doSwap) {
+            rTable.tableData.data = rTable.tableData.data
```
loitly, line 211: I see why it's needed.  But, this workaround looks pretty bad. Maybe the fix is at `loadSchemaDefJoin`.  But, that's just a guess.

loitly (approved):
I tested the listed items and they appear to be fixed.

src/firefly/js/ui/tap/TapUtil.js
robyww: If we could talk the table to reorder the columns then we could avoid it. I also considered doing it in the select but there are other reasons to do a `select *`.

lrebull:
Screenshot 2024-05-22 at 1 23 00 PM
OK, don't kill me, but I see what Gregory is seeing in number 3. Yes, you've removed the bold, so yes, that's fixed. But if you leave it up and change the left menu from 'WISE' to 'Spitzer' while watching the right menu, you can see it move. The screenshot here is what happens when you take a screenshot of the WISE view and the Spitzer view, make one partially transparent, and line them up on top of each other. Look at the right menu -- despite 'Project' being lined up exactly, 'Tables' is not lined up, either horizontally or vertically. True, one could argue it doesn't matter. But I do see the alignment he's calling out in the ticket. If we declare that part of it ignorable, and declare that just removing the bold is sufficient, then this is ok.

lrebull:
4 is done.
6 is done.
I'm confused by the complaint in 7. It seems the same as on ops with respect to column and row ordering. The first column is wider in this version, so that's fine, but I am not sure what the complaint was in the ticket, so I'm not checking this off...?

lrebull:
with respect to https://datalab.noirlab.edu/tap, i get the "tables" menus to load, but I never get the stuff on the lower half of the screen to load for the default menu choices, and the default choice there is the same as ours (allwise!) so I assumed it hadn't loaded and was still showing me the drop-down menus from our TAP service. 

also, new bug. When I tried to pick another option from the datalab choices, I got an error; see screenshot below. "Spacial" --> "Spatial" please. And why won't it recognize ra and dec, especially given that those columns are clearly visible right there? is that a problem at the NOAO end?
  
Screenshot 2024-05-22 at 1 35 13 PM

robyww:
@lrebull I am calling 3 done

PR #1568, opened 2024-06-03
IRSA-5492: Cleanup and work related to current Euclid IFE ticket

kpuriIpac:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-5492
IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/326

- Keeping this firefly PR open for any other firefly changes that need to be made to accomodate the IFE Euclid PR (these firefly changes don't need to go into RC in my opinion) 
   - this change allows Region Explorer to determine search type and customize additional search options (IBESearchType) when the `searchType === Polygon `

Test builds in the [IFE PR](https://github.com/IPAC-SW/irsa-ife/pull/326).

### DIFF
--- a/src/firefly/js/templates/router/RouteHelper.jsx
+++ b/src/firefly/js/templates/router/RouteHelper.jsx
@@ -7,7 +7,7 @@ import {
     dispatchOnAppReady, dispatchSetMenu, FORM_CANCEL, FORM_SUBMIT, getMenu
 } from '../../core/AppDataCntlr.js';
 import {dispatchHideDropDown, dispatchSetLayoutInfo, getDropDownInfo} from '../../core/LayoutCntlr.js';
-import {dispatchAddActionWatcher} from '../../core/MasterSaga.js';
+import {dispatchAddActionWatcher, dispatchCancelActionWatcher} from '../../core/MasterSaga.js';
 import {FireflyRoot} from '../../ui/FireflyRoot.jsx';
 import {useStoreConnector} from '../../ui/SimpleComponent.jsx';
 import {dispatchComponentStateChange, getComponentState} from 'firefly/core/ComponentCntlr.js';
@@ -94,7 +94,9 @@ function useFormWatcher(submitTo='/?results', onCancel) {
                 }
             });
         });
-    },[]);
+
+        return (() => dispatchCancelActionWatcher(id));
+    },[submitTo]);
 }
 
 /**
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -154,7 +154,7 @@ export function EmbeddedPositionSearchPanel({
                     placeholder: 'Coordinates',
                     manageHiPS:false,
                 }} />}
-            {otherComponents}
+            {otherComponents && otherComponents(doGetConeAreaOp())}
             {doGetConeAreaOp() === UPLOAD_CHOICE_KEY &&
                 <UploadTableSelector {...{uploadInfo, setUploadInfo, uploadTable:true}}/>
             }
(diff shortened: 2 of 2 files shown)


src/firefly/js/templates/router/RouteHelper.jsx
```diff
@@ -94,7 +94,7 @@ function useFormWatcher(submitTo='/?results', onCancel) {
                 }
             });
         });
-    },[]);
+    },[submitTo]);
```
kpuriIpac, line 97: @loitly could I request your feedback on this? it's related to this change in the IFE PR in [SourceInspector.jsx](https://github.com/IPAC-SW/irsa-ife/pull/326/files#diff-3817738aa78c72dcf422585b88c6add4bc06ea705344aa9ac81d079485931b5bR166-R168): 

`Search by ID` results can look like either `Region Explorer` or `Source Inspector `results. So I needed to tweak the call to `FormWatcher's` `submitTo` function so that it can change based on user's selection. 

But for this to work, I needed to add `submitTo` as a dependency to the `useEffect` in the `useFormWatcher` hook. Let me know if that makes sense.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -154,7 +154,7 @@ export function EmbeddedPositionSearchPanel({
                     placeholder: 'Coordinates',
                     manageHiPS:false,
                 }} />}
-            {otherComponents}
+            {otherComponents && otherComponents(doGetConeAreaOp())}
```
robyww, line 157: You don’t need to change this file at all. The components are part of the same field group so they can use.

```js
const [getConeAreaOp, setConeAreaOp] = useFieldGroupValue(CONE_AREA_KEY);
```

src/firefly/js/templates/router/RouteHelper.jsx
loitly: https://github.com/Caltech-IPAC/firefly/blob/d91eb076a95139c9ce47a1e4fbcc5be5de0951b6/src/firefly/js/templates/router/RouteHelper.jsx#L60

`submitTo` can be a path(string) or a function that return a path.  So technically, you can achieve what you want to do without any changes.  However, if you want to add this, be aware that `FormWatcher` will re-render when a new `submitTo` is passed.

Also, I noticed this code is missing cleanup.  Please add `return (() => dispatchCancelActionWatcher(id));` at the end of the `useEffect`.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
kpuriIpac: @robyww if you get a chance before you leave (or on Wednesday) to take a look at this (the [IFE PR](https://github.com/IPAC-SW/irsa-ife/pull/326) is updated with latest code changes)...I've kept this code as is, as I was unable to use `useFieldGroupValue` in the `otherComponents` defined in `RegionExplorer.jsx`, as it's not a React Component. 
I may be missing an obvious trick here, but let me know if you figure out how to do this.

src/firefly/js/templates/router/RouteHelper.jsx
kpuriIpac: Added the cleanup for the `useEffect`... about the dependency, the re-render of `FormWatcher` is intentional and is what I was trying to achieve. 

Reason: I am  passing in `submitTo` as a string. But it still needs to change and re-render `FormWatcher` when the user 'changes' this string (via selecting `tile` vs `object` id). Because otherwise, since `FormWatcher` component is already called with previous value of `submitTo`, a re-render does not happen when the user changes their selection (since `submitTo` was not a dependency of the `useEffect`) and they end up going to the wrong path. Re-render makes sure it gets the latest value of `submitTo`...

Let me know if that sounds good?

robyww (approved):

PR #1570, opened 2024-06-06
Firefly-1498: fixed: polygon does not default on first click

robyww:
#### Testing
 - https://firefly-1498-polygon.irsakudev.ipac.caltech.edu/irsaviewer/dce
 - test according to ticket


kpuriIpac (commented):
The polygon bug is fixed, but if you switch back to cone after populating some polygon coordinates, and click on the background MOC, it does not populate the input for cone search now.

robyww:
> The polygon bug is fixed, but if you switch back to cone after populating some polygon coordinates, and click on the background MOC, it does not populate the input for cone search now.

OK, hopefully fixed now. I just made a new build.

PR #1574, opened 2024-06-12
Firefly 1494 column reset

loitly:
More changes here: https://github.com/IPAC-SW/irsa-ife/pull/332

Tests: 
https://firefly-1494-column-reset.irsakudev.ipac.caltech.edu/irsaviewer/
https://firefly-1494-column-reset.irsakudev.ipac.caltech.edu/irsaviewer/dce
https://firefly-1494-column-reset.irsakudev.ipac.caltech.edu/applications/finderchart/

**Updates: 06/20/2024**
Converted all `loading-mask` to JoyUI Skeleton.  Lots of files were touched.  Please help test.

The column states reset after a derived column is added or modified.
https://jira.ipac.caltech.edu/browse/FIREFLY-1494
- IRSA catalog search -> m31 -> Table Options -> hide all columns except for a couple -> Add Column
Hidden columns should stayed hidden.

FinderChart download related issues.  See ticket for details.
https://jira.ipac.caltech.edu/browse/FIREFLY-1502
Besides making sure what's in the ticket is fixed, 
- Disabled the 'Send to background' button when it is not yet available.
We set a default wait on backgroundable tasks so that they can return immediately if it can be completed fast.
- Extended Download Dialog mask to fully cover most of the dialog.  Not only does it looks better, it makes sense that those buttons should not be clickable, again.


src/firefly/js/core/background/BgMaskPanel.jsx
```diff
@@ -52,10 +48,10 @@ export const BgMaskPanel = React.memo(({componentKey, onMaskComplete, style={}})
 
     const msg = jobInfo?.errorSummary?.message || jobInfo?.jobInfo?.progressDesc || 'Working...';
 
-    const Options = () => ( (inProgress || true) &&
+    const Options = () => (
         <Stack direction='row' spacing={1}>
-            <Button variant='solid' color='primary' onClick={sendToBg}>Send to background</Button>
-            <Button onClick={abort}>Cancel</Button>
+            <Button variant='solid' color='primary' disabled={!jobInfo} onClick={sendToBg}>Send to background</Button>
```
jaladh-singhal, line 53: I think we need something other than this check for timing the buttons' disabling more accurately, because of the bug I mentioned above.

jaladh-singhal (approved):
1494: tested, works perfectly

1502: tested, background sending and download works correctly for the test cases mentioned in the ticket except for a few corner cases (?) in multiple positions search. To reproduce: do a search with file in the ticket, click prepare download -> send to background -> works. in multi objects table, select 3rd row (or some other row if can't reproduce), click prepare download -> click send to background 
1. it doesn't get sent to background and it just keeps working until it returns with the download. I wonder if it's due to the early enabling the disabled background  buttons?
2. also sometimes, it starts showing tooltip "select how to layout images" over background buttons which is likely coming from "Page layout" option below the mask.

(I should note, this bug happens inconsistently for some reason so try clicking akk other rows in multi object table one by one until you can reproduce it)

src/firefly/js/tables/ui/BasicTableView.jsx
```diff
@@ -328,20 +331,22 @@ function correctScrollLeftIfNeeded(totalColWidths, scrollLeft, width, triggeredB
...
 
-    const paddings = 8;
-    return maxVals.map((text, idx) => {
-        const header = columns[idx].label || columns[idx].name;
-        const style = header === text ? headerStyle : {fontSize:12};
-        text = text.replace(/[^a-zA-Z0-9]/g, 'O');    // some non-alphanum values can be very narrow.  use 'O' in place of them.
-        const pxNum =  getPxWidth({text, ...style}) + paddings;
-        return pxNum < minWidth ? minWidth : pxNum;
-    });
+const paddings = 8;
```
jaladh-singhal, line 340: is this used by anything other than `colWidthInPixel()`?

if not, it might be worth defining it inside the function or (even better) making it a parameter

robyww (approved):
The updated working `<Skeleton/>` panels look good and seem to work correctly.

PR #1576, opened 2024-06-21
Firefly-1506: Update buttons to material icons

robyww:
#### [Firefly-1506](https://jira.ipac.caltech.edu/browse/FIREFLY-1506): Update buttons to material icons

   - Update the filter and check button to be the same as images, supports dark mode
   - Changed order of filter and select to be the same as images

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1506-dark/firefly
- do a TAP search
- do a selection with the chart
- change to dark mode
- the buttons should show up and be the same as images


jaladh-singhal (approved):
Tested - looks good

PR #1590, opened 2024-07-18
FIREFLY-1509: Polish dark mode for ADQL entry screen

jaladh-singhal:
Fixes [FIREFLY-1509](https://jira.ipac.caltech.edu/browse/FIREFLY-1509)

- Used emotion CSS to plug in two separate theme-based CSS files provided by prismjs package
  - this required small changes in package.json and webpack config 
  - reading css as string looks bit hacky but was necessary because prism doesn't scope its different theme css files
- Cleanups:
  - Wrapped `<code>` with `<pre>` to follow HTML semantics for a code block and Prism recommends so 
  - Removed copy-pasted css for prism in `/externalSource` and using it directly from prismjs npm package
  - Removed copy-pasted prism-sql.js with slight modifications for ADQL functions, because we can access it from npm package and then modify it programmatically

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1509-dark-adql/firefly

Go to ADQL edit screen - check that changing theme, changes color-theme of code blocks and text area


robyww (commented):
it looks nice but I am not sure we should bring in another css package just to do this. I think I need to be convince and know if there are other options.

@jaladh-singhal maybe you, I and @loitly should meet about `emotion`

jaladh-singhal:
> it looks nice but I am not sure we should bring in another css package just to do this. I think I need to be convince and know if there are other options.
> 
> @jaladh-singhal maybe you, I and @loitly should meet about `emotion`

That's a fair point. I should have explained why I decided to use emotion. I wanted to use `primjs/themes/*.css` files already present in npm package (and will receive updates over time). But webpack config we have just loads all css files in <style> tags causing overriding of theme-based css and wasn't giving me any control to pick one over another based on mode in our app. So based on my research and some thinking, I came up with following options:
- dynamically create and remove the style tag for each theme-css file in a useEffect hook 
  - I think it's unclean and react suggests to not manipulate low-level DOM, so didn't decide to go that route
- copy paste both them-css files into one and scope them under light- and dark- classes 
  - this is kind-of what we were doing earlier (only for light theme though) but it won't recieve updates as prismjs package updates, unless someone goes and do the manual tinkering
- read both css files separately into some variable (hence raw-loader to read them as string) and then convert string css to styles understandable by joy-ui components.
  - I couldn't find a way to do so directly so I used emotion css which is already used by joy-ui/mui [under the hood](https://mui.com/system/getting-started/installation/). Because it can parse string to css styles that React understands but I'm happy to change it if there's a better way of parsing without using emotion.

src/firefly/js/ui/tap/AdvancedADQL.jsx
```diff
@@ -25,18 +26,25 @@ import {
...
 
-import '../../externalSource/prismLive/prism.css';
-import '../../externalSource/prismLive/prism-live.css';
+// 'primjs/themes/*.css' uses same classnames, so to avoid css overriding, we import light and dark theme css files
+// into 2 separate strings using `?raw` resourceQuery (see the corresponding rule defined in webpack.config.js)
+import prismLightCss from 'prismjs/themes/prism.css?raw';
+import prismDarkCss from 'prismjs/themes/prism-okaidia.css?raw';
+
+// 'prism-live.css' needs to be imported in same way as above otherwise specificity of CSS selectors will change
+import prismLiveCss from '../../externalSource/prismLive/prism-live.css?raw';
```
jaladh-singhal, line 44: I removed the inline webpack syntax for raw-loader (also removed from package.json). I noticed that webpack supports raw-loading in-built starting with v5 (which is what we're using).

src/firefly/js/ui/tap/AdvancedADQL.jsx
```diff
@@ -25,18 +26,25 @@ import {
...
+import 'blissfuljs'; // bliss is needed for prism-live
 import '../../externalSource/prismLive/prism-live.js';
 
-import '../../externalSource/prismLive/prism.css';
-import '../../externalSource/prismLive/prism-live.css';
+// 'primjs/themes/*.css' uses same classnames, so to avoid css overriding, we import light and dark theme css files
+// into 2 separate strings using `?raw` resourceQuery (see the corresponding rule defined in webpack.config.js)
+import prismLightCss from 'prismjs/themes/prism.css?raw';
+import prismDarkCss from 'prismjs/themes/prism-okaidia.css?raw';
+
```
robyww, line 42: This is more easily understandable.

buildScript/webpack.config.js
```diff
@@ -159,8 +159,13 @@ export default function makeWebpackConfig(config) {
                 inline: localBuild ? undefined : 'fallback',
             }
         },
+        {
+            resourceQuery: /raw/, //see https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax
+            type: 'asset/source',
+        },
         {
             test    : /\.css$/,
+            resourceQuery: { not: [/raw/] }, // to exclude raw css assets from being processed by other loaders
```
robyww, line 168: looks good. I am glad you dug this out.

PR #1591, opened 2024-07-19
FIREFLY-1517: use SQL to create decimate data

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1517

Optimizes the memory usage of `DecimationProcessor`. It should work the same as before with one key difference: the represented point (highlighted row) is now taken from the first row of an `SQL GROUP BY` statement instead of a random row determined by Java logic. This change should not impact functionality. Open to feedback.

This approach eliminates the need to load the entire table into JVM memory, which previously could consume gigabytes of memory. It also significantly speeds up processing for larger tables for the same reason.

Test: https://fireflydev.ipac.caltech.edu/firefly-1517-decimation/firefly/
Load bigger tables, at least 5000 rows to trigger decimation.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -1,4 +1,3 @@
-version: '3.4'
 services:
   firefly:
     image: ipac/firefly:${BUILD_TAG:-latest}  # value taken from shell, or latest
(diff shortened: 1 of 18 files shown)


robyww (commented):
I think it is good.

PR #1592, opened 2024-07-22
IRSA-6248: fix bug in HydraLanding that prevent it from passing in slotProps

loitly:
Fix used by https://github.com/IPAC-SW/irsa-ife/pull/344


wmiipac (approved):
Looks good!

PR #1606, opened 2024-08-14
FIREFLT-1535: Timeseries bug fix

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1535
- changes are very small, but this is my first time working with plotly, so took some time to figure out the workflow and the issue.
- So we currently update the chart when user applies one of 4 options (grid, top/right, reverse and log) using `Plotly.relayout`, and we send it a `layoutUpdate` (or `layoutDelta` object), which is just the difference of the current and old layout based on user selection instead of the entire layout context (which is the `layout` object)
   - from my research and understanding of this topic: `Plotly.relayout` is great for updates like grid, log, etc. but not for complex stuff like axes reversal. 
   - To fix this, I considered using `Plotly.react`, which re-renders the entire plot though. 
      - After some trial and error, I figured I could just send the entire `layout` context to `Plotly.relayout` instead. This fixes the axes reversal issue. And it is still theoretically better than doing a `Plotly.react` here, because even though I send the entire `layout` object and not just the `layoutDelta` / `layoutUpdate`, `Plotly.relayout` internally only attempts to update the properties that differ from the current layout. 
- Let me know your thoughts on this fix. 

**Testing**: 
Timeseries: https://firefly-1535-timeseries-bug.irsakudev.ipac.caltech.edu/irsaviewer/timeseries
- Test according to ticket (upload the file in the [ticket](https://jira.ipac.caltech.edu/browse/FIREFLY-1535) to timeseries)
- Then click on chart settings -> `Chart Options` 
   - Select and deselect `reverse` for both X and Y axes and click on `Apply` multiple times
      - the bug was that after applying reverse once or twice, it would stop working (the other options: `grid`, `top` and `log` would still work as expected). This should be fixed now. 
  - also select and deselect other options for both axes like `grid`, `log` and `right`/`top`. 
  - essentially, try and play around with the `Options` selections under `Chart Options`.

### DIFF
--- a/src/firefly/js/charts/ui/PlotlyWrapper.jsx
+++ b/src/firefly/js/charts/ui/PlotlyWrapper.jsx
@@ -280,7 +280,7 @@ export class PlotlyWrapper extends Component {
                         this.restyle(this.div, Plotly, data, dataUpdate, dataUpdateTraces);
                         break;
                     case RenderType.RELAYOUT:
-                        Plotly.relayout(this.div, layoutUpdate);
+                        Plotly.relayout(this.div, layout);
                         break;
                     case RenderType.RESTYLE_AND_RELAYOUT:
                         this.restyle(this.div, Plotly, data, dataUpdate, dataUpdateTraces);
@@ -308,7 +308,7 @@ export class PlotlyWrapper extends Component {
                             chart.on('plotly_relayout', () => this.showMask(false));
                             chart.on('plotly_restyle', () => this.showMask(false));
                             chart.on('plotly_redraw', () => this.showMask(false));
-                            chart.on('plotly_relayout', (changes) => this.syncLayout(chartId, changes));
+                            chart.on('plotly_relayout', (changes) => {this.syncLayout(chartId, changes);});
                         }
                         else {
                             this.showMask(false);
--- a/src/firefly/js/templates/lightcurve/LcViewer.jsx
+++ b/src/firefly/js/templates/lightcurve/LcViewer.jsx
@@ -73,7 +73,7 @@ export const LcViewer = memo(({menu, dropdownPanels=[], appTitle, slotProps, ...
 
     return (
         <App slotProps={mSlotProps}
-             dropdownPanels={[...dropdownPanels, <UploadPanel {...{fileLocation}}/>]}
+             dropdownPanels={[...dropdownPanels, <UploadPanel {...{fileLocation, name:'LCUpload'}}/>]}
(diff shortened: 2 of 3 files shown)


src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -280,9 +280,10 @@ export class PlotlyWrapper extends Component {
                         this.restyle(this.div, Plotly, data, dataUpdate, dataUpdateTraces);
                         break;
                     case RenderType.RELAYOUT:
-                        Plotly.relayout(this.div, layoutUpdate);
+                        Plotly.relayout(this.div, layout);
                         break;
                     case RenderType.RESTYLE_AND_RELAYOUT:
+                        //ToDo: Remove this? this is probably never called, since when RenderType is RESTYLE_AND_RELAYOUT, we set it to NEW_PLOT anyway
                         this.restyle(this.div, Plotly, data, dataUpdate, dataUpdateTraces);
```
loitly, line 287: No need for this comment.  It's probably not accurate.  This is only true when `useScatterGL` is false.

src/firefly/js/templates/lightcurve/LcViewer.jsx
```diff
@@ -241,10 +241,19 @@ export const UploadPanel = ({initArgs, name= 'LCUpload'}) =>{
...
 };
 
+UploadPanel.defaultProps = {
+    name: 'LCUpload'
+};
+
+// UploadPanel.propTypes = {
+//     name: PropTypes.oneOf(['LCUpload']),
+//     initArgs: PropTypes.object
+// };
```
loitly, line 255: `defaultProps` is deprecated.  Use default parameters instead.
Also, no need to add commented out code.  It's already in history.

src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -280,9 +280,10 @@ export class PlotlyWrapper extends Component {
...
                         break;
                     case RenderType.RELAYOUT:
-                        Plotly.relayout(this.div, layoutUpdate);
+                        Plotly.relayout(this.div, layout);
```
loitly, line 283: This should be fine.  `layoutUpdate` may no longer be necessary since plotly may be doing it internally.

loitly (approved):
Changes works.  Please address a few minor requests before merging.

PR #1608, opened 2024-08-15
IRSA-6084: changing x/y axis does not work when using column search popup

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-6084

When changing the x/y axis, the label does not update. This only happens when using the column search popup (magnifying glass icon).

Test: https://fireflydev.ipac.caltech.edu/dev/firefly/
- Load a chart
- Use column search popup to change x/y axis
- Make both label and values are updated.

### DIFF
--- a/src/firefly/js/charts/ui/options/SpectrumOptions.jsx
+++ b/src/firefly/js/charts/ui/options/SpectrumOptions.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {get, range, isEqual} from 'lodash';
+import {get, range, isEqual, memoize} from 'lodash';
 import {getSpectrumDM, REF_POS} from '../../../voAnalyzer/SpectrumDM.js';
 
 import {getChartData} from '../../ChartsCntlr.js';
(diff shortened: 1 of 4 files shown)


src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -310,7 +310,7 @@ export function spectrumInputs ({chartId, groupKey}) {
                 : <ValidationField fieldKey={SFOptionFieldKeys(activeTrace).value} initialState={{value: sfRefPos}} readonly={true} {...allProps}/>;
         },
     };
-}
+}, (props) => Object.values(props).join('|'));
```
jaladh-singhal, line 313: might be worth defining this repeated resolver function in BasicOptions and reusing it from there?

jaladh-singhal (approved):
The test build link was wrong but I made a build: https://fireflydev.ipac.caltech.edu/irsa-6084-chart-bugs/firefly

It works as expected!

PR #1612, opened 2024-08-28
Firefly-1547: fixed: color table not inverting

robyww:
#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1547-color/firefly
- load any images
- show the color dropdown
- it should look the same in dark mode or light mode


jaladh-singhal (approved):
Works as intended

PR #1614, opened 2024-08-30
FIREFLY-1521: a few chart related bugs

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1521
- Plot label not updating (already fixed)
- User-defined columns are not appearing as options in the chart options dialog
- Input fields in chart options dialog did not update after chart data has changed

Test:  https://fireflydev.ipac.caltech.edu/firefly-1521-chart-new-bugs/firefly
- Upload the sample test file in the ticket.
- Use the Chart Options dialog to change the x/y axis and ensure the chart data and axis labels are updated.
- Add a new column.  Make sure it is a numeric type.  Then, open the Chart Options dialog to verify that the new column is selectable.  
- Make additional updates to the chart, close the dialog, then reopen it to ensure the updates are reflected.
- This process should work with other tables as well.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -579,9 +579,9 @@ function updateChartData(chartId, traceNum, tablesource, action={}) {
     } else {
         if (!isFullyLoaded(tbl_id)) return;
         const tableModel = getTblById(tbl_id);
-        if (!getColValStats(tbl_id)) {
-            dispatchLoadTblStats(tableModel.request);
-        }
+
+        // may not need to call it as often.  if this becomes a performance issue, then optimize.
+        dispatchLoadTblStats(tableModel.request);
 
         const changes = getDataChangesForMappings({mappings, traceNum});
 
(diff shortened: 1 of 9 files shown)


src/firefly/js/charts/ChartUtil.js
```diff
@@ -579,9 +579,9 @@ function updateChartData(chartId, traceNum, tablesource, action={}) {
     } else {
         if (!isFullyLoaded(tbl_id)) return;
         const tableModel = getTblById(tbl_id);
-        if (!getColValStats(tbl_id)) {
-            dispatchLoadTblStats(tableModel.request);
-        }
+
+        // may not need to call it as often.  if this becomes a performance issue, then optimize.
+        dispatchLoadTblStats(tableModel.request);
```
jaladh-singhal, line 584: not sure why condition was removed. Do we want to recompute table statistics because new column was added?

jaladh-singhal (approved):
Tested as per instructions, works as intended!

Also replacing lodash `memoize` with React `useCallback` looks cleaner - I thought of it in your last PR but wasn't sure if it applies to our use case.

src/firefly/js/charts/ChartUtil.js
loitly: Yes, it's for when columns are added/removed. There is logic in dispatchLoadTblStats that determine when to load.

PR #1619, opened 2024-09-09
IRSA-6331: change logic to convert VOTable's description into IPAC table comments

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-6331
- split and wrap description so that it does not exceed 80 characters per line.

Radar: https://irsa-6331-invalid-ipactable-created.irsakudev.ipac.caltech.edu/applications/Radar/
Timeseries: https://irsa-6331-invalid-ipactable-created.irsakudev.ipac.caltech.edu/irsaviewer/timeseries

Test: 
- do a radar search on anything (i did "sv sag")
- find the NEOWISE exposure photometry and send to IRSA Viewer
- save data table from there.
- load time series tool and upload that copy.
- try to do periodogram.
- periodogram fails.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/IpacTableUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/table/IpacTableUtil.java
@@ -177,20 +177,37 @@ public class IpacTableUtil {
         if (attribs == null) return;
         // write attributes first
         for (DataGroup.Attribute kw : attribs) {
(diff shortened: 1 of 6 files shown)


wmiipac (approved):
Tested. The saved table now have fixed line length set at 80 for the header, and periodogram calculation works fine. 
When sent radar result to irsaviewer, in table panel click view table info and choose View full text from "...",  the popup shows correct table header.

PR #1620, opened 2024-09-18
FIREFLY-1556: clean up of wrapperStyle, labelWidth, and inline

robyww:
#### [FIREFLY-1556](https://jira.ipac.caltech.edu/browse/FIREFLY-1556): remove many references to wrapperStyle, labelWidth, or inline are no longer used. Remove from component creation and propTypes.

 - This should not change the behavior of code at all.
 - This work only cleans up the "low hanging fruit"
 - There are still some more complex uses that should be addressed in separate ticket

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1556-cleanup/firefly/
- Everything should work like before


kpuriIpac (approved):
Looks good. Thanks for the cleanup! Tested different searches across firefly, clicked around on various tools (reflecting the file changes). Everything works as expected.

PR #1622, opened 2024-09-20
Firefly-1541: web api 'hipsPanel' url can setup hips services and load multiple hips

robyww:
#### [Firefly-1541](https://jira.ipac.caltech.edu/browse/FIREFLY-1541): Load hips with `hipsPanel`

- The `api=hipsPanel' endpoint should be able to directly load a HiPS

#### Testing
-  https://fireflydev.ipac.caltech.edu/firefly-1541-hipspanel/firefly/?api=hipspanel
        -  try the last two examples
-  https://fireflydev.ipac.caltech.edu/firefly-1541-hipspanel/firefly/?api=hipspanel&uri=https://web.ipac.caltech.edu/staff/desai/test/PilotHiPS
- should load HiPS


loitly (approved):
I tested it, and it works. However, is there a reason why you wouldn't combine it with the existing `hips` command? That's how the other API commands are structured. You use `execute` to decide whether to display the results or simply populate the form. In this case, you could replace `showPanel` with `execute`, defaulting it to `true` to retain existing behavior.

robyww:
Ultimately I will remove the hips command and make the hips panel the final endpoint. Until then there needs to be two since irsa does not use the separate hips panel.

PR #1623, opened 2024-09-20
Firelfy-1558: Save user add TAP servers as a preference and show it in the list.

robyww:
#### Firefly-1558: Save user add TAP servers

 - saved via api
 - saved with "enter my URL" option
 - save as a preference
 - The option should remain through reloads until the user removes it.
 - improve bad server error handling

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1558-tap-service/firefly
- enter a alternate tap server (such as: https://irsadev.ipac.caltech.edu/TAP) 
- reload, it should still be in the list
- enter a bad one  (such as: https://abcdefg.edu/TAP) 
- The error handling has improved
- reload and they both should still be in the list. You can select them just like before.
- try out the `x` on the servers you entered

#### Clearing preferences
- _open_ developer tools
- _on top choose_: Application
- _on left choose_:  Local Storage -> the url
- _on right highlight_: `APP_PREFERENCES`
- click on the `X`

### DIFF
--- a/src/firefly/html/test/test-hips-mocs.html
+++ b/src/firefly/html/test/test-hips-mocs.html
@@ -217,7 +217,7 @@
     </script>
 </template>
 
-<template title="Load a MOC" style='height: 400px' class="tpl lg " >
+<template title="Load a MOC as Fill" style='height: 400px' class="tpl lg " >
     <div id="expected" >
         <div class="source-code indent-3">
             - load a HiPS
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/Counters.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/Counters.java
@@ -11,6 +11,7 @@ package edu.caltech.ipac.firefly.server;
 
 import edu.caltech.ipac.firefly.server.security.SsoAdapter;
 import edu.caltech.ipac.firefly.server.util.VersionUtil;
+import edu.caltech.ipac.firefly.server.visualize.VisContext;
 import edu.caltech.ipac.util.ComparisonUtil;
 import edu.caltech.ipac.util.FileUtil;
 import edu.caltech.ipac.util.StringUtils;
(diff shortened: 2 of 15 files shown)


lrebull:
wow, it is really nice! works great!

kpuriIpac (changes requested):
Looks great overall! 

A couple of issues I found:
- when you "X" a service, there's a console error (`serviceUrl` is set to `empty`/`null` causing this probably).
   - when fixing this, would it be easy enough to either keep the current TAP service selected? (let's say I'm on `CADC`, but go through the list and delete one of my added servers, it would be nice for me to stay on `CADC`). 
   - Or, if I'm on a newly added TAP service, and delete that one itself, then go back to the initial TAP service. 

- Second issue is that when I go to `Enter my URL`, I could press enter (even multiple times) to keep adding one of the default services to the list. and since these are default services, I can't delete them. See screenshot below: 
Screenshot 2024-09-20 at 5 24 55 PM
   - For this, I can suggest this fix, when adding a new service: 
      - Can check in `getTapServiceOptions` if this `serviceUrl` already exists as a default service we offer. If it does, don't add it.  
      - also, check the `userServices` array (line 576 in `TapSearchRootPanel.jsx`). If this `serviceUrl` already exists in this array, don't re-add it. (user could also press enter multiple times, which keeps adding this duplicated service to the list).
      - For both of these checks, I don't think there's any need to notify the user that we're not adding these, as It'll be expected behavior to not add any of the default services, as well as not duplicating the new services they add.

gpdf:
Followup to @kpuriIpac's comment above:

* We should probably make some attempt to decide how precisely to match the user's URL with the ones already in the list: do trailing '/'s matter?  "http:" vs. "https:"?  What if it's one of us trying to type it slightly differently to debug something?  I think on balance we should let the user add the URL unless it's spelled *exactly* like the one in the built-in list.

> I don't think there's any need to notify the user that we're not adding these, as It'll be expected behavior to not add any of the default services

Agreed, but in that case we *should* automatically jump the menu to the duplicated service, so that the user actually gets to the service they wanted to get to.

robyww:
> when fixing this, would it be easy enough to either keep the current TAP service selected?

Good Idea I will fix it.

> Second issue is that when I go to Enter my URL, I could press enter (even multiple times) to keep adding one of the default services to the list. and since these are default services, I can't delete them. See screenshot below:

Yes, I need to fix that. I was actually planning to check for that but forgot!

> We should probably make some attempt to decide how precisely to match the user's URL with the ones already in the list: do trailing '/'s matter? "http:" vs. "https:"? What if it's one of us trying to type it slightly differently to debug something? I think on balance we should let the user add the URL unless it's spelled exactly like the one in the built-in list.

I will see what I can do

robyww:
I have rebuilt the PR with the updates from the feedback.

kpuriIpac (approved):
The issues have been addressed. This works great now! No bugs that I can think of. 
Also, @gpdf's suggestion works quite well:
- for instance, if user starts entering a URL while on a default TAP service, like: `https://irsa.ipac.caltech.edu/TAP`, and they make a one char change anywhere and press enter, this is now saved as a new preference. But since it's not a default TAP service, it's deletable.

PR #1624, opened 2024-09-20
FIREFLY-1553: Upload table chooser popup UI bug fixes

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1553
- The changes are very small, but they fix some existing issues mentioned in the ticket, a quick summary:   
  - When you upload a `moc fits` file via the `UploadTableChooser` popup (Multi-object search), the `Load/Cancel` buttons were not visible unless you resized the dialog to make it bigger. 
  - Also, the `Details` model table on the right was either not visible or just barely visible in the `UploadTableChooser` popup. 
  - This PR fixes both issues. 
- **_Note_**: The styling is applied directly within `FileUploadViewPanel`, so this affects the `Upload` tab as well. I chose to do it this way because it makes the `Upload` tab cleaner as well. 
   - it makes the summary model (the table on the left when you upload a file like `moc fits` file) narrower.
   - since we use is a SplitPane, the user can still make either table narrower or wider.
   - if we want to apply some specific styling to the `FileUploadDropdown` / `FileUploadViewPanel` when called from `UploadTableChooser` instead, perhaps slotProps approach can be used, but I didn't find that necessary in this case. 

**Testing**: 
- **firefly**: https://fireflydev.ipac.caltech.edu/firefly-1553-upload-popup-bug/firefly
- use multi object upload from the TAP panel to test the UploadTableChooser popup 
    - upload a non moc fits file, observe the summary / details tables. 
    - upload a moc fits file. Make sure you see both the Summary and Details model tables, and that you can resize either. 
       - make sure you also see the `Load/Cancel` buttons. 
- Repeat the steps above on the Upload tab directly. 
  - make sure it looks clean, you can resize both tables when there's summary table on the left.

### DIFF
--- a/src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
+++ b/src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
@@ -398,7 +398,7 @@ function makeSummaryModel(report, summaryTblId, acceptList) {
     const columns = [
         {name: 'Index', type: 'int', desc: 'Extension Index'},
         {name: 'Type', type: 'char', desc: 'Data Type'},
-        {name: 'Description', type: 'char', desc: 'Extension Description', width: 60},
+        {name: 'Description', type: 'char', desc: 'Extension Description', width: 30},
         {name: 'AllowedType', type: 'boolean', desc: 'Type in AcceptList', visibility: 'hidden'}
     ];
     const {parts=[]} = report;
@@ -619,7 +619,7 @@ function AnalysisTable({summaryModel, detailsModel, report, isMoc, UNKNOWN_FORMA
 
     // Details table needs to render first to create a stub to collect data when Summary table is loaded.
     return (
-        <Stack {...{direction:'row', flexGrow:1, position:'relative', minHeight:'22rem'}}>
+        <Stack {...{direction:'row', flexGrow:1, position:'relative', minHeight:'1rem'}}>
             {(summaryModel.tableData.data.length>1) ?
                 <MultiDataSet summaryModel={summaryModel} detailsModel={detailsModel} isMoc={isMoc}
                               acceptOneItem={acceptOneItem}/> :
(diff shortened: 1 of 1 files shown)


robyww (approved):
Looks good.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -619,7 +619,7 @@ function AnalysisTable({summaryModel, detailsModel, report, isMoc, UNKNOWN_FORMA
 
     // Details table needs to render first to create a stub to collect data when Summary table is loaded.
     return (
-        <Stack {...{direction:'row', flexGrow:1, position:'relative', minHeight:'22rem'}}>
+        <Stack {...{direction:'row', flexGrow:1, position:'relative', minHeight:'14rem'}}>
```
loitly, line 622: This is the change that matters the most.  It's what pushing the 'Load' button down when there's not enough vertical space.  I would make this smaller.  It does not need to be big because it will resize based on the browser's size.

loitly (approved):
I commented on what I think matters most. The other changes are, in my opinion, debatable. You've adjusted to make it look nicer on smaller screens but less so on larger ones. I don't really have a solution for that because of the fixed pixel size the library requires.

kpuriIpac:
> I commented on what I think matters most. The other changes are, in my opinion, debatable. You've adjusted to make it look nicer on smaller screens but less so on larger ones. I don't really have a solution for that because of the fixed pixel size the library requires.

Yes I agree. Can rethink this in the future I suppose, to make it cleaner for bigger screens as well. But yeah the fixed size requirement is a limitation for that.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
kpuriIpac: good point, done. set the `minHeight` to `1rem`. (may be too small, but like you said it resizes and I tested it and looks ok). Test build is updated if you want to have a look. https://fireflydev.ipac.caltech.edu/firefly-1553-upload-popup-bug/firefly

PR #1629, opened 2024-09-26
Firefly-1545: charts better handle large selections

robyww:
#### [Firefly-1545](https://jira.ipac.caltech.edu/browse/FIREFLY-1545): charts better handle large selections

 - charts now give warning when selection over 30000 and will not show the selection

#### Testing
 - https://fireflydev.ipac.caltech.edu/firefly-1545-charts/firefly
 - Load a big file (over 30000 rows)
 - it will show a heatmap so make a scatter chart
 - Select a large area and click check icon
 - A warning dialog will appear only the first time
 - The toolbar will show a warning icon when scatter selection is over 30000

PR #1631, opened 2024-09-27
Firefly-1545: charts better handle large selections (alternative)

loitly:
Alternative solution to PR https://github.com/Caltech-IPAC/firefly/pull/1629

[Firefly-1545](https://jira.ipac.caltech.edu/browse/FIREFLY-1545): charts better handle large selections

The first commit addresses two issues:
- Lodash's `_.isEqual` is slow when comparing large sets or arrays.
- Firefly should skip unnecessary selection updates.

These changes enable Firefly to handle scatter plots with up to 1 million rows, instead of crashing at a much smaller datasets. 
However, performance is noticeably slow beyond this point.  It turns out updating the highlighted row rather than selection updates that's slow.

The second commit attempt to optimizes chart highlighting:
- Reduced code execution time from 160ms to around 60ms.
- Unfortunately, the remaining execution time is due to Plotly.js, which I suggest we try  to optimize at a later time.

Test: https://fireflydev.ipac.caltech.edu/firefly-1545-charts-alternative/firefly/
- Upload a large table (over 50k rows).
- You should see coverage, chart(heat map), and table displayed.
- Test table selection; there should be no performance issues.
  - When the chart is not displaying a scatter plot, the table size has no impact on performance.
- Go to chart options -> Overplot New Trace -> select plot type = Scatter, set X and Y axes, then click OK.
- Now test both chart selection and table selection.
  - Performance begins to degrade at 500k rows.
  - It worsens significantly at 1 million rows.
  - 
We might need to consider limiting the size of scatter plots we support.

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -44,6 +44,7 @@ import {getColValidator} from './ui/ColumnOrExpression.jsx';
 export const DEFAULT_ALPHA = 0.5;
 
 export const SCATTER = 'scatter';
+export const SCATTERGL = 'scattergl';
 export const HEATMAP = 'heatmap';
 
 export const SELECTED_COLOR = 'rgba(255, 200, 0, 1)';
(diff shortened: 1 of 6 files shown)


robyww (approved):
This looks good. I think we are probably going to get to the point where we have to disable scatter selection (like the original PR) when it gets to a certain size before we really need to disable scatter charts. We will see in the future…

loitly:
> This looks good. I think we are probably going to get to the point where we have to disable scatter selection (like the original PR) when it gets to a certain size before we really need to disable scatter charts. We will see in the future…

Just to clarify, the delay comes from updating the chart, which is primarily handled by PlotlyJS. Several things trigger chart updates, including selection and highlighting. To avoid this, you would need to disable both features.

I think there's potential for performance improvements if we explore the Plotly JS API further. I've noticed that rendering is really fast, but the majority of the time is spent on JS execution.

PR #1634, opened 2024-10-04
FIREFLY-1562: scrollbar appears when there are more tabs than can fit within the available space

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1562
- Add scrollbarWidth='none' to ensure it's not visible even in firefox
- Decrease min-width to accommodate more tabs; setting it slightly bigger for removable tabs
- Introduce more advance logic to calculate the optimal tab size

Tests: 
https://firefly-1562-scrollbar-over-tab.irsakudev.ipac.caltech.edu/applications/sofia/
https://firefly-1562-scrollbar-over-tab.irsakudev.ipac.caltech.edu/irsaviewer/

In both IrsaViewer and Sofia, add multiple tabs to the results view to observe how it performs when there are more tabs than can fit. 

Also, even if the scrollbar is not visible, you can still scroll through the tabs using swipe gestures or by holding the Shift key while scrolling with the mouse.

### DIFF
--- a/src/firefly/js/ui/panel/TabPanel.jsx
+++ b/src/firefly/js/ui/panel/TabPanel.jsx
@@ -12,7 +12,7 @@ import {
 } from '@mui/joy';
 import {tabClasses} from '@mui/joy/Tab';
 import sizeMe from 'react-sizeme';
-import {omit, isString, pick} from 'lodash';
+import {omit, isString} from 'lodash';
 
 import {dispatchComponentStateChange, getComponentState} from '../../core/ComponentCntlr.js';
 import {useFieldGroupConnector} from '../FieldGroupConnector.jsx';
@@ -252,8 +252,8 @@ function TabHeader({children, slotProps}) {
     useEffect(() => {
         const tbEl = toolbarEl.current;
         if (tbEl) {
-            const width = tbEl.getBoundingClientRect().width;
-            setMaxTitleWidth( (width/tabHeaders.length) -4 );   // -4 account for margin and padding
+            const maxWidth = calcOptimalTabSize(tbEl);
+            setMaxTitleWidth(maxWidth);
         }
     }, [toolbarEl?.current?.getBoundingClientRect()?.width]);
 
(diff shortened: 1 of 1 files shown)


src/firefly/js/ui/panel/TabPanel.jsx
```diff
@@ -412,5 +413,16 @@ function getAllTabId(children) {
...
+    const availW = tabEl.getBoundingClientRect().width;
+    const maxW = availW/tabAry.length;
+    const relTabs = tabAry.filter((t) => t.offsetWidth > maxW);
+    if (relTabs.length === 0) return maxW;      // all tabs fit, no adjustment needed.
+
+    // these don't need adjusting. so, ignore them in calculation
+    const ignTabsW = tabAry.map((t) => t.offsetWidth)
+                           .filter((w) => w <= maxW).reduce((sum, w) => sum + w, 0);
+    const adjAvailWidth = availW - ignTabsW;
+    return adjAvailWidth/relTabs.length;
```
robyww, line 427: I am not surprise you had to write this. It was coming.

robyww (approved):
Looks good!

lrebull:
hm. ok, i'm trying to get to a place where i can need the scroll bar. I asked for a ton of results from SOFIA.

(1) i am getting a lot of timeout errors (which @wmiipac and Angela know about -- the ones that look like Unable to parse VOTABLE from [...]Content is not allowed in prolog. Cause: [IOException] Unable to parse VOTABLE from [...]

(2) the tab names are so small that i keep accidentally hitting the tab color swatches, but I'm not sure there's much to be done here -- there is still space where i can hit the name and not the color swatch; i just have to be careful about where i point the mouse.

(3) for me, the shift-mouse scroll and/or mouse swipe tries to go backwards/forwards in the browser history (or even desktops); it doesn't go between tabs in Firefly for me. 

I see from the description here that I should not expect the scroll bar to show up, so that it won't ever block the tab names (which was my original complaint). In terms of tab navigation, here is always a tab navigation pulldown, and I can always resize the windows to make the tables take up the whole window if I need to. So I think the original purpose of the ticket has been met.

We do need to solve the timeouts, but that's beyond the scope of this particular ticket.

wmiipac:
@lrebull I will make a ticket for SOFIA timeout issue

PR #1637, opened 2024-10-07
FIREFLY-1565: Copy filters makes screen blink

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1565
- fix clear `Current Constraints` stopped working

Test: https://fireflydev.ipac.caltech.edu/firefly-1565-copy-filter-refresh-screen/firefly
- Load a table
- Apply a column filter
- Bring up Table Option, then `Advanced Filter`
- Verify that 'Copy to clipboard' works and that it does not reload the the table(blink).
- Verify `Current Constraints: Clear` works.

### DIFF
--- a/docs/next-release-details.md
+++ b/docs/next-release-details.md
@@ -7,14 +7,15 @@
 
 ### _Notes_
 #### This is probably one of new most new feature pack releases of firefly in several years. It also includes a many, many bug fixes, clean up and optimization (not all list below). 
-
+                                                                                                           `
 
 #### Major Features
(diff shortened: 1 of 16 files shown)


kpuriIpac:
@loitly I think you forgot to add the test build link above (I see you built a firefly build, but I think we need an irsaviewer to test timeseries). Should the PR also be against `rc-2024.3` as opposed to `dev`?

lrebull:
this works!

PR #1638, opened 2024-10-08
FIREFLY-1567: Fix upload popup style for HiPS MOC dialog

kpuriIpac:
**Ticket**: [FIREFLY-1567](https://jira.ipac.caltech.edu/browse/FIREFLY-1567)
- fix for a regression bug, caused by my previous PR: https://github.com/Caltech-IPAC/firefly/pull/1624
- that PR didn't take into account the HiPS `Add MOC Layer` upload popup calling `FileUploadViewPanel` directly

**Testing**
Firefly: https://fireflydev.ipac.caltech.edu/firefly-1567-upload-bug/firefly
- do any TAP search (for instance `m81` on `IRSA` & `wise_allwise`)
  - In `Coverage`, under the `HiPS / FITS / MOC` dropdown, click on `Add MOC Layer`
  - go to the `Use my MOC` tab. 
  - Upload a MOC fits file here. You should be able to see the details clearly (as opposed to it being squished, like in the screenshot below). Try this in both a small and large browser size.  
     - here's a screenshot of what it looked like before: 
Screenshot 2024-10-06 at 1 38 17 PM

### DIFF
--- a/src/firefly/js/ui/HiPSImageSelect.jsx
+++ b/src/firefly/js/ui/HiPSImageSelect.jsx
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {Sheet, Stack, Typography} from '@mui/joy';
+import {Box, Sheet, Stack, Typography} from '@mui/joy';
 import React, {useContext, useEffect} from 'react';
 import {object, string} from 'prop-types';
 import {validateUrl} from '../util/Validate.js';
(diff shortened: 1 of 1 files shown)


loitly (approved):
I worked, to some extend.  At least it expands the table summary, which is what this ticket is about.  So, I would say this ticket is fixed.

But, the whole popup is not quite right.  
- It resizes when you switch tabs, switch `where from` radio button, then resize again when you load the meta data.  But, if you resize the popup once, it will keep that size.  Why not start up with a reasonable size and keep it that way instead of flickering?
- At the moment, if you decrease the dialog height, just a little, the action buttons are lost with no way to scroll.  How will they 'Add Mod' or 'Cancel'?

lrebull:
works!

kpuriIpac:
@loitly you're right. There are some existing problems using `FileUploadViewPanel` in a popup in general. 

- For instance, for the first one that you pointed out. try out `Multi-object` in TAP `Spatial Search`. 
   - If you switch to `Upload from URL` and back, the popup is resized. But if you resize it manually (make it smaller), then that size stays. And now in `Upload from URL` mode you can't even see the `Upload` button, unless you make the popup bigger again. At the very least, I think we should make this horizontally scrollable to avoid this. 
   - I'll try and work on a fix, but may need to be a new ticket. Will address how `FileUploadViewPanel` works in general when used within a popup. 
- I've addressed the second point, set a min height so you can see the `Add MOC` and `Cancel` buttons always. And that works for both the `Search` tab and the `Use My MOC` tab. 
   - Updated test build: https://fireflydev.ipac.caltech.edu/firefly-1567-upload-bug/firefly/ 
   - I think this may be good enough for `rc` for now, and the first point I can make a ticket for and continue working on. Let me know what you think.

loitly:
Changes made it looks much better.  Thank you.

PR #1644, opened 2024-10-14
FIREFLY-1492: Timeseries Period Finder Layout Issues

kpuriIpac:
**Ticket**: [FIREFLY-1492](https://jira.ipac.caltech.edu/browse/FIREFLY-1492)
- see screenshot in ticket for a quick summary existing layout issues
   - a lot of these issues occur due to resizing browser size, if you opened the browser with a small enough size and didn't resize, you may not even notice these 
- There are 4 major components on the page, tried to make them equally spaced out 
- The `LcPFOptionsBox` (the box on the top left) looked awkward when browser size is large enough, leaving whitespace between itself and the table below
  - I essentially used `flexGrow: 1` here to make it expand into and take up the whitespace
  - This may look a little awkward, but I tried to make the component internally spaced out as well. Open to better suggestions. 
- Made the top-right chart expand to the right (to use up the whitespace) and also expand/contract when you resize the browser size (earlier this chart would only grow/shrink in size vertically, not horizontally). 
- When you zoom into the top right chart, the increase in precision leads to increasing number of digits (decimals) for the y-axis, until you eventually don't see the beginning of these numbers 
  - this also affects our plotly chars in general, not just here in the `Period Finder...`
  - we could have a new ticket to address this

**Testing**: 
- **Timeseries**:  https://firefly-1492-timeseries-layout-bug.irsakudev.ipac.caltech.edu/irsaviewer/timeseries
- Upload a file, then click on `Period Finder...`
- Before clicking on `Calculate Periodogram`, notice the `LcPFOptionsBox` on the left and the chart on the right are the same height, make the browser size bigger/smaller. 
- Click on `Calculate Periodogram`
- As you make the browser size smaller, the `LcPFOptionsBox` (top left) will become scrollable (this is pre-existing behavior), but as you make it bigger, it'll keep expanding to take up the whitespace. 
  -  the top right chart also expands & shrinks vertically and horizontally as you resize the browser window 
- Also play around with the divider (SplitContent) both horizontally between the 2 top components and the 2 bottom ones, and also vertically (between the 2 left and 2 right components), resizing this way to ensure there are no issues with the layout. 
- Essentially, be sure to test with different browser sizes, the layout should remain consistent throughout.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LCPanels.css
+++ b/src/firefly/js/templates/lightcurve/LCPanels.css
@@ -34,13 +34,6 @@
     cursor: auto;
 }
 
-.phaseFolded {
-    display: inline-flex;
-    width: 100%;
-    height: 100%;
-}
-
-
 .settingBox {
     margin-bottom: 3px;
     padding: 5px 10px 10px 5px;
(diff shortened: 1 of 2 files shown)


robyww (approved):
Looks good.

PR #1645, opened 2024-10-14
FIREFLY-1273: Table Link tag incorrectly encode HREF path separator

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1273

According to the ticket, Firefly incorrectly encodes the path portion of the URL when the table data contains path separators that are used to generate a link. This PR addresses the issue by analyzing the final generated URL and encoding the entire URL as needed, rather than relying on pre-encoded URL components.

Tests: https://fireflydev.ipac.caltech.edu/firefly-1273-encoding-href/firefly/test/tests-table.html?test=Column+links+via+API
- Added 2nd row of data to demonstrate this use case.
- Also, wrote unit tests to show other scenarios


robyww (approved):
Looks good.

Does this need to go into the rc and the release?

PR #1646, opened 2024-10-14
Fits reading not aware of reading gz file: should keep a local uncompressed version

robyww:
#### Testing
- https://firefly-1577-gz.irsakudev.ipac.caltech.edu/applications/spherex
- try cutouts, the wavelengths readout should work.

#### Note
After testing with sphereX and merging. I plan to cherry pick into `rc-2024.3` it should go in both branches but need to be tried on dev first to test with sphereX.


src/firefly/java/edu/caltech/ipac/util/FileUtil.java
```diff
@@ -71,7 +71,7 @@ public class FileUtil
     public static final long K            = 1024;
 
     private static String[] CHARSETS_TO_BE_TESTED = {"ISO-8859-1","windows-1252", "windows-1253","UTF-8", "UTF-16",};
-    public static final int BUFFER_SIZE = (int) (32 * FileUtil.K);
+    public static final int BUFFER_SIZE = (int) (16 * FileUtil.MEG);
```
loitly, line 74: The buffer size you're suggesting is MUCH larger than the 8 KB default which is typically sufficient for many use cases. From the testing I did a while back, increasing the size beyond a certain point showed no performance gain. 
Since this is used throughout, I'd recommend testing to confirm if the larger buffer actually improves performance before committing to it.
Or, isolate it to the piece of code where you see this large buffer size improved performance.

robyww: Seems on modern machines 8KB is pathetic. I don't understand why it can take advantage of a bigger buffer.

However, I can change it back since you test show no improvement.

jaladh-singhal (approved):
Tested, works as expected! I see you added the zip handling in code which was missing earlier.

PR #1652, opened 2024-10-17
FIREFLY-1575: Simbad query returns duplicate columns

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1575
- The column names only differ in terms of case sensitivity (uppercase vs. lowercase).
- Add support for VOTable and Parquet formats.
- For other formats, duplicate column names will have '_idx' appended to ensure uniqueness.

Solution:
When a duplicate column is detected, we will append an index to the column name (e.g., col_a becomes col_a_1). 
The original name (col_a) will be saved as `column.label`, which will be displayed in the UI. If the column's `ID` doesn't already exist, we will assign the modified name (col_a_1) as the column's ID to indicate the uniqueness of the column.
These information will be used to accurately write back the file in its original format.

Test: https://fireflydev.ipac.caltech.edu/firefly-1575-simbad-dulicate-columns/firefly/
- Follow the steps in the ticket to ensure it is working correctly now.

For regression testing, I've attached two sample files to the ticket, simbad.vot and dup-cols.csv, to demonstrate the behavior of this change.

**CSV:**
- Upload the dup-cols.csv file.
- Observe that the second column name is appended with _1.

**VOTable:**
- Upload simbad.vot (a saved result from a Simbad query).
- In the file summary, filter the `name` column by typing 'filter' to display only column names containing 'filter.'
- You'll notice both `FILTER_NAME_U` and `FILTER_NAME_u` appear.
- Internally, `FILTER_NAME_u` is represented as `FILTER_NAME_u_1`, but it is displayed normally to the user.

The internal name is only relevant in the `Advanced Filter`:
- Load the simbad.vot table.
- Go to Table Settings (gear icon) -> Advanced Filter.
- On the left pane, click `FILTER_NAME_u`, and note that the internal name `FILTER_NAME_u_1` is used.

You can also test saving this table in VOTable or Parquet format to verify that the file is written back accurately.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/DuckDbReadable.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/DuckDbReadable.java
@@ -26,6 +26,9 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static edu.caltech.ipac.table.TableUtil.getAliasName;
 
 /**
  * @author loi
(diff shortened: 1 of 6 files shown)


robyww (approved):
Looks good. The tap simbad search then upload to wise works as well.

PR #1653, opened 2024-10-17
Firefly-1581: recognize cutout default from service

robyww:
#### Testing
- https://firefly-1581-cutout-def.irsakudev.ipac.caltech.edu/applications/spherex
- do search
- cutout default should be .5 degree (30 arcmin) which come from the server


jaladh-singhal (approved):
Tested, gets 30' from service, and remembers cutout size if changed by UI.

Code looks good too

PR #1655, opened 2024-10-18
IRSA-6427: encodeUrlString() returning bad hash segment

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-6427


robyww (approved):
Looks good.

PR #1661, opened 2024-10-23
IRSA-6431: Firefly tables in expanded mode was unable to switch tabs.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-6431

When multiple `tbl_groups` were used, tables in the second group were unable to switch tabs in expanded mode.

Test: https://fireflydev.ipac.caltech.edu/irsa-6431-switch-tabs-expanded-mode/firefly/test/tests-table.html?test=Custom+Table+Title
- Expand the second table tabs
- Ensure you are able to switch tabs
  - The drop-down Open Tabs selector should also work

@cwang2016 It would be nice if you can try and test this version to see that it actually fixes the bug you found.


robyww (approved):
Looks good.

PR #1667, opened 2024-11-07
FIREFLY-1592: Directly import IPAC table data into the database

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1592
See ticket for details.

In the past, handling large tables required more memory. This PR should improve that.
Test: https://fireflydev.ipac.caltech.edu/firefly-1592-ipactable-direct-ingest/firefly/
- Upload a large IPAC table

If you need a test file, you can use this one: a 3 million row, 25 column, 1.2GB IPAC table file. https://irsadev.ipac.caltech.edu:9201/test/3mil.tbl  (Upload via URL)

From DevTools, I see that my PR build takes 21.4s, compared to 28.2s for nightly. On the nightly build, it seems to crash (or possibly fails to load) on hips coverage, as it sometimes doesn't show up. Both PR and nightly are running with 4GB of memory.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/DbDataIngestor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/DbDataIngestor.java
@@ -5,17 +5,21 @@
 package edu.caltech.ipac.firefly.server.db;
 
 import edu.caltech.ipac.firefly.data.FileInfo;
+import edu.caltech.ipac.firefly.data.TableServerRequest;
 import edu.caltech.ipac.firefly.server.query.DataAccessException;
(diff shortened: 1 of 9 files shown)


kpuriIpac:
My tests results: 
- I ran both your test build and nightly (I used irsaviewer upload) a few times, disabled cache, and here's the avg times I got: 
   - PR test build: ~23 sec
   - Nightly Irsaviewer: ~30 sec
- So definitely an improvement! Taking a brief look at the code now, but other than that this looks good.

kpuriIpac (approved):
I skimmed through the code, got the gist of it (but don't understand it deeply enough). Code changes in addition to the testing and speed improvement, overall looks good!

PR #1668, opened 2024-11-08
FIREFLY-1607: Server-side implementation of Healpix index

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1607
- create deg2pix UDF
- create HealpixProcessor
- create HealpixProcessorTest
- enable json error log for better duckdb error message


robyww (approved):
Well, I am really impressed. Implementing the healpix pixel computation as a sql function!

I have no idea how any of it works or how you are thinking we will use it; so we need to talk and you walk me though your ideas.

I am guessing somewhere you return healpix index and counts at some level… I don’t know if there is a place that takes a query for a set of indexes.

Anyway we need to talk about it when I get back.  I have some specific ideas what the calls should look like. 

I am fine if you want to merge or you can wait until I get back. After we talk you might need to do a second PR.

PR #1670, opened 2024-11-18
FIREFLY-1615: Improve file type detection

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1615
- upgrade duckdb to latest version 1.1.3
- install magic extension for file type detection
- move TableUtil.Format to FormatUtil.Format
- added detectFormat to FormatUtil.
- deprecate TableUtil.guessFormat; use FileUtil.detectFormat instead
- add FormatUtilTest

Although many files were modified, most changes involved moving Format from TableUtil to FormatUtil. This should not pose significant issues.

From ticket:

> The logic is to first use the file's magic number to determine the format. If that fails, a small sample of the file content is analyzed to detect known file types. Finally, trial and error is applied to selected formats based on the file's MIME type.  This logic does not rely on the file extension. 

Test: https://fireflydev.ipac.caltech.edu/firefly-1615-file-type-detection/firefly/
Upload various file types and verify that Firefly functions as expected.

Also, ticket https://jira.ipac.caltech.edu/browse/FIREFLY-1628
My guess is that the server crashes unexpectedly which causes unreproducible errors.
This PR upgrade DuckDB to the latest version which seems to address some of the crashes.

Test: https://firefly-1615-file-type-detection.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA/
Compare to: https://irsatest.ipac.caltech.edu/applications/Spitzer/SHA/

Use SHA IRS Enhanced Products for testing.
Position = ngc 1333
select `IRS Enhanced Products`
-> Search
Switch to `IRS Enhanced Products` and repeatedly select different rows. Each row fetches a new table, creating a new database. For each order, a new temporary table is also created.
Compare to this, irsatest was crashing after several tries. You can tell through error messages or by the application becoming unresponsive.


src/firefly/java/edu/caltech/ipac/util/FileUtil.java
```diff
@@ -25,264 +34,279 @@
...
-    if (i > 0 &&  i < s.length() - 1) 
-    {
-      ext = s.substring(i+1).toLowerCase();
+    public static String getExtension(String s) {
+        String ext = "";
+        int i = s.lastIndexOf('.');
+        if (i > 0 && i < s.length() - 1) {
+            ext = s.substring(i + 1).toLowerCase();
+        }
+        return ext;
```
robyww, line 123: Did you mean to reformat all of `FileUtil.java`?

src/firefly/java/edu/caltech/ipac/util/FileUtil.java
```diff
@@ -1013,10 +1012,107 @@ private static URL getThisClassURL() {
...
+
+    /**
+     * Detects the format of the given file
+     *
+     * @param fname The full path of the file to detect the format for.
+     * @return The detected {@code Format} of the file, or {@code UNKNOWN} if the format could not be determined.
+     * @throws IOException If an I/O error occurs while accessing the file.
+     */
+    @Nonnull
+    public static Format detectFormat(String fname) throws IOException {
```
robyww, line 1039: I am not sure this `detectFormat` function. It probably needs to need in its own file. It make FileUtil.java have a lot of external dependencies. Before it was only dependent on java.

src/firefly/java/edu/caltech/ipac/util/FileUtil.java
```diff
@@ -1013,10 +1012,107 @@ private static URL getThisClassURL() {
...
+        StopWatch.getInstance().start("detectFormat");
+        String mime = getMimeType(fname);
+        DuckDbAdapter.MimeDesc mimeDesc = null;
+        Format format = mapToFormat(mime, null);
+        if (format == null) {
```
robyww, line 1044: It appears that if it can get the format from the file extension then it does not look inside the file like it use to. I think that is dangerous. I think our own server code somethings save stuff with the wrong extension. Because code that was intended  for a specific purpose got generalized.

src/firefly/java/edu/caltech/ipac/util/StringUtils.java
```diff
@@ -750,6 +750,21 @@ public static <R> R getSafe(FuncWithEx<R> func, R defaultVal) {
...
     }
 
+    /**
+     * A supplier that throws exception
+     */
+    public interface TryWithEx {
+        void run() throws Exception;
+    }
+
+    public static void tryIt(TryWithEx func) {
```
robyww, line 760: This function name is confusing. I studied it and understand now but `tryIt()` seems to vague for a general string function.

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/HealthCheck.java
```diff
@@ -29,6 +30,8 @@ protected void processRequest(HttpServletRequest req, HttpServletResponse res) t
             // memory check
             int mem = Math.max(Math.min(StringUtils.getInt(req.getParameter("mem"), 5), 1024), 1);
             alloc = new byte[mem * 1024 * 1024];
+            // database check
+            DbAdapter.getAdapter().execQuery("select 1", null);
```
robyww, line 34: This makes sense to add.

robyww (commented):
The code looks fine. My big concern, and I might not be understanding the code, is that you are relying on the file extension for type detection. I don't think we did that before.

src/firefly/java/edu/caltech/ipac/util/FileUtil.java
loitly: Yes, the file was poorly formatted, and I had intended to clean up FileUtil. However, based on your suggestion to move my logic elsewhere, I’ll go ahead and revert the entire file.

src/firefly/java/edu/caltech/ipac/util/FileUtil.java
loitly: Tika first checks the magic number and then the file extension. DuckDB's extension, similar to the file command, relies solely on the magic number but can reliably identify Parquet files.

The logic here is to use these two tools to determine the file format. If they fail, we fall back to trial and error based on the resolved MIME type. For example, if the file is identified as XML, we can attempt to process it using a VOTable reader.  I’ve implemented handling for common cases and expect to add more as needed.

This approach works well in most cases. However, it fails when a file lacks a magic number but has a misleading file extension—for instance, a .fits file that is actually an IPAC table.

Do we need to handle this edge case?

src/firefly/java/edu/caltech/ipac/util/StringUtils.java
loitly: I will remove.  I plan to introduce more functional helpers gradually in a separate util class.

src/firefly/java/edu/caltech/ipac/util/FileUtil.java
robyww: > This approach works well in most cases. However, it fails when a file lacks a magic number but has a misleading file extension—for instance, a .fits file that is actually an IPAC table.

This is exactly my concern. I think there are cases that we download a file with the fits extension even it is an ipac table. I don't remember right now where. I know many of our files have a `.ul` extension. 

Before you do anything let me look into this case and see if I can find it.

loitly: Of course.  If we need to handle misleading file extensions, I can explore the best approach. The `.ul` extension is not an issue since it's not a recognized file type.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DuckDbUDF.java
```diff
@@ -17,12 +17,12 @@
 public class DuckDbUDF {
     public static String deg2pix = importDeg2Pix();
     public static final String decimate_key = """
-        CREATE FUNCTION decimate_key(xVal, yVal, xMin, yMin, nX, nY, xUnit, yUnit) AS
+        CREATE OR REPLACE FUNCTION decimate_key(xVal, yVal, xMin, yMin, nX, nY, xUnit, yUnit) AS
```
robyww, line 20: Is this related to the stability issue?

src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
```diff
@@ -158,6 +158,15 @@ private static void displayCacheInfo(PrintWriter writer, CacheManager cm, Server
 
     private static void showDatabaseStatus(PrintWriter writer) {
         DbAdapter.EmbeddedDbStats stats = DbMonitor.getRuntimeStats(true);
+
+        writer.printf(""" 
+            DATABASE INFORMATION
+            --------------------
+            Rows: %,15d       Peak Rows: %,15d 
+            Databases: %,10d       Peak Databases: %,10d      Total Database: %,10d
+
```
robyww, line 167: Glad you added this.

src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
```diff
@@ -163,6 +164,12 @@ private static DataGroup[] voToDataGroups(VOElement docRoot,
         return handler.getAllTable();
     }
 
+    public static boolean isVoTable(File inFile) {
+        try {
+            DataGroup[] tbl = voToDataGroups(inFile.getAbsolutePath(), true, 0);
+            return tbl != null && tbl.length > 0;
+        } catch (Exception e) { return false; }
+    }
```
robyww, line 172: This function appears to never be called. My bigger question is that is seems to go through code the retrieves the file. That is unexpected.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbDataIngestor.java
```diff
@@ -29,54 +31,92 @@
...
+        return ingestTable(dbAdapter, table, false);        //logic is already in the reader
     }
 
+    static FileInfo ingestJsonTable(TableServerRequest req, DbAdapter dbAdapter, File srcFile, boolean searchForSpectrum) throws IOException, DataAccessException {
+        var table = JsonToDataGroup.parse(srcFile, req);
+        return ingestTable(dbAdapter, table, searchForSpectrum);
+    }
 
+    static FileInfo ingestTable(DbAdapter dbAdapter, DataGroup table, boolean searchForSpectrum) throws DataAccessException {
+        if (searchForSpectrum) SpectrumMetaInspector.searchForSpectrum(table,true);
```
robyww, line 118: Is the DataGroup only partially read here? Is the meta data all in place or searchForSpectrum with not work.

My bigger question - are there cases where searchForSpectrum was being call before but is not longer being called because of the streaming ingestion.

src/firefly/java/edu/caltech/ipac/util/FormatUtil.java
```diff
@@ -0,0 +1,224 @@
...
+     *
+     * @param inFile The full path of the file to detect the format for.
+     * @return The detected {@code Format} of the file, or {@code UNKNOWN} if the format could not be determined.
+     * @throws IOException If an I/O error occurs while accessing the file.
+     */
+    @Nonnull
+    public static Format detect(File inFile) throws IOException {
+
+        Format format = null;
+        DuckDbAdapter.MimeDesc mimeDesc = DuckDbAdapter.getMimeType(inFile);
```
robyww, line 93: So this line actually looks inside the file? The function name is a little confusing since it implies it is just looking at the extension.

robyww (approved):
This looks good. There are a few questions I have that are listed below.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DuckDbUDF.java
loitly: No.  This occurs only in edge cases. I included it to ensure the create function never fails and always contains the latest version.

src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
loitly: It’s intended to read only the header, similar to IpacTableReader. However, after reviewing our guesser, I’m confident it does a good enough job detecting VOTable. As a result, this is no longer needed, and I’ll remove it.

src/firefly/java/edu/caltech/ipac/util/FormatUtil.java
loitly: This line uses DuckDB's magic extension to determine the file's MIME type, similar to the Unix `file` command. So yes, it will read the magic number(if exists) along with additional data as needed.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DuckDbUDF.java
loitly: The latest version of DuckDB appears to fix some of the unexpected crashing issue.

PR #1672, opened 2024-11-26
Firefly-1621: fixed race condition when images loading

robyww:
#### [Firefly-1621](https://jira.ipac.caltech.edu/browse/FIREFLY-1621): fixed race condition when images loading

 - Also fixes [Firefly-1617](https://jira.ipac.caltech.edu/browse/FIREFLY-1617)
 - url api examples changed to use https

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1621-race/firefly/?api
- load am image via url api

### DIFF
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -432,7 +432,11 @@ function bootstrap(props, clientAppSpecificOptions, webApiCommands) {
 
     }).then(() => {
         // when all is done, mark app as 'ready'
-        defer(() => dispatchUpdateAppData({isReady: true}));
+        defer(() => {
+            setTimeout(() => {
+                dispatchUpdateAppData({isReady: true})
+            },3);
+        });
         initWorkerContext();
     });
 }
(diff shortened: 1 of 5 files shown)


src/firefly/js/Firefly.js
```diff
@@ -472,9 +472,11 @@ function handleWebApi(webApiCommands, e, doAppRender) {
             // window.history.pushState('home', 'Home', new URL(window.location).pathname); // ?? is this necessary?
             doAppRender();
             dispatchOnAppReady(() =>  {
-                if (apiCompleted) return;
-                execute?.(cmd,params);
-                apiCompleted= true;
+                setTimeout(() => {
```
loitly, line 475: It seems that `isReady` is fired too early.  I think this is where [the problem](https://github.com/Caltech-IPAC/firefly/blob/1de7779b87999160aa1051750045971deeb21377/src/firefly/js/Firefly.js#L265) is.  That's why doAppRender exited before it's completed.  If possible, fixing the problem there would be better.

loitly (approved):
Looks good.

PR #1673, opened 2024-11-27
Firefly-1620: more spectrum types for the guesser

robyww:
#### [Firefly-1620](https://jira.ipac.caltech.edu/browse/FIREFLY-1620): more spectrum types for the guesser

- recognize spectrum from next-sci
- spectrum plot will reverse the low spectrum if negative

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1620-spec-types/firefly
- upload some of the files in the ticket (make sure to check `Attempt to interpret tables as spectra`)

### DIFF
--- a/src/firefly/js/charts/dataTypes/FireflySpectrum.js
+++ b/src/firefly/js/charts/dataTypes/FireflySpectrum.js
@@ -122,7 +122,8 @@ export function spectrumPlot({tbl_id, spectrumDM}) {
 
 export function getSpectrumProps(tbl_id, spectrumDM) {
 
-    spectrumDM = spectrumDM || getSpectrumDM(getTblById(tbl_id));
+    const tbl= getTblById(tbl_id);
+    spectrumDM = spectrumDM || getSpectrumDM(tbl);
     const {spectralAxis={}, fluxAxis={}, isSED, spectralFrame={}, derivedRedshift={}, target={}} = spectrumDM || {};
 
     const x = quoteNonAlphanumeric(spectralAxis.value);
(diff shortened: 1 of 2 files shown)


src/firefly/js/charts/dataTypes/FireflySpectrum.js
```diff
@@ -131,7 +132,18 @@ export function getSpectrumProps(tbl_id, spectrumDM) {
...
     const xErrArrayMinus = spectralAxis.statErrLow;
     const yErrArray = fluxAxis.statErrHigh || fluxAxis.statError;
-    const yErrArrayMinus = fluxAxis.statErrLow;
+    let yErrArrayMinus = fluxAxis.statErrLow;
+    if (yErrArray && fluxAxis.statErrLow) {
+        const errCol= getColumnIdx(tbl,yErrArray);
+        const lowCol= getColumnIdx(tbl,fluxAxis.statErrLow);
+        let allEmpty=true;
+        const reverse= tbl.tableData.data.every( (r) => {
+            if (isNaN(r[errCol]) || isNaN(r[lowCol]) || r[errCol===null] || r[lowCol]===null) return true;
```
jaladh-singhal, line 141:
```suggestion
            if (isNaN(r[errCol]) || isNaN(r[lowCol]) || r[errCol]===null || r[lowCol]===null) return true;
```
I think you meant this?

jaladh-singhal (approved):
Looks good except there's a typo I think

Also, when I upload 2MASS file, all error values are null but I still see the error bars (+), is that expected?

src/firefly/js/charts/dataTypes/FireflySpectrum.js
robyww: yes, I will fix it.

robyww:
> Also, when I upload 2MASS file, all error values are null but I still see the error bars (+), is that expected?

I think that is the way it is suppose to work. I am not totally sure but it makes sense. The upper and lower errors are effectively 0

PR #1674, opened 2024-12-02
FIREFLY-1572: Improve copying of coordinates from image panel

jaladh-singhal:
Fixes [FIREFLY-1572](https://jira.ipac.caltech.edu/browse/FIREFLY-1572)

- Removed extra leading whitespace in copy string of hms readouts
- Added Python SkyCoord readout option
  - Created a separate radio group for radio options since independent from readout options
  - Copy options appear specifically when click lock is on and options are WorldPt compatible
  - For celestial case (fits with WCS), SkyCoord option becomes disabled for Image pixel options - this is done on purpose to prevent layout from jumping 
- ~Show it selectively when firefly is being interfaced by python~

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1572-coord-copy/firefly/

Load a fits image, turn click lock on, copy coords - they shouldn't have leading whitespace

Click on the chip and change copy options to "SkyCoord" option. The copy string should be the code for instantiating a python SkyCoord object. Put it in a notebook/ipython session and the cell should execute successfully.
- Changing to Image pixel options will disable SkyCoord radio option because copy string algorithm ignores it anyway

Try same thing for a HiPS image.

### DIFF
--- a/src/firefly/js/visualize/MouseReadoutCntlr.js
+++ b/src/firefly/js/visualize/MouseReadoutCntlr.js
@@ -117,6 +117,7 @@ const initState= () =>
             pixelSize: 'pixelSize',
             hipsMouseReadout1:'eqj2000hms',
             hipsMouseReadout2:'galactic',
+            mouseReadoutValueCopy: 'str',
             healpixPixel:'healpixPixel',
             healpixNorder:'healpixNorder',
             intFluxValueRadix: '10',
--- a/src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
+++ b/src/firefly/js/visualize/ui/MouseReadoutOptionPopups.jsx
@@ -17,8 +17,8 @@ import {FieldGroup} from '../../ui/FieldGroup.jsx';
 import DialogRootContainer from '../../ui/DialogRootContainer.jsx';
 import {PopupPanel} from '../../ui/PopupPanel.jsx';
 import FieldGroupUtils, {getFieldGroupResults} from '../../fieldGroup/FieldGroupUtils.js';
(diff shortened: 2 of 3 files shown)


robyww (commented):
two UI comments

- readout option should be the same (you new way, showing copy options) even if `click lock` is off
- The copy mode should be saved as a preference
   - see `AppDataCntrl.js` - `dispatchAddPreference()` and `getPreference()`
   - Example CatalogUI.js line 210, then usage at HpxCatalog.js, line 107

src/firefly/js/visualize/MouseReadoutCntlr.js
```diff
@@ -117,6 +117,7 @@ const initState= () =>
             pixelSize: 'pixelSize',
             hipsMouseReadout1:'eqj2000hms',
             hipsMouseReadout2:'galactic',
+            mouseReadoutValueCopy: 'str',
```
robyww, line 120: Concerning preferences... This would be something like:

```js
mouseReadoutValueCopy: getPreference('mouseReadoutCopyPref','str')
```

robyww (commented):
one more thing

jaladh-singhal:
> readout option should be the same (you new way, showing copy options) even if click lock is off
Done

> The copy mode should be saved as a preference
Deferred for later because of UX complexities it will introduce

robyww (approved):
Looks good

lrebull:
Seems to work for me!
SkyCoord('14h03m29.69s +54d21m19.7s', frame='icrs')
SkyCoord('23h59m59.76s +0d00m01.0s', frame='icrs')
SkyCoord('23h59m48.34s -0d01m43.0s', frame='icrs')

PR #1682, opened 2024-12-13
Firefly-1624: improve bottom mouse readout layout

robyww:
#### Testing
- https://firefly-1624-readout.irsakudev.ipac.caltech.edu/applications/sofia
- Test according to ticket

### DIFF
--- a/src/firefly/js/visualize/ui/MouseReadoutBottomLine.jsx
+++ b/src/firefly/js/visualize/ui/MouseReadoutBottomLine.jsx
@@ -10,7 +10,7 @@ import {dispatchChangePointSelection} from '../ImagePlotCntlr.js';
 import {showMouseReadoutFluxRadixDialog} from './MouseReadoutOptionPopups.jsx';
 import {getNonFluxDisplayElements, getFluxInfo} from './MouseReadoutUIUtil.js';
 import {CopyToClipboard} from './MouseReadout.jsx';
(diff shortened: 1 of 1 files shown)


lrebull:
Tried to test this on build provided. searching on AOR 87_0002_1, FIFI-LS, Level 3 and 4 (see screenshot) and am getting these kinds of apparently transient (?) errors. i thought for a bit that maybe there was a non-printable character in my copy-paste of 87_0002_1, but that doesn't seem to be the whole story. 
first search:
Screenshot 2024-12-15 at 1 59 23 PM
first results:
Screenshot 2024-12-15 at 1 59 31 PM

second search results nets me AORs:
Screenshot 2024-12-15 at 2 00 59 PM
(second results) but now that same error is in the FIFI-LS tab :
Screenshot 2024-12-15 at 2 00 53 PM
 
third try worked as it should have the first time. I have no idea what is happening.

It took me a while to find the file i was using in the ticket (very sorry about that!) but the original purpose of the ticket is fixed. You're showing the information from the header. The header, however, is ambiguous. It's actually giving wavelengths in microns, and the readout you have makes it look like it's in meters, which isn't right, but you don't have enough information from the header to do much more than that. So this is fine as it is, and we can close the ticket and merge the branch.

For future us if we ever circle back to this issue, the search is : M42, all other search parameters as default. FIFI-LS results, impose this filter on file_url:
like '%F0247%' and like '%BLU%' and like '%WGR%'
and then go to the data tab and look at HDU #4. (the first few HDUs have undefined wavelengths.)

PR #1683, opened 2024-12-17
FIREFLY-1635: bug fix for CADC prepare download

kpuriIpac:
**Ticket**: [FIREFLY-1635](https://jira.ipac.caltech.edu/browse/FIREFLY-1635)
- there's a regression bug in `ttFeatureWatchers.js`, in some cases the request object does not have a `source` entry, but will have a `serviceUrl` entry instead (from TAP). 
- this fix uses that value when `source` is undefined, fixing the `Prepare Download` not working for CADC bug 

**Testing**: 
Firefly: https://fireflydev.ipac.caltech.edu/firefly-1635-prepare-download-bug/firefly
DCE: https://firefly-1635-prepare-download-bug.irsakudev.ipac.caltech.edu/irsaviewer/dce
- do a CADC search and select one or more rows, click on `Prepare Download` to make sure it works 
   -  the search i do is: select IVOA from the CADC tables, and search for 'm81', 10 rows.
- do a DCE search and repeat the same step (click on `Prepare Download` having selected one or more rows). 
  - i searched the MSX collection, click on a point (or, use this: 2h44m14.43s +57d57m24.8s Equ J2000) and search

### DIFF
--- a/src/firefly/js/templates/common/ttFeatureWatchers.js
+++ b/src/firefly/js/templates/common/ttFeatureWatchers.js
@@ -71,12 +71,9 @@ function setupObsCorePackaging(tbl_id) {
 }
 
 function updateSearchRequest( tbl_id='', dlParams='', sRequest=null) {
-    const hostname= new URL(sRequest.source)?.hostname;
-
+    const hostname= new URL(sRequest.source ? sRequest.source : sRequest.serviceUrl)?.hostname;
     const template= getTapObsCoreOptionsGuess(hostname)?.productTitleTemplate;
     const useSourceUrlFileName= getTapObsCoreOptionsGuess(hostname)?.packagerUsesSourceUrlFileName;
-
-
     const templateColNames= template && getColNameFromTemplate(template);
     const searchRequest = cloneDeep( sRequest);
     searchRequest.template = template;


src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -71,12 +71,9 @@ function setupObsCorePackaging(tbl_id) {
 }
 
 function updateSearchRequest( tbl_id='', dlParams='', sRequest=null) {
-    const hostname= new URL(sRequest.source)?.hostname;
-
+    const hostname= new URL(sRequest.source ? sRequest.source : sRequest.serviceUrl)?.hostname;
```
kpuriIpac, line 74: actually I can just simply this to `sRequest.source ?? sRequest.serviceUrl`. I'll add that in tomorrow, but let me know if it looks good & if testing it works for both CADC / DCE @robyww.

lrebull:
doesn't work for me; see attached movie. screenshot is the error i get (which is before downloading attempt). tbl file with target list also attached.

Screenshot 2024-12-16 at 7 51 25 PM

[bpmg.tbl.txt](https://github.com/user-attachments/files/18159841/bpmg.tbl.txt)


https://github.com/user-attachments/assets/243b908c-90a2-40f1-b3f6-137db472ef74

robyww (approved):
Looks good!

PR #1684, opened 2024-12-18
FIREFLY-1599: cascade combining spectra is broken

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1599
- there were NaNs in the data which is a max value for stats.
- replaced NaNs with null so that it is not counted when collecting stats.

To test: https://fireflydev.ipac.caltech.edu/firefly-1599-comgine-cascade-spectra/firefly/
- upload sample fits file from the ticket -> Load Image
- using `Extract Z-axis from cube` tool to pin a few charts
- switch to `Pinned Charts` tab, then combine them
- remember to select `Apply cascading` option

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/StatisticsProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/StatisticsProcessor.java
@@ -66,8 +66,13 @@ public class StatisticsProcessor extends TableFunctionProcessor {
                 row.setDataElement(columns[0], cname);
                 row.setDataElement(columns[1], desc);
                 row.setDataElement(columns[2], units);
+
+                String maxExp = col.isWholeNumber() ?
+                        "max(\"%1$s\") as \"%1$s_max\"".formatted(cname) :
+                        "max(if(\"%1$s\" = 'nan', null, \"%1$s\")) as \"%1$s_max\"".formatted(cname);    // for duckdb, NaN is greater than any other floating point number; therefore, we will exclude it by setting to null
+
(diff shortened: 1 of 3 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/StatisticsProcessor.java
```diff
@@ -66,8 +66,13 @@ protected DataGroup fetchData(TableServerRequest treq, DbAdapter dbAdapter) thro
                 row.setDataElement(columns[0], cname);
                 row.setDataElement(columns[1], desc);
                 row.setDataElement(columns[2], units);
+
+                String maxExp = col.isWholeNumber() ?
+                        "max(\"%1$s\") as \"%1$s_max\"".formatted(cname) :
+                        "max(if(\"%1$s\" = 'nan', null, \"%1$s\")) as \"%1$s_max\"".formatted(cname);    // for duckdb, NaN is greater than any other floating point number; therefore, we will exclude it by setting to null
+
```
robyww, line 73: This is interesting about `NaN` since in java script a `NaN` comparison is always false.

lrebull:
i confirm it works!

src/firefly/java/edu/caltech/ipac/firefly/server/query/StatisticsProcessor.java
loitly: Yes, this is specific to DuckDB.

```
... DuckDB follows the IEEE 754 specification regarding these special values, with two exceptions:

- NaN compares equal to NaN and greater than any other floating point number.
- Some floating point functions, like sqrt / sin / asin throw errors rather than return NaN for values outside their ranges of definition.
```

PR #1691, opened 2025-01-17
Firefly-1641: new FIREFLY_OPTIONS aproach

robyww:
#### [Firefly-1641](https://jira.ipac.caltech.edu/browse/FIREFLY-1641): Define an alternative and easier way to setting FIREFLY_OPTIONS without using json

- every value will be a separate property
- set a path with an underscore divider
- example- PROPS_OP_coverage_hipsSourceURL=ivo://CDS/P/allWISE/W4

#### Testing 
  - checkout branch
  - modify firefly-docker.env to set any properties
  - use `docker-compose` to build and run

### DIFF
--- a/firefly-docker.env
+++ b/firefly-docker.env
@@ -29,6 +29,23 @@
 #PROPS_DbAdapter__type=hsql
 #PROPS_logger__level=trace
 
+## example setting APP_PROPERTIES with OP syntax, an alternative to using the JSON
+## - path separator is the underscore
+## - any AppProperties entry begins with OP_
+## - this example
+##     - set irsa tap to be irsadev
+##     - irsa tap defaults to using SDSS for hips
+##     - non-IRSA TAP or IRSA Catalogs use ALLWISE band 4 for the coverage HiPS
+##     - when using TAP and IRSA the target panel popup hips will us SDSS and center on 5,5 and be 2 degrees wide
+#PROPS_OP_tap_additional_services_0_label=IRSA
+#PROPS_OP_tap_additional_services_0_value=https://irsadev.ipac.caltech.edu/TAP
+#PROPS_OP_tap_additional_services_0_hipsUrl=ivo://CDS/P/SDSS9/color
+#PROPS_OP_tap_additional_services_0_fovDeg=2D
+#PROPS_OP_tap_additional_services_0_centerWP="\'5;5;EQ_J2000\'"
+#PROPS_OP_coverage_hipsSourceURL=ivo://CDS/P/allWISE/W4
+#PROPS_OP_coverage_hipsSource360URL=ivo://CDS/P/allWISE/W4
+
+
 
 ##---- Empty declarations: values will come from the shell's environment
 ADMIN_PASSWORD
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
@@ -6,6 +6,7 @@ package edu.caltech.ipac.firefly.server;
 import edu.caltech.ipac.firefly.data.Alert;
 import edu.caltech.ipac.firefly.data.ServerParams;
 import edu.caltech.ipac.firefly.data.userdata.UserInfo;
+import edu.caltech.ipac.firefly.messaging.JsonHelper;
 import edu.caltech.ipac.firefly.server.security.SsoAdapter;
 import edu.caltech.ipac.firefly.server.util.Logger;
 import edu.caltech.ipac.util.AppProperties;
@@ -19,6 +20,7 @@ import java.util.List;
 import java.util.Map;
 
 import static edu.caltech.ipac.util.StringUtils.applyIfNotEmpty;
+import static edu.caltech.ipac.util.StringUtils.getDouble;
 
 public class AppServerCommands {
     private static final String SPA_NAME = "spaName";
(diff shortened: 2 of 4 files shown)


firefly-docker.env
```diff
@@ -29,6 +29,23 @@
...
+#     - irsa tap defaults to using SDSS for hips
+#     - non-IRSA TAP or IRSA Catalogs use ALLWISE band 4 for the coverage HiPS
+#     - when using TAP and IRSA the target panel popup hips will us SDSS and center on 5,5 and be 2 degrees wide
+PROPS_OP_tap_additional_services_0_label=IRSA
+PROPS_OP_tap_additional_services_0_value=https://irsadev.ipac.caltech.edu/TAP
+PROPS_OP_tap_additional_services_0_hipsUrl=ivo://CDS/P/SDSS9/color
+PROPS_OP_tap_additional_services_0_fovDeg=2D
+PROPS_OP_tap_additional_services_0_centerWP="\'5;5;EQ_J2000\'"
+PROPS_OP_coverage_hipsSourceURL=ivo://CDS/P/allWISE/W4
+PROPS_OP_coverage_hipsSource360URL=ivo://CDS/P/allWISE/W4
```
robyww, line 46: I will comment out after testing

src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
```diff
@@ -137,21 +139,37 @@ public String doCommand(SrvParam params) throws Exception {
...
+    private static void setToJson(JsonHelper jhelp, String key, String valStr) {
+        if (key.length()>OP_ROOT_LENGTH) {
+            Object val;
+            if (valStr.endsWith("D") && valStr.length()>1) {
+                double num= getDouble(valStr.substring(0, valStr.length()-1));
+                val= Double.isNaN(num) ? valStr : num;
+            }
+            else {
+                val= valStr;
+            }
```
loitly, line 169: Would it make more sense to convert any value into a number, and otherwise treat it as a string? If you have a string of numbers, you can enclose them in double quotes. I've seen this approach used elsewhere as well.

src/firefly/java/edu/caltech/ipac/util/AppProperties.java
```diff
@@ -221,14 +222,27 @@ public static String getProperty(String key, String def, Properties overridePDB)
...
       }
       return retval;
   }
 
+    public static Map<String,String> getWithStartingMatch(String startOfKey) {
+        var retMap= new HashMap<String,String>();
+        for(Map.Entry<Object,Object>  entry : System.getProperties().entrySet()) {
+            var key= entry.getKey().toString();
+            if (key.startsWith(startOfKey)) retMap.put(key,(String)entry.getValue());
+        }
```
loitly, line 238: This threw me off initially. I thought AppProperties was supposed to encapsulate this, but apparently it's implemented in such an odd way that you have to repeat it here. We should refactor AppProperties at some point.

loitly (approved):
I've reviewed the code and have a couple of comments. Otherwise, it looks good. The real test will be seeing how easy it is to use in Jenkins, where other components like Helm and YAML are involved.

src/firefly/java/edu/caltech/ipac/util/AppProperties.java
robyww: I don't understand your concern here. This is simple of function to return a maps of all properties that starts with a string.

src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
robyww: I guess, this approach gives more control though.

firefly-docker.env
```diff
@@ -29,6 +29,23 @@
 #PROPS_DbAdapter__type=hsql
 #PROPS_logger__level=trace
 
+# example setting APP_PROPERTIES with OP syntax, an alternative to using the JSON
+# - path separator is the underscore
```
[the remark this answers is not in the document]
robyww: I have some really concerns about that. Using double underscores properties 5 levels deep is hard to read.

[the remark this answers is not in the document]
robyww: Also the double underscore means something different. It means "turn it into a dot". This is a path separator.

src/firefly/java/edu/caltech/ipac/util/AppProperties.java
loitly: Oh, I was referring to [this](https://github.com/Caltech-IPAC/firefly/blob/1d408ca5c030739095e19e6514633daf326e29e4/src/firefly/java/edu/caltech/ipac/util/AppProperties.java#L209).
The logic of how to resolve a property is repeated here.  It's not a big deal.  There are other part of this class that can use some cleanup as well.

firefly-docker.env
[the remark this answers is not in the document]
loitly: I can see pros and cons in both approaches. I favor using . as a path separator since it aligns with JSON, but it's up to you to decide.

src/firefly/java/edu/caltech/ipac/util/AppProperties.java
robyww: yea, we can clean it up someday. I just removed an underscore. I don't think this is very high priority.

firefly-docker.env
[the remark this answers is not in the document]
robyww: The `.` does not work with environment variables, which is how the properties start out in docker. That is what I originally tried. 
In bash you can't do this:
```bash
a.b.c=123
```

PR #1698, opened 2025-02-03
FIREFLY-1648: fix Dockerfile parsing envVars bugs

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1648

This PR addresses several issues and improvements related to environment variable parsing.

The key changes are:

- Resolved issues with escaping special characters during Dockerfile parsing of environment variables.
- Rewritten the old launchTomcat.sh launch script into entrypoint.py, which is now the primary entrypoint for launching Tomcat.
- Add unit test entrypoint_test.py to make testing easier when modifying entrypoint.py
- Expose System Properties in ServerStatus
- Updated Jenkinsscripts to ensure that special characters in env variables are properly escaped

Test via Jenkins + Helm:  https://firefly-1648-envvars-parsing.irsakudev.ipac.caltech.edu/irsaviewer/
This is built with FIREFLY_PROPS=
```
OP_tap_additional_services_0_label=IRSA
OP_tap_additional_services_0_value=https://irsadev.ipac.caltech.edu/TAP
OP_tap_additional_services_0_hipsUrl=ivo://CDS/P/SDSS9/color
OP_tap_additional_services_0_fovDeg=2
OP_tap_additional_services_0_centerWP=5;5;EQ_J2000

```
I'm not entirely sure what all of this means, but I can confirm that it's hitting irsadev instead of irsa and that HiPS coverage is using SDSS9 color.

Test via Docker Compose:

Added to firefly-docker.env
```
## FIREFLY-1648 tests  -- will remove before merging.
pathPrefix=local       # path to firefly is now /local/firefly/ instead of /firefly/
PROPS="abc=123;semicolon=abc;;xyz;space=abc xyz;quote=abc\"xyz\""
PROPS_OP_tap_additional_services_0_centerWP=5;5;EQ_J2000
```
Start Firefly using Docker Comopse
```bash
cd firefly
docker compose up firefly --build
```
Verify that Firefly is now running at http://localhost:8080/local/firefly/ instead of /firefly/
Go to /admin/status -> click 'Debug Info'
to see that PROPS (multi-props) and PROPS_OP_tap_additional.. (single prop) are set.

You can also run `docker compose run firefly --dry-run` to see what envVars were used and how it will appears in the `CATALINA_OPTS` envVar.

If you have python3 installed, you can run unit tests:
```bash
cd firefly/docker
python3 entrypoint_test.py
```

These are fixes to Jenkins script if interested.  [commit](https://github.com/IPAC-SW/irsa-jenkins/commit/165905782e2edee82c52cb657314b17ee5a1db68) and [commit](https://github.com/IPAC-SW/irsa-jenkins/commit/169a29dc4cf9d2c32b2a1373649909194acfa479)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -34,7 +34,7 @@ RUN yarn install --ignore-platform --frozen-lockfile
 FROM node_module AS dev_env
 
 RUN apt-get update \
-    && apt-get install -y vim procps \
+    && apt-get install -y vim procps python3 \
 # cleanup
     && rm -rf /var/lib/apt/lists/*;
 
@@ -52,7 +52,7 @@ ENV env=local
 WORKDIR ${CATALINA_HOME}
 
 # Tomcat scripts, config files, etc
-COPY firefly/docker/*.sh ./
+COPY firefly/docker/*.sh firefly/docker/*.py ./
 COPY firefly/docker/tomcat-users.xml conf/
 
 # set up directory protections, copy stuff around, add tomcat user and group
@@ -120,7 +120,7 @@ ARG gid=91
 # - this is a big part of the layer so do it early
 # - emacs removed because it is so big: to readd emacs: emacs-nox
 RUN apt-get update && apt-get install -y \
-        vim procps wget unzip \
+        vim procps wget unzip python3 \
         && rm -rf /var/lib/apt/lists/*;
 
 # These are the users replaceable environment variables, basically runtime arguments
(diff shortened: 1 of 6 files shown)


robyww:
I want to look at this in detail but right away I think it is important to test Jenkin with a the first as 
```
OP_tap_additional_services_0_label='IRSA dev with spaces'
```

robyww (approved):
It looks good and appears to work. I think the python makes it easier to work with and to read.

I really like the parameter passing being so clean now.

If you want to keep launchTomcat.sh (it might a good idea for a cycle or two) around for awhile you should put some deprecation message at the top.

I do think if we are going to do this in python we should follow `PEP 8`. That is what we are doing with `firefly_client`. IntelliJ sets that up by default I think. I least mine does.

There are several sections that needs to be cleaned up to be `PEP 8` like.

I personally am not a big fan of `PEP 8` but I think we should be consistent.

PR #1702, opened 2025-02-05
CADC-13454: Firefly Extensions Support

### LINKED ISSUE #1687: [CADC] Accessing Proprietary Tables from TAP Services
Hello Firefly Team,

We (@brianmajor, @shinybrar)at CANFAR are actively working on a generally available Firefly deployment for our user base. While we already have Firefly working with Public TAP Services / Tables we are currently investigating how can Firefly access proprietary tables that require authentication, e.g. the YouCat TAP Service. We would like to know,

- Does firefly currently to support accessing proprietary tables via HTTP Tokens?
- If yes, could you provide documentation or an example on how to configure Firefly to use HTTP Tokens?
- If no, are there any recommended workarounds or plans to include such functionality in future Firefly releases?

Any guidance on this matter would be very helpful.

shinybrar:
Fixes #1687 

In the current architecture of the Firefly web application, the `firefly.war` file is extracted into the `webapps/firefly/` directory at Tomcat startup. If the `webapps/` folder already exists, Firefly completely ignores the extraction process. This makes it problematic to develop extensions for firefly and simply load them into the existing container via,

```
FROM ipac/firefly:latest
COPY MySsoAdapter.jar ${CATALINA_HOME}/webapps/firefly/WEB-INF/lib/
```

To resolve this issue, a modification to the Tomcat startup script has been implemented. The script now automatically copies third-party extensions from a persistent directory (`$CATALINA_HOME/extensions/`) into `WEB-INF/lib/` after the WAR extraction process.

## Implementation Details

- You can define a custom `EXTENSIONS_DIR` via env variable, it defaults to `$CATALINA_HOME/extensions/`

- Startup State Machine Changes
  - Check if `webapps` is empty before extracting WAR files. If missing, create the directory. Extract WAR files only if `webapps` is empty.

  - After the WAR extraction step, check for the existence of `WEB-INF/lib/`. If missing, create the directory. Copy all JAR files from `$CATALINA_HOME/extensions/` into `WEB-INF/lib/`.

- Relevant Codebase Change
```bash

echo -e "============================================================"
echo -e "         Firefly Runtime Configuration"
echo -e "============================================================"
echo "Starting WAR extraction process..."
for archive in "${CATALINA_HOME}/webapps-ref/"*.war; do
    echo "  Discovered archive: ${archive}"
    warName=$(basename "${archive}" .war)
    # If baseURL is set, append it to the context path
    if [[ -n "${baseURL}" ]]; then
      warName="$(sed -r 's/\//#/g;s/^(#)+|(#)+$//g' <<< "${baseURL}")#${warName}"
    fi

    # Set the warDir to the webapps directory
    warDir="${CATALINA_HOME}/webapps/${warName}"

    # If warDir exists, inform user and skip, else extract web archive
    if [[ -d "${warDir}" ]]; then
      echo "  Skipping extraction, ${warDir} already exists"
    else
      echo "  Extracting ${archive} to ${warDir}"
      mkdir -p "${warDir}"
      unzip -oqd "${warDir}" "${archive}"
      sed -E -i.bak 's/##out--//' "${warDir}/WEB-INF/classes/log4j2.properties"
    fi
done
echo "  WAR extraction process complete."
echo "Configuring third party extensions for Firefly..."
# Copy .jar files from CATALINA_HOME/extensions to CATALINA_HOME/firefly/WEB-INF/lib/
if [[ -d ${EXTENSIONS_DIR} ]]; then
  # Check if firefly/WEB-INF/lib/ directory exists
  if [[ ! -d "${CATALINA_HOME}/webapps/firefly/WEB-INF/lib" ]]; then
    echo "  Creating ${CATALINA_HOME}/webapps/firefly/WEB-INF/lib/"
    mkdir -p "${CATALINA_HOME}/webapps/firefly/WEB-INF/lib"
  fi

  for extension in "${CATALINA_HOME}/extensions/"*.jar; do
    cp -v "${extension}" "${CATALINA_HOME}/webapps/firefly/WEB-INF/lib/"
  done
```
- Expected Output during startup
```
============================================================
         Firefly Runtime Configuration
============================================================
Starting WAR extraction process...
  Discovered archive: /usr/local/tomcat/webapps-ref/firefly.war
  Extracting /usr/local/tomcat/webapps-ref/firefly.war to /usr/local/tomcat/webapps/firefly
  WAR extraction process complete.
Configuring third party extensions for Firefly...
'/usr/local/tomcat/extensions/cadc-sso-lib-0.1.jar' -> '/usr/local/tomcat/webapps/firefly/WEB-INF/lib/cadc-sso-lib-0.1.jar'
  Third party extension configuration complete.
============================================================
```


docker/launchTomcat.sh
```diff
@@ -1,14 +1,24 @@
...
 NAME=${BUILD_TIME_NAME:-"ipac/firefly"}
 
 ADMIN_USER=${ADMIN_USER:-admin}
-ADMIN_PASSWORD=${ADMIN_PASSWORD:-`echo $RANDOM | base64 | head -c 8`}
+if [[ -z "${ADMIN_PASSWORD}" ]]; then
+  ADMIN_PASSWORD=$(echo "${RANDOM}" | base64 || true)
+  ADMIN_PASSWORD=$(echo "${ADMIN_PASSWORD}" | head -c 8)
+fi
 USE_ADMIN_AUTH=${USE_ADMIN_AUTH:-"true"}
+EXTENSIONS_DIR=${EXTENSIONS_DIR:-${CATALINA_HOME}/extensions}
```
loitly, line 15: I recommend extending the existing ${CATALINA_HOME}/webapps-ref/ directory to handle more use cases than just serving as a placeholder for the WAR file.

For example, if you have firefly.war, after extraction, we would copy everything from ${CATALINA_HOME}/webapps-ref/firefly/ to ${CATALINA_HOME}/webapps/firefly/.

In your case, instead of copying the JAR file into a separate configurable location, you would place it into ${CATALINA_HOME}/webapps-ref/firefly/WEB-INF/lib.

This approach will also make it possible to side-load other files, not just JARs.
What do you think?

docker/launchTomcat.sh
```diff
@@ -1,14 +1,24 @@
...
 
+# Check if CATALINA_HOME is set, if not, exit the script
+if [[ -z "${CATALINA_HOME}" ]]; then
+  echo "CATALINA_HOME enviroment variable is not set. Exiting..."
+  exit 1
+fi
```
loitly, line 21: I don't believe this is necessary, as the base image we use should always have it set. If it's not set, we should configure it to the correct location rather than allowing it to fail.

loitly (changes requested):
Thank you for identifying this oversight and for your contribution. However, the timing is a bit tricky, as we just merged a [PR](https://github.com/Caltech-IPAC/firefly/pull/1698) that converted the Bash script you modified into Python.

I’d love for you to review the comments I’ve made and consider creating a new PR for the updated Python script. Or, if you’d prefer, we can add the missing feature on our end.

Thanks again for contributing!

docker/launchTomcat.sh
shinybrar: This sounds like a great idea @loitly -- I will translate the effort to the python entrypoint script. 

Just to be certain, we want our approach to be,

1. Extract archives from `${CATALINA_HOME}/webapps-ref/<LIBRARY.war` to `${CATALINA_HOME}/webapps/<LIBRARY>/`.
2. Copy `Java Archives` (and other data productions??) from `${CATALINA_HOME}/webapps-ref/<LIBRARY>/*.jar` to `${CATALINA_HOME}/webapps-ref/<LIBRARY>/WEB-INF/lib/`

Overall I like this approach, it maintains the consistency of the current implementation and it simplifies the config management.

If you give your blessing for this change, I will go ahead and implement the needed changes.

docker/launchTomcat.sh
shinybrar: Sounds good.

docker/launchTomcat.sh
loitly: Yes, feel free to make the change. However, it’s not exactly as you described.

`#1` is correct.

For `#2`, I want to **recursively copy all directories and files** from `${CATALINA_HOME}/webapps-ref/` to `${CATALINA_HOME}/webapps/`.

For example, if `/webapps-ref/` contains:
```
/firefly.war
/firefly/WEB-INF/classes/logging.properties
/firefly/WEB-INF/lib/a.jar
/firefly/css/my.css
/firefly/my.html
```

After expanding `firefly.war` into `${CATALINA_HOME}/webapps/firefly/`, you should **copy all remaining files and directories** listed above into the same `${CATALINA_HOME}/webapps/firefly/` directory.

**Note**: Make sure to handle `PATH_PREFIX` if it's set.

For instance, if `PATH_PREFIX` is set to "staging", then everything above should be copied to `${CATALINA_HOME}/webapps/staging#firefly/`.

shinybrar: That sounds good to me, ill get cracking :D

PR #1703, opened 2025-02-06
Firefly-1655: Generalize IRSA Catalog Search for other services implementing the Gator protocol

robyww:
#### [Firefly-1655](https://jira.ipac.caltech.edu/browse/FIREFLY-1655): Generalize IRSA Catalog Search for other services implementing the Gator protocol

 - Used with LSDB
 - The Gator Protocol tab will only appears if searches are configured with the option `gatorProtocol.services`

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1655-gator-proto/firefly/
- go to `LSDB Test Searches` tab.
- try a search

##### In Jenkins
```
OP_gatorProtocol_title=LSDB Test Searches
OP_gatorProtocol_primary=true
OP_gatorProtocol_searchOptionsMask=cone,polygon,box
OP_gatorProtocol_services_0_serviceId=troyTest
OP_gatorProtocol_services_0_label=LSDB Raen Test
OP_gatorProtocol_services_0_value=https://troyraen.irsakudev.ipac.caltech.edu
OP_gatorProtocol_services_1_serviceId=IRSATEST
OP_gatorProtocol_services_1_label=Irsa Test (not LSDB)
OP_gatorProtocol_services_1_value=https://irsatest.ipac.caltech.edu
OP_gatorProtocol_services_1_showSqlSection=true
OP_gatorProtocol_services_1_searchOptionsMask=Cone,Elliptical,Box,Polygon,Multi-Object,All Sky

```

##### As JSON
```JSON
{
 "gatorProtocol": {
    "title": "LSDB Test Searches",
    "primary": "true",
    "searchOptionsMask": "cone,polygon,box"
    "services": [
      {
        "label": "LSDB Raen Test",
        "serviceId": "troyTest",
        "value": "https://troyraen.irsakudev.ipac.caltech.edu"
      },
      {
        "showSqlSection": "true",
        "label": "Irsa Test (not LSDB)",
        "serviceId": "IRSATEST",
        "value": "https://irsatest.ipac.caltech.edu",
        "searchOptionsMask": "Cone,Elliptical,Box,Polygon,Multi-Object,All Sky"
      }
    ],
  }
}
```

### DIFF
--- a/src/firefly/html/firefly-dev.html
+++ b/src/firefly/html/firefly-dev.html
@@ -15,9 +15,9 @@
             app: {
                 template: 'FireflyViewer',          // set to app mode, instead of api.
                 menu: [
-                    {label:'Images', action:'ImageSelectDropDownCmd'},
-                    {label:'Charts', action:'ChartSelectDropDownCmd'},
-                    {label:'Upload', action: 'FileUploadDropDownCmd'},
+                    {label:'Images', action:'ImageSelectDropDownCmd', primary:true},
+                    {label:'Charts', action:'ChartSelectDropDownCmd', primary:true},
+                    {label:'Upload', action: 'FileUploadDropDownCmd', primary:true},
                     {label:'Workspace', action: 'WorkspaceDropDownCmd'},
                     {label:'TAP & IRSA Searches', action: 'MultiTableSearchCmd'},
                     {label:'Test Searches', action:'TestSearches'},
@@ -109,6 +109,25 @@
                         SurveyKey: '1b',
                         SurveyKeyBand: '4'
                     }
+                },
+                gatorProtocol: {
+                    title: 'LSDB Searches',
+                    searchOptionsMask: 'Cone,Elliptical,Box,Polygon',
+                    services: [
+                        {
+                            serviceId: 'troy',
+                            label: 'Troy Test',
+                            value: 'https://troyraen.irsakudev.ipac.caltech.edu',
+                            searchOptionsMask: 'Cone,Polygon,box'
+                        },
+                        {
+                            serviceId: 'IRSATEST',
+                            label: 'Irsa Test',
+                            value: 'https://irsa.ipac.caltech.edu',
+                            showSqlSection: true,
+                            searchOptionsMask: 'Cone,Elliptical,Box,Polygon,Multi-Object,All Sky'
+                        },
+                    ],
                 }
             })
         };
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -7,7 +7,7 @@ import 'isomorphic-fetch';
 import {Stack, Typography} from '@mui/joy';
 import React from 'react';
 import {createRoot} from 'react-dom/client';
-import {set, defer, once} from 'lodash';
+import {set, defer, once, isArray} from 'lodash';
 import 'styles/global.css';
 
 import {APP_LOAD, dispatchAppOptions, dispatchUpdateAppData} from './core/AppDataCntlr.js';
@@ -22,7 +22,6 @@ import {dispatchUpdateLayoutInfo} from './core/LayoutCntlr.js';
 import {FireflyRoot} from './ui/FireflyRoot.jsx';
 import {SIAv2SearchPanel} from './ui/tap/SIASearchRootPanel';
 import {getSIAv2ServicesByName} from './ui/tap/SiaUtil';
-import {TapSearchPanel} from './ui/tap/TapSearchRootPanel';
 import {mergeServices} from './ui/tap/TapUtil';
 import {dispatchChangeReadoutPrefs} from './visualize/MouseReadoutCntlr.js';
 import {showInfoPopup} from './ui/PopupUtil';
@@ -39,6 +38,7 @@ import {dispatchOnAppReady} from './core/AppDataCntlr.js';
 import {getBootstrapRegistry} from './core/BootstrapRegistry.js';
 import {showLostConnection} from './ui/LostConnection.jsx';
 import {recordHistory} from './core/History.js';
+import {GatorProtocolRootPanel} from './visualize/ui/GatorProtocolRootPanel';
 import {setDefaultImageColorTable} from './visualize/WebPlotRequest.js';
 import {initWorkerContext} from './threadWorker/WorkerAccess.js';
 import {getTAPServicesByName} from './ui/tap/TapKnownServices.js';
@@ -139,7 +139,8 @@ const defAppProps = {
                         layout= {{width: '100%'}}
                           lockTitle='IRSA SIAv2 Search'
                         name='IRSA_USING_SIAv2'/>,
-        ]
+    ]
+
 };
 
 /** @type {FireflyOptions} */
@@ -250,6 +251,7 @@ function installOptions(appSpecificOptions) {
     if (options.imageScrollsToHighlightedTableRow!==visRoot().autoScrollToHighlightedTableRow) {
         dispatchChangeTableAutoScroll(options.imageScrollsToHighlightedTableRow);
     }
+    return options;
 
 }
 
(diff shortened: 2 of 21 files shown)


src/firefly/js/visualize/ui/GatorProtocolRootPanel.jsx
```diff
@@ -0,0 +1,202 @@
...
+                                        onGatorProtoServiceOptionSelect(val);
+                                        addGatorProtoUserService(val.value);
+                                    }}
+                        />
+
+                    ) : (
+                        <ListBoxInputFieldView {...{
+                            sx:{'& .MuiSelect-root':{width:'46.5rem'}},
+                            options:gpOps, value:serviceUrl,
+                            placeholder:'Choose SIAv2 Service...',
```
gpdf, line 142: "SIAv2" doesn't seem to belong here either.  I'm not sure that this placeholder is actually displaying anywhere, though - maybe it's overridden elsewhere in the code?

src/firefly/js/visualize/ui/GatorProtocolRootPanel.jsx
```diff
@@ -0,0 +1,202 @@
...
+                            alignSelf: 'flex-start',
+                            '--Switch-trackWidth': '20px',
+                            '--Switch-trackHeight': '12px',
+                        },
+                    }} />
+                </Stack>
+                <Stack>
+                    {enterUrl ? (
+                        <InputField orientation='horizontal'
+                                    placeholder='Enter SIAv2 Url'
```
gpdf, line 127:
```suggestion
                                    placeholder='Enter IPAC Gator protocol URL'
```

Or you could use the approach at line 156 where there's a variable like `title` that is used to state the type of protocol?

src/firefly/js/visualize/ui/GatorProtocolRootPanel.jsx
```diff
@@ -0,0 +1,202 @@
...
+                                    value={serviceUrl}
+                                    actOn={['enter']}
+                                    tooltip='enter SIAv2 URL'
```
gpdf, line 130:
```suggestion
                                    tooltip='Enter IPAC Gator protocol URL'
```

gpdf (changes requested):
I've put in a few comments where it seems like some "SIAv2" text was inadvertently copied over into this.

At a larger scale, this addresses the near-term need well, and I don't think substantial further effort is justified by the need.  I like the feature that makes the "extra SQL" form field conditional.

A few little comments and questions, but not to be taken as required changes:
* Can a tab be locked to a different Gator service?  So can we have the basic IRSA catalogs tab, which is of course locked, and another labeled tab that's locked to a particular alternate Gator service?  Or does the current version only support "standard IRSA" and "open to Gator selection"?  I just want to know; it's OK either way.
* IF this approach lives for an extended period, I don't know whether we'd want to be able to customize/limit the "search methods" menu.  @troyraen Were you able to implement all of cone, elliptical, box, polygon, and multi-object?

robyww:
@gpdf 
> IF this approach lives for an extended period, I don't know whether we'd want to be able to customize/limit the "search methods" menu. @troyraen Were you able to implement all of cone, elliptical, box, polygon, and multi-object?

I think we could limit this menu.

robyww:
> Can a tab be locked to a different Gator service? So can we have the basic IRSA catalogs tab, which is of course locked, and another labeled tab that's locked to a particular alternate Gator service? Or does the current version only support "standard IRSA" and "open to Gator selection"? I just want to know; it's OK either way.

I put in the hooks for this but did not implement it. I can do that down the road.

src/firefly/js/visualize/ui/GatorProtocolRootPanel.jsx
robyww: I will do that.

gpdf:
> I think we could limit this menu.

OK, let's wait to see what @troyraen says.

troyraen:
> Were you able to implement all of cone, elliptical, box, polygon, and multi-object?

I've implemented cone, box, and polygon. I haven't tried elliptical or multi-object.

gpdf:
> I've implemented cone, box, and polygon. I haven't tried elliptical or multi-object.

I think that suggests that a way to suppress those menu items might be useful.  The alternative is for @troyraen's back end to return an error message; I don't know how error handling works in the Gator protocol.  Is there a way for him to pass back a textual error message (which would say "elliptical queries are not supported yet in LSDB" or some such) and have Firefly display that text?

Whichever is easier!  This is just an interim shim for future work on an IVOA-ish way of doing all this, but we don't know as of right now when we'll be able to schedule that future work.

src/firefly/js/Firefly.js
```diff
@@ -278,15 +280,37 @@ function fireflyInit(props, appSpecificOptions={}, webApiCommands) {
         set(appSpecificOptions, 'table.showPropertySheetButton', appSpecificOptions?.table?.showPropertySheetButton ?? true);
     }
 
-    installOptions(appSpecificOptions);
+    const finalOptions= installOptions(appSpecificOptions);
+    const finalAppProps= setupGatorProtocolPanel(finalOptions,appProps);
```
loitly, line 284: I think this approach needs refinement, if not now, then definitely when the next use case arises.  we should avoid adding this type of logic directly in this file.

robyww:
@gpdf @troyraen I implement setting the search methods. It can be set at two levels. On the service or at the level above then it will apply to all services (except default gator) including one you add on the fly.

robyww:
A new build is available with many of the requested changes https://fireflydev.ipac.caltech.edu/firefly-1655-gator-proto/firefly

src/firefly/js/Firefly.js
robyww: I am not sure where else to add it. I am not sure I want to but in a whole new system for just this.

src/firefly/html/firefly-dev.html
```diff
@@ -109,6 +109,22 @@
                         SurveyKey: '1b',
                         SurveyKeyBand: '4'
                     }
+                },
+                gatorProtocol: {
```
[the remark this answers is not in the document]
robyww: This is for using with lsdb so calling `gator` it gator is confusing. I guess I could change it to `gatorApi` but that is a lot of file  and variable renaming. Do you feel really strongly about this?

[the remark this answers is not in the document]
loitly: Yes, I saw more usage later on, like filename.  It's fine to leave it as is.

src/firefly/js/Firefly.js
loitly: Sure, that's fine for now. However, in the future, I'd like to see an app-options reduction step that processes an ordered list of pure reducing functions (without side effects), where inputs and outputs are clearly defined.

For example, if the menu is affected, it should be passed in as an input and return an updated menu as the output. Any side effects (such as dispatching events) should be handled separately by Firefly.js.

robyww: That fine... just so you know I did not write `installOptions(appSpecificOptions)` for this PR it has been around for a long time. I just change it to return a value. `setupGatorProtocolPanel` does not have side effects.

But I agree it would be better to modify `installOptions` do it like you said.

troyraen (approved):
The new option looks good to me. Thanks! 

One question about how and when the tab title, etc. can be set:
I see that those things are configurable with build parameters, so I assume that means when building the image. Is that the intended access point for these options going forward? We are sending the Gator-over-LSDB code to an outside group for them to run and the code currently pulls a pre-built Firefly image from DockerHub. I'm wondering if I need to update that to build Firefly from source or if there will be some way to set the options at runtime.

Noting an issue that is probably unrelated:
Several of my queries failed with "Catalog is unavailable". For at least one of those catalogs I was able to get one query to succeed. Not sure what's going on there but the same problem is occurring in my preexisting deployment so it seems unrelated to this PR.

robyww:
@troyraen These are options set at runtime. So you can pull a pre-built from dockerhub and run with these parameters. We should talk about how you are starting firefly but if you are using environment variable to set your options you would add `PROP_` to the start of each. When the firefly docker starts if pulls any environment variables the start with `PROP_*` and turns them into start up options.

However, I am not sure how you are starting firefly so we should talk.

troyraen:
> However, I am not sure how you are starting firefly so we should talk.

In general, I'm using a Docker Compose file that pulls the latest Firefly image from DockerHub and starts it with this env var to point at the lsdb-server service:

```
PROPS=irsa.gator.dd.hostname=http://lsdb-server:8000 irsa.gator.hostname=http://lsdb-server:8000
```

The instance running at https://fireflydev.ipac.caltech.edu/troyraen/firefly was set up differently -- Loi did that using Jenkins.

PR #1706, opened 2025-02-11
FIREFLY-1616: The Firefly table does not function correctly when a column contains non-ASCII characters.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1616

- Add charset=UTF-8 to the Content-Type header to ensure the server correctly interprets the encoding.

Test: https://fireflydev.ipac.caltech.edu/firefly-1616-unicode-characters-columns/firefly/

Upload the sample table found in the ticket. It should be working now. Remember to test other functions as well, such as sorting and filtering. Since the modified function is a core function, perform a regression test to ensure everything else works as expected.

### DIFF
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -856,36 +856,40 @@ export function tokenSub(valObs, str='') {
  */
 export async function lowLevelDoFetch(url, options, doValidation, loggerFunc) {
     if (options.params) {
-        const params = toNameValuePairs(options.params);        // convert to name-value pairs if it's a simple object.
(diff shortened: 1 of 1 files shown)


robyww (approved):

cwang2016:
yes, the table viewer shows the column names with the unicode characters accurately. The sorting and filtering on such columns work well, too.
However, the unicode chacaters on the first column of the summary table under 'Upload' tab,  'name', are missed as I uploaded the test file posted in the ticket, https://jira.ipac.caltech.edu/browse/FIREFLY-1616. Please help check.

PR #1708, opened 2025-02-11
CADC-13454: Firefly Entrypoint Extractions

shinybrar:
Fixes https://github.com/Caltech-IPAC/firefly/issues/1687

## Summary
This PR overhauls the startup script for our Tomcat-based application. The new implementation replaces legacy shell commands with Python’s standard libraries, introduces type annotations, improves logging, and refines environment variable processing—all while maintaining backward compatibility with existing options like `--help` and `--dry-run`.

## Key Changes
- **WAR Extraction & Log Modification**
  - **Before:** Used external commands (`unzip`, `sed`) in `extract_war_files`.
  - **Now:** The new `extract()` function uses the `zipfile` module and `pathlib` for robust file handling. It extracts WAR files, creates backups of `log4j2.properties`, and directly modifies log settings.
  - Added support for extracting additional files from the origin, i.e. when extracting `LIBRARY.war`, any file under the tree of `/LIBRARY/..` is also copied to the destination.
- Added type hints and more descriptive docstrings for maintainability.
- Retains with the existing codebase
- Added testing for extraction process


docker/entrypoint.py
```diff
@@ -6,28 +6,83 @@
...
-    - Extract WAR files from webapps-ref to webapps.
-    - Modify log4j to send logs to stdout.
-    - Modify context path (pathPrefix) if given.
-
-    :param webapps_dir: Path to the webapps directory.
-    :param webapps_ref: Path to the webapps reference directory.
-    :param path_prefix: Prefix for the context path.
+
+def extract(source: Path, destination: Path, prefix: str = "/", kind: str = "war"):
+    """Extract a file to a destination directory.
```
loitly, line 18: `extract` should be called only once during the initial startup. it looks like it's getting called on every startup.

docker/entrypoint.py
```diff
@@ -6,28 +6,83 @@
...
+
+    Args:
+        source (Path): Source directory to extract from
+        destination (Path): Destination directory to extract to
+        prefix (str, optional): Prefix extracted filepaths. Default is "/".
+        kind (str, optional): File extension to extract. Default is "war".
```
loitly, line 24: This does not need to be a parameter; it should always be `war`. Can you think of a case where extracting a non-WAR file into webapps is needed?

docker/entrypoint.py
```diff
@@ -6,28 +6,83 @@
...
+            if log4j.exists():
+                # Create backup as log4j.properties.bak
+                print(f"    Modifying {log4j}")
+                shutil.copy(log4j, log4j.with_suffix(".bak"))
+                with open(log4j, "r+", encoding="utf-8") as file:
+                    content = file.read()
+                    content = content.replace("##out--", "")
+                    file.seek(0)
+                    file.write(content)
+                    file.truncate()
```
loitly, line 71: This doesn't seem to be working. The file remains unmodified, and no logs are being sent to the console.

docker/entrypoint.py
```diff
@@ -77,16 +132,17 @@ def add_single_prop_env_vars():
...
+    print(
+        f"""
     ========== Information: You can set environment variables using -e on the docker run line =====
 
     Environment Variables:
             Description                      Name                          Value
             -----------                      --------                      -----
             Admin username                   ADMIN_USER                    {os.getenv('ADMIN_USER')}
-            Admin password                   ADMIN_PASSWORD                {os.getenv('ADMIN_PASSWORD')}
+            Admin password          s         ADMIN_PASSWORD                {os.getenv('ADMIN_PASSWORD')}
```
loitly, line 145: `'s'` Typo?

docker/entrypoint.py
```diff
@@ -117,35 +173,44 @@ def log_env_info(path_prefix, visualize_fits_search_path):
     Command line options:
             --help  : show help message, examples, stop
             --debug : start in debug mode
-    """)
+    """
+    )
+
 
+def show_help(name: str, webapps_ref: Path):
```
loitly, line 180: Good catch!

docker/test_entrypoint_extract.py
```diff
@@ -0,0 +1,66 @@
+"""Test the extract entrypoint function."""
+
+import unittest
+from pathlib import Path
+import shutil
+import tempfile
+from entrypoint import extract
+
+
+class TestExtractFunction(unittest.TestCase):
```
loitly, line 10: Please combine this file with `entrypoint_test.py` then rename it to `test_entrypoint.py`.

loitly (changes requested):
The changes look good, but I have a few concerns that I’ve commented on. Please review and address them.

docker/entrypoint.py
shinybrar: I will get this fixed.

docker/entrypoint.py
shinybrar: 😄 typo indeed!

docker/entrypoint.py
shinybrar: Done.

docker/entrypoint.py
shinybrar: The issue has been fixed, I believe since I was overloading the files at `destination / "WEB-INF/classes/log4j2.properties"` rather than `landing / "WEB-INF/classes/log4j2.properties"`.

shinybrar:
Addressed the PR Comments, and additionally refactored the extraction process into three smaller methods for better readability.

loitly:
Your latest changes worked great! If you don’t mind, please rename `entrypoint_test.py` to `test_entrypoint.py` to align with your original best practice naming convention. Also, the backup file `log4j2.bak` should be named `log4j2.properties.bak`. 
Thanks for your contribution, we really appreciate it!

shinybrar:
@loitly Apologies, on missing these change requests. The PR has been updated with the file rename and the logging filename changes.

loitly (approved):

PR #1714, opened 2025-02-22
IRSA-6654: Nicer Simbad search error

robyww:
#### Testing
- https://fireflydev.ipac.caltech.edu/irsa-6654-simbad-error/firefly
- bring up a HiPS
- zoom in deep and find an area with is pretty blank
- select a small area and use the Screenshot 2025-02-21 at 5 14 20 PM
- from dropdown choose: `Search (cone) using simbad...`
- if there is an error it should be cleaner.

### DIFF
--- a/src/firefly/js/ui/DefaultSearchActions.js
+++ b/src/firefly/js/ui/DefaultSearchActions.js
@@ -1,13 +1,14 @@
 import {getMenu} from '../core/AppDataCntlr.js';
 import {makeSearchActionObj, SearchTypes} from '../core/ClickToAction.js';
 import {flux} from '../core/ReduxFlux.js';
-import {MetaConst} from '../data/MetaConst';
 import {ServerParams} from '../data/ServerParams.js';
 import {sprintf} from '../externalSource/sprintf.js';
-import {getActiveTableId, getMetaEntry, getTableUiByTblId, getTblById, makeFileRequest} from '../api/ApiUtilTable.jsx';
+import {
+    getActiveTableId, getMetaEntry, getTableUiByTblId, getTblById, makeFileRequest, onTableLoaded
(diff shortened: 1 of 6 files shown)


vandesai1:
I wonder why a SIMBAD query with no results is displayed any differently than a WISE query with no results.

vandesai1:
I would change the wording from
"Simbad failed to find any data for 296.72620, -30.34029 and 0.00100 degrees"
to
"No data found for SIMBAD cone search with center 296.72620, -30.34029 and radius 0.00100 degrees."
This way it doesn't seem like SIMBAD failed.

src/firefly/js/ui/DefaultSearchActions.js
```diff
@@ -276,7 +277,17 @@ function searchSimbad(cenWpt,radius) {
...
+    const {tbl_id}= request.META_INFO;
     dispatchTableSearch(request);
+    onTableLoaded(tbl_id).then( (table) => {
+        if (table.error) {
+            const lon= sprintf('%.5f',cenWpt.getLon());
+            const lat= sprintf('%.5f',cenWpt.getLat());
+            const radStr= sprintf('%.5f',Number(radius));
+            const error= `Simbad failed to find any data for ${lon}, ${lat} and ${radStr} degrees`;
+            dispatchTableUpdate({...table,error});
+        }
```
loitly, line 289: @vandesai1 made a good point about inconsistent messages. If you can distinguish between "no data found" vs an actual error, it's better to return an empty table instead of an error.

To return an empty table, update it with { error: null, totalRows: 0 }.

robyww: I will take a closer look at the Simbad error message and see it I can tell.

PR #1715, opened 2025-02-27
FIREFLY-1662: Support for Cutout Service Descriptor URL generation

kpuriIpac:
**Ticket**: [FIREFLY-1662](https://jira.ipac.caltech.edu/browse/FIREFLY-1662)
[IFE PR](https://github.com/IPAC-SW/irsa-ife/pull/383)
- I am mostly looking for feedback in `ObsCorePackager.java`, where support for Cutout Service Descriptors has been added 
- `CutoutServiceDescriptor` logic supports both kinds of service descriptors for cutouts we have come to expect
  - 1. Service Descriptors containing an `x-type=circle` entry
  - 2. Service Descriptors containing separate input params (ra, dec, size) like the current Euclid ones
  - The logic for finding the correct service descriptor (corresponding to the current cutout) among several service descriptors flows largely from our conversations @loitly & @robyww 
    - But I have been able to simplify the logic a bit: 
      - I am **_not_** explicitly doing a sync/async service descriptor check, and I am **_not_** checking the standardID 
        - Reason: Since I am able to identify service descriptors with `utype=adhoc:service`, I can then match this service descriptor's ID with the `service-def` of the cutout entry we're on, to ensure we process the correct service descriptor. 

**Updates 3/4**: 
- Addressed all feedback
- Download Script now works with non-obscore files containing datalink (so Objects Search in euclid works)

**Updates 3/5**
- CADC bug is now fixed, please test it out in the firefly build below 
  - you should now also see a list of products in the download dialog. Try selecting only a few (and unselected 'All data')
  - it should still give you a zipped file with your selection (if those products exist for the row you selected)
- Small changes to `DownloadScript.java` so that when the URL contains a query string, we look at `path` and `file` params in addition to the `return` param to try and construct a file name 
  - this currently helps cutouts in euclid, but we need to consider a more generalized solution in the future  

**Testing**: 
Euclid: https://firefly-1662-service-descriptor-url.irsakudev.ipac.caltech.edu/applications/euclid
Firefly: https://fireflydev.ipac.caltech.edu/firefly-1662-service-descriptor-url/firefly/
- do an Image search (like `57.717836123 -51.248639177`) and an Objects search
- you should be able to download cutouts now (and download any products for Objects search)
  - the semantics list you see in the download dialog now makes use of `fetchSemanticList`
  - try and run the download script to ensure the cutout URLs actually yield valid cutouts. You can even upload them back to firefly to ensure they look correct.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
@@ -2,29 +2,22 @@ package edu.caltech.ipac.firefly.server.query;
 
 import edu.caltech.ipac.firefly.data.DownloadRequest;
 import edu.caltech.ipac.firefly.data.FileInfo;
-import edu.caltech.ipac.firefly.data.HttpResultInfo;
 import edu.caltech.ipac.firefly.data.ServerRequest;
-import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil;
 import edu.caltech.ipac.firefly.server.network.HttpServiceInput;
 import edu.caltech.ipac.firefly.server.packagedata.FileGroup;
 import edu.caltech.ipac.firefly.server.util.Logger;
-import edu.caltech.ipac.table.DataGroup;
-import edu.caltech.ipac.table.MappedData;
+import edu.caltech.ipac.table.*;
 import edu.caltech.ipac.table.io.VoTableReader;
 import edu.caltech.ipac.util.FileUtil;
-import edu.caltech.ipac.util.download.URLDownload;
-import org.apache.commons.lang.ArrayUtils;
+import edu.caltech.ipac.util.StringUtils;
+import org.json.JSONObject;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
+import static edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil.getSelectedData;
 import static edu.caltech.ipac.util.FileUtil.getUniqueFileNameForGroup;
 import static org.apache.commons.lang.StringUtils.substringAfterLast;
 
@@ -36,10 +29,28 @@ public class ObsCorePackager extends FileGroupsProcessor {
             "csv", "tbl", "fits", "json", "pdf", "tar", "html", "xml", "vot", "vot", "vot", "vot", "reg", "png", "xml");
 
     public static final String PRODUCTS = "productTypes";
-    public static final String SCRIPT = "scriptType";
     public static final String ACCESS_URL = "access_url";
+    public static final String SERVICE_DEF = "service_def";
     public static final String ACCESS_FORMAT = "access_format";
     public static final String SEMANTICS = "semantics";
+    public static final String CONTENT_TYPE = "content_type";
+
+    public static final String DATALINK_SER_DEF = "datalinkServiceDescriptor";
+    public static final String ADHOC_SERVICE = "adhoc:service";
+    public static final String DATALINK = "datalink";
+    public static final String CENTER_COLS = "centerCols";
+    public static final String CUTOUT_VALUE = "cutoutValue";
+    public static final String SER_DEF_ACCESS_URL = "accessURL";
+    public static final String SER_DEF_INPUT_PARAMS = "inputParams";
+    public static final String CIRCLE = "circle";
+    public static final String REF = "ref";
+    public static final String VALUE = "value";
+    public static final String TEMPLATE_COL_NAMES = "templateColNames";
+    public static final String USE_SOURCE_FILE_NAME = "useSourceUrlFileName";
+
+    private static final List<String> CUTOUT_UCDs= Arrays.asList("phys.size","phys.size.radius","phys.angSize", "pos.spherical.r");
+    private static final List<String>  RA_UCDs= List.of("pos.eq.ra");
+    private static final List<String>  DEC_UCDs= List.of("pos.eq.dec");
 
     public List<FileGroup> loadData(ServerRequest request) throws IOException, DataAccessException {
         try {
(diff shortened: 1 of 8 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+
+        public CutoutServiceDescriptor(String partialUrl, List<MissingParam> missingParams) {
+            this.partialUrl = partialUrl;
+            this.missingParams = missingParams;
+        }
+    }
+
+    static class MissingParam {
+        String paramName;
+        String refId;
```
robyww, line 417: This should be records. They are much nicer to work with. IntelliJ will do the conversion for you if you make all the object variables `final`. 
```java
record CutoutServiceDescriptor(String partialUrl, List<MissingParam> missingParams) { }

record MissingParam(String paramName, String refId) { }

```

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+            for (ParamInfo param : inputParams) {
+                String ucd = param.getUCD();
+                if (isMatch(ucd, CUTOUT_UCDs) || isMatch(ucd, RA_UCDs) || isMatch(ucd, DEC_UCDs)) continue;
+
+                String key = param.getKeyName();
+                String value = param.getStringValue();
+
+                if (key == null || key.isEmpty()) continue;
+
+                if (param.getRef() != null && (value == null || value.isEmpty())) {
```
robyww, line 375: `StringUtil.isEmpty`

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+
+            //Append other valid params (where value is not null or empty) (excluding RA, DEC, CUTOUT)
+            for (ParamInfo param : inputParams) {
+                String ucd = param.getUCD();
+                if (isMatch(ucd, CUTOUT_UCDs) || isMatch(ucd, RA_UCDs) || isMatch(ucd, DEC_UCDs)) continue;
+
+                String key = param.getKeyName();
+                String value = param.getStringValue();
+
+                if (key == null || key.isEmpty()) continue;
```
robyww, line 373: `StringUtil.isEmpty`

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+
+                    if (isMatch(ucd, CUTOUT_UCDs)) cutoutName = param.getKeyName();
+                    else if (isMatch(ucd, RA_UCDs)) raName = param.getKeyName();
+                    else if (isMatch(ucd, DEC_UCDs)) decName = param.getKeyName();
+                }
+
+                if (raName != null && decName != null && cutoutName != null) {
+                    queryString.append(raName).append("=").append(ra).append("&")
+                            .append(decName).append("=").append(dec).append("&")
+                            .append(cutoutName).append("=").append(cutoutValue);
```
robyww, line 359: A much more readable way
```java
var s= String.format("%s=%s&%s=%s&%s=%s", raName, ra, decName, dec, cutoutName, cutoutValue);
queryString.append(s);
```

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+        for (GroupInfo group : groups) {
+            if (!"inputParams".equals(group.getName())) continue;
+
+            List<ParamInfo> inputParams = group.getParamInfos();
+
+            //Check for xtype = "CIRCLE"
+            for (ParamInfo param : inputParams) {
+                if ("CIRCLE".equals(param.getXType())) {
+                    primaryParamName = param.getKeyName();
+                    queryString.append(primaryParamName).append("=").append(ra).append("+").append(dec).append("+").append(cutoutValue);
```
robyww, line 338: use String.format, see my comment below

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+
+        // Step 2: Locate the "inputParams" group
+        for (GroupInfo group : groups) {
+            if (!"inputParams".equals(group.getName())) continue;
+
+            List<ParamInfo> inputParams = group.getParamInfos();
+
+            //Check for xtype = "CIRCLE"
+            for (ParamInfo param : inputParams) {
+                if ("CIRCLE".equals(param.getXType())) {
```
robyww, line 336: `equalsIgnoreCase`

You should use this everywhere for matching anything in the datalink fire.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+        }
+
+        // Step 2: Locate the "inputParams" group
+        for (GroupInfo group : groups) {
+            if (!"inputParams".equals(group.getName())) continue;
+
+            List<ParamInfo> inputParams = group.getParamInfos();
+
+            //Check for xtype = "CIRCLE"
+            for (ParamInfo param : inputParams) {
```
robyww, line 335: reserver works like "CIRCLE", "accessURL", "inputParams" should be constants at the top

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -240,4 +288,137 @@ else if (obs_title != null) { //option 2, if productTitleTemplate is not availab
...
+
+        // Step 1: Extract accessURL
+        for (ParamInfo param : params) {
+            if ("accessURL".equals(param.getKeyName())) {
+                accessUrl = param.getStringValue();
+                break;
+            }
+        }
+        if (accessUrl == null || accessUrl.isEmpty()) {
+            return new CutoutServiceDescriptor("", new ArrayList<>()); // Invalid service descriptor
```
robyww, line 325:
```java
return new CutoutServiceDescriptor("", Collections.emptyList());
```

robyww (commented):
Looking good so far. I many made little cleanup comments.

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -93,6 +96,11 @@ function getColNameFromTemplate(template) {
 const PrepareDownload = React.memo(() => {
     const tblTitle = getTblById(getActiveTableId())?.title ?? 'unknown';
     const baseFileName = tblTitle.replace(/\s+/g, '').replace(/[^a-zA-Z0-9_.-]/g, '_');
+
+    //todo: ra dec value could change based on Cutout settings (search target center, table row position or entered position)
+    const centerCols = findTableCenterColumns(getActiveTableId());
+    const cutoutValue= useStoreConnector( () => getCutoutSize(DEFAULT_DATA_PRODUCTS_COMPONENT_KEY));
```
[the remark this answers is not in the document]
kpuriIpac: that must be why the cutouts were not working for euclid as well (i was getting only default value of 0.01), I'll fix this.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -82,7 +88,7 @@ private List<FileGroup> computeFileGroup(DownloadRequest request) throws DataAcc
                 if (access_format != null) {
                     if (access_format.contains("datalink")) {
                         ext_file_name = getUniqueFileNameForGroup(ext_file_name, dataLinkFolders);
-                        List<FileInfo> tmpFileInfos = parseDatalink(url, ext_file_name, products, useSourceUrlFileName);
+                        List<FileInfo> tmpFileInfos = parseDatalink(url, ext_file_name, products, useSourceUrlFileName, ra, dec, cutoutValue);
```
loitly, line 91: In my opinion, most of these parameters are specific to `parseDatalink`.  Extracting them from the top just to pass it along here makes the code confusing.
I would have place all of `parseDatalink` logic in that one function, like this:
```
List<FileInfo> tmpFileInfos = parseDatalink(url, request, dgDataUrl.getRow(int idx));
```
When you do this, you will see that almost half of the code in this function is no longer needed.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -56,22 +55,29 @@ private List<FileGroup> computeFileGroup(DownloadRequest request) throws DataAcc
...
                     continue; //no file available to process
                 }
                 String access_format = (String) dgDataUrl.get(idx, "access_format");
                 URL url = new URL(access_url);
 
-                FileInfo fileInf1o1 = new FileInfo(access_url, null, 0);
-                FileInfo fileInf1o2 = new FileInfo(access_url, ".txt", 0);
-
                 String colNames = request.getSearchRequest().getParam("templateColNames");
                 boolean useSourceUrlFileName= request.getSearchRequest().getBooleanParam("useSourceUrlFileName",false);
```
loitly, line 82: In my opinion, the logic for resolving `ext_file_name` is incorrect.
Unless we know what the filename should be, we should defer naming it until the file is actually downloaded. At that point, the `Content-Disposition` header can be inspected to determine the proper filename.

loitly (commented):
I'm just leaving comments as suggestions. Trey should be the one to approve this. My suggestions might require a lot of work, so I completely understand if you decide not to implement them.

robyww (approved):

src/firefly/js/ui/dynamic/ServiceDefTools.js
```diff
@@ -526,3 +527,37 @@ export function ingestInitArgs(fdAry, args) {
     });
 }
 
+//check for and return a datalink service descriptor url and input params, if one is found
+export function checkForDatalinkServDesc(tblModel) {
+    const serviceDescriptors = getServiceDescriptors(tblModel);
+
+    if (!serviceDescriptors) return null;
```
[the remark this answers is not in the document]
kpuriIpac: @robyww I'm using this in EuclidUtils and ttFeatureWatchers to check if there's a datalink service descriptor, and if there is, then returning the accessURL + inputParams (to construct the full URL server side). 

Should I move this function elsewhere then? I figured ServiceDefTools would be right place for it

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -91,8 +93,18 @@ function getColNameFromTemplate(template) {
...
-    const tblTitle = getTblById(getActiveTableId())?.title ?? 'unknown';
+    const tbl_id = getActiveTableId();
+    const tblTitle = getTblById(tbl_id)?.title ?? 'unknown';
     const baseFileName = tblTitle.replace(/\s+/g, '').replace(/[^a-zA-Z0-9_.-]/g, '_');
+
+    const tblModel = getTableModel(tbl_id);
+    const cutoutValue = useStoreConnector( () => getCutoutSize());
+
+    const tblInfo = getTblInfo(tblModel);
+    const cutoutTargetVals = findCutoutTarget('', {}, tblModel, tblInfo?.highlightedRow);
```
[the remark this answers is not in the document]
kpuriIpac: fixed in new code

lrebull:
in the email i get, the backslash that escapes the * in *.zip didn't make it into the email. that's important because unzip doesn't take just the wildcard; it requires "backslashasterisk" not just "asterisk" .. that caused no end of grief for new users of the SHA. This is all assuming you are delivering zip files in this -- i have yet to get a zip file. so far, i'm getting raw fits files. but i've just started testing ... if you're not delivering zip files, then you don't have to tell people how to unzip.

lrebull:
the pop-up gives me a choice of curl OR wget OR plain text, but the email delivers all three.
moreover, I told it to generate just curl, but it's given me all three curl, wget, and plaintext.
obviously this is not a showstopper because it's giving me more than i asked for, but it seems weird.

lrebull:
I am not 100% sure what i am looking at here. I get cutouts but I don’t appear to get a catalog at least not one that appears to match what is shown in the GUI window, and the catalog that is shown in the GUI window doesn’t seem to correspond in terms of location on the sky with the cutouts I have. This is what I did:

1. searched on a region
2. selected a thing to download
3. downloaded stuff
4. loaded images and catalogs from the download back into your fireflybuild
5. nothing from that catalog appears on the image.
6. went back to the euclid window, saved that catalog as a tbl file
7. went back to the firefly window, overlaid that tbl file
8. nothing from that catalog appears on the image either.

this all could just be that i am not enough of an euclid expert to understand what i am looking at

PR #1716, opened 2025-02-27
Firefly-1663: fixed multiple MOC issues

robyww:
#### [Firefly-1663:](https://jira.ipac.caltech.edu/browse/FIREFLY-1663) fixed multiple MOC issues

 - MOC not redrawing properly when more than one MOC in on screen
 - MOC now draws correctly even when tiles wrap
 - MOC drawing is a little faster

#### Test
- https://firefly-1663-moc-issues.irsakudev.ipac.caltech.edu/applications/euclid
- Euclid planned MOC should show correctly in both Spherical and Aitoff

### DIFF
--- a/src/firefly/js/visualize/draw/Drawer.js
+++ b/src/firefly/js/visualize/draw/Drawer.js
@@ -3,13 +3,13 @@
  */
 
 
-import {get, isFunction, isEmpty, isArray} from 'lodash';
-import Point, {makeScreenPt,makeImagePt,pointEquals} from '../Point.js';
+import {isFunction, isEmpty, isArray} from 'lodash';
+import {makeScreenPt,makeImagePt} from '../Point.js';
 import {dispatchAddTaskCount, dispatchRemoveTaskCount, makeTaskId, getTaskCount} from '../../core/AppDataCntlr.js';
 import BrowserInfo, {Browser} from '../../util/BrowserInfo.js';
 import {Style} from './DrawingDef.js';
 import DrawUtil from './DrawUtil.js';
-import Color, {getRGBA, toRGBAString} from '../../util/Color.js';
+import {getRGBA, toRGBAString} from '../../util/Color.js';
 import CsysConverter, {CCUtil} from '../CsysConverter.js';
 import {POINT_DATA_OBJ} from './PointDataObj.js';
 import {DrawingType} from './DrawObj.js';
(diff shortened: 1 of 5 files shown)


lrebull:
I am super late to the party here, for which I apologize. Here is what I see on test, centered on 0,0 ra dec (note not 0,0 gal!!):
Screenshot 2025-02-27 at 9 41 53 AM
and here is what i see on this build:
Screenshot 2025-02-27 at 9 42 01 AM
they sure are different! but i don't know enough to know which is right. the top center has less coverage in the new one, which was not what i was expecting given the comments in Slack. The edges seem less confusing in the new one in any case.

vandesai1:
The MOC is supposed to look something like this in Aitoff projection with center = Galactic.

image

vandesai1:
it used to look like this in our Euclid app:

image

vandesai1:
In this pull request it looks like this:

image

gpdf:
The two screenshots @lrebull posted do seem to be significantly different projections, even of the underlying HiPS image.

lrebull:
@vandesai1 it only sort of looks like what it's supposed to now. 
Screenshot 2025-02-27 at 9 56 43 AM

lrebull:
yes, mine are of different projections. but changing it to match the projection, i still get something new that isn't in vandana's "what it should look like" screenshot:
Screenshot 2025-02-27 at 9 56 43 AM

lrebull:
admittedly, the rest of it looks much better!!

vandesai1:
@lrebull , that's a different ticket, though. This one was about edge effects. The underlying data that was used to plot the Euclid PR image is clearly different than what we were given. I've seen other Euclid PR images that show it slightly differently, for example this one:

image

vandesai1:
So I think that this PR should be approved, and I will double-check with Harry to see if we have been delivered the latest underlying data.

lrebull:
ok, got it. thanks for the help!!

src/firefly/js/visualize/draw/Drawer.js
```diff
@@ -515,31 +426,14 @@ function initOffScreenCanvas(width, height) {
...
+    // let i=0;
     const id= window.setInterval( () => {
-        if (params.done) {
-            window.clearInterval(id);
-        }
+        if (params.done) window.clearInterval(id);
+        // console.time('drawing ' +i);
         drawer.doDrawing(params);
+        // console.timeEnd('drawing ' +i);
+        // i++;
```
kpuriIpac, line 436: a few commented out lines that you may want to remove before merging

kpuriIpac:
The issue with planned moc coverage seems to be resolved from what I can tell (and based on the conversation above & @vandesai1's feedback). I did notice that MOCs (not just the planned MOC) sometimes go beyond the background image (see screenshot below). Just FYI in case this needs to be addressed in a future firefly ticket, or maybe it is intentional? (indicating they wrap around?)  

euclid_moc_pr

kpuriIpac (approved):
Looks good to me!

PR #1717, opened 2025-02-28
Firefly-1666: Update data link to recognize extension columns and imp…

robyww:
#### [Firefly-1666](https://jira.ipac.caltech.edu/browse/FIREFLY-1666): Update data link to recognize extension columns and improve grid arrangement

- Image grid now group, order, and switch correctly

#### Testing
- https://firefly-1666-grid-grouping.irsakudev.ipac.caltech.edu/applications/euclid
- now can select which image in single or table grid mode
- when switch back and forth is should do to the right one
- images are ordered as expected

### DIFF
--- a/src/firefly/js/drawingLayers/Catalog.js
+++ b/src/firefly/js/drawingLayers/Catalog.js
@@ -164,7 +164,8 @@ function saveLastDown(mouseStatePayload) {
  * @param mouseStatePayload
  */
 function highlightChange(mouseStatePayload) {
-    const {drawLayer,plotId,screenPt}= mouseStatePayload;
+    const {drawLayer,plotId,screenPt,shiftDown}= mouseStatePayload;
+    if (shiftDown) return;
     const plot= primePlot(visRoot(),plotId);
     const center= getCenterOfProjection(plot);
     if (lastProjectionCenter && (!pointEquals(center, lastProjectionCenter?.center) || lastProjectionCenter.plotId!==plotId)) return;
--- a/src/firefly/js/drawingLayers/hpx/HpxCatalog.js
+++ b/src/firefly/js/drawingLayers/hpx/HpxCatalog.js
@@ -138,7 +138,8 @@ function saveLastDown(mouseStatePayload) {
  * @param mouseStatePayload
  */
 function highlightChange(mouseStatePayload) {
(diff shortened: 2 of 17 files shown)


vandesai1:
I tried a southern and a northern pointing, and got a different mixture of telescopes/instruments, but they are both in the correct order.

vandesai1:
If I am in the table grid view, I can change the filter nicely with the More menu. 

If I am in the default grid view, and choose MegaCam/R just by highlighting it, and then switch to table grid view, it still shows me all the VIS images. I'm not sure if I was supposed to be able to choose filters in this way, though.

robyww:
> If I am in the default grid view, and choose MegaCam/R just by highlighting it, and then switch to table grid view, it still shows me all the VIS images. I'm not sure if I was supposed to be able to choose filters in this way, though.

That is not supported right now. We can add the in the future.

vandesai1:
The results of the image search seem to work as expected. The results of the Inspect Objects have the label and bandpass in the datalink file, but they don't seem to be being used. There is an option to see the table sort, but it doesn't seem to work.

vandesai1:
@lrebull , I don't think you are supposed to be able to use the 'More' menu when using the row grid, because it's already showing you everything.

I can't reproduce your issue with searching on 18h03m55.62s +65d45m10.0s Equ J2000, choosing the table grid icon, and getting only two choices from the 'More' menu.

Did you just do the default search? Did it return 10 rows?

kpuriIpac:
@vandesai1 I do see what @lrebull sees (only two options) when selecting that icon on an image search: 

grid

vandesai1:
I can also reproduce it with an "image contains target" search.

lrebull:
see attached screencapture (it seemed easier than trying to explain)

https://github.com/user-attachments/assets/bb63e932-8a84-4a46-9063-b0b6f3df437d

lrebull:
> Did you just do the default search? Did it return 10 rows?

no, at no point did i get 10 rows. i got two rows in the table and 9 images for each of the two rows from the table, except when choosing that middle icon from the display choices.

robyww:
@lrebull @vandesai1 @kpuriIpac Just made a build with all bugs fixed.

lrebull:
it is working much better for me now!

kpuriIpac (approved):
looks good! after the changes the bugs all seem fixed too

PR #1718, opened 2025-02-28
IRSA-6701: LVF image search minimized dialog incorrectly shows polygon search

jaladh-singhal:
Fixes [IRSA-6701](https://jira.ipac.caltech.edu/browse/IRSA-6701)

See ife PR: https://github.com/IPAC-SW/irsa-ife/pull/385

Refactored `SearchSummary` to:
- replace hardcoded strings `{'Cone', 'Polygon', 'Multi-Object'}` with `*_CHOICE_KEY`
- get search type label to display from toggleOptions (in default case)
- accept `getSearchType` in place of `getSummaryInfo` because `searchType.value` and `searchType.label` are the only things that need to be customized (in LVF case), rest all follows default logic to generate key value pairs from these two variables 
- simplify redundant logic and overall cleanup

## Testing 
https://irsa-6701-spherex-search-summary.irsakudev.ipac.caltech.edu/applications/euclid/

Do regression testing in Euclid searches to see if search summary is working as before

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -134,7 +134,7 @@ export function EmbeddedPositionSearchPanel({
 
     if (useUpload && !sizeEnabled && enabledSearchTypes.length===2) {
         // because in this case, 'Cone' or 'Polygon' label won't make sense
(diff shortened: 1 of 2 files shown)


robyww (approved):
Looks much better!

kpuriIpac (approved):
Looks good! did regression testing on Euclid, works as expected

PR #1723, opened 2025-03-07
IRSA-6653: and several other smaller ticket items 

kpuriIpac:
Changes to support several Euclid tickets (but generic for firefly) 
- See [IFE PR ](https://github.com/IPAC-SW/irsa-ife/pull/387)


src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -190,10 +190,11 @@ export function DownloadOptionPanel ({groupKey='DownloadDialog', cutoutSize, hel
...
                                    onMaskComplete={() =>hideDownloadDialog()}/>);
 
-    const saveAsProps = {
+    saveAsProps = {
         initialState: {
             value: get(dlParams, 'BaseFileName')
-        }
+        },
+        ...saveAsProps
     };
```
robyww, line 198: `saveAsProps` should really be done a `slotProps` but since we are up against a deadline it will have to be done at another time.

robyww (approved):

src/firefly/js/ui/DownloadDialog.jsx
loitly: DownloadDialog was designed to be composable.  Several of these 'show' flags to render custom components should not be here, but passed in as `children`.

PR #1725, opened 2025-03-12
Firefly-1680: FITS image optimizations

robyww:
#### [Firefly-1680](https://jira.ipac.caltech.edu/browse/FIREFLY-1680): FITS image optimizations

 - Flip the image in place without memory allocation, this also means the interation is cut in half
 - Read fits data during render process, this way the data does not have to be cached. Makes for better GC.
 - build large data histogram with threads
 - getByteStretchArray now only allows one call at a time per session using semaphores
      - radically improving memory managment and not really affecting performance
 - Only cache image on second image read.
     - This indicates the user is changing stretch or the user has gone back to this image.
     - Stretch changes often happen in groups.
     - If the user is not changing stretch more moving back and forth then there is not reason to cache the data

#### Testing
- https://firefly-1680-optimize-fits.irsakudev.ipac.caltech.edu/applications/euclid/
- Everything should work as before
- Try loading full images in a grid

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/astro/net/HorizonsEphPairs.java
+++ b/src/firefly/java/edu/caltech/ipac/astro/net/HorizonsEphPairs.java
@@ -58,9 +58,12 @@ public class HorizonsEphPairs {
 
         try {
             String urlStr = horizonsServer + path + "?sstr="+ URLEncoder.encode(idOrName, "UTF-8");
-            URL url = new URL(urlStr);
-            String jsonStrResult = URLDownload.getDataFromURL(url, null,null,null).getResultAsString();
-            JsonHelper json= JsonHelper.parse(jsonStrResult);
+            var result = URLDownload.getDataFromURL(new URL(urlStr), null,null,null);
+            if (result.getResponseCode() != 200) {
+                throw new FailedRequestException("failed to retrieve ephemeris pairs, response code: " + result.getResponseCode());
+            }
+            var json= JsonHelper.parse(result.getResultAsString());
+
 
             JSONArray rList= json.getValue(new JSONArray(), "result");
             List<HorizonsResults> horizonsResults= new ArrayList<>();
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
@@ -53,6 +53,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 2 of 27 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/visualize/VisServerOps.java
```diff
@@ -69,6 +70,10 @@ public class VisServerOps {
     static {
         VisContext.init();
     }
+    private static final CacheKey BYTE_STRETCH_SEMAPHORE = new StringKey("BYTE_STRETCH_SEMAPHORE");
+    private static final int MAX_SEMAPHORE_WAIT_SECS = 360;
+    private static final int BYTE_STRETCH_MAX_ACCESS_CNT = 1;
+    private static final String BYTE_STRETCH_SEMAPHORE_KEY = "BYTE_STRETCH_SEMAPHORE";
```
loitly, line 76: Are you expecting there will be cases where BYTE_STRETCH_MAX_ACCESS_CNT might need to be greater than 1?
Is that why you're opting for a Semaphore instead of a Lock (e.g., ReentrantLock)?

robyww: This allows for making it 2 or 3 in the future. It is more flexible. There is nothing about the function that requires a lock, it is just a way to slow traffic so a semaphore make the most sense.

loitly (approved):
Tests are passing now. Thanks for the fix.

src/firefly/java/edu/caltech/ipac/astro/net/HorizonsEphPairs.java
```diff
@@ -58,9 +58,20 @@ public static HorizonsResults[] lowlevelGetEphInfo(String idOrName) throws Faile
...
+                throw new FailedRequestException("failed to retrieve ephemeris pairs, response code: " + result.getResponseCode());
+            }
+            String jsonStrResult = result.getResultAsString();
+//            test using HttpServices
+//            OutputStream resultStream = new ByteArrayOutputStream();
+//            HttpServices.Status stat= HttpServices.getData(urlStr, resultStream);
+//            if (stat.isError()) {
+//                throw new FailedRequestException("failed to retrieve ephemeris pairs, response code: " + stat.getStatusCode());
+//            }
+//            String jsonStrResult = resultStream.toString();
```
kpuriIpac, line 72: did you intend to remove this code or keep it for testing purposes?

robyww: I should remove it now

kpuriIpac (approved):
I had to do some reading on FITS and why it needs to be flipped in the first place to understand these changes better. I know it was part of our code already, but I don't think I've touched this part of the code. So it was good to get an idea of why flipping is necessary to display the image to the user...I was aware of us flipping the FITS already, but just reading deeper into it gave me a better sense of how it works. 

Looks good!

PR #1726, opened 2025-03-12
FIREFLY-1378: Chart filter icon does not work on column mapped to a label

loitly:
Tickets:
Chart filter popup does not work when x/yis column
https://jira.ipac.caltech.edu/browse/FIREFLY-1378

HttpServiceInput supports multiple values with the same parmeter key:
https://jira.ipac.caltech.edu/browse/FIREFLY-1679

- Replaces chart filter popup with Table Setting popup but limit to filtering only
- Also, added FIREFLY-1679: Support for multiple params (but same key) for HttpServiceInput

Test:
https://fireflydev.ipac.caltech.edu/firefly-1378-chart-filter-popup/firefly/test/tests-chart.html?test=Chart+with+columns+map+with+Display+Name

A new test has been created to use columns with labels.
Test chart filtering using the filter icon on the chart toolbar. 
Note that the popup now differs from the previous version; it reuses the table component, which promotes code reuse and provides a more robust advanced filter tab.


kpuriIpac (approved):
Looks great. Tested out the filters per the ticket & the test build instructions, works as expected. Code changes in `HttpServices` look good too.

BTW, do you think are we going to (or should) do something similar for `URLDownload` as well?

PR #1734, opened 2025-03-26
FIREFLY-1653: Create a WavelengthInputField component to handle trailing units

jaladh-singhal:
Fixes [FIREFLY-1653](https://jira.ipac.caltech.edu/browse/FIREFLY-1653)

- Added `WavelengthInputField` and `WavelengthRangeInput` components that can handle units changes and validation
- Updated ObsTAP `WavelengthPanel` to use these components (cleaned up TAP/SIA constraints generation logic because these component handle that already)
- Made floatRange validator accept custom error message
- Now SPHEREx instances can define min max bounds on wavelength as well as initial value and initial unit and the component will handle input validation

See IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/396

# Testing
https://firefly-1653-wvl-input-component.irsakudev.ipac.caltech.edu/applications/spherex/

Go to LVF search -> **_Uncheck "Location"_** -> "Spectral Coverage" -> select:
- "By Wavelength" and "contains": 
  - enter `8` -> should show error message because out of bounds
  - enter `4.72` -> change units -> it should update field as per conversion logic (similar to SizeInputField does)
  - "search" -> `2043` rows should appear. Cross check if ObsTAP constraint on `em_min` and `em_max` column was right by switching to "By Filter Bands" -> check "Band 6" checkbox -> search -> `2043` rows should appear
- "By Wavelength" and "overlaps":
  - enter `.5` in max wavelength input -> should show error message
  - enter `5.8` in max wavelength input -> should show error message
  - enter `5` in max wavelength input and `4.45` in min wavelength input -> search -> `2043` rows should appear (since we constrained to only Band 6 by wavelength range)

## Regression Testing 
https://fireflydev.ipac.caltech.edu/firefly-1653-wvl-input-component/firefly

Perform SIA or ObsTAP searches, and see if wavelength constraints are generated correctly when using wavelength range or wavelength input fields. E.g.:
- SIA: CADC m81, 200 nm -> all rows have em_min <= 2e-7 <= em_max
- ObsTAP: just use ADQL edit screen to see constraints

> Note: now that we have bounds validation on wavelength input, we can't work with -Inf and Inf (those are not physically possible for wavelength anyways). I have chosen 0.1 meter to 0.1 Angstrom as default min, max based on EM spectrum classification. We can change it later if scientists disagree. There are also floating point precision issues that need to be taken care of.

### DIFF
--- a/src/firefly/js/ui/WavelengthInputField.jsx
+++ b/src/firefly/js/ui/WavelengthInputField.jsx
@@ -1,23 +1,19 @@
 import React, {memo} from 'react';
 import PropTypes from 'prop-types';
-import {Divider, FormHelperText, Stack} from '@mui/joy';
+import {Divider, FormHelperText, Stack, Typography} from '@mui/joy';
+import {isNaN, memoize} from 'lodash';
 import {ListBoxInputFieldView} from 'firefly/ui/ListBoxInputField';
 import {useFieldGroupConnector} from './FieldGroupConnector.jsx';
 import {InputFieldView} from 'firefly/ui/InputFieldView';
 import Validate from 'firefly/util/Validate';
-import {memoize} from 'lodash';
+import {useFieldGroupValue} from 'firefly/ui/SimpleComponent';
+import {WAVELENGTH_UNITS} from 'firefly/visualize/VisUtil';
 
+// following strings are the keys in WAVELENGTH_UNITS object
 export const BASE_UNIT = 'um';
-const WAVELENGTH_UNITS = {
-    // key: { m: meters equivalent, label: full string that appears in dropdown, symbol: shorter string that appears elsewhere }
-    m: { m: 1, label: 'meters', symbol: 'm' },
-    [BASE_UNIT]: { m: 1e-6, label: 'microns', symbol: 'µm' },
-    nm: { m: 1e-9, label: 'nanometers', symbol: 'nm' },
-    angstrom: { m: 1e-10, label: 'angstroms', symbol: 'Å' },
-};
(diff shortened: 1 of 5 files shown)


robyww (approved):
The code looks good.

#### UI

I think you should use the abbreviations.
- nanometer: `nm`
- angstrom: unicode character: `Å`
- micron: see line below
https://github.com/Caltech-IPAC/firefly/blob/d4de6cd4f75f09a67cc837dc251263e01dd89090/src/firefly/js/visualize/PlotViewUtil.js#L952

You could move all three as constants `VisUtil.js` or `WebUtil.js`

jaladh-singhal:
> I think you should use the abbreviations.

@robyww you mean in the dropdown?

jaladh-singhal:
@robyww Just FYI, modern JS files support UTF-8 encoding out-of-the-box so we don't have to dynamically generate unicode character from a character code. It's not very readable either.

I took your tip and exposed WAVELENGTH_UNITS constant into VisUtil and now this PlotViewUtil example and WavelengthInputField is accessing micron symbol from there, maintaining a single source of these defintions.

loitly (changes requested):
Code looks clean.  Nice work.
However, I found a bug `WavelengthRangeInput` validation.
- Go to SIA → expand Spectral Coverage → select Overlaps
- Enter `10` for max
- Now try entering any value for Min.  No out-of-range error is triggered, even when it should be.

jaladh-singhal:
>Now try entering any value for Min. No out-of-range error is triggered, even when it should be.

@loitly good catch - min wavelength component wasn't doing paired validation, I added it.

And this was happening not just in SIA but everywhere including spherex.

loitly (approved):
It's working now.  Thanks.

PR #1736, opened 2025-04-01
Firefly-1696: changes for url api to work with a routed app and EmbededPositionSearchPanel

robyww:
#### [Firefly-1696](https://jira.ipac.caltech.edu/browse/FIREFLY-1696): changes for url api to work with a routed app and EmbeddedPositionSearchPanel

   - [Firefly-1681](https://jira.ipac.caltech.edu/browse/FIREFLY-1681): fixes: give more details that 1D images are loaded as tables
   - [Firefly-1690](https://jira.ipac.caltech.edu/browse/FIREFLY-1690): fixed distance tool: image pixel offset is wrong on some images when wcs matched and in pixel mode

#### Testing
- https://firefly-1696-euclid-urlapi.irsakudev.ipac.caltech.edu/applications/euclid/
- Load example page: https://firefly-1696-euclid-urlapi.irsakudev.ipac.caltech.edu/applications/euclid/?api
- try the examples
- make variations

#### Testing Firefly-1690
- https://fireflydev.ipac.caltech.edu/firefly-1696-euclid-urlapi/firefly
- Test according to the ticket.

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -3,8 +3,7 @@ import React, {Fragment, useContext, useEffect, useState} from 'react';
 import {oneOf, bool, string, number, arrayOf, object, func, shape, elementType} from 'prop-types';
 import {defaultsDeep} from 'lodash';
 import CoordinateSys from '../../visualize/CoordSys.js';
-import {CONE_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY
-} from '../../visualize/ui/CommonUIKeys.js';
+import {CONE_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY} from '../../visualize/ui/CommonUIKeys.js';
 import {HiPSTargetView} from '../../visualize/ui/TargetHiPSPanel.jsx';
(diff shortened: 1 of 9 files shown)


src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -82,6 +82,7 @@ export const emptyHeaderSx = {
  *
  */
 export function EmbeddedPositionSearchPanel({
+                                                initArgs={},
```
loitly, line 85: Why pass `initArgs` instead of getting them from context?

loitly (approved):
`initArgs` as context looks good.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
robyww: I wanted to leave the option here to control if `EmbeddedPositionSearchPanel` supports init args. I think there are be cases (with DCE) when it should not so I decided to just pass it.

kpuriIpac (approved):
Looks good! Tested 1681 (i see the updated description here - should the `Load Table` button also read `Load as Table`, per the ticket request?) and 1690 as well. Both look good.

lrebull:
I confirm the distance tool seems to work properly now. I still get some mildly unexpected behavior with pixels for Herschel images but I think that is related to the size of the pixels and the 0,0 origin.

PR #1737, opened 2025-04-01
FIREFLY-1691-job-monitor: Replace popup Background Monitor with Tab-Based UI

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1691
This PR replaces the current popup-style background monitor with a redesigned, tab-based Background Monitoring UI.
See ticket for more details.

Default cleanup policy: (configurable)
- Jobs will be archived 7 days after termination.
- Archived jobs will be removed after 7 days.

Auto-send Searches:
- Added an app option to allow regex matching of searches to automatically send to the background.
- Non-backgroundable searches are unaffected.

Test: https://fireflydev.ipac.caltech.edu/firefly-1691-job-monitor/firefly/
Default behavior.  Should work as before, but with the new UI.

Test: https://firefly-1691-job-monitor.irsakudev.ipac.caltech.edu/applications/wise/
(Will not merge)
- Uses "Create Script" for downloads. This will generate jobs with multiple results. The UI differs when there are multiple items to download.
- All background-able searches will be automatically sent to the background:
  - TAP, IRSA Catalog, etc.
  - WISE searches are not background-able and will remain unchanged.


robyww (commented):
This is just initial UI and behavioral feedback. I might have more by the end of the day.

- I like the overall approach. Having the it in the tabs seems cleaner and easier to understand (I am not quite sure why that is).
- I like the overall layout of the table, particularly the header above the table.

#### Suggestions
- Background Monitor should show some sort of working as part of the tab, use the same one as Results (a slower less distracting spinning)
- Maybe Background monitor tab should show a count?
- It should have a clear all button
- The load button is too small. I don't think the load button should be an icon. It should say `Load`
- maybe `COMPLETED` should be clickable and should load?
- We should rename the tab:  Background / History ?
- Menu resizer needs to be fixed, (I will point this out in the code review)
- It needs to show jobs not specificly backgrounded. This is important to the history aspect.
- The searches table need some sort of description, maybe two columns
    - Almost all the searches that firefly does have a `DATA_SERVICE_ID` meta data entry. I think we should show this or at least try it out
    - We should also show the title in request

robyww (commented):
Code looks good.

A couple of notes on `Menu.jsx`

- I think we can remove `MenuItemButton()` since it is not longer used.
- In `getButtonSize()` we should no longer do the background monitor button adjustment. Line 300, `bgMon` variable should be remove.

loitly:
> * Background Monitor should show some sort of working as part of the tab, use the same one as Results (a slower less distracting spinning)
Sure.

> * Maybe Background monitor tab should show a count?
The number may get big.  Let's get for more feedbacks.

> * It should have a clear all button
Clear what?  Filters?

> * The load button is too small. I don't think the load button should be an icon. It should say `Load`
The buttons under 'Control' are distinguished by color; green, orange, then red.
There are 3 types of actions in the same slot: Load Results, Download file, Show multiple files for download.
They are 'green'.  I can make them 'solid' or 'soft' to make them stand out more.
There's also lots of space.  We can have a 'legend' to the right to map each icon to its description.
Let's get more feedbacks.

> * maybe `COMPLETED` should be clickable and should load?
I don't think this is a good idea.  Let's get more feedbacks.

> * We should rename the tab:  Background / History ?
Let's make this configurable.  For Rubin where every Search go directly to 'History', this make sense.
But, for IRSA, they are not 'History'.  The user has to click 'Send to Background' to have the Search show up there.

> * Menu resizer needs to be fixed, (I will point this out in the code review)
Yes.  I was going to ask you. `showBgMonitor` flag is sent all over the places.

> * It needs to show jobs not specificly backgrounded. This is important to the history aspect.
This feature is added and used in [WISE](https://firefly-1691-job-monitor.irsakudev.ipac.caltech.edu/applications/wise/) for testing.  If it is used it in Rubin, all Searches will go directly to 'History'.

> * The searches table need some sort of description, maybe two columns
>   * Almost all the searches that firefly does have a `DATA_SERVICE_ID` meta data entry. I think we should show this or at >   * We should also show the title in request
least try it out
This require adding additional column(s).  These columns need to be populated for 'Download' job type as well.  Let's get more feedbacks.

robyww:
> This feature is added and used in [WISE](https://firefly-1691-job-monitor.irsakudev.ipac.caltech.edu/applications/wise/) for testing.
I tied it and it does not work for me.

robyww:
> The searches table need some sort of description, maybe two columns

I see this a a must do. There is not enough information at this point to really be useful.  You have to click three time to get a job parameters. Once there are over 10 - 15 jobs I don't think it will be usable with out some sort of description.

robyww:
Just looked at Rubin. They use a different font so the columns are not wide enough. Do on search and look at the background tab.

loitly:
> Just looked at Rubin. They use a different font so the columns are not wide enough. Do on search and look at the background tab.

It's not just for Rubin, but all of our SIA searches. I think there's a bug in the SIA panel that's not setting the correct value.
I was going to fix it in my next ticket, but I do that here if you like.

robyww:
I don't think `Job History` is the right name for the tag. I think `Job Monitor` is better with history implied. 

This is overkill. This user does not need or cannot process this much information
Screenshot 2025-04-09 at 10 17 38 AM

I liked what we have before better.
Screenshot 2025-04-09 at 10 19 49 AM

You could do this and just call the button `Send to Job Monitor`. It does not matter that it is already there. I like have a cancel. clicking it could kill the job and remove it from the history.

src/firefly/js/ui/Menu.jsx
```diff
@@ -218,18 +191,16 @@ function MenuTabBar({menuTabItems=[], size, selected, dropDown, displayMask, set
...
-                    <Tab {...{ key: idx, value:action, disableIndicator:true, color, variant,
-                        ref: (c) => setElement(idx+'',c),
-                        sx: (theme) => ({ ...setupTabCss(theme,size) }) }} >
-                        {label}
-                    </Tab>);
+                const tabProps = {key: idx, value:action, disableIndicator:true, color, variant,
+                    ref: (c) => setElement(idx+'',c),
+                    sx: (theme) => ({ ...setupTabCss(theme,size) })
+                };
+                const tab= TabRenderer ? <TabRenderer {...tabProps} /> : <Tab {...tabProps}> {label}</Tab>;
```
robyww, line 203: The ref is not being calls via `TabRenderer`. the setElement builds a object of all the tab widths for the  font size computation. The `Job Manager` tab via `TabRender` is never calling setElement so the size computation is wrong.

src/firefly/js/core/background/JobHistory.jsx
```diff
@@ -0,0 +1,358 @@
...
+    const label = getAppOptions()?.background?.history?.label || 'Job History';
+    const TabRenderer =  React.forwardRef((props, ref) => {
+        const {jobs={}} = useStoreConnector(() => getBackgroundInfo());
+        const loading = Object.values(jobs).some((j) => isActive(j));
+        return (
+            <Tab {...props}>
+                {loading && <ListItemDecorator>{workingIndicator}</ListItemDecorator>}
+                {label}
+            </Tab>
+        );
```
robyww, line 72: Add `ref` to the tab props.

I don't quite understand `forwardRef` (I think react is removing it in react 19). The `ref` object appears to be populated but the `setElement` function is not being called. It works if you add it to the props.
```
 <Tab {...{...props, ref}}>
                {loading && <ListItemDecorator>{workingIndicator}</ListItemDecorator>}
                {label}
 </Tab>
```

robyww (changes requested):
### Fixing the  tab sizing

#### `JobHistory.jsx` changes

See my comments in the code.

#### `Menu.jsx` changes

Readd the change I sugested last time.

In `getButtonSize()` we should no longer do the background monitor button adjustment. Line 274, bgMon variable should be remove.


line 275: should be:
```js
    const bTotal= sumTabWidth + (showHelp?50:0) + (showUserInfo?50:0);

```

the `showBgMonitor` parameter should be removed everywhere. It is passed around and only used in 

`getButtonSize()` and `getButtonDisplayMask()`

robyww:
I almost forgot....

The panel needs to show the policy for keeping jobs. Probably on the top such as
```
Jobs saved for 1 month or the last 100
```
We can discuss the policy but it should be in the UI.

src/firefly/js/ui/Menu.jsx
loitly: My fault;  I correctly use `React.forwardRef` but forgot to set/forward the `ref`.  I'll fix it.
I'll also try to 'remove ALL showBgMonitor', there's so many.  As long as you say it's safe, I'll proceed.

loitly: That was it.  It's fixed now.  Test URLs updated.

lrebull:
My $0.02 of reaction:

- huh. ok, I think I see where you're going with this, but at least for the WISE incarnation you built, i actually can’t get anything to go into it! because the download data is now “create download scripts”, everything is so fast that nothing goes into the background. i guess that means we can turn off that tab from appearing by default in some tools?
- I think I like "Job Monitor" rather than "Job History". It sounds .. less invasive? I don't know.
- the "Job Summary" in the upper left is the same blue as a link, but it isn't a link as far as I can tell. this seems weird and inconsistent.
- presumably, this is limited to this session, not over my lifetime. So "Job summary this session" might be important to have.
- Would there ever be a service ID that isn't IRSA here?
- I am not sure what "type" means in this context. Aren't all of them searches?
- phase - I understand what this means, but I do not understand what the default filter is that it has there. I can't completely see the filter, and the pulldown only has "completed" as a choice presumably because all my jobs have completed at the moment. (?)
- There is no cancel filters button here - does it matter? maybe not.
- if this is also for bundling data (will it ever be? because that's the main purpose of the background monitor on ops now), where will the 'download data' button be? the control column only has the 'display' and 'trashcan' icons for me, no obvious "download" choice. 
- clicking on the job name ("label") gives me job information. this may be counterintuitive and maybe this information should (also?) be retrievable via an i-with-a-circle icon under 'control' on the far right. (that's the icon that retrieves this information when it's shown in a table.)
- how do you 'archive' a job?

robyww (approved):

PR #1738, opened 2025-04-04
FIREFLY-1693: ObsCore improvements 

kpuriIpac:
**Tickets**: [Firefly-1693](https://jira.ipac.caltech.edu/browse/FIREFLY-1693) and [Firefly-1692](https://jira.ipac.caltech.edu/browse/FIREFLY-1692)
IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/400
- Cleaning up the logic in ObsCorePackager to **_remove_** all custom file naming logic, as this will be handled by both the Download Script and the Zipped Packager, by first looking at Content-Disposition header, and then at the URL path as a fallback
   - this keeps our code generic, and lets the service set the name. 
   - @robyww suggested maybe keeping some of the file naming convention, @loitly is making a strong case for not doing that with which I agree. We can discuss this further if needed. For now, I am only setting file name (from the `file` column, if available) for cutouts. But we can move away from doing this as well.   
- merged the `PrepareDownload` from `EuclidUtils` into `ttFeatureWatchers` 
- apps have the option of choosing via props to `PrepareDownload` whether they want to display File Structure (flattened vs folder)
  - I have kept the default as `flattened`. Currently Euclid is the only one giving users this option. 
- When users select `folder`, I have used `"row_" + row_index` as the folder name. Open to other suggestions instead of this, but not sure if looking at any of the columns again for this makes sense, in the spirit of keeping it generic? 

**Testing**: 
   - **DCE**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/irsaviewer/dce
   - **Euclid**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/applications/euclid 
   - **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1693-obscore-improvements/firefly
   - **SPHEREx**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/applications/spherex
   - **SUIT Portal**: (creating the test build shortly)
   - Do a search on all apps (CADC ivoa ObsCore for Firefly and in SPHEREx, search LVF images)
      - Select a few rows from the results and run `Prepare Download` for DCE, CADC and SUIT, and `Generate Download Script` for Euclid & SPHEREx 
      - For Euclid and SPHEREx, select both `flattened` and then `folder` as the File Structure.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/PackagingWorker.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/PackagingWorker.java
@@ -110,7 +110,7 @@ public final class PackagingWorker implements Job.Worker {
 
                     curFileInfoIdx++;
                     try {
-                        zippedBytes += zipHandler.addZipEntry(zout, fi);
+                        zippedBytes += (int) zipHandler.addZipEntry(zout, fi);
                         totalBytes += zippedBytes;
                     } catch (AccessDeniedException e) {
                         denied.add(e.getMessage());
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
@@ -26,6 +26,10 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipException;
 import java.util.zip.ZipOutputStream;
 
+import static edu.caltech.ipac.util.download.URLDownload.getFileNameFromUrl;
+import static java.util.Optional.ofNullable;
+import static org.reflections.util.Utils.isEmpty;
+
 /**
  * @author tatianag
  * @version $Id: ZipHandler.java,v 1.48 2012/08/08 18:30:49 roby Exp $
@@ -64,11 +68,13 @@ public class ZipHandler {
         ZipEntry zipEntry = null;
         long totalBytes = 0;
 
-        String filename = fi.getExternalName();     // file name before it has gone through FileNameResolver
+        String filename = fi.getExternalName();
+
         try {
 
             is = getInputStream(fi.getInternalFilename(), fi, baseDir);  // filename may change if FileNameResolver is set
             String zipEntryComment = "(" + fi.getSizeInBytes() + "b) ";
+            //getExternalName/suggestedName,
             filename = FileUtil.getUniqueFileNameForGroup(fi.getExternalName(), dupMap);
 
             int inBufSize = 4096;
@@ -173,9 +179,11 @@ public class ZipHandler {
                 URLConnection uc = URLDownload.makeConnection(url, cookies, headers);
                 uc.setRequestProperty("Accept", "text/plain");
 
-                if (fi.hasFileNameResolver()) {
+                String extName = ofNullable(fi.getExternalName()).orElse("").trim();
+                if (extName.isEmpty() || extName.endsWith("/")) {
                     String suggestedFilename = URLDownload.getSugestedFileName(uc);
-                    fi.setExternalName(fi.resolveFileName(suggestedFilename));
+                    suggestedFilename = isEmpty(suggestedFilename) ? getFileNameFromUrl(url) : suggestedFilename;
+                    fi.setExternalName(extName + suggestedFilename);
                 }
 
                 is = uc.getInputStream();
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
@@ -9,7 +9,6 @@ import edu.caltech.ipac.firefly.server.packagedata.FileGroup;
 import edu.caltech.ipac.firefly.server.util.Logger;
 import edu.caltech.ipac.table.*;
 import edu.caltech.ipac.table.io.VoTableReader;
-import edu.caltech.ipac.util.FileUtil;
 import edu.caltech.ipac.util.StringUtils;
 import org.json.JSONObject;
 
@@ -18,9 +17,9 @@ import java.net.URL;
 import java.util.*;
 
 import static edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil.getSelectedData;
-import static edu.caltech.ipac.util.FileUtil.getUniqueFileNameForGroup;
-import static org.apache.commons.lang.StringUtils.substringAfterLast;
+import static org.apache.commons.lang.StringUtils.isEmpty;
 import static edu.caltech.ipac.firefly.core.Util.Opt.ifNotNull;
+import static org.apache.commons.lang.StringUtils.substringAfterLast;
 
 @SearchProcessorImpl(id = "ObsCorePackager")
 public class ObsCorePackager extends FileGroupsProcessor {
@@ -35,6 +34,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
     public static final String ACCESS_FORMAT = "access_format";
     public static final String SEMANTICS = "semantics";
     public static final String CONTENT_TYPE = "content_type";
+    public static final String FILE = "file";
 
     public static final String DATALINK_SER_DEF = "datalinkServiceDescriptor";
     public static final String ADHOC_SERVICE = "adhoc:service";
@@ -46,6 +46,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
     public static final String RA = "ra";
     public static final String DEC = "dec";
     public static final String CUTOUT_VALUE = "cutoutValue";
+    public static final String ZIP_TYPE = "zipType";
     public static final String SER_DEF_ACCESS_URL = "accessURL";
     public static final String SER_DEF_INPUT_PARAMS = "inputParams";
     public static final String CIRCLE = "circle";
@@ -54,6 +55,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
     public static final String TEMPLATE_COL_NAMES = "templateColNames";
     public static final String USE_SOURCE_FILE_NAME = "useSourceUrlFileName";
     public static final String POSITION = "position";
+    public static final String GENERATE_DOWNLOAD_FILE_NAME = "generateDownloadFileName";
 
     private static final List<String> CUTOUT_UCDs= Arrays.asList("phys.size","phys.size.radius","phys.angSize", "pos.spherical.r");
     private static final List<String>  RA_UCDs= List.of("pos.eq.ra");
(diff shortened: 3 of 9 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/PackagingWorker.java
```diff
@@ -110,7 +110,7 @@ public String doCommand(SrvParam params) throws Exception {
 
                     curFileInfoIdx++;
                     try {
-                        zippedBytes += zipHandler.addZipEntry(zout, fi);
+                        zippedBytes += zipHandler.addZipEntry(zout, fi, curFileInfoIdx);
```
kpuriIpac, line 113: I'll remove this argument, was testing something, not needed anymore.

src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
```diff
@@ -173,9 +177,32 @@ private static InputStream getInputStream(String filename, FileInfo fi, File bas
...
+                String suggestedFilename = URLDownload.getSugestedFileName(uc);
+
+                //fallback to filename from URL path if content-disposition header is missing
+                if (isEmpty(suggestedFilename)) {
+                    String urlPath = url.getPath();
+                    suggestedFilename = urlPath.substring(urlPath.lastIndexOf('/') + 1);
+                }
+
+                //strip surrounding quotes, if any
+                suggestedFilename = suggestedFilename.replaceAll("^\"|\"$", "");
```
loitly, line 189: These logic should only apply when `getFileNameFromUrl`.  But, it's not enough.  I would also sanitize it to remove illegal characters in file path/name.
A quick and dirty way is to do this:
```
suggestedFilename.replaceAll("[^a-zA-Z0-9._-]", "_");        // all chars not in this set will be converted to '_'
```

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -136,37 +136,19 @@ private List<FileGroup> computeFileGroup(DownloadRequest request) throws DataAcc
...
-                        List<FileInfo> tmpFileInfos = parseDatalink(url, request, dgDataUrl, idx, ext_file_name);
-                        fileInfos.addAll(tmpFileInfos);
-                    }
-                    else { //access_format & access_url are both null, so try and get it from the url's Content_Type
-                        FileInfo fileInfo = new FileInfo(access_url, ext_file_name, 0);
-                        fileInfos.add(fileInfo);
-                    }
-                }
+                boolean isDatalink = (access_format != null && access_format.contains(DATALINK)) ||
+                        (access_format == null && datalinkServDesc != null);
```
loitly, line 143: This simplifies to : datalinkServDesc != null || access_format != null && access_format.contains(DATALINK).
Is that the logic you intended?

loitly (changes requested):
Please make the suggested code cleanup before merging.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
kpuriIpac: Yes, I suppose that would work as well. The second condition there `datalinkServDesc != null` is essentially checking for non-obscore tables containing datalink (like `mer-catalog` in Euclid, for instance). So I'll make the change to simplify this.

src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
```diff
@@ -177,32 +179,12 @@ private static InputStream getInputStream(String filename, FileInfo fi, File bas
...
-                        //else: do nothing, extName is already a valid file path
-                    } else {
-                        //if no externalName set, use the suggested file name, from content-disposition or the URL path
-                        fi.setExternalName(suggestedFilename);
-                    }
+                String extName = ofNullable(fi.getExternalName()).orElse("").trim();
+                if (extName.isEmpty() || extName.endsWith("/")) {
+                    String suggestedFilename = URLDownload.getSugestedFileName(uc);
+                    suggestedFilename = isEmpty(suggestedFilename) ? getFileNameFromUrl(url) : suggestedFilename;
+                    suggestedFilename = suggestedFilename.replaceAll("[^a-zA-Z0-9._-]", "_");
```
loitly, line 186: Move this line into `getFileNameFromUrl()`.

loitly (approved):

src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
kpuriIpac: done, addressed the other changes as well. Rebuilding the test builds to make sure no new bugs were introduced.

robyww (commented):
#### Testing CADC

- I tried to do cutout only and I did not get anything even though I could do a cutout.
- When I ask for the whole file I got filename surrounded by quotes.
Screenshot 2025-04-08 at 10 24 22 AM




#### Testing DCE
- looks good.

kpuriIpac:
> #### Testing CADC
> * I tried to do cutout only and I did not get anything even though I could do a cutout.
> * When I ask for the whole file I got filename surrounded by quotes.
>   Screenshot 2025-04-08 at 10 24 22 AM
> 
> #### Testing DCE
> * Mostly looks good
> * The first row is not package.
>   
>   * the first row is not packaged
>   * if I choose the first three rows I just get two products
>   * If I choose the second and third rows I get the same two products.

Thanks for pointing this out. There was a bug in the cutout service descriptor url generation. Fixed now, along with the quotes. CADC should work as expected now. 

For the DCE bug, can you tell me what search you did where you did not see the first row's downloads show up? I am not able to recreate it, it works for me.

robyww:
>  For the DCE bug, can you tell me what search you did where you did not see the first row's downloads show up? I am not able to recreate it, it works for me.

1. goto DCE and choose DIGIT
    - 20h23m46.89s +42d12m43.0s Equ J2000
   - 200 arcsec
4. select first 3 rows and download - 2 results in zip
5.  select second and third row  and download - 2 results in zip (same as above)
6. select only first row and download - no results

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -97,9 +99,10 @@ function getColNameFromTemplate(template) {
     return template.match(/\${[\w -.]+}/g)?.map( (s) => s.substring(2,s.length-1));
 }
 
-const PrepareDownload = () => {
-    const tbl_id = getActiveTableId();
-    const tblTitle = getTblById(tbl_id)?.title ?? 'unknown';
+export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showFileStructure=false,
+                                               downloadType='package', dataSource, fileName}) => {
```
robyww, line 103: The dialog should show a working message until the  `fetchSemanticList` completes. In Euclid on the first download the list of semantics only shows `all`. 
```js
useEffect(() => {
        fetchSemanticList(tbl_id).then(setSemList);
    }, [tbl_id]);
```

it probably should be more like this.
```js
useEffect(() => {
        fetchSemanticList(tbl_id).then( (result) => {
                     setSemList(result);
                    setWorking(false);
             });
       setWorking(true);
    }, [tbl_id]);
```
Then in the UI show a working indicator.

robyww (changes requested):
I am letting @loitly take the lead of the Java side review. So this is the JavaScript side.

I am also testing.

robyww:
#### Testing Euclid

If I click too fast I get the following. The details are in the review.
Screenshot 2025-04-09 at 9 54 56 AM

robyww:
This is not correct. Is it fixed in @loitly's PR?

Screenshot 2025-04-09 at 9 58 38 AM

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: @robyww I did this, but while testing it, `DownloadDialog` does not update probably because it internally calls `showDownloadDialog`, which evaluates the JSX only once. so when `working` is `true` and let's say I have some text like "Loading....", then when `semList` is updated and `setWorking` is set to `false`, the dialog still does not update. 

I tried a few things to make it work, but it was getting quite messy. 

So I ended up with the logic i have in the latest commit, `DownloadDialog` by itself is only called when the call to `fetchSemanticList` is resolved. So theoretically the user won't see the `Generate Download Script` button until this is ready, but ideally there should be no visible lag. 

This ensures `DownloadDialog` is populated with the `semList` when it's available. Let me know what you think.

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -174,9 +182,12 @@ export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showF
         centerColValues: { ra, dec }
     };
 
+
+    const isRowSelected = selectInfoCls.getSelectedCount()>0;
+
     return (
         <>
-            {loading && <div />}
+            {loading && <ToolbarButton enabled={false} variant={isRowSelected?'solid':'soft'} color='warning' text={downloadType === 'script' ? 'Generate Download Script' : 'Prepare Download'}/>}
```
kpuriIpac, line 190: came up with a solution to show a dummy (disabled) button while the `semList` is loading so as to not have any weird UX movement of a button randomly appearing for the user @robyww.

kpuriIpac:
@robyww @loitly  I addressed all of the feedback. 

- brought back file names if a service requests it, and also using the file name logic for folder names 
- waiting for `fetchSemanticList` to resolve before the `Prepare Download` / `Generate Download Script` button is made clickable 
- did some more refactoring in `ObsCorePackager` to try and make the code more readable (by using separation of concern where possbile)
- the bug with DCE has a new ticket now assigned to @loitly: https://jira.ipac.caltech.edu/browse/FIREFLY-1711

Apps to test downloads on: 
   - **DCE**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/irsaviewer/dce
   - **Euclid**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/applications/euclid 
   - **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1693-obscore-improvements/firefly
   - **SPHEREx**: https://firefly-1693-obscore-improvements.irsakudev.ipac.caltech.edu/applications/spherex

**_Note_**: I have temporarily kept `generateDownloadFileName = true` for firefly, so if you test CADC, you can notice the naming convention being used for folder and file names is written by us in ObsCorePackager. For all other apps, we're using whatever name the service gives us (or from the URL as a fallback).

lrebull:
Looking at the description at the top but not understanding enough of the code to do more detailed following of all the comments, AND empirically playing around with the builds you've linked, here are my $0.02:

- astronomers are going to want the flattened/folders choice in several archives
- the original filenames of the downloaded data often have important meaning(s) and should be preserved
- for bundled (zip files?) downloads, filenaming is less of a big deal, but I agree it should look like it came from the corresponding archive (like a euclid bundle should have euclid in the filename by default).
- for scripts, the default name should look like a script. When I get a bundled file from ops DCE, it's, say, "LGAImages-1.zip" so we know it's a zip file without looking too closely. the extension doesn't appear in the pop-up, so I don't know before it finishes whether i'm going to get a zip file (blahblah.zip) or a script (blahblah.csh). (I note that Euclid and SPHEREx are successfully telling me this clearly.) 
- the second request from the same tile should be blahblah-2 -- at the moment, they're all blahblah-1
- i  am confused. i'm trying the DCE build you linked above. at first, i thought from the description that it was *all* DCE searches, but a regular dce download is giving me the same behavior as on ops (zip files of real data), so not even Loi's updated background monitor, much less any wget/curl scripts. then i read more closely, and thought maybe this is just obscore searches? but that is also giving me zip files (zip files of real data), not download scripts. So i'm super confused. (Euclid and SPHEREx are both giving me a script.)
- am i supposed to be testing the script per se, or just the GUI interface to all of this?

kpuriIpac:
@lrebull I'll address some of your concerns:

- Currently only Euclid & Spherex offer Download Scripts, DCE / CADC etc. are all regular packaged downloads (_zip files of real data_). We could offer users an option when they're downloading for one versus the other (and this is the plan for a future sprint). 
  - Loi's PR on the background monitor work is still open, so we won't see those changes here until that PR is merged. But once my PR and his are merged, we should be able to see the new background monitor UI.  
- We could offer flattened/folders choice in other archives as well, if it's helpful. 
- For zip files, we could possibly say in the UI that the file name will be "LGAImages-1.zip" (including the `zip`) if that would be helpful. 
- For this PR, testing the GUI is more important than testing the script. I'll have a new ticket right after this PR that will work on some improvement & bug fixes for the script. 

@robyww feel free to add to this.

robyww (approved):

PR #1739, opened 2025-04-05
IRSA-6747: Changes in the job-submission UI

jaladh-singhal:
Fixes [IRSA-6747](https://jira.ipac.caltech.edu/browse/IRSA-6747)

See IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/401

- Mostly wording changes
- Make WavelengthOptions more customizable - fixed vs selectable range type, filter definitions label can be changed
- Major change in Search Summary of EmbeddedSearchPositionPanel: instead of key value pairs, it appears as a much shorter string now (no resolver name, and hms or dcm is chosen from HiPS readout preference)

# Testing
https://irsa-6747-specphot-updates.irsakudev.ipac.caltech.edu/applications/spherex

- Check if all wordings have been changed as per ticket
- Check Spectrophotometry form's wavelength options
- Try all 3 cases (Cone, polygon, multi-object) in LVF search form and collapse the form to check search summary. Similarly check 2 cases (point and multi-object) in Spectrophotometry tool

## Regression testing
https://irsa-6747-specphot-updates.irsakudev.ipac.caltech.edu/applications/euclid

https://irsa-6747-specphot-updates.irsakudev.ipac.caltech.edu/irsaviewer

Check Search summary for Euclid and DCE uses of EmbeddedSearchPositionPanel

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -1,5 +1,5 @@
 import {Box, Sheet, Stack, Typography} from '@mui/joy';
-import React, {Fragment, useContext, useEffect, useState} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 import {oneOf, bool, string, number, arrayOf, object, func, shape, elementType} from 'prop-types';
 import {defaultsDeep} from 'lodash';
 import CoordinateSys from '../../visualize/CoordSys.js';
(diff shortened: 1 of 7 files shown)


src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -351,30 +356,30 @@ function SearchSummary({request, targetKey, sizeKey, polygonKey, searchTypeKey,
 
     const target = request?.[targetKey];
     const polyCoords = request?.[polygonKey];
+    const coordPref = getPreference(MR_FIELD_HIPS_MOUSE_READOUT1, MR_EQJ2000_HMS);
     const coords = searchTypeValue === POLY_CHOICE_KEY ? polyCoords
-        : formatWorldPtToString(parseWorldPt(target)); // for cone, point, etc.
+        : formatWorldPtToString(parseWorldPt(target), getDcmOrHmsFromMR(coordPref), false); // for cone, point, etc.
```
robyww, line 361: This is not quite right. If there is a resolved name you always want to show it instead of the ra/dec. I think you need to use `formatWorldPtToStringSimple` and modify that function to take the `dcmOrHms` parameter. I don't think the changes should be in `formatWorldPtToString`. 

That said I think they both should be change to accept a string or a point for the `wp` parameter. I will comment on that below

src/firefly/js/visualize/ui/WorldPtFormat.jsx
```diff
@@ -29,7 +29,7 @@ export function formatWorldPt(wp, pad=3, useBold=true) {
     }
 }
 
-export function formatWorldPtToString(wp,addNewLIne=false) {
+export function formatWorldPtToString(wp,dcmOrHms,useResolverName=true,addNewLIne=false) {
```
robyww, line 32: `formatWorldPtToString()` should reverted so you just use `formatWorldPtToStringSimple`. The way  if wp.objName is defined that is returned alone.
Make `formatWorldPtToStringSimple` take the `dcmOrHms` parameter. I think if should be a boolean such as `dcmOnly`. 

I would like you to change both functions to also take wp as a string.

first two lines in both, something like:
```js
const actualWp= isValidPoint(wp) ? wp : isString(wp) : parseWorldPt(wp) : undefined;
if (!isValidPoint(actualWp)) return '';
```

robyww (changes requested):
Most of the code looks good but I would like you to do some clean up on how we handle target feedback. I put the comment below we can talk if they are unclear.

src/firefly/js/visualize/ui/WorldPtFormat.jsx
jaladh-singhal: Thanks - this simplifies it!

> I think if should be a boolean such as dcmOnly.

I thought about doing it but boolean is not a good type to represent 3 states we have: DCM only, HMS only, DCM and HMS both (it's ambiguous if false means both or HMS only). I instead created an enum type (`EQ_TYPE`) and I'm using it everywhere now instead of hard coded strings.

robyww (approved):

kpuriIpac (approved):
Looks good! Tested changes according to ticket, and tested Euclid / DCE as well. The summary looks much cleaner now.

PR #1745, opened 2025-04-23
FIREFLY-1704: Download Script Enhancement

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1704
**IFE PR**: https://github.com/IPAC-SW/irsa-ife/pull/403

- Changes to download dialog (merging `Title` and `File Name` fields into one in the UI, and using this value as `BaseFileName` on the backend)
- Added getter/setter for suffix in `FileInfo` (currently used only for cutouts as `-cutout` at the end of a filename before the extension, but could be anything)
- Modified Download Script to handle duplicate file names (especially for curl, as wget does it internally) by roughly using the logic/algorithm described in the ticket

**Testing**: 
-  **CADC** on **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1704-download-script/firefly
- **Euclid**: https://firefly-1704-download-script.irsakudev.ipac.caltech.edu/applications/euclid
- **CADC**:  This is to test our original packager, ensure it works as expected. If there is a cutout, you should see `-cutout` at the end of the file name.  
   - since the packager downloads files, duplicate file names will be handled by the OS, I didn't add any explicit logic to check for that here. 
- **Euclid**: This is super important to test, as it'll test the download script changes. Do any image/object search. Select 1 or more rows and download. 
  - download all 3 (URLs, curl and wget). 
  - Run the curl script at the very least. You don't need to run it all the way, just ensure that folders are created and you see some files downloaded. You may even choose to only download cutouts - notice that a 4th argument of "cutout" (as the suffix) will be passed to the `download_file` function in the script. 
    - the downloaded cutout files should also have a `-cutout` suffix at the end of the file name. 
- On **Euclid (& Spherex)**, when you're on any tab other than the `Results`,  refresh the page. Previously, this would lead to all tabs except for `Results` and `Job Monitor` getting removed. This is now fixed.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/DownloadRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/DownloadRequest.java
@@ -74,7 +74,7 @@ public class DownloadRequest extends ServerRequest implements Serializable {
 
     public String getDataSource() { return getParam(DATA_SOURCE); }
 
-    public String getTitle() { return getParam(TITLE); }
+    public String getTitle() { return getParam(BASE_FILE_NAME); }
 
     public String getEmail() { return getParam(EMAIL); }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/FileInfo.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/FileInfo.java
@@ -15,11 +15,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import static edu.caltech.ipac.firefly.data.HttpResultInfo.CONTENT_TYPE;
-import static edu.caltech.ipac.firefly.data.HttpResultInfo.EXTERNAL_NAME;
-import static edu.caltech.ipac.firefly.data.HttpResultInfo.RESPONSE_CODE;
-import static edu.caltech.ipac.firefly.data.HttpResultInfo.RESPONSE_CODE_MSG;
-import static edu.caltech.ipac.firefly.data.HttpResultInfo.SIZE_IN_BYTES;
+import static edu.caltech.ipac.firefly.data.HttpResultInfo.*;
 
 
 public class FileInfo implements HasAccessInfo, Serializable, CacheKey {
@@ -90,6 +86,7 @@ public class FileInfo implements HasAccessInfo, Serializable, CacheKey {
     }
 
 
+
     public static FileInfo blankFilePlaceholder() {
         FileInfo fi= new FileInfo();
         fi.putAttribute(BLANK,true+"");
@@ -143,12 +140,22 @@ public class FileInfo implements HasAccessInfo, Serializable, CacheKey {
 
     public String getExternalName() { return getAttribute(EXTERNAL_NAME); }
 
+    public String getSuffix() { return getAttribute(SUFFIX); }
+
     public String getContentType() { return getAttribute(CONTENT_TYPE); }
 
     public void setInternalName(String filename) {
         if (filename!=null) putAttribute(INTERNAL_NAME,filename);
     }
     public void setExternalName(String filename) { putAttribute(EXTERNAL_NAME,filename); }
+
+    public void setSuffix(String filename) {
+        if (StringUtils.isEmpty(filename)) {
+            attributes.remove(SUFFIX);
+        } else {
+            putAttribute(SUFFIX,filename);
+        }
+    }
 
     public long getSizeInBytes() { return StringUtils.getInt(getAttribute(SIZE_IN_BYTES),0); }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
@@ -4,6 +4,7 @@
 package edu.caltech.ipac.firefly.server.packagedata;
 
 import edu.caltech.ipac.firefly.data.FileInfo;
+import edu.caltech.ipac.util.StringUtils;
 import edu.caltech.ipac.util.download.URLDownload;
 import edu.caltech.ipac.firefly.server.util.Logger;
 import edu.caltech.ipac.util.AppProperties;
@@ -180,10 +181,16 @@ public class ZipHandler {
                 uc.setRequestProperty("Accept", "text/plain");
 
                 String extName = ofNullable(fi.getExternalName()).orElse("").trim();
+                String suffix = fi.getSuffix();
                 if (extName.isEmpty() || extName.endsWith("/")) {
                     String suggestedFilename = URLDownload.getSugestedFileName(uc);
                     suggestedFilename = isEmpty(suggestedFilename) ? getFileNameFromUrl(url) : suggestedFilename;
-                    fi.setExternalName(extName + suggestedFilename);
+                    extName = extName + suggestedFilename;
+                    fi.setExternalName(extName);
+                }
+
+                if (!StringUtils.isEmpty(suffix)) {
+                    fi.setExternalName(FileUtil.appendSuffixBeforeExtension(extName, suffix));
                 }
 
                 is = uc.getInputStream();
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
@@ -9,6 +9,8 @@ import edu.caltech.ipac.firefly.server.ServerContext;
 import edu.caltech.ipac.firefly.server.packagedata.FileGroup;
 import edu.caltech.ipac.table.IpacTableUtil;
 import edu.caltech.ipac.util.AppProperties;
+import edu.caltech.ipac.util.StringUtils;
+import edu.caltech.ipac.util.download.URLDownload;
 
 import java.io.BufferedWriter;
 import java.io.File;
@@ -52,7 +54,7 @@ public class DownloadScript {
             fi
             """.stripIndent();
 
-//====================================================================
+   /*//====================================================================
 //  Consistent behavior for both curl and wget:
 //  - Both curl and wget must follow redirects.
 //  - Command exits with non-zero status on failure or error http status code.
@@ -64,7 +66,7 @@ public class DownloadScript {
 //      * If no file name is provided, use the remote file name from the URL, or from the Content-Disposition header if available.
 //        * curl will overwrite an existing file, whereas wget will create a new file with a numbered suffix to avoid overwriting.
 //====================================================================
-static final String funcTmpl = """
+    static final String funcTmpl = """
         download_file() {
           href="$1"; filePath="$2"; fileName="$3"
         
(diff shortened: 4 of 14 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
```diff
@@ -161,6 +163,10 @@ static UrlInfo resolve(FileInfo fi, File baseDir) {
         String baseDirPath = ifNotNull(baseDir).get(f -> f.getPath().endsWith("/") ? f.getPath() : f.getPath() + "/");
         String href = ifNotNull(fi.getInternalFilename()).get(ServerContext::resolveUrl);
         String extName = ifNotNull(fi.getExternalName()).getOrElse(baseDirPath);
+        String suffix = fi.getSuffix();
+        if (!StringUtils.isEmpty(suffix)) {
+            extName = URLDownload.appendSuffixBeforeExtension(extName, suffix);
+        }
```
kpuriIpac, line 169: This will not work consistently. Logic needs to be moved to passing suffix in to the download script, as discussed earlier. I'm working on changing this.

src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
```diff
@@ -180,10 +181,16 @@ private static InputStream getInputStream(String filename, FileInfo fi, File bas
                 uc.setRequestProperty("Accept", "text/plain");
 
                 String extName = ofNullable(fi.getExternalName()).orElse("").trim();
+                String suffix = fi.getSuffix();
                 if (extName.isEmpty() || extName.endsWith("/")) {
                     String suggestedFilename = URLDownload.getSugestedFileName(uc);
                     suggestedFilename = isEmpty(suggestedFilename) ? getFileNameFromUrl(url) : suggestedFilename;
                     fi.setExternalName(extName + suggestedFilename);
+                    extName = extName + suggestedFilename;
```
loitly, line 189: If you are going to do this, you should flip the lines around so that extName is not computed twice.
```.java
                    extName = extName + suggestedFilename;
                    fi.setExternalName(extName);
```

loitly: You still need to handle duplicate files.

src/firefly/java/edu/caltech/ipac/util/download/URLDownload.java
```diff
@@ -133,6 +133,21 @@ public static String sanitizeFilename(String fName) {
...
+    public static String appendSuffixBeforeExtension(String fName, String suffix) {
+        if (fName == null || suffix == null) return fName;
+
+        int lastDotIndex = fName.lastIndexOf('.');
+        if (lastDotIndex == -1) {
+            return fName + suffix; //no extension found, just append suffix
+        }
+
+        String name = fName.substring(0, lastDotIndex);
+        String extension = fName.substring(lastDotIndex);  //for example, ".fits"
```
loitly, line 145: There's `getBase()` and `getExtension()` in FileUtil already.

loitly (commented):
Just comments.

src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
kpuriIpac: @loitly about duplicates - since we're downloading files, any OS should be able to handle duplication, right? for example when I test with same file name on mac OS, it'll automatically add `-1`, `-2`, `-3` and so on at the end of the file name for duplicate file names. 

kind of like this: 

Screenshot 2025-04-25 at 2 37 23 PM

src/firefly/js/templates/router/RouteHelper.jsx
```diff
@@ -111,10 +111,11 @@ export function useDropdownRoute() {
     useEffect(() => {
         const {path, visible} = getViewInfo();
         if (pathname !== path) {
+            const {menuItems,showBgMonitor}= getMenu();
             const menuItem = getMenuItem(pathname);
             if (menuItem) {     // the requested pathname is in the menu
                 dispatchSetLayoutInfo({dropDown:{view: menuItem.action, menuItem, visible: true}});
-                dispatchSetMenu({selected: menuItem.action});
+                dispatchSetMenu({selected: menuItem.action, menuItems, showBgMonitor});
```
kpuriIpac, line 118: @loitly for when you're back on Monday, you may want to take a look at this fix. Since we would call `dispatchSetMenu` with only the selected tab when you refresh the page, the app would display that tab, but `menuItems` previously existing would get overwritten (you can try that behavior in Euclid or SPherex on nightly). So I'm passing them explicitly here again. 

The reason job monitor tab still showed up is because the `MENU_UPDATE` action creator would call this after `dispatchSetMenu`, which checks for adding background monitor as a menu item.

robyww:
Screenshot 2025-04-28 at 3 50 22 PM

Remember we talked about the flatten in the CCB-sci. Make it a single check box that says `flatten` or maybe `flattened file structure` or something like that.
- if checked all files should be in the same directory
- if not (the default) only create directories if there is more than one file in a datalink result.

kpuriIpac:
> Screenshot 2025-04-28 at 3 50 22 PM
> 
> Remember we talked about the flatten in the CCB-sci. Make it a single check box that says `flatten` or maybe `flattened file structure` or something like that.
> 
> * if checked all files should be in the same directory
> * if not (the default) only create directories if there is more than one file in a datalink result.

I was planning on doing it in a new smaller ticket right after this PR @robyww, but yes I think I can get it done in this PR itself probably. Let me give it a shot, just didn't want to increase the scope of this PR too much.

src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
```diff
@@ -91,6 +93,105 @@ public class DownloadScript {
...
+//  - Command exits with non-zero status on failure or error http status code.
+//  - If UrlInfo.filePath is provided:
+//      * If it ends with '/', treat it as a directory.
+//      * Otherwise, treat it as a file path (including possible subdirectories).
+//  - When downloading:
+//      * Save the file to the specified path.
+//      * If no file name is provided, use the remote file name from the URL, or from the Content-Disposition header if available.
+//        * curl will overwrite an existing file, whereas wget will create a new file with a numbered suffix to avoid overwriting.
+//====================================================================
+static final String funcTmpl = """
```
robyww, line 120: This is a little picky so take if or leave it.

I don't think you need to do string substitution 
Try this: at the top of the function 
```bash
curlCmd="curl -fLJO"
wgetCmd="wget --content-disposition"
if [ "$URL_GETTER" = "curl" ]; then
   cmd=$curCmd
else
   cmd=$wgetCmd
fi
```
then right before the downoad_file function set which way. That will have to be in the java code
```java
writer.println( is(Wget, attribs) ? "URL_GETTER=wget" : "URL_GETTER=curl");
write.printlin("");
writer.println(funcTmpl);
```
No string substitution and it allows the use the change the command with out have to go through the while packaging process again.

robyww (approved):
I think this is basically ready.  

- move the function to FileUtil.java
- look at my script creation comment.
- I am find with you changing the UI in another ticket.

src/firefly/js/templates/router/RouteHelper.jsx
loitly: I think you just want to select the menu. It was working before because `dispatchSetMenu` allowed partial updates, but that caused issues elsewhere. We've recently fixed it so that it now performs a full 'set'.
If you want to select a menu item, you should use `LayoutCntlr.dispatchShowDropDown` instead.  Can you try this to see if that fixes the problem?

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -220,7 +213,7 @@ export function DownloadOptionPanel ({groupKey='DownloadDialog', cutoutSize, hel
```
jaladh-singhal, line 213: Just a minor cosmetic change suggestion because this dialog looks bit crowded:
Current spacing | Suggested spacing
--- | ---
image | image

I hacked above from CSS but the idea is to use correct spacing to establish visual grouping: https://github.com/Caltech-IPAC/firefly/blob/dev/docs/ui-notes.md#spacing-and-layout. As you can notice, "Including Anciliary" is part of "Download Files" so it gets tighter spacing of `1` to convey that visually, and everything else in form gets `spacing={2}`. 

As a rule of thumb, I start any form/ field group with spacing 2 (or sometimes even 3!) and keep reducing it as I go down to children that are Stack of grandchildren.

src/firefly/java/edu/caltech/ipac/firefly/data/FileInfo.java
```diff
@@ -143,13 +140,21 @@ public void setRequestInfo(HttpServiceInput requestInfo) {
...
     public void setInternalName(String filename) {
         if (filename!=null) putAttribute(INTERNAL_NAME,filename);
     }
     public void setExternalName(String filename) { putAttribute(EXTERNAL_NAME,filename); }
 
+    public void setSuffix(String filename) {
+        if (!StringUtils.isEmpty(filename)) {
+            putAttribute(SUFFIX,filename);
+        }
+    }
```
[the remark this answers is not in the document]
kpuriIpac: @loitly I'm not sure I follow this question. Can you explain a bit more / give an example?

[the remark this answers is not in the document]
loitly: How do you remove a 'suffix' once you've set a value to it?  I am guessing you would `setSuffix(null)`.  But, this will not work.  So, maybe:

```java
          if (StringUtils.isEmpty(filename)) {
              attributes.remove(SUFFIX);
          } else {
              attributes.put(SUFFIX,filename);
          }
```
I am also fine if you decide to do nothing.  This class need refracting anyways.

src/firefly/java/edu/caltech/ipac/firefly/data/DownloadRequest.java
```diff
@@ -74,7 +74,7 @@ public void setFilePrefix(String filePrefix) {
 
     public String getDataSource() { return getParam(DATA_SOURCE); }
 
-    public String getTitle() { return getParam(TITLE); }
+    public String getTitle() { return getParam(BASE_FILE_NAME); }
```
kpuriIpac, line 77: @loitly I need your feedback on this change. Earlier we had two fields (title and file name) in the download dialog (with `fieldKey` set to `title` and `BaseFileName` respectively). 

Now, these are both used on the server side (`BaseFileName` as the file name) and `title` is used here. 

After merging the title and file name fields into just one 'title' field on the UI, I also changed the `fieldKey` to `BaseFileName`, because the title will also be used as the file name now. 

So with that in mind, does this change look okay? Any alternate suggestions?

src/firefly/java/edu/caltech/ipac/firefly/data/FileInfo.java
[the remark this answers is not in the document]
kpuriIpac: Okay, I made this change. And yes, we should refactor this in a future sprint.

src/firefly/js/templates/router/RouteHelper.jsx
```diff
@@ -114,7 +119,6 @@ export function useDropdownRoute() {
             const menuItem = getMenuItem(pathname);
             if (menuItem) {     // the requested pathname is in the menu
                 dispatchSetLayoutInfo({dropDown:{view: menuItem.action, menuItem, visible: true}});
-                dispatchSetMenu({selected: menuItem.action});
```
kpuriIpac, line 117: @loitly I was able to fix this by just getting rid of this line. `dispatchSetLayoutInfo` in the line above seems to be enough to make this work.

loitly: You are right.  That line should be enough.  Basically, you want to switch to that menu tab/item.

src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
kpuriIpac: I'll tinker with this in the very next ticket I have (where I change the UI of the download dialog) since that's a smaller ticket anyway if that's ok. Since this PR has been open for some time, I would like to close it to have a working version of the download script across our apps.

PR #1747, opened 2025-04-24
Firefly-1718: Better handle data product catalog that connect to images

robyww:
#### Testing
- Euclid: https://firefly-1718-hightlight-change.irsakudev.ipac.caltech.edu/applications/euclid
- do a `Inspect Objects` search
- On the images: click on a point
- instead of going  to the table row and reloading the images a question will appear at the bottom first.


lrebull:
That's nice! Works great to address what I was seeing.

src/firefly/js/drawingLayers/CatalogUI.jsx
```diff
@@ -260,3 +263,44 @@ CatalogUI.propTypes= {
...
+        borderColor: theme.vars.palette.warning.outlinedColor,
+        backgroundColor: theme.vars.palette.neutral.softBg,
+        visibility: visible ? 'visible' : 'hidden',
+        mb: 2,
+    });
+    
+
+    useEffect(() => {
+        setVisible(true);
+        const id= setTimeout(() => setVisible(false),5000);
```
kpuriIpac, line 285: I think we should increase the timeout to 10 or 15 seconds. When the text came up, by the time I noticed and read it, it disappeared (the first few times). It might be better to leave it on there a little longer.

kpuriIpac (approved):
Looks great! Really neat feature. Just one comment with a suggestion below.

src/firefly/js/drawingLayers/CatalogUI.jsx
robyww: ok. I will do that.

PR #1749, opened 2025-04-28
FIREFLY-1686: Invalid SQL statement generated when expressions are used in charts requiring decimation.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1686

A regression bug has been identified where Firefly generates an invalid SQL statement when expressions are used in charts that require decimation. This bug was introduced in ticket [FIREFLY-1649](https://jira.ipac.caltech.edu/browse/FIREFLY-1649).

Test: https://fireflydev.ipac.caltech.edu/firefly-1686-invalid-sql-chart/firefly/
Steps to reproduce in Firefly:

Load a table with more than 10,000 rows.
Open the heatmap setting, change X and Y axes to use an expression, such as `log(x)` and `log(y)`
Click Apply

@lrebull At your convenience, please verify that the fix is working.


lrebull:
works great! thanks!!

kpuriIpac (approved):

PR #1750, opened 2025-04-28
Firefly-1725: improve center on selected table row so the data products images are never scrolled off screen

robyww:
#### Testing
- https://firefly-1725-recenter.irsakudev.ipac.caltech.edu/applications/euclid
- https://firefly-1725-recenter.irsakudev.ipac.caltech.edu/applications/spherex
- do a cutout grid with either application, cutout should not longer scroll off the screen


lrebull:
It seems better ... i couldn't trigger it on demand before, but i've been fusssing with both of them for several minutes and it's not happened.

jaladh-singhal (approved):
Couldn't reproduce the issue. I'm not very familiar with image handling code but nothing stands out.

PR #1751, opened 2025-04-28
FIREFLY-1720: Generalize UploadTableSelector for handling shape fit columns in Spectrophotometry tool

jaladh-singhal:
Fixes [FIREFLY-1720](https://jira.ipac.caltech.edu/browse/FIREFLY-1720):
- Extract redundant code from `UploadTableSelectorPosCol` and `UploadTableSelectorSingleCol` to `UploadTableSelector`, which:
  - accepts a list of n `columnFields` that is looped over to run the state logic and render components
  - exposes a slot for a stateless `ColumnMappingPanel` component that controls how to render header and column fields in the collapsible that appears after uploading a table file
  - allows choosing whether to render Columns as text or a "column selector" chip button. For the latter, there is now a dynamic tooltip that shows which columns are selected
- Change SpectroPhotometry (at SPHEREx) to render Shape Fit parameters as column fields in multi-object case
- Update other users of these in TAP, Euclid, EmbeddedSearchPositionPanel, etc.
- Cleanup updates in `ColumnFld`

See ife PR: https://github.com/IPAC-SW/irsa-ife/pull/406

Additionally, fixes [FIREFLY-1707](https://jira.ipac.caltech.edu/browse/FIREFLY-1707):
- Updated UploadTableChooser's "Loaded Table" panel to:
  - change TAP wording
  - update styling to prevent scrollbar obscuring "Load table" button

# Testing

This requires some complex testing. Basically, it will be permutations of the following:

## Test Views (UI rendering)
1. Firefly TAP: https://fireflydev.ipac.caltech.edu/firefly-1720-generalise-uptbl-selector/firefly/?__action=layout.showDropDown&view=TAPSearch
    1. Spatial:
        - "Position Column (from the selected table on the right)": `CenterColumns`
        - click "Multi-object" radio option: `UploadTableSelectorPosCol`
    2. Object ID: 
        - "ID Column (from the selected table on the right)": `SingleCol`
        - click "Load object IDs from a table" radio option: `UploadTableSelectorSingleCol`
2. Euclid: https://firefly-1720-generalise-uptbl-selector.irsakudev.ipac.caltech.edu/applications/euclid/
    1. "Images" or "Inspect Objects" tab (EmbeddedPositionSearch) -> "Multi-object": `UploadTableSelectorPosCol`
    2. "Search by ID" tab -> "multi-object" tab: `UploadTableSelectorSingleCol`
3. SPHEREx: https://firefly-1720-generalise-uptbl-selector.irsakudev.ipac.caltech.edu/applications/spherex/
    1. "LVF" tab -> "Location" panel -> "multi-object" (EmbeddedPositionSearch containing TAP form): `UploadTableSelectorPosCol`
    2. "Spectrophotometry" tab -> "multi-object" (EmbeddedPositionSearch containing custom UI): `UploadTableSelectorSpecPhoto`

### Files to use
For SPHEREx spectrophotometry in particular, use [multi-object elliptical.txt](https://github.com/user-attachments/files/19949685/multi-object.elliptical.txt)

For others, can also use this file. Or go to "loaded tables" tab in upload table chooser, load table from other search results.

## Test Behaviors (State logic)
For any of the above view, check for following (there can be more test cases than what I've mentioned in each bullet):

1. **Column mapping in input fields goes into fields payload:** results should return correctly once search form is submitted
2. **Selection of columns to upload through "n columns (m selected)" chip button** (different from column mapping through column fields):
    1. default selection that appears in button tooltip and table after clicking the button: position columns if present in upload table should be selected by default
    2. selection of a column mapping triggers selection of that column for upload: change the value for column field, it should increase the "m selected count" as well as the checkboxes in column selection table 
    3. selected columns goes into upload info payload (`use` flag in columns): TAP ADQL should have these columns in SELECT statement on upload table (`ut`)
3. **Guess Column mapping:**
    1. default value that appears in column fields: ra, dec colums should be mapped by default (others do not have a guesser yet but can be added)
    2. close collapsible if all guessed: in case of ra, dec, it will close the collapsible
6. **Filter column choices** (only `SingleCol` use has it, others don't but it can be added): TAP -> CADC, IVOA -> Object ID -> click search button on the column field, the column table should be filtered by UCD=meta.id


src/firefly/js/ui/tap/ObjectIDSearch.jsx
```diff
@@ -205,27 +199,21 @@ export function makeColsLines(objAry) {
 }
 
 export function UploadTableSelectorObjectID({uploadInfo, setUploadInfo, setSelectInObjList, getUseSelectIn, setWorking}) {
-    const [getSingleCol,setSingleCol]= useFieldGroupValue(UploadSingleColumn);
+    const [getSingleCol,] = useFieldGroupValue(UploadSingleColumn);
```
robyww, line 202: in this use case with `useFieldGroupValue`. I will sometimes use.

```
const singleCol= useFieldGroupValue(UploadSingleColumn)[0]();
```

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -215,16 +216,11 @@ const LoadedTables= (props) => {
     return (
         <Stack width={1} height={1}>
             <Typography level={'title-lg'} color={'neutral'} p={1}>
-                {'Select one of the existing tables below to load into the TAP panel: '}
+                {'Select one of the existing tables below to load: '}
```
robyww, line 219: You don't need the `:` since it is vertically aligned.

src/firefly/js/ui/UploadTableChooser.js
```diff
@@ -172,6 +172,7 @@ function applyDefColumnSelection(columns,defaultColsEnabled) {
         columnsSelected = defaultColumnsSelector(columns,colTypes,colCount);
     }
     else {
+        //Fixme: don't select position columns by default, UploadTableSelector can be used for non position columns fields too
```
robyww, line 175: add `TODO` at the start of this comment.

src/firefly/js/charts/ui/ColumnOrExpression.jsx
```diff
@@ -109,10 +110,9 @@ export function ColumnFld({cols, groupKey, fieldKey, initValue, label, tooltip='
             <ToolbarButton icon={MAGNIFYING_GLASS}
                            tip={`Select ${name} column`}
                            onClick={(e) => {
-                               if (!onSearchClicked || onSearchClicked()) {
-                                   showColSelectPopup(cols, onColSelected, `Choose ${name}`, 'OK',
-                                       val, false, colTblId,doQuoteNonAlphanumeric);
-                               }
+                               onSearchBtnClicked?.(e);
+                               showColSelectPopup(cols, onColSelected, `Choose ${name}`, 'OK',
+                                       val, false, colTblId, doQuoteNonAlphanumeric);
```
robyww, line 115: before `onSearchClicked` has to exist to call `showColSelectPopup`. That seems like it was probably a bug.

robyww (approved):
This looks good!

I did some testing and I don't see any issues.

src/firefly/js/charts/ui/ColumnOrExpression.jsx
loitly: This definitely changed the logic.  Original logic is to skip `showColSelectPopup` when `onSearchBtnClicked` returns false.  I am guessing it's a way to do some sort of validation.  @kpuriIpac should look at this.

src/firefly/js/ui/UploadTableSelector.jsx
```diff
@@ -0,0 +1,460 @@
...
+import {onTableLoaded} from 'firefly/tables/TableUtil';
+
+
+//--------------------------------------------------------------------------------------------------------------------
+// GENERIC components and helpers
+//--------------------------------------------------------------------------------------------------------------------
+
+/**
+ * A generic component that allows users to select an upload table and map its columns to the required fields.
+ *
```
loitly, line 26: Normally, when I read this, I would expect its props to include a list of fields along with their current or default values—both required and optional—and an onChange or onValueChange callback to pass back the selected values. I'm a bit surprised to see that so much of the parent logic is being passed into this component, especially its  state modifier (setUploadInfo).

src/firefly/js/ui/UploadTableSelector.jsx
```diff
@@ -0,0 +1,460 @@
...
+export function ColumnMappingPanel({cols, columnFieldValues, columnFields, panelKey,
+                                       headerTitle='Mapped Columns:', getHeaderColumnMapping, headerPostTitle = '',
+                                       openPreMessage='', sx, slotProps, children}) {
+    const defaultColumnMapping = (colValues) => colValues?.every((val) => !val)
+        ? 'unset'
+        : colValues?.map((val) => val || 'unset').join(', ');
+
+    const panelHeader= (
+        <Stack {...{direction:'row', alignItems:'baseline', spacing:1}}>
+            <span>{headerTitle}</span>
```
loitly, line 227: Interesting how you choose to use `span` only for this one when you are using Typography everywhere else.

src/firefly/js/ui/UploadTableSelector.jsx
jaladh-singhal: I agree the upload info state logic could have been lifted up to parent when this was written. But this is being used at 7-8 places (As you can see from my topmost comment on this PR) so doing that in this PR would have required much more refactoring. Refactoring the position, ID and custom variation of this component into this generic one was a challenge itself which took a lot more time than I anticipated. 

So let's aim for what you suggested in round 2 of refactoring.

src/firefly/js/ui/UploadTableSelector.jsx
jaladh-singhal: This was a copy-over from older deleted files. My interpretation of this choice was: headerTitle can be a `node` (with inline display type) that's why `span` is used, whereas following header bits are `string`. 

Can `Typography` be used to wrap a node type component?  if yes, I'll update it

src/firefly/js/charts/ui/ColumnOrExpression.jsx
jaladh-singhal: I checked uses of `onSearchClicked` in our codebase before changing it. It was being used only in src/firefly/js/ui/UploadTableSelectorSingleCol.jsx L156-160, and it always returns true (see the deleted file below)

I did find it confusing though that how it was maintaining a toggle flag in the state to do so which didn't seem useful. So @kpuriIpac should confirm if I took out something he wanted to keep?

src/firefly/js/ui/UploadTableSelector.jsx
loitly: I thought the same, until I noticed that headerTitle is declared as a String.
```
headerTitle: PropTypes.string,
```

src/firefly/js/charts/ui/ColumnOrExpression.jsx
kpuriIpac: Yes I can't seem to recall the exact reason for it being there, but you're right that it always returned true @jaladh-singhal, probably a bug anyway. This change looks good and everything still works as expected.

kpuriIpac (approved):
Looks great! Tested all apps/behavior as best as I could following the instructions. Looked at the ADQL as well to confirm, everything looks ok. Also tested all multi-object uploads on Euclid and on the 2 searches on SPHEREx, working as expected.

src/firefly/js/charts/ui/ColumnOrExpression.jsx
jaladh-singhal: @kpuriIpac ok - thanks for testing!

src/firefly/js/ui/UploadTableSelector.jsx
jaladh-singhal: Ah, good catch - I'll fix the prop type

PR #1753, opened 2025-05-01
FIREFLY-1731: Download Dialog UI Improvement

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1731
**IFE PR**:  https://github.com/IPAC-SW/irsa-ife/pull/407 

- The download dialog now only has one checkbox option: `Flattened File Structure`, 
  - When selected, we flatten downloaded files , else we default to 'structured' (`folder`) 
  - for IFE apps not using ObsCorePackager, their downloads should look the same as before when flattened is selected and when it's turned off (defaulting to structured)
  - the logical changes in this PR only affect Obs Core files, for these:
    -  flattened will flatten the files, even if you have multiple rows selected
    - for 'structured' (folder), which is the default, we only create folders when there's more than one entry in the datalink, if there's only 1 entry per row, we still flatten files 
- I made some changes in `SwitchInputField` to be able to pass non-boolean values (like `folder` and `flat` in this case). It should still work as expected in other places it's used. 
  - @jaladh-singhal you may want to take a look at this. 
- There was an issue with folder names when **_we_** assign them in `ObsCorePacakger`. 
  - to test this, select 2 or more rows to download in SPHEREx. Earlier (you can test on nightly) you would see in the downloaded script that even though each row has more than 1 product, we still use the same folder name (because it comes from `obs_title`) 
    - so when we assign folder names ourselves in `ObsCorePackager`, we need to check for duplicate folder names there and assign `-1`, -`2`, and so on at the end of the folder name if it already exists. 
    - This is now fixed. 

**Testing**: 
- On firefly (TAP panel), test the `SwitchInputField` toggle (`Enter my URL` right under `Select TAP Service`) to ensure it works as expected. 
- For IFE apps like WISE, SHA, etc. ensure that downloads work as before (you can compare the downloads on ops or nightly) 
- Test the **_logic_** changes with CADC, DCE, Euclid & SPHEREx 
  - For packagers (like with CADC and DCE), if you select flattened checkbox you shouldn't see any folders, otherwise you should see folders created, if there's more than one file per row. This is usually the case for CADC. 
    - For DCE, you may see the same results when flattened is turned on or off, this is because lots of tables have only one fits file per row, so this is expected behavior. 
  - For the download script (euclid, spherex), when flattened is selected you'll see no folder name being passed as an argument (2nd argument) to the `download_file` function, otherwise you should see a folder name here. 
-  CADC on firefly: https://fireflydev.ipac.caltech.edu/firefly-1731-download-dialog/firefly
- Euclid: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/euclid
- SPHEREx: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/spherex
- DCE: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/irsaviewer/dce
- Wise: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/wise
- SHA: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA
- ZTF: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/ztf
- Sofia: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/sofia
- Finderchart: https://firefly-1731-download-dialog.irsakudev.ipac.caltech.edu/applications/finderchart

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
@@ -46,7 +46,6 @@ public class ObsCorePackager extends FileGroupsProcessor {
     public static final String RA = "ra";
     public static final String DEC = "dec";
     public static final String CUTOUT_VALUE = "cutoutValue";
-    public static final String ZIP_TYPE = "zipType";
     public static final String SER_DEF_ACCESS_URL = "accessURL";
     public static final String SER_DEF_INPUT_PARAMS = "inputParams";
     public static final String CIRCLE = "circle";
@@ -90,6 +89,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
             var selectedRows = new ArrayList<>(request.getSelectedRows());
             MappedData dgDataUrl = EmbeddedDbUtil.getSelectedMappedData(request.getSearchRequest(), selectedRows); //returns all columns
             DataGroup dg = getSelectedData(request.getSearchRequest(), selectedRows);
+            Map<String, Integer> prefixCounter = new HashMap<>();
 
             String datalinkServDesc = request.getParam(DATALINK_SER_DEF); //check for a non-obscore table containing a Datalink Service Descriptor
             String pos = request.getParam(POSITION);
@@ -114,16 +114,15 @@ public class ObsCorePackager extends FileGroupsProcessor {
                 String colNames = request.getSearchRequest().getParam(TEMPLATE_COL_NAMES);
                 String[] cols = colNames != null ? colNames.split(",") : null;
                 String fileNamePrefix = getFileNamePrefix(idx, cols, dgDataUrl, url, positionColVals); //this will be used as folder name, and as prefix for individual file names
+                fileNamePrefix = makeUniquePrefix(fileNamePrefix, prefixCounter);
 
-                String zipType = request.getParam(ZIP_TYPE); //flat or folder
-                
                 boolean isDatalink = (datalinkServDesc != null) || (access_format != null && access_format.contains(DATALINK));
 
                 if (isDatalink) {
                     List<FileInfo> tmpFileInfos = parseDatalink(url, request, dgDataUrl, idx, fileNamePrefix, positionColVals);
                     fileInfos.addAll(tmpFileInfos);
                 } else {
-                    String extName = "folder".equalsIgnoreCase(zipType) ? fileNamePrefix + "/" : null; //if flattened, no external name, just set null
+                    String extName = null;
                     FileInfo fileInfo = new FileInfo(access_url, extName, 0);
                     fileInfos.add(fileInfo);
                 }
@@ -138,7 +137,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
 
     public static List<FileInfo> parseDatalink(URL url, DownloadRequest request, MappedData dgDataUrl, int idx, String prepend_file_name, List<String> positionColVals) {
         List<FileInfo> fileInfos = new ArrayList<>();
-        String zipType = request.getParam(ZIP_TYPE); //flattened downloads or folders
+        boolean isFlattenedStructure = request.getBooleanParam("isFlattenedStructure"); //true if flattened, else false for structured logic
         boolean generateDownloadFileName = Boolean.parseBoolean(Objects.toString(request.getParam(GENERATE_DOWNLOAD_FILE_NAME), "false"));
 
         String productTypes = request.getParam(PRODUCTS); //products to download in datalink file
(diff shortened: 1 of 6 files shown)


src/firefly/js/ui/SwitchInputField.jsx
```diff
@@ -62,7 +68,9 @@ SwitchInputField.propTypes = {
     fieldKey: PropTypes.string.isRequired,
     groupKey: PropTypes.string,
     initialState: PropTypes.shape({
-        value: PropTypes.bool
+        value: PropTypes.any // changed from bool → any to support string values
     }),
+    trueValue: PropTypes.any,
+    falseValue: PropTypes.any,
```
jaladh-singhal, line 74: I don't think any of this is necessary. The state of switch component will always be binary. The consumer components should translate string that they want to true/false.

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -262,20 +262,17 @@ export function TitleField({style={}, value, label='Title:', size=30}) {
...
-
+        <Stack width={250} mt={2}>
+            <SwitchInputField
+                fieldKey={fieldKey}
+                initialState={{ value: 'folder' }}
+                trueValue='flat'
+                falseValue='folder'
+                label='Flattened File Structure'
+            />
+        </Stack>
```
jaladh-singhal, line 275: fieldKey can be defined something as `isFlattenedStructure` and the payload will have true or false. 

Then consumer of that payload (I'm assuming your server-side code) can read true as flat and false as folder.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -201,13 +206,11 @@ public static List<FileInfo> parseDatalink(URL url, DownloadRequest request, Map
                         String extension = "unknown";
                         extension = content_type != null ? getExtFromURL(content_type) : getExtFromURL(productUrl);
                         ext_file_name += "." + extension;
-                        ext_file_name = ("folder".equalsIgnoreCase(zipType) || (zipType == null && makeDataLinkFolder))
-                                ? prepend_file_name + "/" + ext_file_name
-                                : ext_file_name;
+                        ext_file_name = ("flat".equalsIgnoreCase(zipType) || (!makeDataLinkFolder)) ?
+                                ext_file_name : prepend_file_name + "/" + ext_file_name;
```
jaladh-singhal, line 210: Yes, here! You are in-fact just doing a binary operation on zipType=='flat', 'folder' doesn't really have any use in your server-side parser. 

This will then become: 
```suggestion
                        ext_file_name = (isFlattenedStructure || (!makeDataLinkFolder)) ?
                                ext_file_name : prepend_file_name + "/" + ext_file_name;
```

Just make sure to replace
`String zipType = request.getParam(ZIP_TYPE);` with `boolean isFlattenedStructure = getBooleanParam('fieldKeyYouDefinedAtClientSide');`

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -201,13 +206,11 @@ public static List<FileInfo> parseDatalink(URL url, DownloadRequest request, Map
...
                     } else {
-                        ext_file_name = ("folder".equalsIgnoreCase(zipType) || (zipType == null && makeDataLinkFolder))
-                                ? (prepend_file_name + "/" + (fileName != null ? fileName : ""))
-                                : fileName;
+                        ext_file_name = ("flat".equalsIgnoreCase(zipType) || (!makeDataLinkFolder)) ?
+                                ext_file_name : (prepend_file_name + "/" + (fileName != null ? fileName : ""));
```
jaladh-singhal, line 213: same here

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -262,20 +262,17 @@ export function TitleField({style={}, value, label='Title:', size=30}) {
...
-                tooltip: 'File Structure',
-                label : 'File Structure:'
-            }}
-            options = {[
-                {label: 'Structured (with folders)', value: 'folder'},
-                {label: 'Flattened (no folders)', value: 'flat'}
-            ]}
-        />
-
+        <Stack width={250} mt={2}>
```
jaladh-singhal, line 267:
```suggestion
        <Stack>
```
[Nitpick] don't constraint the width unless it's overflowing and you don't need margin top, the parent stack has spacing={2} set so that will take care of this.

I wonder if we really need to wrap single child (Switch component) in Stack. I'd just take it out.

jaladh-singhal (changes requested):
I tested SPHEREx, CADC, it's working as expected. The spacing in dialog looks good - thanks for fixing it.

In refactoring your two choice list box to switch, you didn't have to keep the field state and server side param of type ['flat', 'folder']. You can change the type to reflect the binary nature of this state/param. I explain it in my comments how.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
kpuriIpac: Yes good point, this is a nice suggestion. The reason I was initially hesitant is because this change will need to be made in other apps' download processors as well (for example `WiseFileGroupsProcessor.java` line 84. But it's the same logic there as well, so I can change this in those apps' code also.

jaladh-singhal: Yeah that line in Wise is also converting list of two strings to boolean. The whole idea is to type your params how they are behaving, both options are mutually exclusive

src/firefly/js/ui/DownloadDialog.jsx
```diff
@@ -262,21 +262,10 @@ export function TitleField({style={}, value, label='Title:', size=30}) {
...
-                label : 'File Structure:'
-            }}
-            options = {[
-                {label: 'Structured (with folders)', value: 'folder'},
-                {label: 'Flattened (no folders)', value: 'flat'}
-            ]}
-        />
-
-    );
+export function FileStructure({fieldKey='isFlattenedStructure'}) {
```
wmiipac, line 265: if you take param zipType out from DonwloadDialog, you need to update app download processor in ife for ztf, sofia, wise

src/firefly/js/ui/SwitchInputField.jsx
```diff
@@ -40,7 +39,6 @@ function handleOnChange(ev, viewProps, fireValueChange) {
     onChange?.(value);
 }
 
-
```
jaladh-singhal, line 43: these whitespace changes still show up in diff

PR #1756, opened 2025-05-05
FIREFLY-1695: Display error message when upload table can't be read

jaladh-singhal:
Fixes [FIREFLY-1695](https://jira.ipac.caltech.edu/browse/FIREFLY-1695)

- onSubmit of UploadTableChooser (used in Multi-object fields) now uses `determineValidity` like FileUploadPanel's onSubmit, that shows an error popup if invalid
- FileUploadPanel was maintaining a `message` state besides `report` but wasn't using it in generating invalidation popup, now made it do so by registering in `request.additionalParams` and passing it to `determineValidity`.
- Made `message` state to read `FileAnalysis.ErrorResponse` description from server-side for showing more specific error messages.

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1695-upload-error-msg/firefly/

Go to "Upload" tab -> Upload the bad table file in ticket -> click "load" button, the error popup should show VOTable parsing error instead of "Could not recognize the file type" error which is misleading.

Go to TAP -> Spatial -> Multi-object -> Upload the bad table file in ticket -> click "load" button, it should show an error popup with same error message as Upload tab.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java
@@ -228,8 +228,8 @@ public class VoTableReader {
                 voTablePath = tmpFile.getPath();
             } catch (Exception e) {
                 tmpFile.delete();
-                LOG.error(e);
-                throw new IOException("Unable to fetch URL: "+ location + "\n" + e.getMessage(), e);
+                LOG.error(e, "Unable to fetch URL: " + location);
+                throw new IOException(e.getMessage(), e);
             }
         } catch (MalformedURLException ex) { /* ok to ignore.  location may not be a URL */ }
 
@@ -238,8 +238,8 @@ public class VoTableReader {
             // at this point, voTablePath is a file path.
             return getVoTableRoot(new FileInputStream(voTablePath), policy);
         }  catch (Exception e) {
-            throw new IOException("Unable to parse VOTABLE from "+ location + "\n" +
-                    e.getMessage(), e);
+            LOG.error(e, "Unable to parse VOTABLE from "+ location);
+            throw new IOException(e.getMessage(), e);
         }
     }
 
@@ -250,7 +250,8 @@ public class VoTableReader {
             voFactory.setStoragePolicy(policy);
             return voFactory.makeVOElement(new BufferedInputStream(source), null);
         } catch (Exception e) {
-            throw new IOException("Unable to parse VOTable from stream: \n" + e.getMessage(), e);
+            LOG.error(e, "Unable to parse VOTable from stream");
+            throw new IOException(e.getMessage(), e);
         }
     }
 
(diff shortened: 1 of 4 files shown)


loitly (approved):
It’s working as described. However, if possible, it would be better to display the user’s original input instead of the server-side filename/path, which is meaningless to the user.

A related issue, possibly outside the scope of this PR, is with the IRSA Catalog’s Multi-Object search. If a bad file is uploaded, the search still runs but ends up stuck in a loading state.
(IRSA Catalog → Search Method = Multi-Object)

jaladh-singhal:
>However, if possible, it would be better to display the user’s original input instead of the server-side filename/path, which is meaningless to the user.

@loitly yes I had the same thought and also found the relevant code, but wasn't sure what to replace `location` here with?

https://github.com/Caltech-IPAC/firefly/blob/66314c89cdf08f3d92b96636ba49828e21c6f9d4/src/firefly/java/edu/caltech/ipac/table/io/VoTableReader.java#L241-L242

jaladh-singhal:
> A related issue, possibly outside the scope of this PR, is with the IRSA Catalog’s Multi-Object search. If a bad file is uploaded, the search still runs but ends up stuck in a loading state.
(IRSA Catalog → Search Method = Multi-Object)

Good catch, I didn't notice it. I just looked up this (and all ife apps) are using an old `UploadOptionsDialog` instead of newer `UploadTableChooser` that's being used everywhere else in Firefly. I'm not sure what was the reason of leaving it when @kpuriIpac migrated all upload components. @robyww do you remember? (if we decide to use `UploadTableChooser`, that migration can be done in a separate ticket)

robyww:
> I'm not sure what was the reason of leaving it when @kpuriIpac migrated all upload components. @robyww do you remember? 

the migration is not done. The image panel needs to be migrated as well.

loitly:
> @loitly yes I had the same thought and also found the relevant code, but wasn't sure what to replace location here with?
> OR is there a better way to extract only relevant information from server-side error responses at client-side?

I would change server-side code to only return the error message and allow the caller(client) to add context if needed.  So, in your case, the message would look like....

Error in File Retrieve: Unable to parse firefly_1695_bad_table.vot as a VOTABLE.
Reason: The element type "DESCRIPTION" must be terminated by the matching end-tag "</DESCRIPTION>".

kpuriIpac (approved):
Looks good, works for both uploads when you try to load the bad file. Agree with @loitly's comment to simplify the error message by taking out location.

kpuriIpac:
> > A related issue, possibly outside the scope of this PR, is with the IRSA Catalog’s Multi-Object search. If a bad file is uploaded, the search still runs but ends up stuck in a loading state.
> > (IRSA Catalog → Search Method = Multi-Object)
> 
> Good catch, I didn't notice it. I just looked it up, this (and all ife apps) are using an old `UploadOptionsDialog` instead of newer `UploadTableChooser` that's being used everywhere else in Firefly. I'm not sure what was the reason of leaving it when @kpuriIpac migrated all upload components. @robyww do you remember? (if we decide to use `UploadTableChooser`, that migration can be done in a separate ticket)

Yes, I can work on this in a new ticket if we want to use `UpoadTableChooser` here as well @robyww. I may have missed this in the initial migration.

jaladh-singhal:
> I would change server-side code to only return the error message and allow the caller(client) to add context if needed.

@loitly thanks, I logged the location and only returning exception's message. The error on client-side looks much cleaner now. Rebuilt the build

PR #1757, opened 2025-05-06
FIREFLY-1729: Client tables reset scroll on column width change

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1729
- should behave the same for both client and server backed table.

Test: https://fireflydev.ipac.caltech.edu/firefly-1729-table-reload-column-resize/firefly/
- Images -> M81; WISE -> Search
- Image Toolbar -> Show FITS header 'i' icon
- Scroll to the bottom of the table, then resize a column.  Scroll position should remain the same.


jaladh-singhal (approved):
works as expected. code looks good.

PR #1761, opened 2025-05-08
Firefly-1390 table upload ignore overlay

robyww:
#### testing 
- https://fireflydev.ipac.caltech.edu/firefly-1390-table-upload-ignore-overlay/firefly
- load some fits images
- create a file and upload the table below
- You should see some points in the bottom right and toward the middle
- add another image
- the points should be in the same place

Note - there will be no coverage since images point can's show up on HiPS

```
\CatalogOverlayType=IMAGE_PTS
\IMAGE_COLUMN=x;y
|     x|     y|index|what            |
|double|double| int | char           |
|      |      |     |                |
|  null|  null| null| null           |
  2.0     3.0   1     zero-zero
  5.5     6.0   2     half-zero
 100.0    50    3     zero-quarter
  88      150   4     half-half
```


loitly (approved):
Behaves as expected.  The changes look good.

PR #1763, opened 2025-05-09
Firefly-1736: URL API: add a go to tab

robyww:
#### Testing
- try: https://fireflydev.ipac.caltech.edu/firefly-1736-gourl/firefly/?api
- try: https://fireflydev.ipac.caltech.edu/firefly-1736-gourl/firefly/?api=go
- try: https://fireflydev.ipac.caltech.edu/firefly-1736-gourl/firefly/?api=go&to=mon
- try: https://fireflydev.ipac.caltech.edu/firefly-1736-gourl/firefly/?api=go&to=Upload
- try other stuff

### DIFF
--- a/src/firefly/js/Firefly.js
+++ b/src/firefly/js/Firefly.js
@@ -522,14 +522,14 @@ function mergeAppOptions(ops, overrideOps) {
 }
 
 function renderRoot(root, viewer, props, webApiCommands) {
-    const e= document.getElementById(props.div);
-    if (!e) {
+    const element= document.getElementById(props.div);
+    if (!element) {
         showInfoPopup('HTML page is not setup correctly, Firefly cannot start.');
         logger.error(`DOM Element "${props.div}" is not found in the document, Firefly cannot start.`);
         return;
     }
 
-    const rootToUse = root ?? createRoot(e);
+    const rootToUse = root ?? createRoot(element);
     initWebApi(webApiCommands);
     const webApi= isUsingWebApi(webApiCommands);
     const doAppRender= () => {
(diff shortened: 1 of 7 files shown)


loitly (approved):
I tested the changes and they worked well. However, I’d like to suggest an alternative URL scheme that I believe makes more sense overall.

From your examples, the general pattern seems to be:
```ini
api=<command>&<parameters>
```
But the values used for api aren’t really commands, they’re nouns that represent core components like modules or functions.

With that in mind, I suggest renaming the current `go` command to `menu`, since that's what it really represents.

Also, the current `to` parameter uses arbitrary values that don’t map to anything in the application, making them hard to trace. I suggest replacing it with the existing `view=[value]` used in the app's URL when a menu is clicked. This would provide a direct and consistent connection between the API and the application.

So, instead of: `?api=go&to=SIAv2 Searches`
it would become: `?api=menu&view=SIAv2Search`

robyww:
> Also, the current to parameter uses arbitrary values that don’t map to anything in the application, making them hard to trace. I suggest replacing it with the existing view=[value] used in the app's URL when a menu is clicked. This would provide a direct and consistent connection between the API and the application.

> So, instead of: ?api=go&to=SIAv2 Searches
it would become: ?api=menu&view=SIAv2Search

The point of this is to make it unnecessary for the user to have to know the internal action name. That is why it matches on the label, partial label, or action.

That is one of the reasons I did not use `view`. Also- I have always thought `view` was confusing so I did not want to expose that to the end user.

I am happy to change the api to `menu`. I tried three of four different names already so that is a good suggestion.

PR #1765, opened 2025-05-12
FIREFLY-1327: Update job info dialog

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1327

Implemented as described in the ticket, with two exceptions:
- Owner has been moved to the bottom section to provide a clearer separation between the top and bottom.
- The default proportional font is used, so the dates are not perfectly aligned as shown in the ticket. Would you like me to switch all values to a monospaced font for better alignment?  But, this will cause the dialog's font to deviate from the rest of the application.

Test: https://fireflydev.ipac.caltech.edu/firefly-1327-jobinfo-dialog/firefly/
Submit a few TAP or IRSA Catalog searches, then go to `Job Monitor` and click the Job Info (i-icon) to view details.

### DIFF
--- a/src/firefly/js/core/background/JobInfo.jsx
+++ b/src/firefly/js/core/background/JobInfo.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {getJobInfo} from './BackgroundUtil.js';
+import {getJobInfo, isTapJob, isUWS} from './BackgroundUtil.js';
 import {useStoreConnector} from '../../ui/SimpleComponent.jsx';
 import {KeywordBlock} from '../../tables/ui/TableInfo.jsx';
 import {PopupPanel} from '../../ui/PopupPanel.jsx';
@@ -12,6 +12,7 @@ import {uwsJobInfo} from 'firefly/rpc/SearchServicesJson.js';
 import {Box, Skeleton, Stack} from '@mui/joy';
 import {TableErrorMsg} from 'firefly/tables/ui/TablePanel.jsx';
 import {showInfoPopup} from 'firefly/ui/PopupUtil';
+import {PrismADQLAware} from '../../ui/tap/AdvancedADQL';
 
 const dialogID = 'show-job-info';
 
@@ -28,14 +29,15 @@ export function isJobInfoOpen() {
 }
 
 export function showJobInfo(jobId) {
+
     const popup = (
-        <PopupPanel title='Job Information' >
-            <Stack key={jobId} sx={popupSx}>
-                <JobInfo jobId={jobId} sx={{overflow: 'auto'}}/>
-                <HelpIcon helpId={'basics.bgJobInfo'} sx={{ml: 'auto'}}/>
-            </Stack>
-        </PopupPanel>
-    );
+            <PopupPanel title='Job Information'>
+                <Stack key={jobId} sx={popupSx}>
+                    <JobInfo jobId={jobId} sx={{overflow: 'auto'}}/>
+                    <HelpIcon helpId={'basics.bgJobInfo'} sx={{ml: 'auto'}}/>
+                </Stack>
+            </PopupPanel>
+        );
     DialogRootContainer.defineDialog(dialogID, popup);
     dispatchShowDialog(dialogID);
 }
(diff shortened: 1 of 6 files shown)


src/firefly/js/tables/ui/TableInfo.jsx
```diff
@@ -157,18 +158,18 @@ const ShowObjectTreeBtn = ({data, title}) => {
...
 }
 
-export function Keyword({label, value, title, asLink}) {
-    label = label && label + ':';
-    if (label || value) {
-        value = String(value);
+export function Keyword({label=null, value, title, asLink}) {
+    if ((label !== null) || value) {
+        label = label ? label + ':' : '\u00A0';
+        value = value ? String(value) : '\u00A0';
```
jaladh-singhal, line 172: have you tried using `'&nbsp;'` here instead? 

I tend to avoid using raw unicode characters where possible for readability.

jaladh-singhal (commented):
Looks good - left minor comments in code (sorry for nitpicking)

Two probably important things from ticket that I notice missing are:
1. Copy icon button next to job id.
2. the "query" field in Parameters should be easier to find. This can be done either by moving the "query" to the top of parameters key-value pairs or by giving the query value a different font color (maybe 'primary' aka blue or 'success' aka green?) or by doing both.

robyww (commented):
Could you detect a select statement and do some simple formatting like we do with the ADQL panel. 

You would have to parse apart the `select`, `columns`, and `from`. There are some utils functions buried in TAP (`makeColsLines()`)

I think it would look better and me more skimable.

loitly:
I updated the test instance with the suggested changes.  Let me know what you think.
https://fireflydev.ipac.caltech.edu/firefly-1327-jobinfo-dialog/firefly/

src/firefly/js/tables/ui/TableInfo.jsx
loitly: '&nbsp;' doesn't work.  It treats it as plain text.

robyww (approved):
The AQDL looks good!

src/firefly/js/core/background/JobInfo.jsx
```diff
@@ -79,12 +81,13 @@ export function JobInfo({jobId, ...props}) {
 }
 
 export function UwsJobInfo({jobInfo, sx, isOpen=false}) {
-    const {results, parameters, errorSummary, meta, jobInfo:aux, ...rest} = jobInfo;
+    const {results, parameters, errorSummary, jobInfo:aux} = jobInfo;
     const hrefs = results?.map((r) => r.href);
     const hasMoreSection = hrefs || parameters || errorSummary || aux;
     return (
         <Stack spacing={.5} p={1} sx={sx}>
```
jaladh-singhal, line 88:
```suggestion
        <Stack spacing={1} p={1} sx={sx}>
```
Looks tight, JobInfoDetails, TAPDetails, etc. already have their children with 0.5 vertical space (at least visually). To make these visual groups separate, consider using higher vertical spacing.

src/firefly/js/core/background/JobInfo.jsx
```diff
@@ -130,6 +133,29 @@ function JobInfoDetails({jobId, ownerId, phase, creationTime, startTime, endTime
...
+    const lang = jobInfo?.parameters?.lang;
+    const params = jobInfo?.parameters || {};
+    const adql = params[Object.keys(params).find((k) => k.toLowerCase() === 'query')];
+
+    if (!adql || (lang && lang.toUpperCase() !== 'ADQL')) return null;
+
+    return (
+        <Stack spacing={0}>
+            <KeywordBlock label='ADQL QUERY'/>
+            <PrismADQLAware text={adql} sx={{marginBlock: '-8px'}}/>
```
jaladh-singhal, line 146:
```suggestion
            <PrismADQLAware text={adql} sx={{marginBlock: '-8px', fontSize: 'fontSize.sm'}}/>
```
Everyone else using sm font size in this compact panel except this Prism block

jaladh-singhal (approved):
Thanks for updating it. Looks great.

Left 2 cosmetic change suggestions to make it even more polished (feel free to ignore if you don't like them). Incorporating them will make the dialog look like this (from dev tools hacking):
image

On a side note, the ADQL block in this dialog is not theme responsive - in dark mode, it still stays light (unlike ADQL examples in TAP ADQL panel)

src/firefly/js/core/background/JobInfo.jsx
loitly: Did you mean fontSize: `var(--joy-fontSize-sm)`? Using `fontSize.sm` didn’t work for me.

jaladh-singhal: Oh, I got confused. The right syntax should be `fontSize: 'sm'` - sx should recognize it

PR #1766, opened 2025-05-13
Firefly-1740,48: clean up the color,strech,and otherlay sync button and add to other places

robyww:
#### [Firefly-1740](https://jira.ipac.caltech.edu/browse/FIREFLY-1740), [Firefly-48](https://jira.ipac.caltech.edu/browse/FIREFLY-48): Add `color,strech,and otherlay` lock button to stretch panels

  - [Firefly-1740](https://jira.ipac.caltech.edu/browse/FIREFLY-1740): done
  - [Firefly-48](https://jira.ipac.caltech.edu/browse/FIREFLY-48): done
  - Fixes bug: an issue `Hue preserving stretch`. Scaling coefficients not working
  - Changes `Hue preserving stretch` change  to only work with refresh button like the rest of the panel
  - A little clean on datalink related processing

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1740-color/firefly
- [Firefly-1740](https://jira.ipac.caltech.edu/browse/FIREFLY-1740): `color,strech,and otherlay` lock button is in both color and stretch dropdown. And it color and stretch dialog
- [Firefly-48](https://jira.ipac.caltech.edu/browse/FIREFLY-48): no bounds checking on lower/upper range in data mode
- `Hue preserving stretch` Scaling coefficients now work: you will see the data change when the are changed
-  `Hue preserving stretch` only refreshed when button is pushed.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/DirectStretchUtils.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/visualize/DirectStretchUtils.java
@@ -81,7 +81,7 @@ public class DirectStretchUtils {
         int bPosQuarter=0;
         Band[] bands= state.getBands();
         ThreeCComponents tComp= get3CComponents(frGroup,sv.totWidth,sv.totHeight,state);
-        RGBIntensity rgbI= get3CRGBIntensity(state.getRangeValues(),frGroup,bands);
+        RGBIntensity rgbI= get3CRGBIntensity(state,frGroup,bands);
          new ArrayList<Stretch3CTile>(sv.tileLen);
 
         var sTileList =doTileStretch(sv,tileSize, Stretch3CTile::new,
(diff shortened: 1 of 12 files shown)


lrebull:
1740 looks good; i had a question about where we should (should not?) put the arrow that exists in the color table pulldown. (the 42 second movie is in slack but i can copy it over here if you want)

FIREFLY-48 (wow that's an old ticket!!) looks good too as far as i can tell

hue preserving stretch stuff:

scaling coefficients seem to have two minus signs, one of which is on top of the 1 in 10.699 so it looks very odd. but that appears to be the way it is on ops too so oh well.
Screenshot 2025-05-13 at 4 24 10 PM

i don't know how to reconcile both the statements above: "you will see the data change when the are changed" AND "only refreshed when button is pushed" - which one of these are supposed to be what happens? am i supposed to see the data change right away, or am i supposed to only see it refresh when the 'refresh' button is pushed?

robyww:
@lrebull 
> am i supposed to see the data change right away, or am i supposed to only see it refresh when the 'refresh' button is pushed?

Only when the `refresh` button was pushed. I change this because we have had a long time issue with `Hue preserving stretch` overwhelming the server as the scale bar is moved. This is one of the most heavy calls to the server, sometimes required moving a half gig of data. This should not happen when the user drags. It could cause the call to happen 10 or 20 times. It was a really bad problem that I have been meaning to address for a long time.

Also, the whole rest of the dialog only updates with `refresh`. So `Hue preserving stretch` was not only a problem is was inconsistent. It needed to be fixed.

PR #1768, opened 2025-05-15
Firefly-1739: better handle table filtering when nothing is selected

robyww:
#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1739-filtering/firefly
- search two or three catalogs
- do a selection that does not include any - you should get a warning
- do a selection that just covers some
   - you should get a warning
   - the covered tables will be filtered
   - the non-covered tables will be filtered to zero rows


lrebull:
Works! Marvelous!

Please change popup from this:
_2 tables not covered
2 tables are not covered in this selection, they will be filtered to zero rows_
to this
_Tables omitted
This selection omits 2 tables. They will be filtered down to have zero rows._

PR #1770, opened 2025-05-19
Firefly-1745: Job info size and position stays the same as the use clicks around

robyww:
#### [Firefly-1745](https://jira.ipac.caltech.edu/browse/FIREFLY-1745): Job info size and position stays the same as the use clicks around

- Also added a `Box` around the ADQL display

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1745-job-info/firefly
- make 2 or three jobs
- go to Job Monitor
- click on `i` and move/resize the dialog
- click on other rows. the dialog should stay in place.


loitly (approved):
Works perfectly.  Thanks.

PR #1771, opened 2025-05-21
IRSA-6794: Add Search Type Options optionally inside Embed panel

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/IRSA-6794

See **IFE PR** for details: https://github.com/IPAC-SW/irsa-ife/pull/413

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -1,6 +1,6 @@
 import {Box, Sheet, Stack, Typography} from '@mui/joy';
 import React, {useContext, useEffect, useState} from 'react';
-import {oneOf, bool, string, number, arrayOf, object, func, shape, elementType} from 'prop-types';
+import PropTypes, {oneOf, bool, string, number, arrayOf, object, func, shape, elementType} from 'prop-types';
 import {defaultsDeep} from 'lodash';
 import CoordinateSys from '../../visualize/CoordSys.js';
 import {CONE_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY} from '../../visualize/ui/CommonUIKeys.js';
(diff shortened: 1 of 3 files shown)


src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -412,6 +412,7 @@ function SpatialSearch({rootSlotProps: slotProps, insetSpacial, uploadInfo, setU
                 tooltip: 'Chose type of search', initialState: {value: initToggle},
                 options: searchTypeToggleOptions
             }} />}
+            {slotProps.searchTypeOptions}
```
robyww, line 415: -  This is probably the wrong name. `searchTypeOptions` is confusing with `searchTypeOp` right below.  From `EmbeddedPositionSearchPanel`'s perspective it is just a widget that goes in `SpatialSearch`. I think a better name is `midDetailsWidget`
- This should be part of the `SpatialSearch` slotProps
- so `spatialSearch.MidDetailsWidget`
   - change line 405 to be
 ```js
      const { searchTypeKey=CONE_AREA_KEY, midDetailsWidget, sx } = slotProps.spatialSearch ?? {};
```
- This should be added to `EmbeddedPositionSearchPanel.propTypes`
- I did not put this in the UI comments but the spacing it too tight when you add and extra component. I think you change change the `spacing` with the slotProp: `spacialSearch.sx`

robyww (changes requested):
This concept is fine here, but I gave several suggestions below better integrate it into the `EmbeddedPositionSearchPanel`

There are all very small, quick changes.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
jaladh-singhal: I second what @robyww said. Just a suggestion: `midDetailsWidget` can be defined as `children` because it's basically a slot in this component to render another component.

kpuriIpac:
@robyww @jaladh-singhal done, with with a mix of suggestions and used `children` passed to `SpatialSearch`

robyww (approved):

jaladh-singhal (approved):
Looks better!

PR #1782, opened 2025-05-29
FIREFLY-1756: Download Script 'null' in file name fix

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1756
- @loitly when `DownloadScript` is called here: `new DownloadScript.UrlInfo(r.href(), null, null)` in `EmailNotification.java` (line 141), the suffix `null` was being formatted into the string `null`, which then added `null` to the filename as a suffix
  - because of line ~236 in `DownloadScript.java`
  - so i added a check for suffix being null
  - this can be made better though, and I need to do more cleanup of the download script post release based on your other feedback as well. Anything else you would like me to address pre-release? 

**Testing**:
**Wise**: https://firefly-1756-download-bug.irsakudev.ipac.caltech.edu/applications/wise
**Sofia**: https://firefly-1756-download-bug.irsakudev.ipac.caltech.edu/applications/sofia
- test according to ticket instructions (Luisa's latest comment is helpful)
- in the download script you should not see a suffix 'null' anymore, instead it'll be empty  
- @lrebull now if you run the script, you should not see **null** in the filename anymore


lrebull:
confirm this is now fixed!!

loitly (approved):
Looks good!

PR #1786, opened 2025-06-03
Firefly-1667: when HiPS is changed, the associated MOC is added

robyww:
####[ Firefly-1667](https://jira.ipac.caltech.edu/browse/FIREFLY-1667): when HiPS is changed, the associated MOC is added to the drawing layers

- _Note:_ MOCs are always added not replaced

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1667-moc-adding/firefly
- Be aware the the MOCs are only added not replaced.  The user can always delete a MOC


lrebull:
OK, I still think this is a little weird from the user perspective, but that's ok! This is better than it was, and it works as you describe.

PR #1788, opened 2025-06-04
FIREFLY-1771: Fix Existing Table Load missing rows

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1771
- fixes the above ticket
- add a loading animation for bigger tables (similar to the one on the upload tab) 

**Testing**: 
**Euclid**: https://firefly-1771-row-mismatch.irsakudev.ipac.caltech.edu/applications/euclid
- test Euclid Catalogs according to ticket
   - here's what I did: 
     - Select the following catalog & search: `euclid_q1_spectro_zcatalog_spe_galaxy_candidates`
     - Back on Euclid Catalogs, now select: `euclid_q1_phz_photo_z`, but don't click on Search
     - Now under `Object ID search`, select `Load object IDs from a table` and select the first table you searched for from the `Loaded Tables` tab
- For both the ID columns, select `object_id`
- Do a search. You should get back several thousand rows (50,000  if you never changed row limit)
   -previously, you would only get back 100 (default page size)

### DIFF
--- a/src/firefly/js/ui/UploadTableChooser.js
+++ b/src/firefly/js/ui/UploadTableChooser.js
@@ -1,5 +1,5 @@
-import {Stack, Typography, Box} from '@mui/joy';
-import React from 'react';
+import {Stack, Typography} from '@mui/joy';
+import React, {useState} from 'react';
 
 import {dispatchHideDialog, dispatchShowDialog} from '../core/ComponentCntlr.js';
 import {FileAnalysisType, Format} from '../data/FileAnalysis.js';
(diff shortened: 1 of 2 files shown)


lrebull:
the 100 row limit is fixed here!
i am not sure what the "loading animation for bigger tables" would be so i need help if you want me to test that.

robyww (approved):
Looks good!

kpuriIpac:
> the 100 row limit is fixed here! i am not sure what the "loading animation for bigger tables" would be so i need help if you want me to test that.

It is essentially this: 

Screenshot 2025-06-04 at 10 28 31 AM

Previously, if you clicked on `Load Table` here, and say the table was big enough that it took 3-5 seconds to load, you would just not see the dialog being greyed out the the small loading circle. It would just stay as is, probably feeling like you're stuck. And eventually when the table is loaded, the popup would close. Since we were doing this only for 100 rows earlier, this was almost instant so you would never notice the lag. So this was something we anyway needed to add here.

PR #1791, opened 2025-06-07
FIREFLY-1754: Fix poor positioning of App Hints

jaladh-singhal:
Fixes [FIREFLY-1754](https://jira.ipac.caltech.edu/browse/FIREFLY-1754)

- Moved AppHint code from Landing page to its own file
- Made AppHint accept an `anchorNode` prop that is used to calculate its position relative to that anchor
- Stored tabs' DOM refs at `MenuTabBar` in the layout info store and then reading them in `LandingPage` as AppHint anchor nodes.
- Removed hardcoded position settings (that broke with job monitor refactoring) in different apps. See:
  - IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/421
  - SUIT PR: https://github.com/lsst/suit/pull/62

## Testing
Go to each app and see if app hints are positioned correctly (if they are not present, go to dev tools -> "Application" tab -> clear all preferences by pressing none icon, reload the page):

Note: For job monitor hint to appear, you need to perform some catalog search and once results are avaialabe reload the page while staying at results/landing-page

1. Firefly: https://fireflydev.ipac.caltech.edu/firefly-1754-bgmon-hint/firefly
2. Firefly template in IV: https://firefly-1754-bgmon-hint.irsakudev.ipac.caltech.edu/irsaviewer
3. Routed template in spherex: https://firefly-1754-bgmon-hint.irsakudev.ipac.caltech.edu/applications/spherex
4. Hydra template in wise: https://firefly-1754-bgmon-hint.irsakudev.ipac.caltech.edu/applications/wise

### DIFF
--- a/src/firefly/js/ui/AppHint.jsx
+++ b/src/firefly/js/ui/AppHint.jsx
@@ -1,0 +1,109 @@
+import {useStoreConnector} from 'firefly/ui/SimpleComponent';
+import {dispatchAddPreference, getPreference} from 'firefly/core/AppDataCntlr';
+import React, {useLayoutEffect, useRef, useState} from 'react';
+import {Button, Snackbar} from '@mui/joy';
+import TipsAndUpdates from '@mui/icons-material/TipsAndUpdates';
+import {object, oneOf, string} from 'prop-types';
+
+
+export const APP_HINT_IDS = {
+    TABS_MENU: 'tabsMenu',
+    BG_MONITOR: 'bgMonitor'
+};
+
+export const HINT_TIP_PLACEMENTS = {
+    START: 'start',
+    MIDDLE: 'middle',
+    END: 'end'
+};
+
+/**An app hint needs to be shown only the first time user loads an app. So this is controlled by a flag saved as app preference**/
+export const appHintPrefName = (appTitle, hintId) => `showAppHint__${appTitle}--${hintId}`;
+
+export function AppHint({appTitle, id, anchorNode, hintText, tipPlacement=HINT_TIP_PLACEMENTS.MIDDLE, sx={}}) {
+    const showAppHint = useStoreConnector(() => getPreference(appHintPrefName(appTitle, id), true));
+    const appHintRef = useRef();
+
+    // AppHint is rendered inside LandingPage, so we cannot yet compute top/bottom from the anchor node but only left/right
+    const [anchorRelativePosSx, setAnchorRelativePosSx] = useState({});
+    useLayoutEffect(() => {
+        if (anchorNode) {
+            const anchorRect = anchorNode.getBoundingClientRect();
+            const appHintRect = appHintRef.current?.getBoundingClientRect() ?? {width: 0};
+            const posSx = {left: 'auto', right: 'auto'};
+            switch (tipPlacement) {
+                case HINT_TIP_PLACEMENTS.START:
+                    posSx.left = anchorRect.left;
+                    break;
+                case HINT_TIP_PLACEMENTS.END:
+                    posSx.right = `calc(100vw - ${anchorRect.right}px)`;
+                    break;
+                case HINT_TIP_PLACEMENTS.MIDDLE:
+                default:
+                    posSx.left = anchorRect.left + (anchorRect.width/2) - (appHintRect.width/2); // to center the hint
+                    break;
+            }
+            setAnchorRelativePosSx(posSx);
+        }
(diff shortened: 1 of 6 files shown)


src/firefly/js/ui/AppHint.jsx
```diff
@@ -0,0 +1,109 @@
...
+/**An app hint needs to be shown only the first time user loads an app. So this is controlled by a flag saved as app preference**/
+export const appHintPrefName = (appTitle, hintId) => `showAppHint__${appTitle}--${hintId}`;
+
+export function AppHint({appTitle, id, anchorNode, hintText, tipPlacement=HINT_TIP_PLACEMENTS.MIDDLE, sx={}}) {
+    const showAppHint = useStoreConnector(() => getPreference(appHintPrefName(appTitle, id), true));
+    const appHintRef = useRef();
+
+    // AppHint is rendered inside LandingPage, so we cannot yet compute top/bottom from the anchor node but only left/right
+    const [anchorRelativePosSx, setAnchorRelativePosSx] = useState({});
+    useLayoutEffect(() => {
```
loitly, line 29: Good use of useLayoutEffect.

loitly (approved):
Changes look good.

src/firefly/js/ui/Menu.jsx
```diff
@@ -128,10 +134,15 @@ function AdjustableMenu({menuTabItems, helpItem, selected, dropDown, showUserInf
...
 
-    const setElement= useCallback( (key,e) => {
-        if (!e) return;
+    const setElement= useCallback( (key,el) => {
+        if (!el) return;
+
+        // save the first and last tab elements in the layout info store, needed for positioning the app hints
+        if (key==='0') dispatchUpdateLayoutInfo({menuTabNode: {first: el}});
+        if (key===menuTabItems.length-1+'') dispatchUpdateLayoutInfo({menuTabNode: {last: el}});
+
```
robyww: I am concerned you are to aggressively calling calling redux here. We make too many layout info update anyway. I am sort of surprise it works at all but since it does here a a couple of sugestion.

- only call if the element not set
```js
  if (getLayouInfo().menuTabNode.last!==el) dispatchUpdateLayoutInfo({menuTabNode: {last: el}});
```
- set the first and last in one call.
   - set the first in a ref
   - make the call when you are ready to call the last then set both

Both of these change will drastically reduce the call to redux.

src/firefly/js/ui/Menu.jsx
```diff
@@ -192,13 +203,13 @@ function MenuTabBar({menuTabItems=[], size, selected, dropDown, displayMask, set
...
     const tabItems= [
-        <ResultsTab {...{key:'results-tab', size, color, variant, ref: (c) => setElement('results-tab ',c)}}/>,
+        <ResultsTab {...{key:'results-tab', size, color, variant, ref: (el) => setElement('results-tab ', el)}}/>,
         ...menuTabItems
             .filter( isItemEnabled)
             .map(({action,label,TabRenderer,title}, idx) =>
             {
                 const tabProps = {key: idx, value:action, disableIndicator:true, color, variant,
-                    ref: (c) => setElement(idx+'',c),
+                    ref: (el) => setElement(idx + '', el),
```
robyww, line 224: this is a better name!

robyww (approved):
Good job. This was sort of hard problem.

I am approving, but I would like you to make the changes I suggested in `Menu.jsx` before you merge.

src/firefly/js/ui/Menu.jsx
jaladh-singhal: > I am concerned you are to aggressively calling calling redux here. We make too many layout info update anyway. I am sort of surprise it works

I was calling it aggressively earlier when I had this code in consumer of `setElement` - I was getting maximum depth reached. Then I moved it to inside `setElement` which is wrapped in `useCallback` so runs only when el changes.

`if (getLayouInfo().menuTabNode.last!==el) ` - isn't useCallback already doing that because `key` and `el` are its arguments?

robyww: I added a logpoint. The callback is being called 2 times for every button. So I think the dispatch is happening 4 times. All this happens for every tab click. Before there was not that much overhead, but if it is doing a dispatch then I am more concerned.

jaladh-singhal: Oh, I see - will make this change then

PR #1792, opened 2025-06-09
FIREFLY-1778: Update the cleanup script so that it only targets its designated work directories

loitly:
See ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1778


robyww (approved):
Looks good.

PR #1796, opened 2025-06-12
FIREFLY-1781: Fix spherex list of target search never returns

jaladh-singhal:
Fixes [FIREFLY-1781](https://jira.ipac.caltech.edu/browse/FIREFLY-1781)

At L156, ADQL query creation was using original table name instead of alias when there are no selected columns (which is the case with spherex that has column constraints turned off). Now alias `tableAsName` is used whenever `isUpload` is true.

## Testing
https://firefly-1781-lvf-multiobject.irsakudev.ipac.caltech.edu/applications/spherex

In spectral image search, select "Multi-object" -> upload the tables in the ticket -> the query should complete instead of getting stuck or failing:
- `example2input-short.tbl`: Luisa's original table doesn't have any covered targets so it should come back with no data found
- `modded-example2input-short.tbl`: I created this table with 3 targets covered by spherex, so it should come back with data.

### DIFF
--- a/src/firefly/js/ui/tap/TapSearchSubmit.js
+++ b/src/firefly/js/ui/tap/TapSearchSubmit.js
@@ -132,14 +132,15 @@ export function getAdqlQuery(tapBrowserState, additionalClauses, allowColumnCons
     }
 
     const helperFragment = getHelperConstraints(tapBrowserState);
+    const tableAsName = getAsEntryForTableName(tableName); // alias is used when upload table is present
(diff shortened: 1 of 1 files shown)


lrebull:
yes, this is now working! thanks!

robyww (approved):
Looks good. I had to stair at it for awhile to see exactly what was broken and why we had not found it in TAP. I see now in TAP we never upload `*`. Very subtle. Good Job!

PR #1797, opened 2025-06-12
FIREFLY-1759: Enable downloads for Datalink tables

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1719
**IFE PR**: https://github.com/IPAC-SW/irsa-ife/pull/423

- Code should be fairly straightforward, just needed a bit of work to get it going and to make existing code adapt to work directly with datalink table (as opposed to obscore/non-obscore tables **_containing_** datalink, which is what our primary use case was)

**Testing**: 
- Euclid: https://firefly-1759-datalink-downloads.irsakudev.ipac.caltech.edu/applications/euclid
- do an image or object search, then extract one or more rows
- select one or more rows in the extracted Products table and the download script 
   -  you may even filter the table and then try this. For example, try searching for cutouts and download all cutouts. Ensure you see cutouts in the download script in this case. 
- try changing the cutout selection (cutout size and/or search target) directly from the products table. Any of the search targets should work as expected same as they do for the image/objects table. 
- Regression testing: select a few rows and download products on the images/objects results table as well, to ensure there are no new issues introduced there.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
@@ -15,6 +15,7 @@ import org.json.JSONObject;
 import java.io.IOException;
 import java.net.URL;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import static edu.caltech.ipac.firefly.server.db.EmbeddedDbUtil.getSelectedData;
 import static org.apache.commons.lang.StringUtils.isEmpty;
@@ -35,6 +36,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
     public static final String SEMANTICS = "semantics";
     public static final String CONTENT_TYPE = "content_type";
     public static final String FILE = "file";
+    public static final String SOURCE = "source";
 
     public static final String DATALINK_SER_DEF = "datalinkServiceDescriptor";
     public static final String ADHOC_SERVICE = "adhoc:service";
@@ -59,6 +61,11 @@ public class ObsCorePackager extends FileGroupsProcessor {
     private static final List<String> CUTOUT_UCDs= Arrays.asList("phys.size","phys.size.radius","phys.angSize", "pos.spherical.r");
     private static final List<String>  RA_UCDs= List.of("pos.eq.ra");
     private static final List<String>  DEC_UCDs= List.of("pos.eq.dec");
+
+    private static final List<String> SERVICE_DESC_COL_NAMES = List.of(
+            "id", "access_url", "service_def", "error_message", "semantics",
+            "description", "content_type", "content_length"
+    );
 
     record CenterCols(String lonCol, String latCol) {}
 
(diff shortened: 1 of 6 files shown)


lrebull:
LOL my images tab no longer has a "Generate Download Script" button!!
Screenshot 2025-06-12 at 4 14 07 PM
euclid_Objects_1-curl.sh seems ok
Objects_1_Productsrow1-curl.sh has comments at the top that include 'null': "Download null from IRSA" instead of "Download Euclid from IRSA" which is what the other one has.

lrebull:
ok, very weird.. tried again to reproduce the missing "generate download script" button, including several times with new private window, and now it seems to be working ok. 🙄 so it's 'just me' again??

kpuriIpac:
> ok, very weird.. tried again to reproduce the missing "generate download script" button, including several times with new private window, and now it seems to be working ok. 🙄 so it's 'just me' again??

if it helps..I think it had happened with me *once* while I was working, I immediately tried to recreate it but wasn't able to, so just dismissed it. I'll keep an eye out, though.

src/firefly/js/metaConvert/TableDataProductUtils.js
```diff
@@ -181,13 +188,19 @@ export function extractDatalinkTable(table,row,title,setAsActive=true) {
...
     if (rowWP) dataTableReq.META_INFO[MetaConst.ROW_TARGET]= rowWP.toString();
+    const serviceId = getMetaEntry(table.tbl_id,MetaConst.DATA_SERVICE_ID); //use service id as the viewer id for MultiProductViewer
 
     dispatchTableSearch(dataTableReq, {setAsActive, logHistory: false, showFilters: true, showInfoButton: true});
+    onTableLoaded(dataTableReq.tbl_id).then( () => {
+        const {tbl_ui_id, leftButtons=[]}= getTableUiByTblId(dataTableReq.tbl_id) ?? {} ;
+        //todo: dataSource for euclid (as an example) is 'euclid' - is there a better alternative than serviceId to use generically as the dataSource for the Download Script?
+        leftButtons.unshift(() => <PrepareDownload tbl_id={dataTableReq.tbl_id} viewerId={serviceId} downloadType={'script'} dataSource={serviceId}/>);
+        dispatchTableUiUpdate({tbl_ui_id, leftButtons});
+    });
```
robyww: I don't think this is the right place for this code. We should not put UI code in a Util.

why don't you return a promise
so
```javascript
export async function extractDatalinkTable(table,row,title,setAsActive=true)
```


that at the end
```javascript
return onTableLoaded(dataTableReq.tbl_id);
```

then the caller can do
```javascript
await extractDatalinkTable(table,row,table.title + `: Products row ${row+1}`);
<ui code>
```

src/firefly/js/ui/tap/Cutout.js
```diff
@@ -168,14 +167,18 @@ export function findPreferredCutoutTarget(dataProductsComponentKey=DEFAULT_DATA_
...
+            else positionWP = makeWorldPtUsingCenterColumns(table, row);
             foundType = ROW_POSITION;
         }
     } else if (cutoutType === ROW_POSITION) {
-        positionWP = makeWorldPtUsingCenterColumns(table, row);
+        if (isDatalinkTable(table)) {
+            //positionWP to come from metadata
+            positionWP = parseWorldPt(getMetaEntry(table, MetaConst.ROW_TARGET, undefined));
+        }
+        else positionWP = makeWorldPtUsingCenterColumns(table, row);
```
robyww: usually if you have an if with brackets you should do the same with else.  I just noticed this because at first I missed the else when looking at the code.

src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
```diff
@@ -145,7 +161,11 @@ public static List<FileInfo> parseDatalink(URL url, DownloadRequest request, Map
         String cutoutValue = request.getParam(CUTOUT_VALUE);
 
         try {
-            DataGroup[] groups = VoTableReader.voToDataGroups(url.toString(), false);
+            DataGroup[] groups;
+            if (data != null) { //if the request is from a datalink table, we will get the DataGroup object created directly from the table's selected rows
+                groups = new DataGroup[]{data};
+            }
+            else groups = VoTableReader.voToDataGroups(url.toString(), false); //for obscore/non obscore tables containing datalink, we use the acesss_url of the datalink
```
robyww: same here with the else. It threw me off again.

robyww (approved):
Looks good. Just do the clean up a mentioned and it should be ready to merge.

src/firefly/js/voAnalyzer/VoDataLinkServDef.js
```diff
@@ -97,7 +97,7 @@ function getSDDescription(table, ID) {
     const descriptionCol = getColumnIdx(table, 'description');
     if (descriptionCol === -1 || serviceDefCol === -1) return;
     if (table.totalRows > 50) return;
-    const sdRow = table.tableData.data.find((dAry) => dAry[serviceDefCol] === ID);
+    const sdRow = table?.tableData?.data?.find((dAry) => dAry[serviceDefCol] === ID);
```
kpuriIpac, line 100: added this check @robyww because doing a column filter search on the data products table was crashing the app if the result returned nothing

PR #1800, opened 2025-06-17
FIREFLY-1677: Better messaging for Position Cols in TAP Spatial Search

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1677
- See ticket for changes requested
- Also made the text a bit smaller
- The text for empty columns might be a bit too long. Should we consider a shorter version or does it look ok @gpdf?

**Testing**: https://fireflydev.ipac.caltech.edu/firefly-1677-tap-msg-improve/firefly
- Test according to ticket
- if you select a table with pre-filled columns, you should see the default message (`These are the recommended columns...`)
- If you now manually select any columns, you should see the user selected message (`User-specified coordinate columns...`)
- if you remove a column selection or get a table with no pre filled columns (try `TAP_SCHEMA` under `IRSA`), you should get an empty columns message (`Unable to identify coordinate columns...`)

### DIFF
--- a/src/firefly/js/ui/UploadTableSelector.jsx
+++ b/src/firefly/js/ui/UploadTableSelector.jsx
@@ -1,4 +1,4 @@
-import React, {useContext, useEffect} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 import {Box, Chip, Stack, Tooltip, Typography} from '@mui/joy';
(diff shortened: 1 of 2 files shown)


wmiipac (approved):
The logic and messages are looking good. 
Tested UI, it is working as expected.

PR #1802, opened 2025-06-25
Firefly 1354 timeseries save table filename

wmiipac:
View [ticket](https://jira.ipac.caltech.edu/browse/FIREFLY-1354) for details. I added two test files in the ticket for testing purpose.

To test:
https://firefly-1354-ts.irsakudev.ipac.caltech.edu/irsaviewer/ts.html

1. upload ztt-ts.tbl or wise-ts.tbl 
2. follow the steps in Luisa's video, by applying filter on the chart, then save the filtered table
3. in the ztf case the table will be saved as table_ztf-Input-Data.tbl, for wise upload table name is table_wise-Input-Data.
4. if an  unspecified mission data is uploaded, there will be no mission name in the table name: table_Input-Data. I am open to suggestions.

### DIFF
--- a/src/firefly/js/templates/lightcurve/LcManager.js
+++ b/src/firefly/js/templates/lightcurve/LcManager.js
@@ -281,7 +281,7 @@ export function* lcManager(params={}) {
 function updateRawTableChart(timeCName, fluxCName, converterId) {
 
     if (timeCName && fluxCName) {
-
+        const missionName = getConverter(converterId).missionName;
         const title =getConverter(converterId).showPlotTitle?getConverter(converterId).showPlotTitle(LC.RAW_TABLE):'';
 
         const chartX = get(getChartData(LC.RAW_TABLE), ['tablesources', 0, 'mappings', 'x']);
@@ -307,7 +307,7 @@ function updateRawTableChart(timeCName, fluxCName, converterId) {
                 mode: 'markers'
             }],
             layout: {
-                title,
+                title: `${missionName}` + ' ' + `${title}`,
                 yaxis: {autorange: 'reversed', showgrid: true, title: {text: fluxCName}}
             }
         };
--- a/src/firefly/js/templates/lightcurve/ptf/PTFMissionOptions.js
+++ b/src/firefly/js/templates/lightcurve/ptf/PTFMissionOptions.js
@@ -130,7 +130,7 @@ export function ptfRawTableRequest(converter, source, uploadFileName='') {
         uploadFileName
 
     };
-    return makeFileRequest(`${mission}`+'-Input Data', source, null, options);
+    return makeFileRequest(`${mission}`+' Input Data', source, null, options);
 
 }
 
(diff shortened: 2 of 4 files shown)


src/firefly/js/templates/lightcurve/LcManager.js
```diff
@@ -281,7 +281,7 @@ export function* lcManager(params={}) {
 function updateRawTableChart(timeCName, fluxCName, converterId) {
 
     if (timeCName && fluxCName) {
-
+        const missionName = getConverter(converterId).missionName;
```
kpuriIpac, line 284: suggestion: should we capitalize mission name so that `ztf` becomes `ZTF`, wise becomes `WISE`, and so on?

kpuriIpac (approved):
Changes look good, I made a comment about mission names possibly being capitalized. But even after that, a potential filename would look like:

`table_WISE-Input-Data.tbl`

Per the guide on confluence, I think the part `WISE-Input-Data` is fine as `WISE` is the mission name and `Input Data` is the table name, and since they're space separated, we use hyphens in the file name as mentioned in the guide. 

But I think firefly's `makeFileRequest` adds the `table_` prefix which not only looks awkward as is but also uses underscore (which when combined with hyphens in the file name is odd). If you want, I can take a look at that if it falls outside scope of this ticket. 

For unspecified missions, I think it's fine to have `table` in there (but possibly still better to just have the actual table name, in this case `Input Data`), but that again is probably a fix in `makeFileRequest` in firefly.

lrebull:
still broken for me -- see video -- it happens when i launch the timeseries tool from ztf search results.

https://github.com/user-attachments/assets/1c21f94c-049d-446d-94ba-0fd9c8982288

wmiipac:
@lrebull Now I understand you are send the ts data from Gator to ts tool not with the ts upload panel. I will look into the api call sent to ts from backend ( the g2p cgi call).

wmiipac:
> Per the guide on confluence, I think the part `WISE-Input-Data` is fine as `WISE` is the mission name and `Input Data` is the table name, and since they're space separated, we use hyphens in the file name as mentioned in the guide.

@kpuriIpac I thought about using upper case for mission name, but in the case of sending ts api from catalog search we have to use convertId which is defined in lower case. For the consistency, we just have them in lower case, if there is no strong objection

kpuriIpac:
> @kpuriIpac I thought about using upper case for mission name, but in the case of sending ts api from catalog search we have to use convertId which is defined in lower case. For the consistency, we just have them in lower case, if there is no strong objection

Yes, I'm good with that!

wmiipac:
@kpuriIpac currently save table and chart files are prefixed with `table_` and `chart_` ,  could change the underscore to dash

kpuriIpac:
> @kpuriIpac currently save table and chart files are prefixed with `table_` and `chart_` , could change the underscore to dash

I think since we change any spaces in file names to hyphens (like `Input Data` becomes `Input-Data`) when saving files, I think that change would make sense?

lrebull:
looks great to me.

PR #1803, opened 2025-06-26
 FIREFLY-1677: Better messaging for Position Cols in TAP Spatial Search

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1677
- See ticket for changes requested
- Also made the text a bit smaller
- The text for empty columns might be a bit too long. Should we consider a shorter version or does it look ok @gpdf?

**Testing**: https://fireflydev.ipac.caltech.edu/firefly-1677-better-tap-msg/firefly
- Test according to ticket
- if you select a table with pre-filled columns, you should see the default message (`These are the recommended columns...`)
- If you now manually select any columns, you should see the user selected message (`User-specified coordinate columns...`)
- if you remove a column selection or get a table with no pre filled columns (try `TAP_SCHEMA` under `IRSA`), you should get an empty columns message (`Unable to identify coordinate columns...`)

### DIFF
--- a/src/firefly/js/ui/UploadTableSelector.jsx
+++ b/src/firefly/js/ui/UploadTableSelector.jsx
@@ -1,4 +1,4 @@
-import React, {useContext, useEffect} from 'react';
+import React, {useContext, useEffect, useState} from 'react';
 import {Box, Chip, Stack, Tooltip, Typography} from '@mui/joy';
 import PropTypes from 'prop-types';
 import {defaultsDeep, omit} from 'lodash';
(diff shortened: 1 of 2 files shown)


wmiipac (approved):
The logic and messages are looking good.
Tested UI, it is working as expected.

gpdf:
We've just been talking about this, completely coincidentally, on the Rubin side this morning.

I'll review on the K8s build in a couple of minutes.

One comment: as a future ticket, we might want to add additional logic here.  

1. I see we are not pre-filtering the list of columns in the selectors by UCD, though that would make the UX a lot better. (The "longitude" column could filter on "pos.eq.ra OR pos.galactic.lon OR pos.ecliptic.lon" and the "latitude" column on "pos.eq.dec OR pos.galactic.lat OR pos.ecliptic.lat".  If you can't do ORs you can just do the equatorial values.)
2. We might want to add conditional logic based on whether the `indexed` attribute is set for the columns, too.  (For example: if there are ra/dec columns that _are_ labeled indexed, and you change to ones that are not, the UI could warn you that that's unlikely to be a good idea.)

gpdf (approved):
Looks great, absolutely usable as it is.

Please do make ticket(s) for followup work as suggested.

PR #1809, opened 2025-07-15
FIREFLY-1748: Cleanup and fix failing tests

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1748
**IFE PR**: https://github.com/IPAC-SW/irsa-ife/pull/426
- fix failing unzip test in download script, re-enabled the test
- Per an older PR suggestion (https://github.com/Caltech-IPAC/firefly/pull/1745), change usage of `BaseFileName` to `Title` (since we merged these fields in the UI), and use `Title` as the file name 
  - this change has been made for the IFE apps as well, refer to the IFE PR above for the same

**Testing**
- [Spherex](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/spherex), [Euclid](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/euclid), [Wise](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/wise), [SHA](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/Spitzer/SHA), [ZTF](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/ztf), [Sofia](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/sofia), [Finderchart](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/applications/finderchart), [DCE](https://firefly-1748-cleanup-fixes.irsakudev.ipac.caltech.edu/irsaviewer/dce)
- For all apps, just do any (only one) search, try and select a few resulting rows and download, ensure the downloaded file name matches the name in the `Title` field of the download dialog

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/DownloadRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/DownloadRequest.java
@@ -70,11 +70,11 @@ public class DownloadRequest extends ServerRequest implements Serializable {
 
     public boolean isSelectAll () { return _selectInfo.isSelectAll(); }
 
-    public String getBaseFileName() { return getParam(BASE_FILE_NAME); }
+    public String getBaseFileName() { return getParam(TITLE); } //use Title as the file name also
 
     public String getDataSource() { return getParam(DATA_SOURCE); }
 
-    public String getTitle() { return getParam(BASE_FILE_NAME); }
+    public String getTitle() { return getParam(TITLE); }
 
     public String getEmail() { return getParam(EMAIL); }
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/util/DownloadScript.java
@@ -176,6 +176,8 @@ static final String funcTmpl = """
       # move file to main directory and cleanup
       mv "$downloadedFile" "$destPath"
       echo ">> Saved as $destPath"
+      
+      [zip-logic-added-here]
 
       cd "$startDir"
       rm -rf "$tmpDir"
--- a/src/firefly/js/core/background/BackgroundCntlr.js
+++ b/src/firefly/js/core/background/BackgroundCntlr.js
@@ -198,11 +198,11 @@ function bgSetInfo(action) {
 function bgPackage(action) {
     return (dispatch) => {
         const {dlRequest={}, searchRequest, selectionInfo, bgKey='', downloadType} = action.payload;
-        let {fileLocation, wsSelect, BaseFileName} = dlRequest;
+        let {fileLocation, wsSelect, Title} = dlRequest; //use Title as the file name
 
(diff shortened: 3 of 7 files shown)


loitly (approved):
> - If the title is set to `abc.zip`, the downloaded file ends up as `abc.zip.zip`. This might be acceptable, but just noting it in case it's unintended.

Yes, this is unintended. I'll see if it's easy enough to fix this.

kpuriIpac:
> I thought the goal is to make the download `Title` matches the Job History `Title`. I didn't expect it would require changing the actual parameter name and touching so many files. That said, if this change is necessary, it is fine.
> 

Yes so the reason for this is that we used to have two separate fields on the UI, a `title` and a `file name` (corresponding to `Title` and `BaseFileName`). And on the backend we have 2 functions `getTitle` and `getBaseFileName` (in `DownloadRequest.java`), now both of these return the same thing. So I could just get rid of `getBaseFileName` and replace all its usage with `getTitle`, but the only reason I left it as is is because in our logic it makes sense that when you do a `getBaseFileName`, what you get back (title) is to be treated as a `file name`.

kpuriIpac:
> * How is the download script generated? I no longer see an option.  I only seem to receive it when I request email notification. Is that the expected behavior?

wait, could you clarify this part @loitly? Are you not able to download and then run the script? (if you do `curl` or `wget`)

loitly:
> > * How is the download script generated? I no longer see an option.  I only seem to receive it when I request email notification. Is that the expected behavior?
> 
> wait, could you clarify this part @loitly? Are you not able to download and then run the script? (if you do `curl` or `wget`)

In WISE, when I click Download, there's no option to get a script,  It just packages and downloads the zip file directly.
If I go to the Job Monitor and click Download, it also downloads the zip file directly.

The only way to get a script is by enabling email notification. The email includes a link to download the script, which can then be used to download the zip file.

This is the observed behavior. If all of this is intended, then it's fine.

kpuriIpac:
> In WISE, when I click Download, there's no option to get a script, It just packages and downloads the zip file directly. If I go to the Job Monitor and click Download, it also downloads the zip file directly.
> 
> The only way to get a script is by enabling email notification. The email includes a link to download the script, which can then be used to download the zip file.
> 
> This is the observed behavior. If all of this is intended, then it's fine.

Oh right, yes that's intended. Wise is one of the apps where we don't do download script by default (eventually we may, I think @robyww prefers that as well). But yes as of now it's the intended functionality.

kpuriIpac:
> > * If the title is set to `abc.zip`, the downloaded file ends up as `abc.zip.zip`. This might be acceptable, but just noting it in case it's unintended.
> 
> Yes, this is unintended. I'll see if it's easy enough to fix this.

@loitly I'm merging for now, as this is technically easy to do but I would like to discuss logic at Thursday's meeting. Because we could normally allow periods in filenames, so we can discuss what the edge cases could be.

PR #1810, opened 2025-07-16
FIREFLY-1796: Optimize Redis usage when querying jobs by user

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1796

### Optimizations Implemented

* **Improved `getUserJobs`**: now uses a compound key to filter Redis keys by user before fetching `JobInfo`, significantly reducing lookup overhead.
* **Added `job_cache` versioning**: helps identify when a migration is required.
* **Migration support**: added logic to convert existing keys to compound-key format.
* **Minimized `getAllJobs` usage**:

  * This expensive call is now only used in `admin/status`.
  * Removed from other parts of the code, especially pinging logic.
* **Optimized `listUserJob`**:

  * Calls `getUserJobs()` once, updates results as needed, and returns the final list.
  * Returns full `JobInfo`, avoiding additional server calls for job details.
  * Supports `LAST` parameter to limit the number of jobs returned (default: 500 via config).
  * Also implemented `PHASE` and `AFTER` filters for future use.
* **Improved `importJobHistory`**:

  * Now receives `getUserJobs()` result as input.
  * Only fetches full `JobInfo` from UWS services for **active jobs** or if the job is not found in Firefly's history.

---

Test URL: https://fireflydev.ipac.caltech.edu/firefly-1796-optimize-redis-bgjobs/firefly

### Testing Notes from local testing

* **Job Cache loaded with 101,104 jobs** (mocked).

* **Performance Benchmarks**:

```
Redis is configured to use both AOF and RDB persistence.
Therefore, it will use at least 2-3 times the memory in disk space. 
1m jobs requires 600M RAM
  - 11s to get all jobs
  - 1.9s to get user with 100k jobs
  - 1s to get user with 1k or fewer jobs
500k jobs requires 270M RAM
  - 4.7s to get all jobs
  - 1.1s to get user with 100k jobs
  - 450ms to get user with 1k or fewer jobs
100k jobs requires 64M
  - 1s to get all jobs
  - 100ms to get user with 1k or fewer jobs
```
---

### Manual Test

* Start Firefly.
* Inject the 1,000 job userKey into the browser cookie.
* Open **Job Monitor** → loaded quickly with **no additional network call**.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/api/Async.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/api/Async.java
@@ -18,8 +18,11 @@ import edu.caltech.ipac.firefly.core.background.JobManager;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.time.Instant;
+import java.util.Comparator;
 import java.util.List;
 
+import static edu.caltech.ipac.firefly.core.background.JobManager.JOB_LIST_DEFAULT_LIMIT;
 import static edu.caltech.ipac.firefly.data.ServerParams.JOB_ID;
 import static edu.caltech.ipac.util.StringUtils.*;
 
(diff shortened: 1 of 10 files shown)


robyww (approved):

kpuriIpac (approved):
Ran the tests mocking the different users, all looks good and works as expected. Code looks good as well.

PR #1812, opened 2025-07-22
Firefly-1801: line and point extractions supports multiple images

robyww:
note this PR was closed accidentally. Github seems to have trouble reopening closed PRs. So i create and new PR and merged. https://github.com/Caltech-IPAC/firefly/pull/1814

#### Firefly-1801](https://jira.ipac.caltech.edu/browse/FIREFLY-1801): line and point extractions supports multiple images
 
 - create one wide table
 - create multiple tables

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1801-extraction/firefly
- Load more than one images
- do a line or point extract then click on `Pin Chart/Table`.
- this will give you three options. Try with all three
- Mullti HDU images work as well. In these cases the table can get pretty wide.

### DIFF
--- a/src/firefly/js/visualize/ui/extraction/ExtractionDialog.jsx
+++ b/src/firefly/js/visualize/ui/extraction/ExtractionDialog.jsx
@@ -253,13 +253,6 @@ function PointExtractionPanel({canCreateExtractionTable, pv, pvCnt}) {
     const plane= getImageCubeIdx(plot)>-1 ? getImageCubeIdx(plot) : 0;
     const {x:chartX,y:chartY,chartXAxis:lastChartChartXAxis=chartXAxis}=plot?.attributes?.[PlotAttribute.SELECT_ACTIVE_CHART_PT] ?? {};
 
-
-    // const bottomUI = plotlyData ?
-    //         (<RadioGroupInputFieldView value={chartXAxis} buttonGroup={true}
-    //                                    options={ [ {label: 'Image X', value: 'imageX'}, {label: 'Image Y', value: 'imageY'} ]}
-    //                                    onChange={(ev) => setChartXAxis(ev.target.value)} />) :
-    //         undefined;
-
     useEffect(() => {
         const getData= async () => {
             if (imPtAry && imPtAry.length && plot) {
--- a/src/firefly/js/visualize/ui/extraction/ExtractionDialog.jsx
+++ b/src/firefly/js/visualize/ui/extraction/ExtractionDialog.jsx
@@ -4,11 +4,10 @@
 
 
 import {Box, Button, Divider, Stack, Tooltip, Typography} from '@mui/joy';
-import {isUndefined} from 'lodash';
+import {isUndefined,isArray} from 'lodash';
 import React, {useEffect, useState} from 'react';
 import sizeMe from 'react-sizeme';
 import {getAppOptions} from '../../../api/ApiUtil.js';
-import {isDefined} from '../../../util/WebUtil';
 import {makeImagePt} from '../../Point';
 import {allowPinnedCharts} from '../../../charts/ChartUtil';
 import {ensureDefaultChart} from '../../../charts/ui/ChartsContainer.jsx';
@@ -28,7 +27,6 @@ import {FieldGroup} from '../../../ui/FieldGroup';
 import HelpIcon from '../../../ui/HelpIcon.jsx';
 import {ListBoxInputField, ListBoxInputFieldView} from '../../../ui/ListBoxInputField.jsx';
 import {PopupPanel} from '../../../ui/PopupPanel.jsx';
-import {RadioGroupInputFieldView} from '../../../ui/RadioGroupInputFieldView.jsx';
 import {useFieldGroupValue, useStoreConnector} from '../../../ui/SimpleComponent.jsx';
 import {ValidationField} from '../../../ui/ValidationField';
 import {intValidator} from '../../../util/Validate';
@@ -48,8 +46,8 @@ import {
     isMultiHDUFits, primePlot
 } from '../../PlotViewUtil.js';
 import {computeDistance, computeScreenDistance, getLinePointAry} from '../../VisUtil.js';
-import {genPointChartData, genSliceChartData, genZAxisChartData} from './ExtractionChart';
-import {keepLineExtraction, keepPointsExtraction, keepZAxisExtraction} from './ExtractionTable';
+import {genPointChartData, genSliceChartData, genZAxisChartData} from './ExtractionChart.jsx';
+import {keepDataExtraction, keepZAxisExtraction} from './ExtractionTable.jsx';
 
 
 const DIALOG_ID= 'extractionDialog';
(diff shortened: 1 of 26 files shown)


lrebull:
OOOhhh soo cool! Note that this also solves one of the things that bugged Steve -- the continuous layering of the same line segment on top of all the images, once per extracted line. Now we have just one line segment layer across all images. 

Suggested wording changes. (note commas too; they're important!)

> Extract Image Type
> You can either extract from the currently selected single image, or all loaded images that have a WCS and a limited amount of HDUs (with WCS information). 
> Do you want to extract from all images?
> [Yes / into one table]  [Yes / into multiple tables] [No]

Now, column headers. I know what you're doing, and why, but a new person would be confused by a column header that includes the string "(ZTF R...rence..age g)" or "(2MASS...deg mo...ics J)"  This doesn't seem to be magically expanded anywhere, like the 'Details' tab still has it printed exactly like I have here, so I bet the saved tbl file will have it like this too.
We need to find a better solution to truncation of long image titles. I'm not saying I know what it is, but I am saying that this implementation isn't it.

robyww:
> Now, column headers. I know what you're doing, and why, but a new person would be confused by a column header that includes the string "(ZTF R...rence..age g)" or "(2MASS...deg mo...ics J)" This doesn't seem to be magically expanded anywhere, like the 'Details' tab still has it printed exactly like I have here, so I bet the saved tbl file will have it like this too.
We need to find a better solution to truncation of long image titles. I'm not saying I know what it is, but I am saying that this implementation isn't it.

This is a really hard problem to get uniqueness and keep it short. I worked pretty hard on it.  It will be hard to find a better solution. However, I can ask the user if he would prefer the short name or the full name. That will solve the problem for people who really care but it will make for wide table columns.

lrebull:
I like giving users the choice!

src/firefly/js/visualize/ui/extraction/ExtractionDialog.jsx
```diff
@@ -256,11 +254,11 @@ function PointExtractionPanel({canCreateExtractionTable, pv, pvCnt}) {
...
-    const bottomUI = plotlyData ?
-            (<RadioGroupInputFieldView value={chartXAxis} buttonGroup={true}
-                                       options={ [ {label: 'Image X', value: 'imageX'}, {label: 'Image Y', value: 'imageY'} ]}
-                                       onChange={(ev) => setChartXAxis(ev.target.value)} />) :
-            undefined;
+    // const bottomUI = plotlyData ?
+    //         (<RadioGroupInputFieldView value={chartXAxis} buttonGroup={true}
+    //                                    options={ [ {label: 'Image X', value: 'imageX'}, {label: 'Image Y', value: 'imageY'} ]}
+    //                                    onChange={(ev) => setChartXAxis(ev.target.value)} />) :
+    //         undefined;
```
kpuriIpac: did you mean to leave this in for now? if yes, then we could add a `Todo`

src/firefly/js/visualize/ui/extraction/ExtractionTable.jsx
```diff
@@ -84,46 +91,186 @@ export function keepZAxisExtraction(pt, pv, plot, filename, refHDUNum, extractio
...
+    const extractionSizeY= axis==='y' ? 1: pointSize;
+    const params= {baseImPtAry, pv, extractionSizeX, extractionSizeY, combineOp, save, doOverlay, isLine};
+    if (!canDoMultiImage) {
+        return Promise.resolve([makeDataExtractionTable(params)]);
+    }
+
+    if (filteredPvAry.length < pvAry.length) {
+        let msg= 'Cannot extract from all images';
+        if (pvAry.some( (testPv) => !hasWCSProjection(testPv))) {
+            msg+= ' Some image do not have a WCS Projection.';
```
kpuriIpac: minor: `'Some images do not have a WCS Projection.'` 
`images` should be plural is the suggested change

src/firefly/js/visualize/ui/extraction/ExtractionTable.jsx
```diff
@@ -84,46 +91,186 @@ export function keepZAxisExtraction(pt, pv, plot, filename, refHDUNum, extractio
...
+        }
+        if (pvAry.some( (testPv) => getMatchingHDUCount(testPv)>MAX_HDU)) {
+            msg+= ` Some images have more than ${MAX_HDU} HDUs.`;
+        }
+        msg+= ' These image can be extracted in single image mode';
```
kpuriIpac: same here, plural `images`

src/firefly/js/visualize/ui/extraction/ExtractionTable.jsx
```diff
@@ -84,46 +91,186 @@ export function keepZAxisExtraction(pt, pv, plot, filename, refHDUNum, extractio
...
+                You can either extract from the currently selected single image, or all loaded images that have a WCS
+                and a limited amount of HDUs (with WCS information).
+            </Typography>
+            <Typography color='warning'>
+                Do you want to extract from all images?
+            </Typography>
+        </Stack>
+    );
+    const answers= {allSame: 'Yes / into one table', allDiff: 'Yes / into multiple tables', single: 'No'};
+
```
kpuriIpac: I think this should be: 

`Yes - into one table`, `Yes - into multiple tables` **_OR_**
`Yes (into one table)`, `Yes (into multiple tables)` 

as `/` may suggest a choice between the two, but I'll defer this to @lrebull

kpuriIpac (approved):
Looks great! Tested per instructions, code looks good as well. Just some very minor comments.

src/firefly/js/visualize/ui/extraction/ExtractionTable.jsx
robyww: I mostly just used it like that to make the `yes` stand out more. You have a good point. maybe @lrebull need to weigh in again on this.

gpdf:
I don't have the ability to access the VPN with the way that I'm getting to the Internet on this trip.  So, just to make sure I'm understanding:

> this will give you three options. Try with all three

and 

> Extract Image Type
> You can either extract from the currently selected single > image, or all loaded images that have a WCS and a limited > amount of HDUs (with WCS information).
> Do you want to extract from all images?
> [Yes / into one table] [Yes / into multiple tables] [No]

This choice is presented as a new interstitial pop-up when the user requests "pin table"?  Would you mind uploading a screenshot of the pop-up (though I think I have it visualized)?

You mentioned creating very wide tables when there are multiple HDUs involved and multiple images.  I think it's very likely the case that in that situation the cases most desired may be "all HDUs for one image" (1) and "the current HDU for every image" (2) with "all (matching) HDUs for every image" (3) being useful but less often needed.

Use case 1: look at both flux and flags for one image to diagnose saturation at the central point of a bright star

Use case 2: look at flux in many bands to look for color effects

Use case 3: look at flux *and variance* in the above case to check for significance of effects seen

So I think if you could offer that option as well it would be less overwhelming in a fair number of situations relevant to SPHEREx and Rubin.

Also: since I can't get to the VPN a screen shot of the new option for choosing names would be helpful in evaluating.

Otherwise I won't see this until Saturday at the earliest.

lrebull:
oh, i see. that is better! 
one thing: third radio button should be "All images (into multiple table**s**)"
(you lost the plural tables)

robyww:
opps reopened

robyww:
dddddddddddddddd

robyww:
ddd

robyww:
reopent

PR #1813, opened 2025-07-24
FIREFLY-1798: Fix downloads for routed apps and improve obs core tbl recognition 

kpuriIpac:
**Ticket**: [FIREFLY-1798](https://jira.ipac.caltech.edu/browse/FIREFLY-1798) 
- **[IFE PR](https://github.com/IPAC-SW/irsa-ife/pull/428)**
- The `Prepare Download` button wouldn't show up in our routed apps in some places (like TAP search, Catalogs, etc.) 
- So the first step was for routed apps to trigger the `ObsCore` watcher (which takes care of generating the download buttons) 
- Some improvements to our ObsCore Watcher logic 
   - Now recognize ObsCore like tables (tbls containing a datalink service descriptor), so that the download script/packager works button shows up for these tables as well
  - Let apps pass in `obsCoreDownloadProps` (such as `script` or `packager`), and pass those on to `PrepareDownload` now to generalize logic (this helps get rid of  `PrepareDownload` calls in `Euclid` and `Spherex`)

**Testing**: 
- [Euclid](https://firefly-1798-euc-cats-download-bug.irsakudev.ipac.caltech.edu/applications/euclid): 
  - Do an `Image` and `Inspect Objects` search, the results should have the `Generate Download Script` button. Try downloading some data. 
  - Do a Euclid Catalogs search, if it has cutouts, you should see a `Prepare Download` button
  - Do a TAP search (try CADC -> ivoa obscore table -> m81), you should see a `Prepare Download` button
- [Spherex](https://firefly-1798-euc-cats-download-bug.irsakudev.ipac.caltech.edu/applications/spherex)
  - Repeat the TAP search here as well
  - Do an LVF search to ensure you see the download script button and it works like before 
- [Firefly](https://fireflydev.ipac.caltech.edu/firefly-1798-euc-cats-download-bug/firefly) and [DCE](https://firefly-1798-euc-cats-download-bug.irsakudev.ipac.caltech.edu/irsaviewer/dce): Do a CADC search on firefly, and do any search on DCE. Ensure you see the `Prepare Download` button like before (just regression testing)

### DIFF
--- a/src/firefly/js/ui/tap/TapSearchSubmit.js
+++ b/src/firefly/js/ui/tap/TapSearchSubmit.js
@@ -82,7 +82,7 @@ export function onTapSearchSubmit({request, serviceUrl, tapBrowserState, additio
             ...treq.META_INFO,
             ...additionalTapMeta,
             ...metaInfo,
-            [MetaConst.DATA_SERVICE_ID] : getServiceId(serviceUrl),
+            [MetaConst.DATA_SERVICE_ID] : metaInfo?.[MetaConst.DATA_SERVICE_ID] ?? getServiceId(serviceUrl)
         };
         dispatchTableSearch(treq, {backgroundable: true, showFilters: true, showInfoButton: true, ...tblOptions});
     };
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ObsCorePackager.java
@@ -367,7 +367,7 @@ public class ObsCorePackager extends FileGroupsProcessor {
                             break;
                         }
                     }
-                    String resolvedValue = (String) dgDataUrl.get(idx, key);
+                    String resolvedValue = Objects.toString(dgDataUrl.get(idx, key), null);
                     if (resolvedValue != null && !resolvedValue.isEmpty()) {
                         if (serDescUrl.partialUrl.endsWith("?"))
                             finalUrl.append(missing.paramName).append("=").append(resolvedValue);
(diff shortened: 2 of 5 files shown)


src/firefly/js/ui/tap/TapSearchSubmit.js
```diff
@@ -82,7 +82,7 @@ export function onTapSearchSubmit({request, serviceUrl, tapBrowserState, additio
             ...treq.META_INFO,
             ...additionalTapMeta,
             ...metaInfo,
-            [MetaConst.DATA_SERVICE_ID] : getServiceId(serviceUrl),
+            [MetaConst.DATA_SERVICE_ID] : metaInfo?.[MetaConst.DATA_SERVICE_ID] ?? getServiceId(serviceUrl)
```
kpuriIpac, line 85: @robyww @jaladh-singhal: 

To understand the reason for this, take a look at my changes in [spherex.js](https://github.com/IPAC-SW/irsa-ife/pull/428/files#diff-65c2264b316640641ed515807f667c10077afeb6636f5f5414ea9ad665fa52b5) and [lvf.jsx](https://github.com/IPAC-SW/irsa-ife/pull/428/files#diff-be81298e831121013e75fd93c0f893287f3d3c5a0d2e92fbb08f280a73887769) in the IFE PR. 

You'll notice I define `obsCoreDownloadProps` within `spherex` (because doing it just under `dataServiceOptions` without specifying a `data service ID` would apply these changes to all of the Spherex app - so even TAP search in spherex would get the download script button, as opposed to the `Prepare Download `button, which users might expect there). 

And then in `LVF.jsx`, I can set the `[MetaConst.DATA_SERVICE_ID]: 'spherex'`, this works, but it gets overwritten in `TapSearchSubmit`, which is the reason for this change.

robyww: This makes sense. We have been generalizing `TapSearchSubmit.js` for a while.

robyww (approved):
Looks good!

PR #1815, opened 2025-07-28
IRSA-6721: SPHEREx Spectrophotometry: Integrate UI with UWS service

jaladh-singhal:
Fixes [IRSA-6721](https://jira.ipac.caltech.edu/browse/IRSA-6721)

See ife PR: https://github.com/IPAC-SW/irsa-ife/pull/429

- Error handling improvements:
  - Fixed error retrival when job completes with ERROR and error is not in the job info but at an /error endpoint  
  - Allow error parsing even when job POST failed by updating the logic in `parseError` in DaliUtil.java
  - Updated the `parseError` function client-side to avoid splitting messages containing URLs or JSON
- Other minor improvements needed for ife PR

## Testing
https://fireflydev.ipac.caltech.edu/irsa-6721-integrate-phot-uws/firefly/

Needs Regression testing to make sure I didn't break the Async TAP which also subclasses UWSJobProcessor:
- I tried changing an ADQL query to a bad table name and the error comes back correctly.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
@@ -130,7 +130,7 @@ public class DaliUtil {
      * @return a string containing the parsed error message
      */
     public static String parseError(HttpMethod method, String url) {
-        boolean isErrorDoc = HttpServices.isOk(method); // url is an error document
+        boolean isErrorDoc = HttpServices.isOk(method); // url is an error document (at /error) if isOK() is true
         String httpErrMsg = isErrorDoc ? "" : String.format("Request to %s failed: %s", url, method.getStatusText());
 
         String errMsg;
@@ -145,8 +145,7 @@ public class DaliUtil {
                 }
             } else {
                 // error is VOTable doc
-                InputStream is = HttpServices.getResponseBodyAsStream(method);
-                try {
+                try (InputStream is = HttpServices.getResponseBodyAsStream(method)) {
                     String voError = VoTableReader.getError(is, url);
                     if (voError == null) {
                         voError = isErrorDoc
@@ -154,8 +153,6 @@ public class DaliUtil {
                                 : httpErrMsg; // give generic HTTP error message over the VOTable error
                     }
                     errMsg = voError;
-                } finally {
-                    FileUtil.silentClose(is);
                 }
             }
         } catch (Exception e) {
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
@@ -121,34 +121,47 @@ public class DaliUtil {
         return new DataAccessException(msg, cause);
     }
 
-    public static String parseError(HttpMethod method, String errorUrl) {
+    /**
+     * Parse the error message from the given HttpMethod. Also takes into account whether the URL is an
+     * error document or not.
+     *
+     * @param method the HttpMethod that contains the error response
+     * @param url    the URL of the request
+     * @return a string containing the parsed error message
+     */
+    public static String parseError(HttpMethod method, String url) {
+        boolean isErrorDoc = HttpServices.isOk(method); // url is an error document
+        String httpErrMsg = isErrorDoc ? "" : String.format("Request to %s failed: %s", url, method.getStatusText());
 
         String errMsg;
-        if (HttpServices.isOk(method)) {
-            String contentType = HttpServices.getResHeader(method, "Content-Type", "");
-            boolean isText = contentType.startsWith("text/plain");
-            try {
-                if (isText) {
-                    // error is text doc
-                    errMsg = HttpServices.getResponseBodyAsString(method);
-                } else {
-                    // error is VOTable doc
-                    InputStream is = HttpServices.getResponseBodyAsStream(method);
-                    try {
-                        String voError = VoTableReader.getError(is, errorUrl);
-                        if (voError == null) {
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
```diff
@@ -121,34 +121,47 @@ public static DataAccessException createDax(String title, String url, Exception
...
+    /**
+     * Parse the error message from the given HttpMethod. Also takes into account whether the URL is an
+     * error document or not.
+     *
+     * @param method the HttpMethod that contains the error response
+     * @param url    the URL of the request
+     * @return a string containing the parsed error message
+     */
+    public static String parseError(HttpMethod method, String url) {
+        boolean isErrorDoc = HttpServices.isOk(method); // url is an error document
```
loitly: You mean url to /error endpoint?

src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
```diff
@@ -121,34 +121,47 @@ public static DataAccessException createDax(String title, String url, Exception
...
+                    String voError = VoTableReader.getError(is, url);
+                    if (voError == null) {
+                        voError = isErrorDoc
+                                ? "Non-compliant VOTable in error document at " + url
+                                : httpErrMsg; // give generic HTTP error message over the VOTable error
                     }
+                    errMsg = voError;
+                } finally {
+                    FileUtil.silentClose(is);
                 }
```
loitly: You can simplify this by using try-with-resources.

```
try (InputStream is = HttpServices.getResponseBodyAsStream(method)) {
  // you logic here.  no silenClose is needed.
}
```

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -138,8 +138,11 @@ public DataGroup fetchDataGroup(TableServerRequest req) throws DataAccessExcepti
                     updateJob(ji -> ji.setPhase(Phase.SUSPENDED));
                     throw new DataAccessException("Job temporarily paused by the system");
                 } else if (phase == Phase.ERROR) {
-                    updateJob(ji -> ji.setError(uwsJob.getError()));
-                    throw new DataAccessException("Job has failed; detailed error information may be available");
+                    JobInfo.Error error = uwsJob.getError() != null
+                            ? uwsJob.getError() // error is a part of the job resource
+                            : getError(jobUrl); // error resource maybe present at /error endpoint of the job
```
loitly: Good!  Use job's errorSummary first.

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -307,7 +308,7 @@ public static Phase getPhase(String jobUrl) throws DataAccessException {
         }
     }
 
-    public static String getError(String jobUrl)  {
+    public static JobInfo.Error getError(String jobUrl)  {
         String errorUrl = jobUrl + "/error";
         HttpServices.Status status = HttpServices.getWithAuth(errorUrl, method -> {
```
loitly: We should use the same logic as above.  Return the errorSummary if exists; otherwise return the message from `/error`.  If you want to remove the function because it's not used, that's fine by me as well.

src/firefly/js/tables/TableUtil.js
```diff
@@ -1569,11 +1569,21 @@ export function splitVals(values='') {
...
+        const [_, type, cause] = error.cause.match(colonRegex) || [];
         return {message, type, cause};
     } else {
-        const [_, error, cause] = message?.match(/(.+?):(.+)/) || [];     // formatted error messages; 'error:cause'
+        // Check if message contains URLs or JSON - if so, don't split
+        const hasUrl = /https?:\/\//.test(message);
+        const hasJson = /[{[]/.test(message) && /[}\]]/.test(message);
+        if (hasUrl || hasJson) {
+            return {message: 'Table Error', cause: message};
+        }
```
loitly, line 1583: This is fine for now.  But, we may need to formalize how table errors are returned from the server. Relying on checks for 'https' or JSON-style braces may be unreliable, as those values could appear in the actual error message content.

loitly (approved):
Code looks good.

src/firefly/js/tables/TableUtil.js
jaladh-singhal: Agreed. I could conjure up an ugly regex to escape the `:` in urls. But for JSON it was getting way too complicated. In an ideal scenario, server should return the `message`, `cause` as 2 entities or use something other than `:` if need to return a single string containing both.

src/firefly/java/edu/caltech/ipac/firefly/server/query/DaliUtil.java
jaladh-singhal: Yes. That's what UWS specification is calling "error document". whatever is present in errorSummary in job info is simple error message.

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
jaladh-singhal: Yes it has a single use in L143. Are you saying, I can just move that conditional operator check (if errorSummary else `/error` endpoint) here?

jaladh-singhal: I went ahead and moved it so that it's at single place now.

PR #1817, opened 2025-07-31
Firefly-1806: add MOC auto display mode

robyww:
#### [Firefly-1806](https://jira.ipac.caltech.edu/browse/FIREFLY-1806): add MOC auto display mode
  - Auto will show MOC filled with a transparency based on FOV
  - Auto will show MOC as outline when the user zooms in
  - PR also include some standard http status message updates.

#### Testing
- https://firefly-1806-moc-auto-mode.irsakudev.ipac.caltech.edu/irsaviewer/dce
- https://firefly-1806-moc-auto-mode.irsakudev.ipac.caltech.edu/applications/euclid
- https://firefly-1806-moc-auto-mode.irsakudev.ipac.caltech.edu/applications/spherex
- in any of the apps look at how the MOC is displayed and how the transparency changes as you zoom in

### DIFF
--- a/src/firefly/js/drawingLayers/HiPSMOC.js
+++ b/src/firefly/js/drawingLayers/HiPSMOC.js
@@ -6,7 +6,7 @@ import {get, set, isEmpty, cloneDeep, isString} from 'lodash';
 import {makeDrawingDef, TextLocation, Style} from '../visualize/draw/DrawingDef.js';
 import DrawLayer, {DataTypes, ColorChangeType}  from '../visualize/draw/DrawLayer.js';
 import {makeFactoryDef} from '../visualize/draw/DrawLayerFactory.js';
-import {primePlot, getDrawLayerById, getPlotViewIdListInOverlayGroup} from '../visualize/PlotViewUtil.js';
+import {primePlot, getDrawLayerById, getPlotViewIdListInOverlayGroup, getFoV} from '../visualize/PlotViewUtil.js';
 import DrawLayerCntlr, {DRAWING_LAYER_KEY, dispatchUpdateDrawLayer, dlRoot} from '../visualize/DrawLayerCntlr.js';
 import MocObj, {createDrawObjsInMoc, setMocDisplayOrder, MocGroup} from '../visualize/draw/MocObj.js';
 import {getUIComponent} from './HiPSMOCUI.jsx';
@@ -22,12 +22,13 @@ import {cloneRequest} from '../tables/TableRequestUtil';
 import {dispatchAddActionWatcher} from '../core/MasterSaga';
 import {MetaConst} from '../data/MetaConst';
 import {getNextColor} from '../visualize/draw/DrawingDef';
-import {rateOpacity} from '../util/Color.js';
+import {rateOpacity, toRGBA, toRGBAString} from '../util/Color.js';
 import {CoordinateSys} from '../visualize/CoordSys.js';
 
 const ID= 'MOC_PLOT';
 const TYPE_ID= 'MOC_PLOT_TYPE';
 const MocPrefix = 'MOC - ';
+const MIN_AUTO_FILL_DEPTH= 6;
 const factoryDef= makeFactoryDef(TYPE_ID, creator, null, getLayerChanges, onDetach, getUIComponent, null, asyncComputeDrawData);
 export default {factoryDef, TYPE_ID};
 
@@ -53,6 +54,7 @@ function getVisiblePlotIdsByDrawlayerId(id, getState) {
 
 function loadMocFitsWatcher(action, cancelSelf, params, dispatch, getState) {
     const {id, mocFitsInfo}= params;
+    let autoUsesOnlyOutline= false;
     if (action.payload.drawLayerId === id && (action.payload.visible)) {
         const {fitsPath,tbl_id,tablePreloaded} = mocFitsInfo || {};
 
@@ -71,6 +73,9 @@ function loadMocFitsWatcher(action, cancelSelf, params, dispatch, getState) {
 
                       // in this case we may have 1 test row loaded. test to see if it is greater than the filter
                 if (preloadedTbl.tableData.data[0][0] > maxDepth)  filterObj= {}; //abort filtering
+
+                autoUsesOnlyOutline= (preloadedTbl.tableData.data[0][0]> 4*(4**(MIN_AUTO_FILL_DEPTH)));
+
 
                 tReq= cloneRequest(preloadedTbl.request,
                     { startIdx : 0, pageSize : MAX_ROW, inclCols: mocFitsInfo.uniqColName, ...filterObj});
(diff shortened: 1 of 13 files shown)


lrebull:
ooh! cool!
ok, my $0.02:

- the first thing i tried was the 2MASS LGA because it's the first on the list, and the little footprints are so faint that i can't distinguish them from the generalized hips background (the background texture in the wise hips map is comparable to the color/size of some of the footprints; see screenshot for a different example). at least for this one, they need to start out a little bit more opaque, otherwise i can't even see that they are there at all. (i was waiting and waiting and waiting for the MOC to load, wondering why it was taking so darn long, when i finally realized that no, actually, it had loaded some unknown time ago.) attempting to generalize this, we could try to come up with some logic like 'if the footprints are smaller than xx square arcmin [or xx arcmin on a side?], make them opaque'..?


Screenshot 2025-07-31 at 10 00 51 AM

argh ok getting pulled off to TWO different other things. let me save this now and i'll come back to it as soon as i can.

lrebull:
- i can't see them at all whatsoever on the BRAVA search page, or HerMES, or HOP. H-ATLAS, however, is _perfect_. HERITAGE and LocalGroup are pretty great too. MAGCLOUDSCPLUS is back to not being able to tell what is footprint and what is noise in the background image. I haven't systematically tested each and every data set, but I think you get the idea here...
- SPHEREx works great!
- Euclid does too!

src/firefly/js/drawingLayers/HiPSMOC.js
```diff
@@ -405,12 +448,31 @@ function updateMocData(dl, plotId) {
...
      }
 }
 
+function getAutoDrawStyle(plot, color) {
+    const [r,g,b]= toRGBA(color);
+    const fov= getFoV(plot);
+    if (!fov) return {style:Style.FILL, color:toRGBAString([r,g,b,1])};
+    if (fov<26) return {style:Style.DESTINATION_OUTLINE, color:toRGBAString([r,g,b,1])};
+
+    let alpha= .7;
```
jaladh-singhal: if fov >= 26 and fov <=30, alpha becomes .7 which breaks the transparency increasing when zoomming in behavior. Perhaps this should be .18.

Or L461 can be `if (fov<30)` to prevent this 26-30 gap

src/firefly/js/drawingLayers/HiPSMOC.js
```diff
@@ -405,12 +448,31 @@ function updateMocData(dl, plotId) {
...
+    else if (fov>150) alpha= .68;
+    else if (fov>130) alpha= .65;
+    else if (fov>110) alpha= .58;
+    else if (fov>90) alpha= .55;
+    else if (fov>70) alpha= .45;
+    else if (fov>50) alpha= .35;
+    else if (fov>40) alpha= .25;
+    else if (fov>30) alpha= .20;
+    return {style:Style.FILL, color: toRGBAString([r,g,b,alpha])};
+}
```
jaladh-singhal: Actually...I was wondering if there's some formula to map fov to our alpha range (0.2-0.7) and asked ChatGPT:

```js

  // Linearly interpolate alpha between these breakpoints:
  const MIN_FOV   = 26;
  const MAX_FOV   = 170;
  const MIN_ALPHA = 0.2;
  const MAX_ALPHA = 0.7;

  // Tiny FOV → outline
  if (fov < MIN_FOV) {
    return {
      style: Style.DESTINATION_OUTLINE,
      color: toRGBAString([...baseRGBA, 1])
    };
  }

  // Fill with interpolated alpha 
  const t = Math.min(Math.max((fov - MIN_FOV) / (MAX_FOV - MIN_FOV), 0), 1); // normalized fov in [0,1]
  const alpha = MIN_ALPHA + t * (MAX_ALPHA - MIN_ALPHA);

  return {
    style: Style.FILL,
    color: toRGBAString([...baseRGBA, alpha])
  };

```

And then it's easier to play with the alpha values. I think MIN_ALPHA = 0.1 will actually be better to show transparently filled to outline transition.

jaladh-singhal (changes requested):
Looks great! I agree with Luisa's comment about DCE stuff.

I noticed a bug: the fill color suddenly becomes opaque (instead of continuously becoming transparent) when you keep zooming in and are between 30 and 26 degrees FOV. I identified the reason for it in the logic (see my comment below).

src/firefly/js/drawingLayers/HiPSMOC.js
robyww: I considered do something like this but it got forgotten after the proof-of-concept started working. I will try it.

src/firefly/js/drawingLayers/HiPSMOC.js
robyww: ok I will fix it.

src/firefly/js/drawingLayers/HiPSMOC.js
```diff
@@ -405,12 +453,44 @@ function updateMocData(dl, plotId) {
...
              }
          }
      }
 }
 
+function getAutoDrawStyle(plot, color,autoUsesOnlyOutline=false) {
+    const [r,g,b]= toRGBA(color);
+    const fov= getFoV(plot);
+    if (!fov) return {style:Style.FILL, color:toRGBAString([r,g,b,1])};
+    if (autoUsesOnlyOutline || fov<30) return {style:Style.DESTINATION_OUTLINE, color:toRGBAString([r,g,b,1])};
```
jaladh-singhal, line 466: This number should come from the constant so that it doesn't break if someone changes that in getAlpha (you probably need to move the constants out of the functions):
```suggestion
    if (autoUsesOnlyOutline || fov<MIN_FOV) return {style:Style.DESTINATION_OUTLINE, color:toRGBAString([r,g,b,1])};
```

src/firefly/js/drawingLayers/HiPSMOC.js
```diff
@@ -454,25 +459,27 @@ function updateMocData(inDrawLayer, plotId) {
...
+    const alpha= getAlpha(fov);
     return {style:Style.FILL, color: toRGBAString([r,g,b,alpha])};
 }
 
+function getAlpha(fov) {
+    const MIN_FOV   = 30;
+    const MAX_FOV   = 170;
+    const MIN_ALPHA = 0.2;
+    const MAX_ALPHA = 0.7;
```
jaladh-singhal: something worth trying, I have a hunch this will give the cleanest transition from transparently filled to only outlined:

```suggestion
    const MIN_FOV   = 25;
    const MAX_FOV   = 170;
    const MIN_ALPHA = 0.01;
    const MAX_ALPHA = 0.7;
```

jaladh-singhal (approved):
Thanks for the update. looks good and the bug is gone.

lrebull:
wow, this is SUPER COOL. works GREAT! tried all three and it is nice!!

src/firefly/js/drawingLayers/HiPSMOC.js
jaladh-singhal: @robyww Before you merge: I noticed your last commit changed `minFov` in getAlpha to 25, but not in L466, they must be the same. That's why I suggested defining them as `MIN_FOV` constant and using it at these both places to prevent getting them out of sync.

```js
    minFov = 25;
    if (autoUsesOnlyOutline || fov<minFov) return {style:Style.DESTINATION_OUTLINE, color:toRGBAString([r,g,b,1])};
    const alpha= getAlpha(fov, {minFov});
```

robyww: They don't have to be the same. if FOV < 30 then the `getAlpha()` never gets called. By making minFov less then 30 then it approaches the "fill disappear effect" that you suggested. However, it will never actually gets there, it switches over first. We don't want to get to the "fill disappear" because there will be a FOV that seems buggy.

robyww: We could accomplish a similar effect by making minFov 30 and minAlpha 5. I might plan with it more at another time.

PR #1820, opened 2025-08-06
IRSA-7144: Fixed mishandling of UTINYINT JDBCType.

loitly:
https://jira.ipac.caltech.edu/browse/IRSA-7144

> The version of the DuckDB JDBC driver used by Firefly does not handle UTINYINT correctly. It maps the type to java.sql.JDBCType.UTINYINT, which is not a valid JDBC type and causes errors.

Test: https://fireflydev.ipac.caltech.edu/irsa-7144-parquet-utinyint/firefly/

Loading the two files from the ticket should work as expected. Everything else should continue to work normally.


kpuriIpac (approved):
Code looks good. Tested the changes with the 2 attached files and did a few other searches as well, everything works as expected.

PR #1821, opened 2025-08-09
FIREFLY-1811: ZipHandler fix to better handle GZIP 

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1811
- Added some more logic to directly check the bytes for indicating gzip (works even if we have URL instead of a file directly) as `FileUtil.isGZipFile()` didn't work even after creating a tmp file from the URL. 
  - I can try `FileUtil.isGZipFile()` if you have some suggestions. But I still think this might be a little better as it'll work for both URLs and files, and directly tests the bytes for gzip. 

**Testing**: 
**Spherex**: https://firefly-1811-packager-gzip-fix.irsakudev.ipac.caltech.edu/applications/spherex
- Do a Spectophotometry Search: `21h24m24.39s +33d39m07.6s Equ J2000` should only take 3-5 minutes. 
  - download the 1 row generated. Unzip the packaged download. Inside, you should get a good filename and extension (.xml) and the file should just open up (won't be zipped). 
- Test [Firefly](https://fireflydev.ipac.caltech.edu/firefly-1811-packager-gzip-fix/firefly) (CADC) and [Euclid](https://firefly-1811-packager-gzip-fix.irsakudev.ipac.caltech.edu/applications/euclid) (Catalogs with `Prepare Download`) to ensure the changes in ZipHandler didn't break any existing behavior

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/packagedata/ZipHandler.java
@@ -4,18 +4,13 @@
 package edu.caltech.ipac.firefly.server.packagedata;
 
 import edu.caltech.ipac.firefly.data.FileInfo;
-import edu.caltech.ipac.util.StringUtils;
-import edu.caltech.ipac.util.download.URLDownload;
 import edu.caltech.ipac.firefly.server.util.Logger;
 import edu.caltech.ipac.util.AppProperties;
 import edu.caltech.ipac.util.FileUtil;
+import edu.caltech.ipac.util.StringUtils;
+import edu.caltech.ipac.util.download.URLDownload;
 
-import java.io.BufferedInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.PrintWriter;
+import java.io.*;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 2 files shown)


robyww (changes requested):
I think the `isGZipFile()` function is correct and doing the same thing the you are in the code.  I think what you really want to do is to add a second method that takes in InputStream. Then make the first method take the second method

```java 
    public static boolean isGZipFile(File f) {
        FileInputStream in=null;
        try {
            in= new FileInputStream(f);
            return isGZipFile(in);
        } catch (IOException e) {
            return false;
        } finally {
            FileUtil.silentClose(in);
        }
    }
```
_then_
```java
public static boolean isGZipFile(InputStream is) {
      <more>
      <make sure you include the reset if you are doing it this way>
}
```

kpuriIpac:
> I think the `isGZipFile()` function is correct and doing the same thing the you are in the code. I think what you really want to do is to add a second method that takes in InputStream. Then make the first method take the second method
> 
> ```java
>     public static boolean isGZipFile(File f) {
>         FileInputStream in=null;
>         try {
>             in= new FileInputStream(f);
>             return isGZipFile(in);
>         } catch (IOException e) {
>             return false;
>         } finally {
>             FileUtil.silentClose(in);
>         }
>     }
> ```
> 
> _then_
> 
> ```java
> public static boolean isGZipFile(InputStream is) {
>       <more>
>       <make sure you include the reset if you are doing it this way>
> }
> ```

Done, can you take a look at the changes @robyww? Made the old `isGZipFile(File f)` call the new `isGZipFile(InputStream is)`, but from ZipHandler, I still directly call `isGZipFile(InputStream is)` (because this one will work directly for streams from local files or from URLs). If it looks good, I'll go ahead and merge. The test builds are updated as well.

robyww (approved):

lrebull:
This worked! much better than before!!!
i have tweaks for the "would be nice" category -- would you like them here or back on the confluence page, destined for a new ticket?

lrebull:
i'm putting them on the confluence page, just seemed easier.

PR #1825, opened 2025-08-19
Firefly-1815: upload a file via an action, usually firefly_client

robyww:
### [Firefly-1815](https://jira.ipac.caltech.edu/browse/FIREFLY-1815): Upload via a action, usually from firefly_client

- Also added URL API support

### Testing From Python firefly_client

- use https://fireflydev.ipac.caltech.edu/firefly-1815-upload/firefly/
- Test from python using `firefly_client`
#### Test loading to the upload panel
```python
import os
from firefly_client import FireflyClient
ffurl= 'https://fireflydev.ipac.caltech.edu/firefly-1815-upload/firefly/'
fc = FireflyClient.make_client(ffurl, channel_override='up')
file= 'path to a fits and/or table file'
key= fc.upload_file(file)
fc.dispatch('app_data.externalUpload', {'fileOnServer':key, 'displayName':os.path.basename(file)})
```
#### Test immediate loading, change last line to add `'immediate':True`
```python
fc.dispatch('app_data.externalUpload', {'immediate':True, 'fileOnServer':key, 'displayName':os.path.basename(file)})
```
#### Test a url
```python
fc.dispatch('app_data.externalUpload', {'url':'http://web.ipac.caltech.edu/staff/roby/data-products-test/1904-66_SFL.fits' })
```

### Testing URL API
- https://fireflydev.ipac.caltech.edu/firefly-1815-upload/firefly/?api=load
- Try the examples or make your own

### DIFF
--- a/src/firefly/js/ui/FileUploadProcessor.jsx
+++ b/src/firefly/js/ui/FileUploadProcessor.jsx
@@ -1,31 +1,41 @@
+import {Stack, Typography} from '@mui/joy';
+import {dispatchHideDialog} from 'firefly/core/ComponentCntlr';
 import {FileAnalysisType, Format, TableDataType} from 'firefly/data/FileAnalysis';
-import {showInfoPopup, showYesNoPopup} from 'firefly/ui/PopupUtil';
-import {getAppHiPSForMoc, isMOCFitsFromUploadAnalsysis} from 'firefly/visualize/HiPSMocUtil';
 import {MetaConst} from 'firefly/data/MetaConst';
-import {isLsstFootprintTable} from 'firefly/visualize/task/LSSTFootprintTask';
-import {makeFileRequest, makeTblRequest} from 'firefly/tables/TableRequestUtil';
-import {getFieldVal} from 'firefly/fieldGroup/FieldGroupUtils';
+import LSSTFootprint from 'firefly/drawingLayers/ImageLineBasedFootprint';
 import {createNewRegionLayerId} from 'firefly/drawingLayers/RegionPlot';
-import {dispatchCreateImageLineBasedFootprintLayer, dispatchCreateRegionLayer, getDlAry} from 'firefly/visualize/DrawLayerCntlr';
-import {getDrawLayersByType, getPlotViewAry, primePlot} from 'firefly/visualize/PlotViewUtil';
-import {dispatchPlotImage, visRoot} from 'firefly/visualize/ImagePlotCntlr';
+import {getFieldVal} from 'firefly/fieldGroup/FieldGroupUtils';
+import {makeFileRequest, makeTblRequest} from 'firefly/tables/TableRequestUtil';
 import {dispatchTableFetch, dispatchTableSearch} from 'firefly/tables/TablesCntlr';
 import {getSelectedDataSync, uniqueTblId} from 'firefly/tables/TableUtil';
-import LSSTFootprint from 'firefly/drawingLayers/ImageLineBasedFootprint';
-import WebPlotRequest from 'firefly/visualize/WebPlotRequest';
-import RangeValues from 'firefly/visualize/RangeValues';
+import {
+    acceptAnyTables, acceptDataLinkTables, acceptImages, acceptMocTables, acceptNonDataLinkTables, acceptNonMocTables,
+    acceptRegions, acceptTableOrSpectrum, acceptUWS, DATA_LINK_TABLES, IMAGES, MOC_TABLES, REGIONS, SPECTRUM_TABLES,
+    TABLES, UWS
+} from 'firefly/ui/FileUploadUtil';
+import {showInfoPopup, showYesNoPopup} from 'firefly/ui/PopupUtil';
+import {
+    dispatchCreateImageLineBasedFootprintLayer, dispatchCreateRegionLayer, getDlAry
+} from 'firefly/visualize/DrawLayerCntlr';
+import {getAppHiPSForMoc, isMOCFitsFromUploadAnalsysis} from 'firefly/visualize/HiPSMocUtil';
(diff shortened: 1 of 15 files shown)


src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -489,3 +507,92 @@ export function getSelectedRows(type, summaryTblId, report, currentSummaryModel,
...
+    };
+    return summaryModel;
+}
+
+
+async function loadFile(fileOnServer,displayName) {
+    const {status, analysisResult}= await upload(fileOnServer, 'details', {name:displayName});
+    if (Number(status)!==200) {
+        console.log('show error here');
+        //todo
```
jaladh-singhal: Can you show here the status and potentially some error from failed promise?

src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -489,3 +507,92 @@ export function getSelectedRows(type, summaryTblId, report, currentSummaryModel,
...
+    };
+    resultSuccess(request, fileOnServer);
+}
+
+
+export function initHandleExternalUpload() {
+    const handleExternalUpload= (action) => {
+        const {type,payload}= action ?? {};
+        if (type!==EXTERNAL_UPLOAD) return;
+        const {fileOnServer,displayName, immediate= false}= payload;
```
jaladh-singhal: If I omit `displayName` in payload "undefined" shows up in the UI. Maybe we can set it to some generic string or better option but more work: split the `fileOnServer` with `/` and extract the last part (firefly might have some util func for this already)?

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
```diff
@@ -556,7 +530,23 @@ function ImageDisplayOption({summaryTblId, currentReport, currentSummaryModel, a
...
+    const {fileOnServer, displayName}= fileOnServerOptions;
+
+    useEffect(() => {
+        const up= async () => {
+            if (uploadSrc!==ON_SERVIER_ID || !fileOnServer || getVal(ON_SERVIER_ID)===fileOnServer) return;
+            const {status, analysisResult}= await upload(fileOnServer, 'details', {name:displayName});
+            const value= fileOnServer;
+            const fld= Number(status)===200
+                ? {analysisResult, message:undefined, value}
+                : {analysisResult:undefined,message:'placeholder Upload fail message', value};
```
jaladh-singhal: This error message also needs to be changed. Can show status or/and rejected promise message.

jaladh-singhal (changes requested):
This is really great!!! ❤️  
I tested it out from python and it works well for both immediate and non-immediate case. Now I can easily write a generic function for the python client using this.

Code looks good - error handling needs some polishing (see my comments).

---
I noticed a bug with menu tabs switching which I remember used to exist but is now more noticeable with subsequent uploads:
- Do `fc.dispatch('app_data.externalUpload', {'fileOnServer':key, 'displayName':os.path.basename(file)})`, firefly client will open the "Upload" tab with preloaded file option. Now without changing anything in UI, do `fc.dispatch('app_data.externalUpload', {'immediate':True, 'fileOnServer':key, 'displayName':os.path.basename(file)})`, firefly client will take you to Results page but the tab remains stuck at "Upload". You can ofc fix it by clicking on "Results" tab but it's odd behavior since the UI now has a tab view (I know results is still a dropdown in our logic).
- The same bug can be reproduced for other firefly_client functions too: Open firefly client, go to any tab other than results (let's say "Images"), and then do `fc.show_fits(plot_id="wise-fullimage", URL='http://irsa.ipac.caltech.edu/ibe/data/wise/allsky/4band_p1bm_frm/6a/02206a/149/02206a149-w1-int-1b.fits')`, it will show "Results" page but tab will remain stuck at "Images" until you manually click the "Results"

Can firefly be made globally aware that if any externally dispatched action opens Results page, also change the tab to "Results"? (This can probably be separated in another ticket)

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
robyww: I think it would be better to have somethink like `Upload File`. The name on the server is meaningless to the user. I will work on that.

src/firefly/js/ui/FileUploadProcessor.jsx
robyww: I think it would be better to have somethink like `Upload File`. The name on the server is meaningless to the user. I will work on that.

src/firefly/js/visualize/ui/FileUploadViewPanel.jsx
robyww: Yes, I missed that. thanks

src/firefly/js/ui/FileUploadProcessor.jsx
robyww: I meant to go back to that. Thanks for catching it.

robyww:
> Do fc.dispatch('app_data.externalUpload', {'fileOnServer':key, 'displayName':os.path.basename(file)}), firefly client will open the "Upload" tab with preloaded file option. Now without changing anything in UI, do fc.dispatch('app_data.externalUpload', {'immediate':True, 'fileOnServer':key, 'displayName':os.path.basename(file)}), firefly client will take you to Results page but the tab remains stuck at "Upload". You can ofc fix it by clicking on "Results" tab but it's odd behavior since the UI now has a tab view (I know results is still a dropdown in our logic).

I will see it I can fix it in this ticket. Otherwise I will make a new one.
**Update**: I will fix it.

jaladh-singhal (approved):
Error messaging updates look good - the info popup is helpful. I also tested the issue I reported with tabs not switching correctly - it's fixed now.

Also the URL API with `load` command is a great addition!

PR #1827, opened 2025-08-20
Firefly-1819: better handle ui for cases with no radius

robyww:
#### Testing
- https://firefly-1819-hips-search-panel.irsakudev.ipac.caltech.edu/applications/euclid
- https://firefly-1819-hips-search-panel.irsakudev.ipac.caltech.edu/applications/spherex
- UI should be more correct now when the is not radius input.

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -165,6 +165,7 @@ export function EmbeddedPositionSearchPanel({
         sRegion= undefined,
         toolbarHelpId= undefined,
         showHelpLines=true,
+        selectionHelpText= undefined,
         coordinateSys : csysStr = 'EQ_J2000',
         sx:hipsTargetViewSx={},
     }= slotProps.hipsTargetView ?? {};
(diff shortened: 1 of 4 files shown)


lrebull:
the UI itself seems to look good, but then i attempted a region search on the image search in spherex and i have hung it, i think. it’s not frozen, entirely, but i cannot send it to background and it doesn’t seem to be coming back either.

lrebull:
same for a euclid search.
my spherex search was “larger than usual” but my euclid search was much smaller than usual (60")

lrebull:
ahhhhh it was a 'new private window' problem, exacerbated by a “observation boundary contains shape” when i needed “observation boundary intersects shape” problem. sorry about that. 

all works great now!

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -165,6 +165,7 @@ export function EmbeddedPositionSearchPanel({
         sRegion= undefined,
         toolbarHelpId= undefined,
         showHelpLines=true,
+        selectionHelpText= undefined,
         coordinateSys : csysStr = 'EQ_J2000',
         sx:hipsTargetViewSx={},
     }= slotProps.hipsTargetView ?? {};
```
jaladh-singhal, line 171: This is something for later, I also ran into it in my last ticket to add showHelpLines: EmbeddedPositionSearchPanel should directly expose the props of HiPSTargetView in slotProps instead of defining them repeatedly here. We can then do something like `<HiPSTargetView {...defaults, ...slotProps?.hipsTargetView, ...overrides}/>` (where `overrides` are the stuff that EmbeddedPositionPanel must need for it to work)

src/firefly/js/visualize/ui/TargetHipsPanelToolbar.jsx
```diff
@@ -22,29 +37,45 @@ export const HelpLines= ({whichOverlay}) => {
...
-                    <Typography level='body-xs' sx={nowrap}>Use the Selection Tools (</Typography>
-                    {selectButton}
-                    <Typography level='body-xs' sx={nowrap}>) to choose a search polygon. Click to change the center.
-                    </Typography>
-                </> )}
-        </Stack>
-    );
 
+    const simpleText= selectionHelpText
+        ? selectionHelpText
```
jaladh-singhal, line 42: Great! this allows custom text that scientists may want

jaladh-singhal (approved):
Code looks good!

Tested spectral image search, spectrophotmetry tool and Euclid image search, they all show correct help text and toolbar items, when in no radius mode.

PR #1828, opened 2025-08-22
FIREFLY-1807: Firefly fails to detect broken Redis connection; misleading server-status and client UI

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1807
Changes:
- use ping for a more direct test
- update client more aggressively
- prevent exception during JobManager init.

Test: https://fireflydev.ipac.caltech.edu/firefly-1807-detect-redis-outage/firefly/
The recommend approach to testing is to run Redis locally and start Tomcat with -Dredis.host=127.0.0.1. 
This allows you to manually stop and restart Redis to observe how Firefly reacts to connection loss and recovery.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/RedisService.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/RedisService.java
@@ -141,13 +141,14 @@ public class RedisService {
             return; // state unchanged
         }
         failSince = ok ? null : Instant.now();
-        sendConnectionStatus(true);     // notify clients of the update
+        sendConnectionStatus();     // notify clients of the update
     }
 
-    public static void sendConnectionStatus(boolean always) {
-        if (!always && failSince == null) {
-            return; // no need to send status if not lost
-        }
+    public static void sendConnectionStatusIfFailed () {
+        if (failSince != null) sendConnectionStatus();
+    }
+
+    public static void sendConnectionStatus() {
         boolean lost = failSince != null;
         String reason = lost ? "A critical system component is currently unavailable" : "";
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/RedisService.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/RedisService.java
@@ -82,7 +82,7 @@ public class RedisService {
             pconfig.setBlockWhenExhausted(true);                // wait; if needed
             pconfig.setMaxWait(Duration.of(5, ChronoUnit.SECONDS));
             JedisPool pool = new JedisPool(pconfig, redisHost, REDIS_PORT, Protocol.DEFAULT_TIMEOUT, REDIS_PASSWORD);
-            pool.getResource().close();     // test connection
+            pool.getResource().ping();     // test connection
             return pool;
         } catch(Exception ignored) {}
         return null;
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/cache/DistributedCache.java
```diff
@@ -113,11 +113,12 @@ public boolean isCached(CacheKey key) {
         return false;
     }
 
+    @Nonnull
     public List<StringKey> getKeys() {
         try(Jedis redis = RedisService.getConnection()) {
             return keys(redis).stream().map(StringKey::new).toList();
         } catch (Exception ex) { LOG.error(ex); }
-        return null;
+        return List.of();
```
robyww, line 121: Interesting.... never used this method before.

src/firefly/java/edu/caltech/ipac/firefly/server/ServCommand.java
```diff
@@ -22,6 +23,7 @@ public abstract class ServCommand extends ServerCommandAccess.HttpCommand {
     public void processRequest(HttpServletRequest req, HttpServletResponse res, SrvParam sp) throws Exception {
         JSONObject json=new JSONObject();
         String jsonData;
+        RedisService.sendConnectionStatus(false);      // notify the client if redis is down
```
robyww: call this functino `RedisService.sendConnectionStatus` with a `false` is confusing since most of the time in just return.

I think you need a `ensureConnectionStatus()`

src/firefly/java/edu/caltech/ipac/firefly/core/RedisService.java
```diff
@@ -107,40 +107,48 @@ static void startLocal() {
...
+            return; // state unchanged
+        }
+        failSince = ok ? null : Instant.now();
+        sendConnectionStatus(true);     // notify clients of the update
+    }
+
+    public static void sendConnectionStatus(boolean always) {
+        if (!always && failSince == null) {
+            return; // no need to send status if not lost
+        }
```
robyww: I think you need to break up this function and add the `!always && failSince == null` login to an `ensureConnectionStatus()` and take out the boolean in `sendConnectionStatus()`
somthing like
```java
public static void ensureConnectionStatus() {
    if (failSince == null) return;
   sendConnectionStatus();
}
```
calling `sendConnectionStatus()` in every `ServCommand.processRequest()` is confusing.

src/firefly/java/edu/caltech/ipac/firefly/core/background/JobManager.java
```diff
@@ -92,16 +94,25 @@ public class JobManager {
...
-            LOG.info("Migrating job cache keys to new format");
-            int count = migrateRedisKeys();
-            LOG.info("Migrated " + count + " job cache keys to new format");
-            CacheManager.getDistributed().put(JOB_CACHE_VERSION_KEY, JOB_CACHE_VERSION); // set the version
-        }
+                JobManager::checkJobs, KEEP_ALIVE_INTERVAL, KEEP_ALIVE_INTERVAL, TimeUnit.SECONDS);   // check every 30 seconds
+
+        ScheduledExecutorService migrator = Executors.newSingleThreadScheduledExecutor();
+        migrator.scheduleWithFixedDelay(() -> {
+            try (Jedis jedis = RedisService.getConnection()) {
```
robyww, line 101: This is probably temporary code that should be removed in a year or so. You might want to document a plan for that.

robyww (approved):
Code looks good. However, I am having trouble seeing where the actual bug fix is. Maybe you can point it out to me in the stand up.

I do think you should at `RedisService.ensureConnectionStatus()` for better readability.

kpuriIpac (approved):
Tested locally according to instructions, stopping/restarting redis and seeing the expected updates on the client.

PR #1830, opened 2025-08-25
FIREFLY-1795: fix alert banner for both light and dark mode

kpuriIpac:
**Ticket**: [FIREFLY-1795](https://jira.ipac.caltech.edu/browse/FIREFLY-1795) 
- @robyww  I checked with @loitly  - IRSA doesn't have a reason for the current hardcoded background color for the alerts banner, so removing the old background color allows JoyUI to use the default warning color which works for both dark & light mode as expected. 
  - I experimented with `soft` & `solid` variants, I think `soft` looks better (and also similar to what we used to have before)

**Testing**: 
- **Euclid**: https://firefly-1795-alert-banner.irsakudev.ipac.caltech.edu/applications/euclid
- **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1795-alert-banner/firefly
- Just see the alert banners on both light and dark modes.  
  - I'll remove the hardcoded `msg` string before merging this PR.

### DIFF
--- a/src/firefly/js/ui/DropDownContainer.jsx
+++ b/src/firefly/js/ui/DropDownContainer.jsx
@@ -25,6 +25,7 @@ import {TapSearchPanel} from 'firefly/ui/tap/TapSearchRootPanel.jsx';
 import {DLGeneratedDropDown} from './dynamic/DLGeneratedDropDown.js';
 import {useStoreConnector} from 'firefly/ui/SimpleComponent.jsx';
 import {JobMonitor} from '../core/background/JobMonitor';
+import {isHtml} from 'firefly/tables/TableUtil';
 
 
 export const flexGrowWithMax = {width: '100%', maxWidth: 1400};
@@ -112,21 +113,35 @@ DropDownContainer.propTypes = {
 };
 
 export function Alerts() {
+    let { msg } = useStoreConnector(getAlerts);
+    
+    if (!msg) return null;
 
-    const {msg} = useStoreConnector(getAlerts);
-    if (msg) {
-        /* eslint-disable react/no-danger */
-        return (
-            <Alert variant='outlined' color='warning'
-                   sx={{
-                       p:1/2,
-                       justifyContent:'center',
-                       borderRadius:0,
-                       backgroundColor:'#FFF7C2'            // retro look
-                   }}
-            >
-                <div dangerouslySetInnerHTML={{__html: msg}} />
-            </Alert>
-        );
-    } else return null;
+    const looksHtml = isHtml(msg);
+
+    if (!looksHtml) { //escape special chars then render safely as HTML below
+        msg = msg
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;')
+            .replace(/\n/g, '<br/>');
+    }
+
+    return (
+        <Alert variant='soft' color='warning' className='ff-alert'
+               sx={{p: 0.5, borderRadius: 0, justifyContent: 'center', width: '100%', textAlign: 'center',
+                '& p': { m: 0 }, '& ul, & ol': { m: 0, pl: 2 },
+                '& a': {
+                    color: 'inherit',
+                    textDecoration: 'underline',
+                    fontWeight: 600,
+                }
+            }}
+        >
+            {/* eslint-disable-next-line react/no-danger */}
+            <div style={{ width: '100%' }} className='ff-alert-content' dangerouslySetInnerHTML={{ __html: msg }} />
+        </Alert>
+    );
 }
+
+


robyww:
The code looks good so far but I would also like you to tweek if so it better supports html.
```html
<style>
    .myWarning{
        background-color: var(--joy-palette-primary-softBg);
        text-align: center;
        font-size:var(--joy-fontSize-lg);
        margin-top: 5px
    }
    div div:has(.myWarning) {
        padding:0;
    }
    div:has(.myWarning) {
        width: 100%;
    }
</style>
<div class= 'myWarning'>
    alert: be very afraid
</div>
```
Rubin had to use something like the above. Notice the two `has` entries. I would prefer that they did not have to do something this advanced.  Also the `.myWarning` style should now be unnecessary but you should probably check. 

They care about using html because they want to add links in there alerts.

If you need to test for html there is a `TableUtil.js`, `isHtml()`

kpuriIpac:
@robyww I addressed your changes and updated the test builds (they now include a dummy link as well). Can you take a look and let me know what you think? 

It handles links when present in the HTML, and generally added a bit more generic styling, so hopefully Rubin won't need to change much else (but I added `ff-alert` and `ff-alert-content` classes if they would **_really_** like to still customize this further, if you think that's unnecessary I can remove those).

src/firefly/js/ui/DropDownContainer.jsx
```diff
@@ -111,22 +112,51 @@ DropDownContainer.propTypes = {
...
+    let { msg } = useStoreConnector(getAlerts);
+
+    msg = 'This is a temporary warning msg for this PR. Follow this link: <a href="https://www.google.com"> Go to Google</a>';
+    if (!msg) return null;
+
+    const looksHtml = isHtml(msg);
+
+    let safeHtml;
+    if (looksHtml) {
+        safeHtml = cleanHtml(msg);
```
robyww: I have mixed feeling about using `cleanHtml`. It is safer but this is not user exposed html. It strips any style block. That give the deployment less control. There is not point then of having styles like `ff-alert-content` if we use `cleanHtml`

src/firefly/js/ui/DropDownContainer.jsx
```diff
@@ -111,22 +112,51 @@ DropDownContainer.propTypes = {
...
+        safeHtml = enforceLinkTargets(safeHtml); //only touches <a> tags
```
robyww: I really like this idea.

src/firefly/js/ui/DropDownContainer.jsx
```diff
@@ -111,22 +112,51 @@ DropDownContainer.propTypes = {
...
+                '& p': { m: 0 }, '& ul, & ol': { m: 0, pl: 2 },
+                '& a': {
+                    color: 'inherit',
+                    textDecoration: 'underline',
+                    fontWeight: 600,
+                }
+            }}
+        >
+            {/* eslint-disable-next-line react/no-danger */}
+            <div className='ff-alert-content'  dangerouslySetInnerHTML={{ __html: safeHtml }} />
```
robyww: make this `div` to be `width:100%`. That would fix most need to have styles.

src/firefly/js/ui/DropDownContainer.jsx
```diff
@@ -111,22 +112,51 @@ DropDownContainer.propTypes = {
...
+    let safeHtml;
+    if (looksHtml) {
+        safeHtml = cleanHtml(msg);
+        safeHtml = enforceLinkTargets(safeHtml); //only touches <a> tags
+    } else { //escape special chars then render safely as HTML below
+        safeHtml = msg
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;')
+            .replace(/\n/g, '<br/>');
```
robyww: good idea

robyww (approved):
For details see my comments below. I don't think we want to use safeHtml in this case. Make the cleanups and I think it is ready to merge.

The alert UI is much better now.

loitly:
> I noticed there’s a gap between the alerts and the rest of the page, which is especially noticeable in SPHEREx. We should address this.

**Never mind.  The gap is coming from SPHEREx. With the new lighter color and no border, it just blends into the white background more.**

Also, the additional enforcement (e.g., enforceLinkTargets) may not be necessary and could even be restrictive. In our use case, the admin is the one creating the message, so they should have control over the link target if needed.

kpuriIpac:
> Also, the additional enforcement (e.g., enforceLinkTargets) may not be necessary and could even be restrictive. In our use case, the admin is the one creating the message, so they should have control over the link target if needed.

Got it. I can get rid of `safeHtml = enforceLinkTargets(safeHtml);`, the idea was to ensure if there's a link, we set `target` to `_blank` (so that it opens a new tab always - in case the link that the admin puts in doesn't have this attribute set), but I can remove it too.

src/firefly/js/ui/DropDownContainer.jsx
kpuriIpac: Makes sense, what do you think the final change should be? Keep the call to `cleanHtml` and get rid of styles? (`ff-alert`)

kpuriIpac:
@loitly @robyww I implemented the feedback. Let me know what you think. Any final changes?

src/firefly/js/ui/DropDownContainer.jsx
loitly: I was expecting @robyww  to respond since this was his request.
@kpuriIpac  I think you misunderstood.  He meant that you should not add `cleanHtml`, as it would strip out any embedded styling in the HTML. Without that styling, `ff-alert` wouldn’t serve its purpose. The right thing to do is remove `cleanHtml` and keep `ff-alert`.

PR #1832, opened 2025-08-26
Firefly-1820,Firefly-1821: fix target panel issues

robyww:
#### Testing
- I could not exactly reproduce the issue in the video but I think I found the root causes
- test according the the tickets
- https://firefly-1820-target.irsakudev.ipac.caltech.edu/applications/spherex


lrebull:
I agree, as far as I can tell, both of these are fixed.

PR #1835, opened 2025-09-03
IRSA-7239: target entry bug

robyww:
#### Testing
- https://irsa-7239-target.irsakudev.ipac.caltech.edu/irsaviewer/dce


lrebull:
yep, this fixes it!

PR #1836, opened 2025-09-05
FIREFLY-1788: Add LaTeX support to unit labels in charts and input fields + upgrade plotly to v3

jaladh-singhal:
Fixes [FIREFLY-1788](https://jira.ipac.caltech.edu/browse/FIREFLY-1788)

**Dependency Updates:**

* Added following to package.json:
  - [better-react-mathjax](https://www.npmjs.com/package/better-react-mathjax) - to manage MathJax as per React lifecycle because otherwise MathJax directly updates DOM to render LaTeX which can cause flashing, non-rendered latex, etc. issues
  - [mathjax-full](https://www.npmjs.com/package/mathjax-full) - community-standard to render LaTeX on webpages (supported by Plotly out-of-the box)
  - [plotly.js-dist-min](https://www.npmjs.com/package/plotly.js-dist-min) - to upgrade plotly to v3 and to remove previous manual script copying logic
 
* Note: the initial bundle size is almost unaffected because of tree shaking and code splitting. You'll see two new JS chunks (in Networks tab and in WAR build): 
   * `plotly.js` (uses dynamic `import()`) - only gets dynamically loaded when the charts container is loaded in the UI
   * `tex-to-svg.js` (emitted from mathjax-full as a URL resource) - is loaded at app root but as async script so that it doesn't block UI.

**Config and LaTeX Setup:**
- Created `LatexUtil` with the LaTeX provider, config, styling, type definitions and util functions  
- For the UI components: Updated `FireflyRoot` to use the `LatexProvider` and global latex styles. And now anywhere in app, latex delimited strings wrapped within `<MathJax>` will be rendered.
- For Plotly charts: plotly automatically picks up global MathJax setup (but leaves less room for fine tuning on lower level so I have some workarounds in config and styling) 

**Spectrum Unit Conversion Improvements:**

* Refactored spectrum unit conversion functions to produce scientific labels and unit options using LaTeX formatting. `getUnitInfo` now has become `getXLabel`, `getYLabel`, and `getUnitOptions`. Also updated its consumer `SpectrumOptions` to render the Latex within `<MathJax/>`.
* Updated and expanded unit conversion tests in `SpectrumUnitConversion-test.js` 

**Extras:**
- Updated some plotly options that got restrucutred in v3 API
- Allow node-type in label of column input and listbox fields
* Updated Jest configuration to support both `raw` and `url` resource queries, and updated documentation to clarify usage of Jest CLI options for unit testing. 

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1788-latex-support/firefly/

Upload the 4 spectrum files (for different cases) in https://jira.ipac.caltech.edu/browse/FIREFLY-1694 - check that latex is rendered in axis labels in chart and in the chart options dialog. Changing units and spectral frame in the dialog, should update the chart with rendered latex without fail or any weirdness.

For developers: go to Networks tab, filter by JS and optionally enable throttling. Load firefly, check order of JS scripts being loaded: firefly-loader and firefly-dev -> tex-to-svg -> plolty (when a chart is requested)


package.json
```diff
@@ -27,8 +28,10 @@
     "json-bigint": "https://github.com/Caltech-IPAC/json-bigint",
     "local-storage": "~2.0",
     "lodash": "~4.17",
+    "mathjax-full": "^3.2.2",
```
loitly: Installing `better-react-mathjax` will also pull in `mathjax-full` as dependency.  Do we need to explicitly declare here?

package.json
```diff
@@ -27,8 +28,10 @@
...
     "md5": "~2.3.0",
     "moment": "~2.30",
+    "plotly.js-dist-min": "^3.0.3",
```
loitly: Are we now picking up minor versions automatically, rather than just patch updates?

loitly (approved):
Great work! I tested the UI and it works well. The code is clean, and I appreciate the unit tests.

package.json
jaladh-singhal: Yes because of `import mathjaxSrc from 'mathjax-full/es5/tex-svg.js?url';` in `LatexUtil.jsx` otherwise linter complains that this node module is not found.

package.json
jaladh-singhal: This is not deliberate. I just did `yarn add <package-name>` and it uses uses `^` by default. This is because yarn (and even npm) believes most packages follow semver and minor versions will be backward compatible so you can safely upgrade them.

As for "picking up versions **automatically**", we never automatically do that since we have yarn.lock with exact versions that anyone installing firefly will get. This `~` or `^` in package.json only matters when we (developers of firefly) do `yarn upgrade` once in a year or so.

loitly: Yes, I was referring to the second case where we run `yarn instal` after updating package.json to refresh dependencies and yarn.lock. Maybe we shouldn’t do that anymore and just install one package at a time.

jaladh-singhal: Yes it's generally better to let yarn/npm add packages rather than editing them in package.json yourself. I updated this in the documentation a while ago: https://github.com/IPAC-SW/irsa-ife/blob/dev/docs/developer-guide.md#updating-packagejson

jaladh-singhal: That being said `yarn install` and `yarn upgrade` will still not break our code because patch or minor version upgrades are backward compatible and we are doing them so infrequently that it's good if we get newer minor version (due to `^`) in the upgrade/install. But if you feel strongly about it, I'm ok with `~` too - we should just document our policy on dependencies better.

robyww: We have avoided putting revision in package.json. 
I think this is bettter.
```json
"plotly.js-dist-min": "^3.0",
```

robyww (approved):
This looks good. I think the `LatexUtil.jsx` is a good idea to provide some indirection. I confirm the plotly was lazy loading. The chart annotation looks very nice!

package.json
jaladh-singhal: @robyww the revision/patch number gets automatically added when you do `yarn add packagename`. I changed it to `yarn add -T packagename` to use `~` as @loitly suggested but still yarn puts the patch number in. That being said, this unnecessary appearance of patch number in package.json doesn't really matter since `~` prefix ensures that the package is not locked to the patch number:

From https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-tilde-t:

> Using --tilde or -T installs the most recent release of the packages that have the same minor version. The default is to use the most recent release with the same major version. For example, yarn add foo@1.2.3 --tilde would accept 1.2.9 but not 1.3.0.

PR #1840, opened 2025-09-11
fix position not updating for downloads

kpuriIpac:


src/firefly/js/ui/tap/Cutout.js
```diff
@@ -199,7 +199,11 @@ export function findCutoutTarget(dataProductsComponentKey=DEFAULT_DATA_PRODUCTS_
 
     if (pointIn(drawObj.pts, positionWP)) return result;
     const fallback = makeWorldPtUsingCenterColumns(table, row);
-    return {requestedType: result.requestedType, foundType: fallback ? ROW_POSITION : undefined, positionWP: fallback};
+    return {
+        requestedType: result?.requestedType,
+        foundType: result?.foundType ?? (fallback ? ROW_POSITION : undefined),
+        positionWP: result?.positionWP ?? (fallback || undefined)
+    };
```
kpuriIpac, line 206: Never mind, I don't need this code. I'll close this PR. 

The only use case was that if the user enters a position (user entered position) that fails the `pointIn(drawObj.pts, positionWP)` (returns `false`), then we fallback to the `ROW_POSITION`. But I think that's intentional, right? In that case, I don't need to add this code. 

Can you confirm @robyww? 

(I can confirm the bug is fixed even without this change, and just with the IFE change. I built the Jenkins builds as well with only the `rc-2025.4` branch and IFE feature branch)

kpuriIpac:
these changes are not needed

PR #1843, opened 2025-09-15
IRSA-7251 SPHEREx bug fixes

kpuriIpac:
**Tickets**: 
- [IRSA-7251](https://jira.ipac.caltech.edu/browse/IRSA-7251)  
  - fixes the unflattening/flattening downloads issue
    -  If `flattened` is turned off, the download script will always create folders (even if there's only 1 file per row/datalink entry)  
- [IRSA-7250](https://jira.ipac.caltech.edu/browse/IRSA-7250)
  - Change "All data" → "All Data" in the download dialog (and capitalized `Calibration` for when that's available as an option) 
  - when downloading `url-list`, the filename will now have `urllist` (instead of `urlsonly`)
  - In the script header (so for `wget` and `curl`): 
    - Changed: `Download SPHEREx from IRSA` → `Script to download SPHEREx data from IRSA`
    - and added this line `The script below defines a subroutine to download the files`
 - (please see my comments in ticket for items not addressed in this PR)  
- [IRSA-7249](https://jira.ipac.caltech.edu/browse/IRSA-7249)
  - Remove table navigation for the main table on the left for Spherex Spectrophotometry search results 
  - Code change for this change is in this IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/439 

**Testing**
- DCE: https://irsa-7251-spherex-bugs.irsakudev.ipac.caltech.edu/irsaviewer/dce
- Euclid: https://irsa-7251-spherex-bugs.irsakudev.ipac.caltech.edu/applications/euclid
- SPHEREx: https://irsa-7251-spherex-bugs.irsakudev.ipac.caltech.edu/applications/spherex
- Firefly: https://fireflydev.ipac.caltech.edu/irsa-7251-spherex-bugs/firefly
- Since changes have been made to the download script logic, I would test all apps where we have downloads 
  - CADC on Euclid, DCE on IRSAviewer, and Euclid & SPHEREx downloads 
- Notice folder (directory) creation for Spherex now when `flattened` option is turned off 
- @lrebull you may even test changing cutout sizes and noticing that being respected in the URLs generated in the download script (this was fixed in my last PR)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/ScriptAttributes.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/ScriptAttributes.java
@@ -12,5 +12,5 @@ package edu.caltech.ipac.firefly.core.background;
 /**
 * @author Trey Roby
 */
-public enum ScriptAttributes {URLsOnly, Unzip, Curl, Wget}
+public enum ScriptAttributes {URLList, Unzip, Curl, Wget}
 
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/EmailNotification.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/EmailNotification.java
@@ -109,7 +109,7 @@ public class EmailNotification implements JobCompletedHandler {
                 String sugName = ifNotNull(jobInfo.getAux().getTitle()).getOrElse("Download");
                 String curlScript = getDownloadURL(makeScript(jobInfo, Curl), makeScriptFilename(Curl, sugName), jobInfo.getMeta().getAppUrl());
                 String wgetScript = getDownloadURL(makeScript(jobInfo, Wget), makeScriptFilename(Wget, sugName), jobInfo.getMeta().getAppUrl());
-                String directUrls = getDownloadURL(makeScript(jobInfo, ScriptAttributes.URLsOnly), makeScriptFilename(URLsOnly, sugName), jobInfo.getMeta().getAppUrl());
+                String directUrls = getDownloadURL(makeScript(jobInfo, ScriptAttributes.URLList), makeScriptFilename(URLList, sugName), jobInfo.getMeta().getAppUrl());
                 String msg = zipSuccess.formatted(name, curlScript, wgetScript, directUrls, contact);
                 EMailUtil.sendMessage(new String[]{email}, null, null, subject, msg);
             } catch (Exception e) {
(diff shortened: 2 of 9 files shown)


src/firefly/java/edu/caltech/ipac/firefly/core/background/ScriptAttributes.java
```diff
@@ -12,5 +12,5 @@
 /**
 * @author Trey Roby
 */
-public enum ScriptAttributes {URLsOnly, Unzip, Curl, Wget}
+public enum ScriptAttributes {URLList, Unzip, Curl, Wget}
```
loitly, line 15: What was the reason for changing this enum? I believe it’s also referenced in several JS/JSX files, but I don’t see those updates here. Are you sure you tested all of them?

loitly (commented):
I tested a couple of applications and they seem to work fine. However, I still see quite a few references to `URLsOnly` in the code, especially in JS/JSX files. Generally, the enum name shouldn’t need to change just for display purposes.  You could try overriding `toString` if needed. I’m not entirely convinced about this change, but if you can confirm that the remaining `URLsOnly` references are safe, I’m okay with it.

src/firefly/java/edu/caltech/ipac/firefly/core/background/ScriptAttributes.java
kpuriIpac: Yes I believe I changed it everywhere it's used, so the change should be safe. 

But you're right, I only needed this change to have in the call to `makeScriptFilename` where we pass this enum and then it's converted to a string to be used in the filename, so the URL file name would have a suffix like `-urlsonly` instead of `urllist`. 

I changed the enum itself so that I wouldn't have to touch the code in `makeScriptFilename` function (in `DownloadScript.java`).

kpuriIpac: But I'm open to reverting this change (keeping the enum the same) for only passing in a different one for URLs when calling `makeScriptFilename`

lrebull:
flattening issue - fixed!
capitalization in pop-up - fixed!
url-list - fixed!
"download spherex data from irsa" - fixed!
subroutine comment - checking with Vandana
6&7 - wishlist
subdirectory called SPHERExLVFSurveyData - ticket created - Anastasia says this is a SPHEREx-set thing.

looks good!

waiting on the spectrophot table results to test the table nav stuff.

lrebull:
yep, can verify table nav is gone on the left! :)

kpuriIpac:
> 6 & 7 wishlist

Yes we should make a new ticket for these two items too. We can tackle them for the next major release (November)

lrebull:
spawned new tickets 
IRSA-7275 and IRSA-7274

src/firefly/java/edu/caltech/ipac/firefly/core/background/ScriptAttributes.java
loitly: Generally, you shouldn’t rename an enumeration just to change how it’s labeled. However, if the intent is to refactor because the existing name is misleading, then it makes sense to rename it globally.
I noticed references on the client side, in BackgroundUtil.js and ScriptDownloadDialog.jsx, and there may be more. Please test after making the modifications.

loitly (approved):

kpuriIpac:
Okay done, made the changes @loitly and replaced all instances of `URLsOnly` with `URLList`. Updated the test builds as well for good measure. 

Can you take a quick look at https://github.com/IPAC-SW/irsa-ife/pull/439 (**IFE PR**) and approve that as well? Thanks!

PR #1849, opened 2025-09-24
FIREFLY-1802: Fixed deleted jobs return after importing history

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1802

This PR ensures that deleted jobs do not reappear after importing job history from preconfigured UWS services.

**Changes**
- When a user deletes a job from the UI, the job phase is now transitioned to ARCHIVED instead of being immediately expired.
- Archived jobs are not displayed in the UI.
- Have a dedicated archive job cleanup expiry time that is longer than the general job expiry.
- This guarantees that deleted jobs remain hidden long enough to prevent them from reappearing after an import when the general expiry time is set too short.

Test: https://firefly-1802-deleted-jobs-come-back.irsakudev.ipac.caltech.edu/applications/spherex/
- Run one or more Spectrometry Tool queries to generate job history.
- Delete a job from the UI and confirm that it disappeared.
- Verify that deleted jobs are not shown in job listings after reloading the page.

**Note:** This instance was deployed with the Spectrometry Tool service configured as one of the UWS services to import job history from.


robyww (approved):
looks good.

PR #1852, opened 2025-09-26
Firefly-1840: Support S3

robyww:
#### [Firefly-1840](https://jira.ipac.caltech.edu/browse/FIREFLY-1840): Add S3 support

 - Added s3 async multithreaded retrieve
 - Now we cna create a S3 or http request type
 - upload will recognize s3
 - WebPlotRequest supports setting both then deferring download decision
 - A lot of refactoring

#### Review
critical files to look at: most of the new logic is here, the rest is refactoring and supporting new modes
- RetrieveUtil.java
- S3Download.java
- I am also looking for feedback for `RetrieveUtil. detemineOptimalResourceType()`

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1840-s3/firefly
- use SIAv2 search with (cloud_access column in results): 
  - 273.11730, 68.20761, 100 arcs, Collection: euclid_DpdMerBksMosaic
  - The Euclid images are big but retrieved pretty fast
- Testing use uploaded table with images that use `s3://` syntax
- [s3.tbl.zip](https://github.com/user-attachments/files/22567728/s3.tbl.zip)
- Test Uploading directly using the upload panel: use S3 uri's from the table above


buildScript/dependencies.gradle
```diff
@@ -101,6 +101,11 @@ dependencies {
 
     // embedded redis server; version ~ 6.2
     implementation 'com.github.codemonstur:embedded-redis:1.4.3'
+
+    implementation 'software.amazon.awssdk:s3:2.33.11'
+
+    implementation 'software.amazon.awssdk:s3-transfer-manager:2.33.11'
```
loitly: The s3-transfer-manager already includes s3:2.33.11 along with several other dependencies, so there’s no need to reference s3:2.33.11 specifically.

I would also add a comment on top, e.g.
```
// streamlines file transfers to and from Amazon S3 with concurrent, multipart uploads and downloads.
implementation 'software.amazon.awssdk:s3-transfer-manager:2.33.11'
```

src/firefly/java/edu/caltech/ipac/firefly/visualize/WebPlotRequest.java
```diff
@@ -641,6 +631,21 @@ public String getURL() {
...
+        setParam(S3_BUCKET, bucket);
+        setSafeParam(S3_KEY, key);
+    }
+
+    public String getS3Region() { return getParam(S3_REGION); }
+    public String getS3Bucket() { return getParam(S3_BUCKET); }
+    public String getS3Key() { return getSafeParam(S3_KEY); }
+
+    public void setS3Uri(String s3Uri) { setParam(S3_URI, s3Uri); }
+    public String getS3Uri() { return getSafeParam(S3_URI); }
```
loitly, line 632: I’m not sure how you’re using this, but just to clarify: an S3 URI only contains the bucket and key. It’s the S3 URL that requires the region.

src/firefly/java/edu/caltech/ipac/util/download/AnyUrlParams.java
```diff
@@ -0,0 +1,77 @@
...
+import edu.caltech.ipac.firefly.server.network.HttpServiceInput;
+
+import java.io.File;
+import java.net.URL;
+import java.util.Map;
+
+public class AnyUrlParams extends BaseNetParams {
+
+    private final URL url;
+    private final S3Download.S3Ref s3Ref;
```
loitly: In my view, the `URL` and `S3Ref` should be a single URI or String parameter. Splitting them will eventually require adding more types, such as gs://.

src/firefly/java/edu/caltech/ipac/util/download/S3Download.java
```diff
@@ -0,0 +1,341 @@
...
+
+    public static FileInfo getPublicDataSync(S3Ref ref, File outfile, Options options) throws FailedRequestException {
+        return getDataSync(ref,outfile,S3Download::applyAnonymouse,options);
+    }
+
+
+    public static FileInfo getDataSync(S3Ref ref,
+                                   File outfile,
+                                   ApplyCredentials applyCredentials,
+                                   Options options) throws FailedRequestException {
```
loitly: When would we want to use the `sync` method?

src/firefly/js/metaConvert/vo/ObsCoreConverter.js
```diff
@@ -201,10 +201,10 @@ async function getObsCoreSingleDataProduct({table, row, activateParams, serviceD
 
 
         const positionWP= getSearchTarget(table.request,table) ?? makeWorldPtUsingCenterColumns(table,row);
-        const request= makeObsCoreRequest(dataSource, positionWP, titleStr,table,row);
+        const request= makeObsCoreRequest(dataSource, cloudAccess, positionWP, titleStr,table,row);
         const primDPType= doFileAnalysis ?
             await uploadAndAnalyze({request,table,row,activateParams,serviceDescMenuList,originalTitle:request.getTitle()}) :
-            createGuessDataType(titleStr,'guess-0',dataSource,prodType,undefined,activateParams, positionWP,table,row,size);
+            createGuessDataType(titleStr,'guess-0',dataSource,cloudAccess, prodType,undefined,activateParams, positionWP,table,row,size);
```
loitly, line 207: It might be a bit late, but these functions really should have taken the entire data structure from getObsCoreRowMetaInfo() as a single parameter. Right now, there are too many repeating parameters without any grouping.

loitly (approved):
I’ve reviewed the code and it looks solid overall. I tested with an S3 URI and confirmed it was able to retrieve the file. 
My concern is about introducing additional fetching utility classes that overlap with URLDownload as well as existing interfaces. I think the model should instead use a top-level service/manager class that defines inputs and outputs, while delegating the actual work to a low-level HTTP client or library(AWS s3, or Google gcs). As it stands, I’m not clear on when to use LockingRetrieve, RetrieveUtil, or URLDownload.

src/firefly/java/edu/caltech/ipac/util/download/RetrieveUtil.java
```diff
@@ -0,0 +1,223 @@
...
+        String serverHost= null;
+        String resHost= null;
+        try {
+            serverHost= new URI(ServerContext.getRequestOwner().getBaseUrl()).toURL().getHost();
+            resHost= new URI(urlStr).toURL().getHost();
+        } catch (URISyntaxException | MalformedURLException ignore) { }
+        if (ref==null) return ResourceType.URL;
+        if (serverHost==null || resHost==null) return ResourceType.S3;
+        return FileUtil.makeHostDomain(serverHost).equals(FileUtil.makeHostDomain(resHost))
+                ? ResourceType.URL : ResourceType.S3;
```
loitly: My interpretation is that `cloud_access` should be used when running in a cloud environment; otherwise, use `url_access`.  
One approach is to introduce a property (e.g., runtime_platform) and use it when available, falling back to auto-detection if it’s not set. There are ways to detect whether you’re running in AWS or GCP. This is also useful when both URIs are provided in `cloud_access`.

robyww: `url_access` is almost certainly not preferred in many non-cloud cases. That is why there were looking to add more properties into the `cloud_access` object. I don’t totally understand all the uses cases though. I want to have a another meeting with @alaity47 to discuss it.

src/firefly/js/metaConvert/vo/ObsCoreConverter.js
robyww: I do need to clean this up. Probably in a future PR,  but I can look at it and see.

src/firefly/java/edu/caltech/ipac/util/download/S3Download.java
robyww: I meant to document this.  It was my first pass at using s3. It is not used but I want to keep it around for reference. I will rename the function. It was a place holder based on the s3 api.

src/firefly/java/edu/caltech/ipac/util/download/AnyUrlParams.java
robyww: S3 can have 2 or 3 fields specified in 2 different ways. The S3Ref encapsulates that. 
S3 can have `region` - `bucket` -`key` or `bucket` - `key`, the string version `s3://` only specifies two.

I could generalize it farther to include a URL. I will look at that.

src/firefly/java/edu/caltech/ipac/firefly/visualize/WebPlotRequest.java
robyww: see my comment below.

buildScript/dependencies.gradle
robyww: I will fix this.

jaladh-singhal (approved):
I only tested this PR and didn't review code (since Loi already did) - it works great!

Euclid SIA results load full images, S3 URLs in table and url input are recognized. Some comments:
- Testing this build on LAN, 1406M Euclid images take a lot longer than I expected. Switching to Caltech WiFi with VPN, they load faster. But I'm not sure if it's reliable observation since LAN might be facing some issues. (I tried to avoid caching by choosing different rows for each case)
- We should indicate in "upload from URL" UI that URLs can have s3 protocol too, not just https - scientists can probably come up with right wording.

src/firefly/js/metaConvert/vo/ObsCoreConverter.js
robyww: Just FYI: makeObsCoreRequest parameters come from more that a obscore table row. It can also be data link and some of the parameters are computed. The table and row parameters might be an obscore table with the access url or the reference obscore table when the access_url is from datalink. Or the dataSource might be computed in a service descriptor. (in the case cloud_access would be undefined).

PR #1853, opened 2025-09-29
FIREFLY-1866: Parenthesize composite-key JOINs in ADQL-generation

gpdf:
[FIREFLY-1866](https://jira.ipac.caltech.edu/browse/FIREFLY-1866)

This is:
* a pedagogical improvement because it keeps composite keys clearly identified in the output
* a workaround for a problem in Rubin Qserv which doesn't correctly handle `AND`s in `JOIN ON` clauses


loitly (approved):
The changes look good.

PR #1855, opened 2025-10-01
IRSA-7274: Download Dialog & Checkbox improvements

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/IRSA-7274
- When a user selects `cutout` from the download dialog, we now notify them of the current cutout size (and point them to the cutout dialog if they need to change it) 
- Made some improvements to `CheckboxGroupInputField` that allow `All Data` to basically operate as a `Select All` button
  - this logic does exist elsewhere like for checkboxes in tables, but that doesn't use `CheckboxGroupInputField` 
  - For `CheckboxGroupInputField`, this logic was partially in place (e.g. it does expect `_all_` and `_none_` as options). 
     - Now with the added logic, it allows you to easily assign a `Select all` type option, which when selected, selects all other options, and is automatically unselected when you unselect any of the other options. 
     - this logic is now generic and can be applied anywhere `CheckboxGroupInputField` is used

**Testing**: 
- [Euclid](https://irsa-7274-cutouts-downloads-imrpove.irsakubedev.ipac.caltech.edu/applications/euclid) & [Spherex](https://irsa-7274-cutouts-downloads-imrpove.irsakubedev.ipac.caltech.edu/applications/spherex): Do a search, select a few rows, open the download dialog
  - play around with the `All Data` selection (select/unselect it, manually select all other options when `All Data` is unselected, etc.) 
  - When `All Data` is selected, or when you manually select `Cutout`, you should see the cutout size show up dynamically
-  [Firefly](https://fireflydev.ipac.caltech.edu/irsa-7274-cutouts-downloads-imrpove/firefly): Mostly regression testing, test any piece of UI that uses `CheckboxGroupInputField`, a couple of instances are: 
   - in the Plot/Chart options under `X label` or `Y label` like `grid, reverse, top, log` (select/unselect any of these and apply to make sure it works as expected)
   - The checkboxes under `Select Data Set` on the `Images` tab. 
      - As an example, try selecting the `Mission`: `Wise` and then `Wise ALLWISE Atlas` (the first option). Make sure all of the nested options under it are also selected. 
        - play around with the checkboxes to ensure they work as expected.

### DIFF
--- a/src/firefly/js/templates/common/ttFeatureWatchers.js
+++ b/src/firefly/js/templates/common/ttFeatureWatchers.js
@@ -13,14 +13,15 @@ import {getCatalogWatcherDef} from '../../visualize/saga/CatalogWatcher.js';
 import {getUrlLinkWatcherDef} from '../../visualize/saga/UrlLinkWatcher.js';
 import {getActiveRowToImageDef} from '../../visualize/saga/ActiveRowToImageWatcher.js';
 import {getMocWatcherDef} from '../../visualize/saga/MOCWatcher.js';
-import {useStoreConnector} from 'firefly/ui/SimpleComponent';
+import {useFieldGroupValue, useStoreConnector} from 'firefly/ui/SimpleComponent';
 import {findCutoutTarget, getCutoutSize, ROW_POSITION, tblIdToKey,} from 'firefly/ui/tap/Cutout';
 import {getTableModel} from 'firefly/voAnalyzer/VoCoreUtils';
 import {fetchSemanticList} from 'firefly/metaConvert/vo/DatalinkFetch';
 import {checkForDatalinkServDesc} from 'firefly/ui/dynamic/ServiceDefTools';
-import {CheckboxGroupInputField} from 'firefly/ui/CheckboxGroupInputField';
-import {Stack} from '@mui/joy';
+import {CheckboxGroupInputField, SelectAllCheckbox} from 'firefly/ui/CheckboxGroupInputField';
+import {FormControl, FormLabel, Stack, Typography} from '@mui/joy';
 import {ToolbarButton} from 'firefly/ui/ToolbarButton';
+import {FieldGroup} from 'firefly/ui/FieldGroup';
 
 export const getAllStartIds= ()=> [
     getMocWatcherDef().id,
@@ -105,6 +106,45 @@ function getColNameFromTemplate(template) {
     return template.match(/\${[\w -.]+}/g)?.map( (s) => s.substring(2,s.length-1));
 }
 
+function ProductTypesBlock({tbl_id, dynamicOptions, cutoutValue}) {
+    const [getProductTypes] = useFieldGroupValue('productTypes', tbl_id);
+    const productTypes = getProductTypes();
+
+    const isCutoutSelected = productTypes?.split(',').map((val) => val.trim()).includes('#cutout') ?? false;
+
+    return (<Stack spacing={2}>
+            <FormControl>
+                <Stack direction='row' alignItems='center' spacing={1}>
+                    <FormLabel>Products to Download:</FormLabel>
+                    <SelectAllCheckbox
+                        fieldKey='productTypes'
+                        groupKey={tbl_id}
+                        label='Select all'
+                        options={dynamicOptions}
+                    />
+                </Stack>
+                <Stack spacing={1} mt={1}>
+                    <CheckboxGroupInputField
+                        fieldKey='productTypes'
+                        groupKey={tbl_id}
+                        options={dynamicOptions}
+                        //initialState={{ value: dynamicOptions.map((o) => o.value).toString() }} //this is if we want to keep default all  vals selected
+                        initialState={{ value: '' }}
+                        alignment='horizontal'
+                    />
+                    {isCutoutSelected && (
+                        <Typography level='body-sm' sx={{ ml: 1 }}>
+                            Note: the current cutout size is {cutoutValue} deg
+                            (you may change this via the cutout dialog).
+                        </Typography>
+                    )}
+                </Stack>
+            </FormControl>
+        </Stack>);
+}
+
+
+
 export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showFileStructure=false,
                                                downloadType='package', dataSource, fileName}) => {
     const tbl_id = table_id || getActiveTableId();
@@ -142,20 +182,16 @@ export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showF
 
     //dynamically generate options from semList
     const dynamicOptions = useMemo(() => {
-        let options;
         if (semList.length > 0) {
-            //use semList if available
-            options = semList.map((value) => ({
-                label: labelMap[value] || value.replace(/^#/, ''), //strip "#" if using raw value as fallback for a cleaner label in the UI
+            return semList.map((value) => ({
+                label: labelMap[value] || value.replace(/^#/, ''),
                 value
             }));
-        } else {
-            options = []; //fallback only to "All data" only below
         }
-        //ensure '*' ("All data") is always included, even as a fallback when semList is empty
-        options.push({ label: 'All Data', value: '*' });
-        return options;
+        return [];
     }, [semList, labelMap]);
+
+
 
     const tblModel = getTableModel(tbl_id);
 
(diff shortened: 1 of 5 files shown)


src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -152,11 +181,13 @@ export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showF
         } else {
             options = []; //fallback only to "All data" only below
         }
-        //ensure '*' ("All data") is always included, even as a fallback when semList is empty
-        options.push({ label: 'All Data', value: '*' });
+        //ensure '_all_' ("All data") is always included, even as a fallback when semList is empty
+        //and make it the first option 
+        options.unshift({ label: 'All Data', value: '_all_' });
```
loitly: I don’t like the idea of presenting an “All data” checkbox that looks identical to the other checkboxes. Typically, a “Select All” control isn’t presented in the same way as the rest of the checkboxes.

I also don’t like the weak coupling of  `_all_` as a reserved value of the CheckboxGroup.

loitly (approved):
I’m approving this PR since it works as expected. However, I’m not sold on the implementation or the presentation of the “All Data” checkbox. I think it would be good for someone else to review that part as well.

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: Yes you're right. I tried separating the `All Data` out onto a separate line and keeping the other options nested under it, but I ran into some weird bugs with separating out the checkbox groups that way. Let me give it a try again, though. 

The other option is removing `All Data` as an option for now. What I don't like is how it works now (before this PR), where when you can select `All Data` and, say, 2 other options (out of 6 total options). But since you have `All Data` selected, it's anyway going to  download all the data (even though you only have 2 other options selected). 

So in my opinion, we either fix `All Data` to work like a `Select all` checkbox or we remove it altogether as an option (which is logic that we do have elsewhere on firefly as well).

loitly: If you need a reusable `Select All` widget, you can export one from `CheckboxGroupInputField`. It should only take `fieldKey` and the optional props `groupKey` and `label`. That’s all the information required to fully control the checkbox group. You can then render it anywhere you like.  An option for your case is next to the `Products to Download` label.

kpuriIpac:
@loitly new changes are ready for review (tomorrow or Monday when you get a chance). A summary of changes: 

- I added a `SelectAllCheckbox` component. 
- Modified existing logic in `handleOnChange` to deal with turning on/off the `Select all` checkbox and other checkboxes
   - as we discussed, the changes to your selections persist so even when you switch rows the download dialog selections remain the same
- While this still makes use of `_all_`, an example of generic usage of `SelectAllCheckbox` is how I've implemented it in `ttFeatureWatchers`
  - one call to `SelectAllCheckbox` for the `Select All` checkbox (customizable name) and one call to `CheckboxGroupInputField` directly for the rest of the options, both using the same `fieldKey` 
- So now any usage of `CheckboxGroupInputField` can be coupled with a `Select all` type checkbox, but for other existing uses of `CheckboxGroupInputField` where it's called without one, everything should still work the same. 

Let me know what you think. And if it looks good then, should I document some of this usage in the code as comments above `SelectAllCheckbox`? 

BTW: testing instructions are same as above. I just changed `All Data` text to `Select all` instead. And there's a spacing issue in the dialog (awkward/too much space), I'll fix that in a small commit next.  
  - Also, right now the `Select all` checkbox shows up next to the label `Products to Download`. I can try it on the next line as well.

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -105,6 +106,50 @@ function getColNameFromTemplate(template) {
...
+    const productTypes = getProductTypes();
+
+    const isCutoutSelected = (() => {
+        if (!productTypes) return false;
+        const vals = productTypes.split(',').map((val) => val.trim());
+        return vals.includes('#cutout') || vals.includes('_all_');
+    })();
+
+    return !isDatalink && (
+        <Stack spacing={2}>
```
loitly: Minor:  I thought you show `Products to Download` when it is a datalink table.
Also, since `isDatalink` is static, does it make sense to reserve a prop for it?  Logic could be 
```js
isDatalink && <ProductTypesBlock/>
```

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -105,6 +106,50 @@ function getColNameFromTemplate(template) {
...
+                        fieldKey='productTypes'
+                        groupKey={tbl_id}
+                        label='Select all'
+                    />
+                </Stack>
+
+                <Stack spacing={1} m={1}>
+                    <CheckboxGroupInputField
+                        fieldKey='productTypes'
+                        options={dynamicOptions.filter((op) => op.value !== '_all_')}
```
loitly: There should not be any reference to `_all_`.

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -190,37 +237,33 @@ export const PrepareDownload = React.memo(({table_id, tbl_title, viewerId, showF
...
-                                        label='Products to Download: ' />
-                            </Stack>}
-                        </DownloadOptionPanel>
-                    </DownloadButton>
-                </Stack>
+                <FieldGroup groupKey={tbl_id}>
+                    <Stack>
+                        <DownloadButton key={keysToWatch}
+                            buttonText = {downloadType === 'script' ? 'Generate Download Script' : 'Prepare Download'}>
+                            <DownloadOptionPanel {...{
```
loitly, line 233: Minor:  it's a good idea to move this into a separate component.  It makes it easier to read as well as identifying props(input).

src/firefly/js/ui/CheckboxGroupInputField.jsx
```diff
@@ -62,33 +62,67 @@ CheckboxGroupInputFieldView.propTypes= {
...
     const checked = ev.target.checked;
-    const curValueArr = getCurrentValueArr(value);
-    const idx = curValueArr.indexOf(val);
-    if (checked) {
-        if (idx < 0) curValueArr.push(val); // add val to the array
+    const allValues = (viewProps.options ?? []).map((option) => option.value);
+
+    //expand '_all_' into the full options list and use that as the list of curr values array
+    const normalizedValue = convertValue(viewProps.value, viewProps.options);
+    const curValueArr = getCurrentValueArr(normalizedValue);
```
loitly: Logic should not be here.  It should be in the `SelectAllCheckbox` component.

loitly (changes requested):
Please keep the 'Select All' logic confined to SelectAllCheckbox component.

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: Yes that's intended to be `!isDatalink`. But I can see why that's confusing based on our conversation yesterday. We want to show `Products to Download` for rows pointing to datalink (so rows containing an `access_url`). For example when you do an Image search on Euclid. So by looking at the datalink we can get a list of the products available to download, and then we display them to the user. 

But when you actually have a datalink table directly, like so: 

ops


(you get to this by extracting the current row and doing `Show table of all Euclid products`) 

Since this table is directly datalink, and users can see the semantics directly (and download each one), there's no point of showing the `Products to Download` here. 

I agree about the prop logic, it's unnecessary there. I'll change to: 

`!isDatalink && <ProductTypesBlock/>`

src/firefly/js/ui/CheckboxGroupInputField.jsx
kpuriIpac: Ok yes I understand this now. Separation of concern as well! I'll implement this.

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: Oh and BTW I notice the bug there. `Generate Download Script` & `Prepare Download` both showing up for the datalink table. Unfortunately on ops. I had fixed this bug, but after having made some more changes (and fixing a spherex bug related to this), it must have creeped back in. Creating a bug ticket for myself.

kpuriIpac:
Thanks for the feedback again @loitly. This revision should be a lot cleaner: 

- Separated all `SelectAllCheckbox` logic from pre-existing `CheckboxGroup` logic. 
  - and added a JSDoc comment for `SelectAllCheckbox`
- no reference to `_all_` anymore outside of `SelectAllCheckbox` logic (e.g. in `ttFeatureWatchers` even when creating the `dynamicOptions` list, not injecting `_all_` into the list anymore)
  - actually, in the latest commit I made a bug fix which also included getting rid of `_all_` altogether. Not needed in `SelectAllCheckbox`. 
- all 3 test builds updated 

I'll work on fixing the styling (bit too much space between the product types and the flattened file structure option, even more when you see the cutout size line). 

Let me know what you think?

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: Not sure about this one, let's talk briefly on Monday morning before the meeting?

src/firefly/js/templates/common/ttFeatureWatchers.js
```diff
@@ -105,6 +106,49 @@ function getColNameFromTemplate(template) {
...
 
+function ProductTypesBlock({tbl_id, dynamicOptions, cutoutValue}) {
+    const [getProductTypes] = useFieldGroupValue('productTypes', tbl_id);
+    const productTypes = getProductTypes();
+
+    const isCutoutSelected = (() => {
+        if (!productTypes) return false;
+        const vals = productTypes.split(',').map((val) => val.trim());
+        return vals.includes('#cutout');
+    })();
```
loitly: Out of curiosity, is there a reason this is defined as a function and then called right away?  Isn't this the same as ...

```js
  const isCutoutSelected = productTypes?.split(',').map((val) => val.trim()).includes('#cutout');
```

src/firefly/js/ui/CheckboxGroupInputField.jsx
```diff
@@ -62,6 +62,44 @@ CheckboxGroupInputFieldView.propTypes= {
...
+        const checked = ev.target.checked;
+        const nextVal = checked ? allValues : [];
+        const {valid, message} = viewProps.validator(nextVal.toString());
+        fireValueChange({value: nextVal.toString(), message, valid});
+    };
+
+    return (
+        <Stack direction='row' alignItems='center'>
+            <Checkbox
+                size='sm'
```
loitly, line 101: The code looks good. Thanks for the changes.
If you want to make it more reusable, consider removing the root Stack component and adding a props parameter. Then, spread those props into the Checkbox. This will allow users of the component to customize its styling more easily.

loitly (approved):
Tested the changes and it works as expected.
 Okay to merge when ready.

src/firefly/js/templates/common/ttFeatureWatchers.js
kpuriIpac: Yes,I was doing some more stuff in it in previous iterations (like checking for #_all_ in addition to #cutout) - but it's now simple enough to have it be the one line variable instead of function like you suggested. So I'll make the change.

kpuriIpac:
@loitly  I only made minor changes, but if you want to take a look at the latest commit, here's what it essentially does:

- minor cleanup based on your last suggestion
- get rid of the `All data` or `Select all` checkbox when there are no product types to display
  - for example, in DCE (https://irsa-7274-cutouts-downloads-imrpove.irsakubedev.ipac.caltech.edu/irsaviewer/dce) you would see this before: 
  - Screenshot 2025-10-09 at 5 01 35 PM
  - which obviously does not make sense as there are no options, so in case `options` is empty now I don't display this checkbox anymore

kpuriIpac:
@robyww the bug I found in firefly is that if we initialize `PrepareDownload` for a table for which there's no `MetaConst.DATA_SERVICE_ID` for a table, then the `viewerId` in ttFeatureWatchers will be `undefined`. 
  - In this case, we default to the default `dataProductsComponentKey` in the Cutout.js whose value is `DataProduct.defaultComponentDataKey`
  - but i think the actual default `dataProductsComponentKey` that we use is the same as the `dataTypeViewerId`  or `dpId` which happens to be `DataProductsType`...traced this all the way back to line ~17 in `MetaDataMultiProductViewer`
    - So, because of these two values not being the same, when we update cutouts in firefly (you can try IVOA CADC search here: https://fireflydev.ipac.caltech.edu/irsa-7274-cutouts-downloads-imrpove/firefly), you'll see the cutout updated in the UI. But when you `Prepare Download` and select Cutout, you won't see the updated cutout value.  
  - The **_fix_**: changing the `dataProductsComponentKey` in cutout.js to this value (`DataProductsType`) will resolve this bug, but I wanted to check with you if this makes sense. 

If this is too verbose, we can discuss briefly tomorrow morning on zoom. I'd like to just put this fix into the PR as well before merging then.

PR #1865, opened 2025-10-16
FIREFLY-1824: Allow option for Job's date/time to be displayed in local timezone.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1824

Add a toggle to switch between **UTC** and **local timezone**.
Update all job date/time fields to display according to the selected timezone.

Test: https://firefly-1824-bgmonitor-timezone.irsakubedev.ipac.caltech.edu/applications/spherex/

- Go to Job Monitor.
- If no job history exists, create a new job.
- Toggle between UTC and Local Time. Verify that all date/time values in the Job History table update accordingly.
- Open Job Details.  Confirm that date/time fields there also reflect the selected timezone.

### DIFF
--- a/src/firefly/js/core/background/JobMonitor.jsx
+++ b/src/firefly/js/core/background/JobMonitor.jsx
@@ -34,8 +34,13 @@ import NotificationsOffIcon from '@mui/icons-material/NotificationsOff';
 import NotificationsActiveIcon from '@mui/icons-material/NotificationsActive';
 import {FormWatcher} from '../../templates/router/RouteHelper';
 import {logger} from '../../util/Logger';
+import {SwitchInputField} from 'firefly/ui/SwitchInputField';
+import {getFieldVal} from 'firefly/fieldGroup/FieldGroupUtils';
 
 export const jobMonitorPath = '/jobMonitor';
+export const jobMonitorGroupKey = 'jobMonitor';
+export const useLocalTimeKey = 'useLocalTime';
+
 const jobIdColIdx = 7;  // the index of the jobId column in the table
 
 export function JobMonitor({initArgs, help_id, slotProps, ...props}) {
@@ -148,7 +153,7 @@ function JobSummary({jobs, overflow, ...props}) {
         </Stack>
     );
     return (
-        <Stack direction='row' gap={5} {...props}>
+        <Stack direction='row' gap={5} justifyContent='space-between' {...props}>
             <Typography level='title-md' color='primary'>Job Summary</Typography>
             <Stack direction='row' gap={10}>
                 <Entry label='Total' value={total + (overflow ? '(+)' : '')}/>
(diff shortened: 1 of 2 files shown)


src/firefly/js/core/background/JobMonitor.jsx
```diff
@@ -198,17 +221,18 @@ function Notification({email='', notifEnabled, ...props}) {
 
 function JobMonitorTable({help_id, ...props}) {
     const jobMap = useStoreConnector(() => getBackgroundInfo()?.jobs || {});
+    const useLocalTime = useStoreConnector(() => getFieldVal(jobMonitorGroupKey, useLocalTimeKey));
```
robyww, line 224: The better way to do this is. 
```jsx
    const useLocalTime=  useFieldGroupValue(useLocalTimeKey)[0]();
```
Also you don't need to group key since it comes from the context so it is more reusable.

src/firefly/js/core/background/JobMonitor.jsx
```diff
@@ -148,18 +153,36 @@ function JobSummary({jobs, overflow, ...props}) {
...
                 {(archived>0) && <Entry label='Archived' value={archived} title={getPhaseTips(Phase.ARCHIVED)}/>}
             </Stack>
+            <LocalOrUTC/>
         </Stack>
     );
 }
 
+export function LocalOrUTC() {
+    const parts = Intl.DateTimeFormat('en', { timeZoneName: 'short' }).formatToParts(new Date());
+    const tzAbbr = parts.find((p) => p.type === 'timeZoneName').value;
```
robyww, line 171: I think it might be better to just say `Local` or `Local Time` here.  I had to do a double take when looking at it.

robyww (approved):
It looks good. I suggested  a couple of small changes.

lrebull:
Looks really nice! it all seems to work as described.

lrebull:
FYI (and i think this is peripheral to your ticket?) i can't load spectrophotometry results -- i get this. Kartik metnioned the other day that this was fixed in his pr, but that doesn't help me for showing results. 
Screenshot 2025-10-16 at 9 35 33 AM

src/firefly/js/core/background/JobMonitor.jsx
loitly: This is not inside a FieldGroup.  What will the groupKey be in this case?
I don't like `useFieldGroupValue`.  Unless `getFieldVal` is removed or deprecated, I’ll keep using it.  It’s simple and straightforward.

src/firefly/js/core/background/JobMonitor.jsx
loitly: @lrebull I don’t have a strong preference here. What do you think?

lrebull: i would prefer that it told me what time zone it was working in. it's been the case that i've been sitting in eastern time but my laptop is on pacific time, for example.

PR #1866, opened 2025-10-17
Firefly-1835: Upgrade to react 19

robyww:
#### Firefly-1835: Upgrade to react 19

- react 19
- replace `react-sizeme` with newer JS API `ResizeObserver`
   - `react-sizeme` does not support react 19.
- remove unsupported `defaultProps`
- `propTypes` are still in the code but react now ignores them.
- removed uses for `forwardRef`
   - the ref is now passed as a prop without using `forwardRef`
- updated some other packages, as many as easily possible.
- updated supported browser list (no driver here, just an annual task)

#### Testing
- Everything should work as before
- https://firefly-1835-react19.irsakubedev.ipac.caltech.edu/irsaviewer
- https://fireflydev.ipac.caltech.edu/firefly-1835-react19/firefly


src/firefly/js/charts/ui/ChartPanel.jsx
```diff
@@ -2,8 +2,8 @@ import React, {useEffect} from 'react';
 import PropTypes from 'prop-types';
 import {Box, ChipDelete, Divider, Stack} from '@mui/joy';
 import {get, isEmpty, isUndefined} from 'lodash';
+import {wrapResizer} from '../../ui/NewResizer';
```
loitly: I suggest not using the word 'New' in the name, since it may lose relevance over time.  Since it exports a function rather than the component itself, a name like ResizeHelper (or something similar) would be more suitable.

loitly (approved):
Changes look good.  I spot-tested the applications, and everything seem to be working fine.

kpuriIpac (approved):
Tested the apps with a few basic searches and uploads. Everything looks good.

src/firefly/js/charts/ui/ChartPanel.jsx
robyww: I agree, I was using the name in an experiment. I meant to change it.

PR #1867, opened 2025-10-21
IRSA-7299: Update upload id for online help

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/IRSA-7299
**IFE PR**: https://github.com/IPAC-SW/irsa-ife/pull/445
- Updated the upload id from `catalogs.owncatalogs` to `upload.intro`
- This applies to the upload tab as well, and not just the upload dialogs 
  - @lrebull if there's a place where the help link should instead go to `catalogs.owncatalogs` when you test the apps below, please let me know
- **_Note_**: I suppose for [Firefly](https://fireflydev.ipac.caltech.edu/irsa-7299-help-fix-upload-dialog/firefly) we don't have an `upload.intro` for the online help, so maybe I should switch back to the default `basics.searching` for that? 

- Fixes [FIREFLY-1826](https://jira.ipac.caltech.edu/browse/FIREFLY-1826)
  - @loitly the bug fix for this is in DataType's `formatHeader` 
     - Bug summary: Essentially, if you have a spectrum table like this (from the ticket)
        - photometry_tbl
     -  notice the `det_id` column. When saving this table as an IPAC table, we convert `short` to `integer`, that part works as expected. But when we call `fitValueInto` inside `formatHeader`, the width for this column is returned as 6 (because of the `det_id` string length), and `integer` is truncated into `intege`, hence the bug.
     - This fix takes care of that, while still being generic. 
        - When the `width` is longer than than the `val length` (as is the usual case), for example, for the `ra` column (not in the screenshot), the width is 11 due to the entries in the column, so we still get a nicely formated `"ra         "` as expected in that case. 
       - But in a case like this specific column, it will make sure the data type value is not truncated. 

**Testing**: 
- [Firefly](https://fireflydev.ipac.caltech.edu/irsa-7299-help-fix-upload-dialog/firefly), [Irsaviewer](https://irsa-7299-help-fix-upload-dialog.irsakubedev.ipac.caltech.edu/irsaviewer), [DCE](https://irsa-7299-help-fix-upload-dialog.irsakubedev.ipac.caltech.edu/irsaviewer/dce), [Euclid](https://irsa-7299-help-fix-upload-dialog.irsakubedev.ipac.caltech.edu/applications/euclid), [Spherex](https://irsa-7299-help-fix-upload-dialog.irsakubedev.ipac.caltech.edu/applications/spherex)
- Just test the help link anywhere upload is used takes you to `/onlinehelp/#id=upload.intro`
  - includes both upload tabs and dialogs   
- For [FIREFLY-1826](https://jira.ipac.caltech.edu/browse/FIREFLY-1826), test according to ticket. Saving the table as an IPAC table should fix the bug in the `det_id` column. And if you try to re-upload this column to firefly, that should work as well. 
  - if Spherex searches take too long, I attached a euclid spectrum table to the ticket. I changed the `NDITH` column to be datatype `short` (instead of `integer`). So if you upload this file to firefly, and try to save the table as an IPAC table, you can test this bug fix that way (and to cross verify, if you upload it on the nightly firefly build: https://fireflydev.ipac.caltech.edu/firefly you'll  encounter the bug again)

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
+++ b/src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java
@@ -62,9 +62,6 @@ public class IpacTableWriter {
     }
 
     private static void save(PrintWriter out, DataGroup dataGroup) throws IOException {
-
-
-        shrinkToFitData(dataGroup);
         List<DataType> headers = Arrays.asList(dataGroup.getDataDefinitions());
         int totalRow = dataGroup.size();
 
@@ -126,11 +123,14 @@ public class IpacTableWriter {
                     dt.setTypeDesc(typeToDesc(dt.getDataType()));
                 });
 
+        //shrink data *after* headers have been modified, to get the correct width for each column
+        shrinkToFitData(dataGroup, modHeaders);
+
         // print column headers
         IpacTableUtil.writeHeader(out, modHeaders);
 
         for (int i = 0; i < totalRow; i++) {
-            IpacTableUtil.writeRow(out, headers, dataGroup.get(i));
+            IpacTableUtil.writeRow(out, modHeaders, dataGroup.get(i));
         }
         out.flush();
     }
@@ -138,8 +138,8 @@ public class IpacTableWriter {
     /**
      * this method will shrink the Column's width to fit the maximum's width of the data
      */
-    private static void shrinkToFitData(DataGroup dataGroup) {
-        for (DataType dt : dataGroup.getDataDefinitions()) {
+    private static void shrinkToFitData(DataGroup dataGroup, List<DataType> modHeaders) {
+        for (DataType dt : modHeaders) {
             String[] headers = {dt.getKeyName(), dt.getTypeDesc(), dt.getUnits(), dt.getNullString()};
             int maxWidth = Arrays.stream(headers).mapToInt(s -> s == null ? 0 : s.length()).max().getAsInt();
             for (int i=0; i<dataGroup.size(); i++) {
--- a/src/firefly/js/ui/FileUploadDropdown.jsx
+++ b/src/firefly/js/ui/FileUploadDropdown.jsx
@@ -27,7 +27,7 @@ export const FileUploadDropdown= ({sx, onCancel, onSubmit=resultSuccess, keepSta
                                       }) =>{
     const [submitText,setSubmitText]= useState('Load');
     const [doMask, changeMasking]= useState(() => false);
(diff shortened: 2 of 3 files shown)


lrebull:
Help anchor: this is definitely fixed for euclid and spherex, which were the ones that were causing the "trouble" with testers. Irsa viewer and dce also work great. so, i need to add the upload.intro anchor to firefly generic help -- that's definitely on me. it should NOT go to basics. 

"intege" type: I confirm this is fixed.

thank you!

kpuriIpac:
> Help anchor: this is definitely fixed for euclid and spherex, which were the ones that were causing the "trouble" with testers. Irsa viewer and dce also work great. so, i need to add the upload.intro anchor to firefly generic help -- that's definitely on me. it should NOT go to basics.
> 
> "intege" type: I confirm this is fixed.
> 
> thank you!

Ok so I will leave that as `upload.intro` for firefly as well, once that anchor is created it'll show up.

lrebull:
IN case anyone cares (probably not) ... I went through several minutes of being extra confused . . . i got into the firefly docs, and there sure as heck IS already an upload.intro anchor there.
Screenshot 2025-10-22 at 10 21 26 AM

and there is really an upload.html#intro destination for it too. BUT apparently I never actually put the toc_upload section into the fireflyToc. sigh. 

fixed now and checked back into git

src/firefly/java/edu/caltech/ipac/table/DataType.java
```diff
@@ -446,7 +446,7 @@ public String formatFixedWidth(Object value) {
      */
     public String formatHeader(String val) {
         val = val == null ? "" : val;
-        return fitValueInto(val, getWidth(), false);
+        return fitValueInto(val, Math.max(getWidth(), val.length()), false);
```
loitly: This fix doesn’t fully address the issue. The problem occurs where the headers are modified after shrinkToFitData is called.  [here](https://github.com/Caltech-IPAC/firefly/blob/fb9ea7052271175f9fd8f19336ce6581df6210d9/src/firefly/java/edu/caltech/ipac/table/io/IpacTableWriter.java#L67)

I suggest moving that function call to just before the writing step. You may also need to add a second parameter to the function so you can pass in the updated headers for the calculation.

loitly (approved):
Please fix the bad IPAC table export bug as I suggested before merging.

kpuriIpac:
> Please fix the bad IPAC table export bug as I suggested before merging.

@loitly I addressed the feedback and updated the test builds. Can you have a quick look before I merge?

loitly (approved):
IPAC table changes look good.  Thanks for the fix.

PR #1868, opened 2025-10-21
Firefly-1886: Improve extract tool behavior for MEFs

robyww:
#### [Firefly-1886](https://jira.ipac.caltech.edu/browse/FIREFLY-1886): Improve extract tool behavior for MEFs

 - if Aperture is set beyond 1x1, then evaluate the HDU name for has to combine

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1886-extract/firefly
- The details are in the design ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1885

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/ExtractFromImage.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/ExtractFromImage.java
@@ -27,7 +27,7 @@ import static edu.caltech.ipac.firefly.core.Util.Try;
         {
                 @ParamDoc(name="extractionType", desc="should be one of z-axis, line, or points"),
                 @ParamDoc(name="pt", desc="image point"),
-                @ParamDoc(name="wpt", desc="world point"),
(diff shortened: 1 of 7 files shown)


loitly (approved):
Code changes look good. However, I have a question:
Right now, the prompt for secondary combine appears when selecting “Pin Chart/Table.”
Would it make more sense to show this prompt when “Aperture” is selected and it’s not 1x1 instead?
This would let users preview the chart before pinning it and eliminate the need to select “Pin Chart/Table” twice.

robyww:
The reason is the check box would be is meaningless until you pin. The options (size and combine type) are using immediately. 

Also for line and point extract we already show a more complex dialog because there are many more extraction options. So it is part of the flow.

lrebull:
FYI, i am not seeing any of these changes on irsaviewer dev or test when using spherex data. should i be seeing these changes? i thought i had seen them at one point, but i am not currently seeing them on dev or test irsa viewer. do they only appear in spherex?

PR #1869, opened 2025-10-24
FIREFLY-1884: Fix Plot Parameter dialog reflecting incorrect state of plotted column

jaladh-singhal:
Fixes [FIREFLY-1884](https://jira.ipac.caltech.edu/browse/FIREFLY-1884)

Bug was happening because when alias is provided, it couldn't find that string in options list. Updated it to use the corresponding option value instead of original string.

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1884-charts-dialog-state-bug/firefly

Check that the issue described in ticket should no longer be reproducible.

### DIFF
--- a/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
+++ b/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
@@ -76,6 +76,16 @@ export function getUnitOptions(unit) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vandesai1:
I uploaded the file in the https://fireflydev.ipac.caltech.edu/firefly-1884-charts-dialog-state-bug/firefly and it now works as expected and desired!

vandesai1 (approved):

robyww (approved):

PR #1870, opened 2025-10-24
Firefly-1887: Support SPHEREx phtometry cutouts

robyww:
#### [Firefly-1887:](https://jira.ipac.caltech.edu/browse/FIREFLY-1887) Support SPHEREx photometry cutouts

  - Enhance ImageTableRowViewer, works with request promises, improved working, improved updates
  - Fixed a (bad) bug in managing worker thread. Sometimes it is deleted before finishing work resulting in render failure

#### Testing
  - https://firefly-1887-cutouts.irsakudev.ipac.caltech.edu/applications/spherex
  - use the SPHEREx Spectrophotometry tool
  - **Warning!** the datalink performance is very bad right now in dev. The backend team is working on it. This will make the performance seem bad the first time you access an image.

### DIFF
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -959,6 +959,22 @@ export function tokenSub(valObs, str='') {
 // }
 
 
+/**
+ * Call a function if the promise has not resolved after a period of time.
+ * The function wraps a promise. If the promise has not resolved in the waitUntilMS milliseconds then the
+ * whileWaitingFunc function is called. The passed promise is returned.
+ * @param {Promise} p - promise to return
+ * @param {Number} waitUntilMS - the timeout to wait to call the whileWaiting function
+ * @param {Function} whileWaitingFunc - if defined call with the passed promise
+ * @return {Promise<*>}
+ */
+export function callWhileWaitingToResolve(p, waitUntilMS, whileWaitingFunc) {
+    if (!p) return undefined;
+    const delay = (ms) => new Promise((resolve) => setTimeout(() => resolve('timeout'), ms));
+    Promise.race([p, delay(waitUntilMS) ])
+        .then( (winner) => winner==='timeout' && whileWaitingFunc?.(p) );
+    return p;
+}
 
 
 
--- a/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
+++ b/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
@@ -4,21 +4,25 @@
 
 import React, {useEffect, useRef, useState} from 'react';
 import {string, number, func, bool, object, oneOfType, node} from 'prop-types';
-import {difference, isEmpty, isNil, isString, xor} from 'lodash';
+import {debounce, difference, isEmpty, isNil, isString, xor} from 'lodash';
 import Slider from 'react-slick';
 import 'slick-carousel/slick/slick.css';
 import 'slick-carousel/slick/slick-theme.css';
 import {getComponentState} from '../../core/ComponentCntlr';
 import {CutoutButton, SHOWING_CUTOUT, SHOWING_FULL} from '../../ui/CutoutSizeDialog';
+import {IfWorkingMask} from '../../ui/panel/MaskPanel';
 
 import {useFieldGroupValue, useStoreConnector} from '../../ui/SimpleComponent.jsx';
 import {getTblById, isFullyLoaded} from '../../tables/TableUtil.js';
-import {getCutoutSize, getCutoutTargetOverride, getPreferCutout} from '../../ui/tap/Cutout';
+import {getPreferCutout} from '../../ui/tap/Cutout';
+import {callWhileWaitingToResolve} from '../../util/WebUtil';
+import {onPlotComplete} from '../PlotCompleteMonitor';
+import {UserZoomTypes} from '../ZoomUtil';
 import {MultiImageViewer} from './MultiImageViewer.jsx';
 import {dispatchReplaceViewerItems, getMultiViewRoot, getViewer, IMAGE, NewPlotMode,} from '../MultiViewCntlr.js';
 import {
     dispatchChangeActivePlotView, dispatchDeletePlotView,
-    dispatchPlotImage, dispatchRecenter, dispatchWcsMatch, visRoot, WcsMatchType
+    dispatchPlotImage, dispatchRecenter, dispatchWcsMatch, dispatchZoom, visRoot, WcsMatchType
 } from '../ImagePlotCntlr.js';
 import {SORT_ASC, SortInfo, UNSORTED} from '../../tables/SortInfo.js';
 import {CloseButton} from 'firefly/ui/CloseButton';
@@ -37,10 +41,15 @@ const MAX_IMAGE_CNT= 7;
 const IMAGE_CNT_KEY= 'imageCount';
 const IMAGE_TABLE_ERROR = 'Unable to load images because the required data table couldn\'t be retrieved.';
 
+const FORCE_RELOAD='FORCE_RELOAD';
+const TABLE_ROW='TABLE_ROW';
+const PAGE='PAGE';
+const ALL='_ALL_';
 
 export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutoutSizeAS, tbl_id, defaultWcsMatchType,
                                        defaultImageCnt= 5, imageExpandedMode, insideFlex=true, cutoutWpt,
                                         closeExpanded, maxImageCnt= MAX_IMAGE_CNT, tblErrorMsg=IMAGE_TABLE_ERROR}) {
+    const [requestPromiseMap, setRequestPromiseMap] = useState(new Map());
     const table= useStoreConnector(() => getTblById(tbl_id));
     const tblLoaded = useStoreConnector(() => isFullyLoaded(tbl_id));
     const imageCnt= useFieldGroupValue(IMAGE_CNT_KEY,viewerId)[0]() ?? defaultImageCnt;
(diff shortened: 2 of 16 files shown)


lrebull:
(you were SO not kidding about the performance.. yikes. and it's not getting faster even if i'm accessing the same images, like changing the cutout size.)

all the images come up for me ridiculously tiny. see screenshot. they should be at the 'zoom to fit' scale.
toosmall

when i change the cutout size, it changes which image is centered. this doesn't seem optimal. it should retain the same image highlighting and ordering as it had before i changed the cutout size.

how is it deciding on orientation? because it didn't put north-up, but it also didn't align with the tile orientation.

robyww:
@lrebull I have seen the small size if you click on the browser while it is loading the first time. It is having trouble computing the size in the case. I don’t think that will be a long term problem.

when you change the cutout size it recenters on the spectrum point. That seems like the right thing.

gpdf:
> how is it deciding on orientation? because it didn't put north-up, but it also didn't align with the tile orientation.

I haven't been able to look at it (no VPN capability with me) but I feel strongly that users will expect that all the cutouts be oriented the same way, so that as you move from one spectrometry point to the next the features on the sky are in the same places.  North-up would be the easiest to implement, I think?

lrebull:
they are all oriented the same way, but not north up. evidently. at least the first time i loaded it.

gpdf:
Maybe it's something like the "first" or "last" image in the spectrum is displayed in its "natural" row-column orientation, and then everything else is locked to it?

robyww:
They are suppose to be north up. Maybe the north up is getting unlock some way. I am seeing them north up.

lrebull:
it may be another instance of "the first time i used it, strange things happen", like the zoom not being right. because now the zoom is right every time i load it.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -84,8 +92,7 @@ export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutout
 
         if (hRowChangedByUI.current) {
             const midSlideIdx = table?.highlightedRow;
-            layoutImages(viewerId, Number(imageCnt), table, makeRequestFromRow, midSlideIdx, cutoutWpt);
-            changeActivePlot(viewerId, table, midSlideIdx, activePlotChangedByUI); //synchronise active plot as per highlighted row
```
jaladh-singhal, line 88: I don't see why you removed this?

I remember there were some cases in which changing the highlighted row from table view (not highlighting a new point from chart view) of spectrum, wasn't synchronizing the chart.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -110,13 +117,12 @@ export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutout
...
         else if((isNext && table.highlightedRow > midSlideIdx) || (!isNext && table.highlightedRow < midSlideIdx)) {
             // keep highlighted row/plot as it is but move slider, until it comes to the middle of slider
-            layoutImages(viewerId, Number(imageCnt), table, makeRequestFromRow, midSlideIdxAfterSliding, cutoutWpt);
+            void layoutImages(viewerId, Number(imageCnt), table, makeRequestFromRow, midSlideIdxAfterSliding, cutoutWpt, setReqPromise, PAGE);
             return true;
         }
         else { //table.highlightedRow === midSlideIdx; move highlighted row/plot as well as slider
             changeHighlightedRow(table, rowToHighlight, hRowChangedByUI);
-            layoutImages(viewerId,  Number(imageCnt), table, makeRequestFromRow, rowToHighlight, cutoutWpt);
-            changeActivePlot(viewerId, table, rowToHighlight, activePlotChangedByUI);
```
jaladh-singhal, line 119: same here

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -384,13 +408,38 @@ function layoutImages(viewerId, imageCnt, table, makeRequestFromRow, midSlideIdx
...
         dispatchReplaceViewerItems(viewerId, newPlotIdAry, IMAGE );
     }
 
     const {mpwWcsPrimId} = visRoot();
     const root= plotIdRoot(viewerId);
 
+
+
+    const finishedPromiseAry= exclusiveNewPlotIds.map( (plotId) => onPlotComplete(plotId));
+    await Promise.all(finishedPromiseAry);
```
jaladh-singhal: Why is it necessary to wait for all images in the slider view here and not let each image viewer handle it's loading state individually?

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -167,6 +176,24 @@ ImageTableRowViewer.propTypes= {
...
+}
+
+const WorkingAnimation= ({message}) => (
+        <Box sx={{ position:'absolute', left:0, right:0, top:34, bottom:20}}>
+            <Skeleton variant='rectangular' sx={{width:1, height:1}}/>
+            <Typography {...{level:'h4', width:1, position:'absolute', top:'3rem', textAlign:'center', color:'neutral'}}>
+                {message}
+            </Typography>
+        </Box>
+    );
```
jaladh-singhal: This is cool addition - esp the use of Suspense.

But I'm not convinced that this is necessary here (also see my previous comment). The image slider was already showing Skeleton box for each image making it clear that this image is being loaded (since toolbar already says "Image count").

If you still want to keep it, I'd advise not to do this single rectangle skeleton that spans across entire width of slider hiding the slider arrows! You can instead use `GridMask` (a Skeleton wrapper Loi added a while ago) to show imageCount number of skeleton boxes, and place it so that the arrows on the left/right edges of the image slider visible.

jaladh-singhal (changes requested):
This takes time to load as you mentioned but it works great! I tested different image counts and connections between image slider, spectrum table, and spectrum chart - they look good.

However, I noticed a UX issue with your changes to ImageTableRowViewer component. Steps to reproduce:
1. Do a spectrophotometry search, wait for image cutouts to load.
2. Click on the right slider arrow to slide the highlighted image/chart-point, it will come to the mid. Now clicking on right arrow further, will load a new image on right with image sliding animation. All good!
3. Now from the spectrum chart, click on a point that's farther away from the current images view, somewhat close to the right end of spectrum. Instead of seeing the image sliding animation and then 5 skeleton boxes, you see a single skeleton box with "searching images" which blocks the slider arrows underneath too.

I find it awkward - the loading UI should be consistent in all different clicks in chart or slider. We want to establish it to user that this is an image slider UI at all times.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
robyww: I removed it because it is is now done in ImageTableRow view after the `onPlotComplete` promse.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
robyww: There are race conditions that are not cleanly solvable. Sometime the active plot would be set to one associated with row zero. It is much, much cleaner to wait until the all load before setting the active plot or recentering.

jaladh-singhal: Ah - I see.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
robyww: see next comment

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
jaladh-singhal: You mean L438-439?

I suspected that they were doing this but wanted to confirm.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -384,13 +408,38 @@ function layoutImages(viewerId, imageCnt, table, makeRequestFromRow, midSlideIdx
...
+    if (loadType===FORCE_RELOAD) {
+        vr= visRoot();
+        const plotId= exclusiveNewPlotIds.find( (plotId) => getPlotViewById(vr, plotId).serverCall==='success');
+        dispatchChangeActivePlotView(plotId);
+        dispatchZoom({ plotId, userZoomType: UserZoomTypes.FIT, });
+        dispatchRecenter({plotId});
+    }
+    else {
+        const plotId= newPlotIdAry.find( (id) => id.endsWith(midSlideIdx+''));
+        if (plotId) dispatchChangeActivePlotView(plotId);
```
jaladh-singhal, line 450: Since it's a replacement of `changePlotView` at 2 places above, shouldn't we call the `changePlotView` here which wraps `dispatchChangeActivePlotView` and handles some edge cases. IIRC `activePlotChangedByUI` ref must be set correctly to handle random order clicks, otherwise image slider and spectrum interaction suffers in some corner cases.

robyww: Keeping the ref might no longer be necessary. since the race conditions are fix by using the promise array. I will test trying to take it out altogether.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
robyww: It is very much needed with the data link taking so long. I think it will be necessary even if they get the response faster. Otherwise the whole area appears to hang. That all happens before the `dispatchPlotImage` which is when you see the image working feedback. This is during the `makeRequestFromRow` call. Notice.`makeRequestFromRow` now returns a promise to a request since it involves a server call.

I will try `GridMask`. However, I don't think there is a good way a the above level to mask part of the area.

robyww: Update: I think I can us the same approach inside the `ImageSlider` component. So I think this will address you concern.

src/firefly/js/util/WebUtil.js
```diff
@@ -959,6 +959,22 @@ export function tokenSub(valObs, str='') {
...
+/**
+ * Call a function if the promise has not resolved after a period of time.
+ * The function wraps a promise. If the promise has not resolved in the waitUntilMS milliseconds then the
+ * whileWaitingFunc function is called. The passed promise is returned.
+ * @param {Promise} p - promise to return
+ * @param {Number} waitUntilMS - the timeout to wait to call the whileWaiting function
+ * @param {Function} whileWaitingFunc - if defined call with the passed promise
+ * @return {Promise<*>}
+ */
+export function callWhileWaitingToResolve(p, waitUntilMS, whileWaitingFunc) {
```
jaladh-singhal, line 971: Thanks for generalizing this - I can see its use at other places with race conditions.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -143,12 +151,16 @@ export function ImageTableRowViewer({viewerId, makeRequestFromRow, defaultCutout
...
-        />
+        <div style={{width:'100%', height:'100%', display:'flex', position: 'relative'}}>
+            <MultiImageViewer
+                {...{viewerId, Toolbar, insideFlex, defaultImageCnt, maxImageCnt, tableId:tbl_id,
+                    makeRequestFromRow,defaultCutoutSizeAS, defaultWcsMatchType, cutoutWpt,
+                    wcsMatchType, activePlotId, makeCustomLayout, mouseReadoutEmbedded: false,
+                    forceRowSize:1, canReceiveNewPlots: NewPlotMode.create_replace.key}}
+            />
+            <IfWorkingMask {...{promise:requestPromiseMap.get(ALL), message:'Searching Images',
+                           gridSize:{rows:1, cols:Number(imageCnt)}, sx:{top:34, bottom:20} }}/>
```
jaladh-singhal, line 173: I was about to say to add `left: 25, right: 25` to `sx` so that slider arrows don't hide. But then I tried through dev tools in browser, and saw that the slider arrows get visible but clicking on them does nothing because it is still searching all 5 images in the new section of spectrum. So it's deliberate to block/hide slider arrows in such cases?

I think this makes sense to me (also what you were saying yesterday) that the images first need to be searched through datalinks (grey skeleton) before retrieving and rendering them (yellow box with info). We might need to document it if users complain about image slider arrows appear/disappearing that it's deliberate (it's just me for now).

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
jaladh-singhal: Thanks for removing it. [For now,] I couldn't reproduce out-of-sync issues by clicking on images (as opposed to through slider arrows) so we are doing fine without this ref.

jaladh-singhal (approved):
Thanks for the updates - it looks better than before esp the individual "searching image"!

The image searching to retrieving/rendering lag is making click interactions weird in some cases (like the image selection doesn't shift on clicking slider arrow until it starts retrieving the image). But I hope it will get better once backend team makes datalinks faster and can be revisited then.

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
robyww: I did decide to keep the arrows hidden. I thought it would just make the flow less complex.

robyww:
@jaladh-singhal I will merge it soon. I am working another issue. I don't think `react-slick` can keep up with the calls to `slickGoTo` so I am going to `debounce` it. it happens if I click 4 or 5 rows on the table very fast. `react-slick` ends up in the wrong place.

jaladh-singhal:
@robyww go ahead - it's good to merge from my end (it's a complexly async component so I won't be surprised if we find more issues later but the bulk of it is working fine!)

PR #1798, opened 2025-06-13
FIREFLY-1694,1105,1784: Improve unit recognition of SpectrumDM

jaladh-singhal:
Fixes [FIREFLY-1694](https://jira.ipac.caltech.edu/browse/FIREFLY-1694), [FIREFLY-1105](https://jira.ipac.caltech.edu/browse/FIREFLY-1105), [FIREFLY-1784](https://jira.ipac.caltech.edu/browse/FIREFLY-1784)

Following task-list is derived from the text and comments in above tickets (+ conversations about this work in meetings):

- [x] Make firefly recognize flux units in `erg/s/cm^2` and convert to equivalent units: `W/m^2/Hz <-> Jy <-> erg/s/cm^2/Hz`. `erg/s/cm^2/A` can't be converted to flux density in frequency space so added one more wavelength space unit `W/m^2/um`.
- [x] Show units in options dialog even when firefly can't recognize them: instead of dropdown of convertible units, this will be a readonly ~field with a tip~ text.
- [x] Make firefly recognize alias of same unit: `Angstrom = angstrom = A`, `erg/s/cm^2/Angstrom = erg/s/cm2/Angstrom = erg/s/cm**2/Angstrom = erg.s**-1.cm**-2.Angstrom**-1`, etc.
- [x] Make firefly display the recognized units on chart axes in shorter latex syntax i.e., if it's recognized by firefly, it will match the dropdown label + other label addition that firefly makes like redshift correction
- [x] Make firefly recognize units prefixed with a scalar factor, also convert them so it will be a new entry in dropdown
- [x] Fix bug in axis label doesn't update with correct units when spectrum is single trace
- [x] Change disabled input fields to just simple text
- [x] Add lamda, F_lambda symbols that appear in chart to dialog too indicate they map to column name
- [x] Add unit tests
- [x] wording changes in spectra file upload

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1694-1105/firefly/

Upload test files (attached in the 1694 ticket), first on irsaviewer ops and then on this build. Open chart dialogs for each file and notice how they differ between irsaviewer and this build. Change unit dropdowns (and redshift correction ones), it should update the column field with correct formula and chart axes label with correct label (once you hit apply)

PR #1872, opened 2025-10-28
FIREFLY-1851: Fix axes labels not getting updated when columns to plot change

jaladh-singhal:
Fixes [FIREFLY-1851](https://jira.ipac.caltech.edu/browse/FIREFLY-1851)

> This was a regression issue that I introduced in https://github.com/Caltech-IPAC/firefly/pull/1798 but fortunately left a comment for my future self then (present self now) :)

Reverted to unset axis label fields in "chart options" so that when chart is updated the title gets generated (via existing code):
https://github.com/Caltech-IPAC/firefly/blob/441fb8995257795a82c2ce7b8e8af52fe9dd92ca/src/firefly/js/charts/dataTypes/FireflyGenericData.js#L174-L182

(Also true for [histogram](https://github.com/Caltech-IPAC/firefly/blob/5325b17db8edaf10a30d31dc8713644b21e0a579/src/firefly/js/charts/dataTypes/FireflyHistogram.js#L123-L132), [heatmap](https://github.com/Caltech-IPAC/firefly/blob/5325b17db8edaf10a30d31dc8713644b21e0a579/src/firefly/js/charts/dataTypes/FireflyHeatmap.js#L223-L230))

## Testing
https://fireflydev.ipac.caltech.edu/firefly-1851-chart-axes-labels-bug/firefly

- Check the examples mentioned in ticket (scatter type), should be fixed
- Increase radius and check the heatmap instance for above too
- Click "Add chart" and make a histogram, check that changing column and applying causes chart to update
- Upload a spectrum and check it is working as before

### DIFF
--- a/src/firefly/js/charts/ui/options/BasicOptions.jsx
+++ b/src/firefly/js/charts/ui/options/BasicOptions.jsx
@@ -21,7 +21,7 @@ import {useStoreConnector} from '../../../ui/SimpleComponent.jsx';
 import {updateSet} from '../../../util/WebUtil.js';
 import {hideColSelectPopup} from '../ColSelectView.jsx';
 import {addColorbarChanges} from '../../dataTypes/FireflyHeatmap.js';
-import {colorsOnTypes, getChartProps, toRGBA, TRACE_COLORS, uniqueChartId} from '../../ChartUtil.js';
+import {colorsOnTypes, getChartProps, toRGBA, TRACE_COLORS, uniqueChartId, isSpectrum} from '../../ChartUtil.js';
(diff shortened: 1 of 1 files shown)


lrebull:
It works great! Thank you!

loitly (approved):
The code looks good. I tested everything listed in the PR, and it all worked as described.
However, I noticed an issue when choosing “Overplot new Trace” instead of “Add new Chart.” 
In that case, the behavior doesn’t make sense. It seems the label only updates once based on the initial chart type and doesn’t change afterward. 
Switching between traces doesn’t fix it. I don’t remember if this is how it used to behave.

PR #1873, opened 2025-10-30
Firefly-1889: update table row viewer with improved masking

robyww:
#### [Firefly-1889](https://jira.ipac.caltech.edu/browse/FIREFLY-1889): update table row viewer with improved masking

- fix issues with TableRowViewer masking that was not right in my last PR
- Create a more `PlotView` like look during working
- refactor working task management to work better with promises
  - task are added with promises that are removed when the promises fulfills. The is not longer a remove task call.
  - removed code that is now unnecessary

#### Testing
- https://firefly-1889-masking.irsakubedev.ipac.caltech.edu/applications/spherex
- Every should work with better `PlotView` like masking when sliding.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/util/download/ConcurrentDownload.java
+++ b/src/firefly/java/edu/caltech/ipac/util/download/ConcurrentDownload.java
@@ -152,7 +152,7 @@ public class ConcurrentDownload {
 
     private static int getWorkerCnt(long contentLen) {
         long lenMeg= contentLen/FileUtil.MEG;
-        if (lenMeg<90) return 1;
+        if (lenMeg<50) return 1;
         if (lenMeg<200) return 2;
         if (lenMeg<400) return 3;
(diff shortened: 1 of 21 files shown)


kpuriIpac:
I looked at the old PR and this PR. It's of course a complex component, code would probably be better reviewed by @jaladh-singhal when he's back (not blocking for this PR, I just mean in a future iteration of work with this code). 

I think the masking is working better, but I did notice a few small issues: 

searching

This looks. fine as `Searching Images` is centered here. But after these images load, and you slide right by a few images really quick, you may get something like: 



I am not suggesting this necessarily needs fixing, but just wanted to highlight it (the fact that only the image on the far right says `Searching Image`)

Looks good other than that!

kpuriIpac (approved):
Looks good! Spent time trying to study code in the old PR and this one. Might need a deeper review eventually as well, but looks good to me.

PR #1875, opened 2025-11-04
FIREFLY-1880: Enable uploads and pinning spectra in spherex

kpuriIpac:
**Tickets**: [FIREFLY-1880](https://jira.ipac.caltech.edu/browse/FIREFLY-1880) , [FIREFLY-1878](https://jira.ipac.caltech.edu/browse/FIREFLY-1878)

**IFE PR**: https://github.com/IPAC-SW/irsa-ife/pull/447

- Passing in viewerId to `dispatchPlotImage` is the crucial fix here to ensure the spectrum cutouts don't show up in the `Pinned Images` tab as well when you upload new images. 
  - Because uploads uses our default viewerId  (`DEFAULT_FITS_VIEWER_ID`), and if we don't pass in the custom `viewerId` to `dispatchPlotImage` for `ImageTableRowViewer`, it'll also default to the default viewerId.

### DIFF
--- a/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
+++ b/src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
@@ -408,6 +408,7 @@ async function doLayoutImages({viewerId, imageCnt, table, makeRequestFromRow,
             wpRequest.setRotateNorth(true);
             dispatchPlotImage({
                 plotId, wpRequest,
+                viewerId,
                 setNewPlotAsActive: false,
                 holdWcsMatch: true,
                 pvOptions: {userCanDeletePlots: false}
--- a/src/firefly/js/visualize/ui/TriViewImageSection.jsx
+++ b/src/firefly/js/visualize/ui/TriViewImageSection.jsx
@@ -88,7 +88,7 @@ export const makeMultiProductViewerTab= ({id,dataProductTableId}) => {
 };
 
 
-function BadgeLabel({labelStr}) {
+export function ImageBadgeLabel({labelStr}) {
     const badgeCnt= useStoreConnector(() => getViewerItemIds(getMultiViewRoot(),DEFAULT_FITS_VIEWER_ID)?.length??0);
     return badgeCnt===0 ?  labelStr:
         (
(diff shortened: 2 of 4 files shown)


src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
```diff
@@ -408,6 +408,7 @@ async function doLayoutImages({viewerId, imageCnt, table, makeRequestFromRow,
             wpRequest.setRotateNorth(true);
             dispatchPlotImage({
                 plotId, wpRequest,
+                viewerId,
```
robyww, line 411: I guess it is a good think you spent some time with my ImageTableRowViewer PR! good catch!

src/firefly/js/visualize/ui/TriViewImageSection.jsx
```diff
@@ -86,7 +86,7 @@ export const makeMultiProductViewerTab= ({id,dataProductTableId}) => {
 };
 
 
-function BadgeLabel({labelStr}) {
+export function BadgeLabel({labelStr}) {
```
robyww: since you are now exporting this component why don't you change the name to something more specific. Such as `ImagePinnedBadgeLabel`

robyww (approved):

src/firefly/js/visualize/ui/ImageTableRowViewer.jsx
kpuriIpac: yeah this is what was causing the spectrum cutouts showing up in the `Pinned Images` tab as well! spent longer than I thought I would trying to catch this hah.

src/firefly/js/visualize/ui/TriViewImageSection.jsx
kpuriIpac: You're right. I had initially changed it to `ImageBadgeLabel` and then changed it back to `BadgeLabel` but then just imported it as `ImageBadgeLabel` (because there's already a `BadgeLabel` for charts as well). I think I'll change the Chart one's name to `ChartBadgeLabel` as well.

PR #1877, opened 2025-11-05
Firefly-1891: add NED or Simbad type to resolution display

robyww:
#### [Firefly-1891](https://jira.ipac.caltech.edu/browse/FIREFLY-1891):  add NED or Simbad type to resolution display

 - Also: firefly a little smarter about default tri-view tab, looks for catalog with service descriptors

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1891-resolve/firefly
- do a name resolution in any target field, you should see the type from NED or Simbad

### DIFF
--- a/src/firefly/js/ui/PositionFieldDef.js
+++ b/src/firefly/js/ui/PositionFieldDef.js
@@ -70,7 +70,9 @@ export function coordToString(csys) {
 export function formatTargetForHelp(wp) {
     if (!wp) return '';
     if (!wp.objName || !wp.resolver) return formatPosForHelp(wp);
-    return `<div style="font-size:14px;"><b>${wp.objName}</b> <span style="font-size:10px;">resolved by</span> ${wp.resolver.desc}</span>
+    const objTypeHtml= wp.objType ? `<span style="font-size:10px;">, type: ${wp.objType}</span>` : '';
+    return `<div style="font-size:14px;"><b>${wp.objName}</b> <span style="font-size:10px;">resolved by</span> 
+             ${wp.resolver.desc}${objTypeHtml}
             <div>
               ${formatPosForHelp(wp)}
             </div>`;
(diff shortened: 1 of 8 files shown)


lrebull:
this is really nice! I was trying a variety of names to see if i could get the same results as a direct simbad query. it was working well until I tried "ZTF J213903.90+573103.6" 
this one should get resolved as "TT*" or similar, but it doesn't -- see attached screenshots. I have noticed that it doesn't spit back *exactly* what Simbad has even for other sources that it is reporting a type correctly, so this may be a "feature not a bug" but I thought I would mention it since I found it by stabbing more or less randomly at about 10-12 sources, implying that maybe ~10% of sources are affected??

simbad
firefly

kpuriIpac:
Looks good. 

One minor observation: When you enter a target (let's say m101 on TAP) you'll see `m101 resolved by NED, type: G`, but if you now move to `Images` or `IRSA Catalogs` tab, you'll now see `m101 resolved by NED` unless you re-enter the target or search for a new target (in that case you'll see the `type` again).

src/firefly/js/ui/PositionFieldDef.js
```diff
@@ -70,7 +70,9 @@ export function coordToString(csys) {
 export function formatTargetForHelp(wp) {
     if (!wp) return '';
     if (!wp.objName || !wp.resolver) return formatPosForHelp(wp);
-    return `<div style="font-size:14px;"><b>${wp.objName}</b> <span style="font-size:10px;">resolved by</span> ${wp.resolver.desc}</span>
+    const objTypeHtml= wp.objType ? `<span style="font-size:10px;">, type: ${wp.objType}</span>` : '';
```
kpuriIpac, line 73: I'm guessing it has something to do with `wp.objType` initially being empty when you switch to a new tab unless you enter or select a new target.

robyww: yes, something is wrong. I will look into it.

robyww: probably i missed something in caching

PR #1878, opened 2025-11-06
FIREFLY-1835: hotfix for Jest

tgoldina:
Hotfix for React 19 PR that fixes JS unit tests (`gradle :firefly:jsTest`).

Jest now requires `transformIgnorePatterns` for packages like slug, chroma-js, and lodash-es because the newer versions of these packages now have "type": "module" in their package.json and use ES module syntax (import/export). 

By default, Jest ignores all files in node_modules and doesn't transform them, but Jest cannot run ES module code without transformation.


robyww (approved):
Looks good. I know almost nothing about jest except how to mock out stuff. (see `initTest.js`). I appreciate you looking into this and solving the issue.

PR #1883, opened 2025-11-11
Firefly-1867: Allow user to set the nan pix.

robyww:
#### [Firefly-1867](https://jira.ipac.caltech.edu/browse/FIREFLY-1867): Allow user to set the nan pix. 
 - allow use to change NaN pixel
 - add support for reversing color bars
 - includes some refactoring because of long parameter list, additional javadocs, and clean up of `ColorPicker.jsx`

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1867-nan-pix/firefly
- load in a image with some NaN pixels. sample fits file: (unzip and upload): [sample.fits.zip](https://github.com/user-attachments/files/23483559/sample.fits.zip)
- use color drop down to control or see the nan-pixel color
- use color drop down to try reverse color bars


src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
```diff
@@ -23,62 +28,46 @@ import {RangeSliderView} from '../../ui/RangeSliderView.jsx';
...
-    { id: 19, icon: ColorTable19, tip: 'Standard (ds9)' },
-    { id: 20, icon: ColorTable20, tip: 'Staircase (ds9)' },
-    { id: 21, icon: ColorTable21, tip: 'Color (ds9)' }
-];
+    { id: '0',  tip:'Gray Scale', disableReverse:true },
+    { id: '1',  tip: 'Reverse Gray Scale', disableReverse:true  },
+    { id: '2',  tip: 'Color Cube' },
+    { id: '3',  tip: 'Spectrum' },
+    { id: '4',  tip: 'For False Color' },
+    { id: '5',  tip: 'For False Color - Reversed', ignore:true },
```
jaladh-singhal: this has a side effect on color bar slider - slide from 3 to 4, check mark moves to next colorbar. now slide from 4 to 5, check mark disappears, now slide from 5 to 6 check mark reappears and moves to next colorbar. The ignored color bar should change indexing used by slider or just delete it if we don't need it.

src/firefly/js/visualize/ImagePlotCntlr.js
```diff
@@ -445,9 +446,10 @@ export function dispatchChangeImageVisibility({plotId, imageOverlayId, visible,
  * @memberof firefly.action
  */
 export function dispatchColorChange({plotId, cbarId, bias, contrast,
-                                        useRed=true, useGreen=true, useBlue=true,
+                                        useRed=true, useGreen=true, useBlue=true, nanPixelColor= [255,0,0],
```
jaladh-singhal: shouldn't it be `nanPixelColor= [0,0,0]` like elsewhere?

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
```diff
@@ -221,40 +223,61 @@ const AdvancedColorPanel= ({allowPopout}) => {
...
+                                   text={ct.icon ? undefined : 'Default Color Map'}
+                                   CheckboxOnIcon={reversed ? <CheckBoxIcon/> : <CheckIcon/>}
+                                   horizontal={false}
+                                   hasCheckBox={true} checkBoxOn={checkBoxOn}
+                                   slotProps={{tbCheckBox: {sx:{height:20, width:22}}}}
+                                   onClick={() => changeBiasContrastColor(ct.id, bias, contrast)}/>
+                    {doReverse && <ReverseButton
+                        tip={ct.tip + '- reversed'}
+                        slotProps={{iconButton: {sx : {p:0, border:0}}}}
+                        onClick={() => changeBiasContrastColor(getRevereId(ct.id), bias, contrast)}/>}
```
jaladh-singhal: As a user, I didn't realize these were two buttons!

Because I clicked on "ReverseButton" to reverse the colorbar, I kept pressing the "ReverseButton" again and again to un-reverse it but nothing happened. Then I clicked on "CheckBoxIcon" which changed to "CheckIcon" which made me feel that is the toggle (but in fact I clicked on "ToolbarButton" which is showing the color bar image as I can tell from code). Then clicking on "CheckIcon" again doesn't reverse the colorbar, because I have to go click on "ReverseButton". 

I think it is common for user to expect one button to control the toggle between reversed and default colorba here. So to make UX less confusing, how about we:
- use a constant CheckboxOnIcon, just `CheckIcon` like it is on dev - which shows the state#1 that "this colorbar is selected"
- make the `ReverseButton` a switch with "Reversed/Default" labels ([like here](https://mui.com/joy-ui/react-switch/#label)) - which will show state#2 that "the selected colorbar is either default or reversed". This switch should only appear conditionally when a colorbar is selected (this will also remove the need to repeat the reverse icon from top to bottom)

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
```diff
@@ -342,14 +365,68 @@ const AdvancedColorPanel= ({allowPopout}) => {
...
+            pickerType:'compact',
+            cb:(ev) => {
+                const {r, g, b} = ev.rgb;
+                setNanColorLocked(true);
+                changeColor([r,g,b]);
+            },
+        });
+    };
+    return (
+        <ToolbarButton {...{
```
jaladh-singhal, line 495: This looks pretty sqished with dividers, consider adding `sx: {my: 0.5}` to give it some room.

From css dev tools hacking: 

Image

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
```diff
@@ -342,14 +365,68 @@ const AdvancedColorPanel= ({allowPopout}) => {
...
+            tip:'set NaN color',
+            slotProps:{tbCheckBox:{
+                sx: { backgroundColor: colorStr, width: 17.5, height: 20, borderRadius: 3.5}
+            }},
+            text:`NaN Pixel Color ${nanColorLocked?' (Locked)': ' (Auto computed)'}`,
+            hasCheckBox:true, checkBoxOn:defaultNanColorLocked, CheckboxOnIcon: checkIcon,
+            onClick:() => {
```
jaladh-singhal, line 502: in the dropdown mode of `AdvancedColorPanel`, as opposed to the popped out mode, it becomes bit confusing to understand what happened since the dropdown disappears and a Nan color picker appears. I know it's by design but I had to switch to popped out mode to understand how to use Nan color lock for the first time.

I wonder if we can disable this color panel dropdown from hiding until user goes to Nan color picker and makes a choice? Not a dealbreaker though

jaladh-singhal (changes requested):
Nan color locking is working as expected! I found some usability problems with colorbar reversing though - see my comments.

Rest all looks good and improved!

docs/next-release-details.md
```diff
@@ -1,15 +1,47 @@
 # Notes for next Release
 
 ## Version 2025.5
-- 2025.5.0 - (tentative planed Nov 15, 2025),  _docker tag_: `nightly`
+- 2025.5.0 - (tentative planed Nov 17, 2025),  _docker tag_: `nightly`
 
-## Plans: S3, and more
+#### This release includes Job Monitor cleanup, S3 support, new FITS projection support and more color support
```
jaladh-singhal: ...and lot of chart/spectra fixes!

src/firefly/js/visualize/ImagePlotCntlr.js
robyww: i will fix

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
robyww: yea that is a problem. I am probably going to have change the slider to show all 48 color bar both color and reversed.

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
robyww: I will look into this. It will be tricky to implement right and might have other problems. It might be better to show the forward and backward in horizontal pairs. 

I will take your feedback and play with more ideas.

docs/next-release-details.md
robyww: I will add that as well

robyww:
@jaladh-singhal I just did an update with work on the UI. I took some of you suggestions but there were some that I think would come with their own set of UI issues.

I made the reverse button always reverse the color map
  - it reverses the colors that the user sees
  - it shows na "is reversed" state.
  - the checkbox no longer changes.

> make the ReverseButton a switch with "Reversed/Default" labels ([like here](https://mui.com/joy-ui/react-switch/#label)) - which will show state#2 that "the selected colorbar is either default or reversed". This switch should only appear conditionally when a colorbar is selected (this will also remove the need to repeat the reverse icon from top to bottom)


A reverse button that only shows when the users mouse is there would be confusing for several reason that we can talk about if you want.

> I wonder if we can disable this color panel dropdown from hiding until user goes to Nan color picker and makes a choice?

I think it is better for it to hide. I don't like to do that unless absolutely necessary. I think once a user does it one of twice it will be better than having to close the dropdown in some way.

jonesmg:
I just tried this out. I could set the NaN color easily and this worked as expected. This is a big improvement to the display options.

I also reversed the color scale and found the press in, then release, toggle buttons intuitive to use, with the first action flipping the color scale, and the second flipping it back. 

One thing I found a little odd was that the default gray scale (and its reverse) are two different colorbars, rather than having a reverse toggle. I see that it is also like this on ops, and I can imagine an argument for this, i.e., these are the two color scales that people are most likely to want and by having them as two separate color bars, the user doesn’t even need to realize that you can reverse the color scales. However, it was nevertheless a little odd to me that there were two. Not a big deal though.

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
```diff
@@ -221,40 +282,62 @@ const AdvancedColorPanel= ({allowPopout}) => {
...
+    const r= isReversedColor(colorTableId+'');
+    const allColorTableIDs=[];
+    ctArray.forEach(({id,disableReverse,ignore}) => {
+        if (ignore) return;
+        allColorTableIDs.push(id);
+        if (disableReverse) return;
+        allColorTableIDs.push(id+REVERSED_END_CHAR);
+    });
+    const cSlideVal= allColorTableIDs.findIndex( (v) => v===colorTableId);
+    const cBarTitle= (ctArray.find( (ct) => ct.id===colorTableId || ct.id+REVERSED_END_CHAR===colorTableId)?.tip ?? '') + (r ? '- reversed' : '');
```
jaladh-singhal:
```suggestion
    const cBarTitle= (ctArray.find( (ct) => ct.id===colorTableId || ct.id+REVERSED_END_CHAR===colorTableId)?.tip ?? '') + (r ? ' - reversed' : '');
```
Add space before `-` too

jaladh-singhal (approved):
Thanks for updating the reversing UI - it's much cleaner now. I like that now:
- reverse icon acts like a push button (similar to switch) and the checkmark icon doesn't change
- the colorbar slider is connected to reversed colorbas too - solving the indexing issue I mentioned before for "false color" colorbar
- the colorbar flips when reverse button is pushed! (kinda felt like an easter egg :))

src/firefly/js/visualize/ui/ColorTableDropDownView.jsx
robyww: thanks for spotting that

PR #1884, opened 2025-11-14
FIREFLY-1837: Add an option to save charts as JSON 

kpuriIpac:
**Ticket**: [FIREFLY-1837](https://jira.ipac.caltech.edu/browse/FIREFLY-1837)
- Add an option to save charts as JSON (in addition to existing PNG option)  

**Testing**: 
- Firefly: https://fireflydev.ipac.caltech.edu/firefly-1837-chart-save/firefly
- Do any search with at least a few hundred or thousand rows 
- Save the chart as JSON format and look at the JSON file
- Overplot new traces of either heatmap/histogram and then save the chart again, you should see this information in the JSON file 
- To visually test the JSON with a live plotly example (to verify the chart looks the same as it does on firefly), you can use this `codepen` example I have setup: https://codepen.io/kartikeya-puri/pen/bNEPNRR
  - just drop in the entire json file into the `fig` variable in the JS window, and you should then see the chart
- Regression testing: Save the charts as PNG and make sure that works like before

### DIFF
--- a/src/firefly/js/charts/ui/PlotlySaveDialog.jsx
+++ b/src/firefly/js/charts/ui/PlotlySaveDialog.jsx
@@ -2,27 +2,27 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import React from 'react';
+import React, {useEffect} from 'react';
 import {Box, Button, Stack} from '@mui/joy';
 import {getWorkspaceConfig} from 'firefly/visualize/WorkspaceCntlr.js';
 import {PopupPanel} from 'firefly/ui/PopupPanel.jsx';
 import DialogRootContainer from 'firefly/ui/DialogRootContainer.jsx';
 import {dispatchHideDialog, dispatchShowDialog} from 'firefly/core/ComponentCntlr.js';
-import {isString} from 'lodash';
+import {isArray, isString} from 'lodash';
 import CompleteButton from 'firefly/ui/CompleteButton.jsx';
 import HelpIcon from 'firefly/ui/HelpIcon.jsx';
 import {FieldGroup} from 'firefly/ui/FieldGroup.jsx';
 import {ValidationField} from 'firefly/ui/ValidationField.jsx';
 import {downloadBlob, makeDefaultDownloadFileName} from 'firefly/util/fetch.js';
+import {RadioGroupInputField} from 'firefly/ui/RadioGroupInputField';
+import {useFieldGroupValue} from 'firefly/ui/SimpleComponent';
+import {isDefined} from 'firefly/util/WebUtil';
 
-
-const DIALOG_ID= 'plotDownloadDialog';
+const DIALOG_ID = 'plotDownloadDialog';
 
 export function showPlotLySaveDialog(Plotly, chartDiv) {
-    // if (fileLocation === WORKSPACE) dispatchWorkspaceUpdate(); //todo keep if we add workspace support, it probably goes somewhere else
-
-    const isWs = getWorkspaceConfig(); //todo - keep if we add workspace support
-    const  popup = (
+    const isWs = getWorkspaceConfig();
+    const popup = (
         <PopupPanel title={'Save Chart'}>
             <Box sx={{
                 minWidth: '32rem',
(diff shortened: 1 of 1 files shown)


src/firefly/js/charts/ui/PlotlySaveDialog.jsx
```diff
@@ -41,47 +41,166 @@ export function showPlotLySaveDialog(Plotly, chartDiv) {
...
+    return name?.toLowerCase?.().endsWith(ext) ? name : `${name}${ext}`;
 }
 
+/**
+ * Get a plain Plotly figure: {data, layout, frames, config}
+ * @param Plotly
+ * @param chartDiv
+ */
+function getFigure(Plotly, chartDiv) {
+    if (chartDiv && (Array.isArray(chartDiv.data) || chartDiv.layout)) {
```
robyww: here an other places, we generally use lodash `isArray`

src/firefly/js/charts/ui/PlotlySaveDialog.jsx
```diff
@@ -41,47 +41,166 @@ export function showPlotLySaveDialog(Plotly, chartDiv) {
...
+    }
+
+    //whitelist data keys per trace type + shared attrs common to all trace types
+    const cleanData = (Array.isArray(fig?.data) ? fig.data : []).map((trace) => {
+        const t = trace?.type || 'scatter';
+        const traceAttrs = getTraceAttrs(schema, t);
+        const out = {type: t};
+        for (const [k, v] of Object.entries(trace || {})) {
+            if (k === 'type') continue;
+            if ((traceAttrs && k in traceAttrs) || (allAttrs && k in allAttrs)) out[k] = v;
```
robyww: i think is would be easier to read if you used defined. such as
```js
 if (isDefined( traceAttrs?.[k]) || isDefined( allAttrs?.[k]))
```

robyww (approved):
This looks good. I did not know you could get a schema our of plot.ly. That is pretty cool.

src/firefly/js/charts/ui/PlotlySaveDialog.jsx
```diff
@@ -41,47 +41,166 @@ export function showPlotLySaveDialog(Plotly, chartDiv) {
 }
 
 function generateTitleFromAxis(layout) {
-    const xaxis= layout.xaxis?.title?.text;
-    const yaxis= layout.yaxis?.title?.text;
-    return (isString(xaxis) || isString(yaxis)) ? `${isString(xaxis)?xaxis:''} ${isString(yaxis)?yaxis:''}` : undefined;
+    const xaxis = layout?.xaxis?.title?.text;
+    const yaxis = layout?.yaxis?.title?.text;
+    return (isString(xaxis) || isString(yaxis)) ? `${isString(xaxis) ? xaxis : ''} ${isString(yaxis) ? yaxis : ''}` : undefined;
```
loitly: This is hard to read and can produce leading or trailing spaces.
Since `title.text` is either nullish or a string, you can do this:
```js
return [xaxis, yaxis].filter(Boolean).join(' ');
```

src/firefly/js/charts/ui/PlotlySaveDialog.jsx
```diff
@@ -41,47 +41,166 @@ export function showPlotLySaveDialog(Plotly, chartDiv) {
...
+    if (chartDiv && (Array.isArray(chartDiv.data) || chartDiv.layout)) {
+        return {
+            data: Array.isArray(chartDiv.data) ? chartDiv.data : [],
+            layout: chartDiv.layout ?? {},
+            frames: chartDiv.frames ?? [],
+            config: chartDiv.config ?? {}
+        };
+    }
+    //else, nothing found
+    return { data: [], layout: {}, frames: [], config: {} };
```
loitly: Same here.  The `if` condition only make it confusing.  I think you really just want:
```js
        return {
            data: chartDiv?.data ?? [],
            layout: chartDiv?.layout ?? {},
            frames: chartDiv?.frames ?? [],
            config: chartDiv?.config ?? {}
        };
```

loitly (approved):
Tested the new feature and it worked perfectly. Good to go.

kpuriIpac:
> This looks good. I did not know you could get a schema our of plot.ly. That is pretty cool.

I didn't either! That was in in the initial ChatGPT suggestion. But the overall code it gave me was really bloated, so I had to strip it down and spend time studying it (which is good because it allowed me to familiarize myself more with Plotly).

PR #1885, opened 2025-11-17
FIREFLY-1832: Make SizeInputField more flexible

jaladh-singhal:
Fixes [FIREFLY-1832](https://jira.ipac.caltech.edu/browse/FIREFLY-1832), [FIREFLY-1882](https://jira.ipac.caltech.edu/browse/FIREFLY-1882)

Also see IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/451

* Added a new generic `QuantityInputField` and supporting helpers to standardize the logic for handling quantities with units, including validation and unit conversion. This abstracts most of the logic from WavelengthInputField that I created recently and applies that to SizeInputField which had older & unnecessarily complex validation and state handling logic.
* Refactored `WavelengthInputField.jsx` and `SizeInputField.jsx` to use the new `QuantityInputField`, which enables:
  - changing the default  name used by quantity: "Size" -> "Angle" for Position Angle field in spherex
  - locking units to single option
  - not showing error in `SizeInputField` when no initial `value` is provided but `nullAllowed=false`

## Testing
SPHEREx: https://firefly-1832-flexible-sizeinput.irsakubedev.ipac.caltech.edu/applications/spherex

Go to Spectrophotometry tab -> select "Sersic Galaxy Profile" -> check that "Position Angle" and "Effective Radius" doesn't have issues mentioned in linked tickets and have fixed unit.

Also go to Spectral Image Search tab and check if location and spectral constraints (using SizeInput and WavelengthInput) are working as before. 

Firefly: https://fireflydev.ipac.caltech.edu/firefly-1832-flexible-sizeinput/firefly

Regression test all instances of SizeInputField - I tried TAP and IRSA catalogs and didn't see any problem but we need more rigorous testing

### DIFF
--- a/src/firefly/js/ui/QuantityInputField.jsx
+++ b/src/firefly/js/ui/QuantityInputField.jsx
@@ -1,0 +1,251 @@
+/**
+ * Generic Quantity Input Field component and helpers.
+ *
+ * "Quantity" refers to a physical quantity that consists of a scalar (numeric value) and a unit,
+ * for e.g., length, angular size, etc.
+ */
+
+import React, { memo } from 'react';
+import PropTypes from 'prop-types';
+import {Divider, Stack, FormHelperText, Typography} from '@mui/joy';
+import { ListBoxInputFieldView } from 'firefly/ui/ListBoxInputField';
+import { InputFieldView } from 'firefly/ui/InputFieldView';
+import {useFieldGroupConnector} from './FieldGroupConnector.jsx';
+import Validate from 'firefly/util/Validate';
+
+/**
+ * Generic units picker used by QuantityInputFieldView.
+ *
+ * If there is more than one unit option, it shows a dropdown ListBoxInputFieldView to select the unit otherwise
+ * it just shows the single unit label.
+ *
+ * @param {object} props
+ * @param {string} props.unit
+ * @param {function} props.onChange
+ * @param {Array<{label:string,value:string}>} props.unitOptions
+ */
+const QuantityUnitsPicker = ({ unit, onChange, unitOptions = [], ...props }) => (
+    <Stack direction='row' alignItems='center'>
+        <Divider orientation='vertical' />
+        {unitOptions.length > 1
+            ? <ListBoxInputFieldView
+                value={unit}
+                onChange={onChange}
+                options={unitOptions}
+                multiple={false}
+                slotProps={{
+                    input: {
+                        variant: 'plain',
+                        sx: {minHeight: 'unset'},
+                    },
+                }}
+                {...props}/>
+            : <Typography sx={{px: 1}} {...props}>{unitOptions[0]?.label || unit}</Typography>
+        }
+    </Stack>
+);
+
+/**
+ * Generic component for a Quantity (= Numeric Value + Unit) Input.
+ *
+ * It's a controlled component in which state is sent through `{value, displayValue, unit, valid, message}` props and
+ * is received back through `onChange()` prop.
+ *
+ * @param {object} props
+ * @param {number} props.min
+ * @param {number} props.max
+ * @param {object} props.sx
+ * @param {object} props.slotProps
+ * @param {object} props.inputStyle
+ * @param {string} props.orientation
+ * @param {string} props.label
+ * @param {boolean} props.showFeedback
+ * @param {string} props.placeholder
+ * @param {string} props.tooltip
+ * @param {boolean} props.connectedMarker
+ *
+ * <hr>Stateful props:
+ * @param {string} props.value quantity value in `quantityBaseUnit` (for internal storage)
+ * @param {string} props.displayValue quantity value that is input by user (shown in the UI)
+ * @param {string} props.unit quantity unit input by user (shown in the UI)
+ * @param {boolean} props.valid whether `value` is valid (i.e. satisfies min/max bounds, nullAllowed, etc.)
+ * @param {string} props.message message to display if invalid
+ * @param {function(object, {value: string, displayValue: string, unit: string}): void} props.onChange lifts the updated
+ * value, displayValue, unit to the parent that controls the state.
+ *
+ * <hr>Quantity-specific props (required):
+ * @param {string} props.quantityName the name of the quantity (for display purposes)
+ * @param {function} props.convertQuantityUnits converts numeric value between units. Signature: (valueStr, fromUnit, toUnit) => string
+ * @param {function} props.formatQuantity formats the value (in quantityBaseUnit) to given unit. Signature: (valueInBaseUnit, outputUnit) => string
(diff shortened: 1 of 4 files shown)


src/firefly/js/ui/QuantityInputField.jsx
```diff
@@ -0,0 +1,251 @@
...
+                options={unitOptions}
+                multiple={false}
+                slotProps={{
+                    input: {
+                        variant: 'plain',
+                        sx: {minHeight: 'unset'},
+                    },
+                }}
+                {...props}/>
+            : <Typography sx={{px: 1}}>{unitOptions[0]?.label || unit}</Typography>
```
loitly: Should we also pass `props` to Typography so the label text can be customized?

src/firefly/js/ui/QuantityInputField.jsx
```diff
@@ -0,0 +1,251 @@
...
+ * @param {function(object, {value: string, displayValue: string, unit: string}): void} props.onChange lifts the updated
+ * value, displayValue, unit to the parent that controls the state.
+ *
+ * <hr>Quantity-specific props (required):
+ * @param {string} props.quantityName the name of the quantity (for display purposes)
+ * @param {function} props.convertQuantityValue converts numeric value between units. Signature: (valueStr, fromUnit, toUnit) => string
+ * @param {function} props.quantityAsStr formats the value (in quantityBaseUnit) to given unit. Signature: (value, unit) => string
+ * @param {string} props.quantityBaseUnit the canonical base unit key to use for `value` prop. Must be one of the `quantityUnitOptions` values.
+ * @param {Array} props.quantityUnitOptions options array for the units picker (label and value pairs)
+ */
```
loitly, line 82: I find the naming a bit confusing. I understand it may be influenced by the base `InputField`, but it’s still not very intuitive. For example, instead of [value, quantityBaseUnit] and [unit, displayValue], I would expect [value, unit] and [displayUnit, displayValue], or even better [inputUnit, inputValue].
Also, I don’t think every prop needs a `quantity` prefix.  
That said, the documentation does help clarify things. I don’t expect any changes.  Just leaving a comment.

loitly: Another example is `quantityAsStr`. It could follow a more standard naming pattern, such as getValueLabel. This would align with the common trio of props: value, inputValue, and getValueLabel, which represent the internal value, the value shown in the input box, and the formatted value used for display.

loitly (approved):
I tested the changes and they look good. I did notice an existing bug where a value like `123abc` still passes validation. I am pretty sure it is being interpreted as `123`. I am not sure if this needs to be addressed.

kpuriIpac (approved):
Tested both Spherex and Firefly according to instructions, looking for anything that may have been off. Looks good to me!

src/firefly/js/ui/QuantityInputField.jsx
jaladh-singhal: Good catch - I'll update this

src/firefly/js/ui/QuantityInputField.jsx
jaladh-singhal: > I understand it may be influenced by the base InputField, but it’s still not very intuitive.

yes this is what happened. Your naming scheme does make more sense!

> Also, I don’t think every prop needs a quantity prefix.

I think I did it so as to namespace these props under quantity-specific props bucket (as opposed to other visual or stateful props) which are also required! Earlier I was intending to use an object with custom type like QuantitySpec but then I didn't plan to nest them because it makes overriding of some of these props by consumers cumbersome.

If I take out "quantity" prefix, do you have better name suggestions for quantityName, convertQuantityValue, quantityUnitOptions too?

jaladh-singhal:
> I tested the changes and they look good. I did notice an existing bug where a value like 123abc still passes validation. I am pretty sure it is being interpreted as 123. I am not sure if this needs to be addressed.

@loitly yes this has existed before and is there now because I didn't change the value parsing logic. We are using `parseFloat` which doesn't care about trailing alphabets if it found some numbers in the beginning (whereas the opposite of this will throw error). Ideally we can write a custom parser on top of parseFloat using some regex to throw error when any non-digit character is present.

src/firefly/js/ui/QuantityInputField.jsx
loitly: I would use these.
```
quantityName 			name
convertQuantityValue 	valueConverter
quantityAsStr			valueAsLabel
quantityBaseUnit 		unit
quantityUnitOptions 	unitOptions
displayValue			inputValue
unit					inputUnit
```

loitly:
> > I tested the changes and they look good. I did notice an existing bug where a value like 123abc still passes validation. I am pretty sure it is being interpreted as 123. I am not sure if this needs to be addressed.
> 
> @loitly yes this has existed before and is there now because I didn't change the value parsing logic. We are using `parseFloat` which doesn't care about trailing alphabets if it found some numbers in the beginning (whereas the opposite of this will throw error). Ideally we can write a custom parser on top of parseFloat using some regex to throw error when any non-digit character is present.

Table PagingBar is using Validate.intValidator and it seems to work correctly.

src/firefly/js/ui/QuantityInputField.jsx
jaladh-singhal: @loitly I started changing to these but this is time taking and there are several issues:
- `displayValue -> inputValue`: `displayValue` is what FieldGroupConnector.jsx is already using so there may be unintended effects of changing it
- `unit -> inputUnit`: `unit` is being used in 30+ places in 20+ files using Size and Wavelength input components so it will affect all of them to change it to `inputUnit`
- `quantityBaseUnit -> unit`: `unit` alone is ambiguous IMO esp when you look it from unit conversion point of view which heavily depends on a base unit in conversion functions. I get the symmetry you are trying to create with `[value, unit]` vs `[inputValue, inputUnit]` but that would have worked if `value` had been `baseValue` in first place, which we can't change due to existing field group machinery
- `quantityName -> name`: `name` also may be confusing with `label` and similar visual props so I'd just keep it quantityName.

I can change the functions easily for this PR (but keeping the verb-first rule):
- `quantityAsStr -> formatQuantity`: I meant to change `...AsStr` but forgot
- `convertQuantityValue -> convertQuantityUnits`: because it's really converting units irrespective of base value or input value

I know I'm choosing the laziest path right now 😅  in interest of merging it ASAP for cutting RC but you're more than welcome to improve the naming in next iterations of this component!

jaladh-singhal:
> Table PagingBar is using Validate.intValidator and it seems to work correctly.

@loitly thanks for this pointer. I dug up why Validate.floatValidator used here was not doing the same and found that unit conversion happens before validator gets the value, and unit conversion was relying on JS `parseFloat` which is more forgiving with alphabets present in number. I changed `parseFloat` to `toFloat` from the validate library and now you can see the validator catches those invalid inputs.

PR #1890, opened 2025-12-03
Firefly-1897: support for web gpu when available in browser

robyww:
#### [Firefly-1897](https://jira.ipac.caltech.edu/browse/FIREFLY-1897):  support for web gpu when available in browser
 - isolated webgpu in single file (not including shaders)
 - each shader is in a wgsl file, in the wsdl directory
 - isolate gpu.js code in single file
 - moved mask processing to worker and web gpu
 - updated BrowserInfo to be more accurate with browsers, more utilities, log if unsupported browser version
 - refactoring and clean up: rawData directory files
 - removed remove rawData/rawAlgorithm/unused files
 - Min supported browser versions now pass to application at build time
   - log if unsupported browser but not shown to user

### Important information for reviewers
 - focus on `*.wgsl` and `WebGpuJobs.js` 
 - remember that some of the code is running in a web worker
    - all `*.gwsl`, `ManageRawDataThread.js`, `RawDataCommon.js` (main and worker), 
    - `GpuJsJobs.js`, `WebGpuJobs.js`, and a couple of others
  - We will remove `GpuJsJobs.js` in 2 or 3 years when we remove `gpu.js`
 - Ignore the deleted files, they are removed from the `unused` directory. We no longer need to keep them.
 - refactoring made for many small changes
 - There is a lot of clean up in `RawDataOps.js` this will be hard to review.

#### Applications to Test - _all should run as before_
- a message in the console will tell you if the gpu interface is `webgpu` or `gpujs`
- `webgpu` runs on chrome always. It runs on the latest firefox and safari only on OSX 26 and later
- Firefly: https://fireflydev.ipac.caltech.edu/firefly-1897-webgpu/firefly
- FinderChart: https://firefly-1897-webgpu.irsakubedev.ipac.caltech.edu/applications/finderchart
- Euclid: https://firefly-1897-webgpu.irsakubedev.ipac.caltech.edu/applications/euclid
- SPHEREx: https://firefly-1897-webgpu.irsakubedev.ipac.caltech.edu/applications/spherex


src/firefly/js/util/BrowserInfo.js
```diff
@@ -199,4 +228,15 @@ function supportsCssColorMix() {
...
+
+function supportsWebGpu() {
+    const foundGpu= Boolean(navigator.gpu);
+    if (isBrowser(Browser.CHROME)) return isVersionAtLeast(113) && foundGpu;
+    if (isBrowser(Browser.EDGE)) return isVersionAtLeast(113) && foundGpu;
+    // if (isBrowser(Browser.SAFARI)) return isVersionAtLeast(26) && BrowserInfo.platformMajorVerison>=26 && foundGpu;
+    if (isBrowser(Browser.SAFARI)) return isVersionAtLeast(26) && foundGpu;
+    if (isBrowser(Browser.FIREFOX)) return isVersionAtLeast(141) && foundGpu && Boolean(BrowserInfo.fireflyWebGpuEnabled);
+    if (isBrowser(Browser.OPERA)) return isVersionAtLeast(99) && foundGpu;
+    return foundGpu;
```
jaladh-singhal, line 240: instead of this hardcode version checking across browsers, why not just check if web gpu object is present in the browser's API?

Something like this:
```js
try {
const adapter = await navigator?.gpu?.requestAdapter();
return Boolean(adapter);
} catch(e) {
return false;
}
```

I believe this will be future proof as other browsers add support and we won't need to edit these version numbers. Also it will save us from the edge case when version is satisfied but web gpu adapter can't be retrieved.

src/firefly/js/visualize/rawData/wgsl/imageMask.wgsl
```diff
@@ -0,0 +1,59 @@
...
+    if (idx >= pixelCount) { return; }
+    let v = pixelData[idx];
+    let i= idx*32;
+//  var bit: u32 = 1u;
+//  for (var j = 0u; (j < 32u); j++) {
+//     outBuf[i+j] = select(NO_COLOR, color, (v & bit) == 0u);
+//     bit <<= 1u;
+//  }
+// the following does the commented out loop (above) but is more optimal
+
```
jaladh-singhal, line 26: This is interesting! 😄

src/firefly/js/visualize/rawData/GpuJsJobs.js
```diff
@@ -0,0 +1,208 @@
+import {isArray, once} from 'lodash';
+import {createCanvas, getGlobalObj} from '../../util/WebUtil';
+import {getGpuJsImmediate} from './GpuJsConfig';
+import {TILE_SIZE} from './RawDataCommon';
+
+let gpu= undefined;
+let GPU= undefined;
+let onGpuJsContextLost= undefined;
```
jaladh-singhal, line 8: [nitpick] unassigned `let` statements assign `undefined` by default so not really needed.

src/firefly/js/visualize/rawData/RawDataCommon.js
```diff
@@ -20,34 +18,40 @@ export const FULL= 'FULL';
...
+    if (BrowserInfo.isSafari() && !BrowserInfo.isVersionAtLeast(17)) return false;
     return isGPUAvailableInWorker();
 }
 
+export const logGpuState= once(() => {
+    const gpuType= BrowserInfo.supportsWebGpu() ? 'webgpu' : 'gpu.js';
+    const outStr= shouldUseGpuInWorker()
+        ? `Images: gpu in worker, gpu: ${gpuType}`
+        : `Images: gpu in main thread: ${gpuType}`;
+    console.log(outStr);
```
jaladh-singhal, line 31: should this rather be `console.debug` so that they don't clutter the console but can be seen when debug log level is enabled?

src/firefly/js/visualize/rawData/RawImageTilesGPU.js
```diff
@@ -1,187 +1,95 @@
...
-            return { ...inData, workerTmpTile };
-        }
-        else {
-            return { ...inData, workerTmpTile:canvas };
-        }
+import {gpujs_createRawDataTile3CRGB, gpujs_createRawDataTileImageRGB} from './GpuJsJobs';
+import {process3CTileViaWebGPU, processMaskTileViaWebGPU, processTileViaWebGPU} from './WebGpuJobs';
+
+const getTilingFuncs= once(() => {
+    if (BrowserInfo.supportsWebGpu() ) {
```
jaladh-singhal: this fallback is great! But what happens if a browser meets the version requirement (like chrome) but still gpu adaptor couldn't be found/initialized? See my related comment in BrowserInfo.js which might be the solution?!

src/firefly/js/visualize/rawData/RawDataOps.js
```diff
@@ -57,65 +36,30 @@ function createTileFromImageData(buffer, width,height) {
...
+    let rawTileDataAry;
+    const workerBitMapTile= undefined;
+
+    if (tileFromWorker[0].workerBitMapTile) { // just move workerBitMapTile to rawImageTile
+        rawTileDataAry= tileFromWorker.map( (obj) => ({...obj, rawImageTile:obj.workerBitMapTile, workerBitMapTile}));
+    }
+    else { // worker returned array data because no worker gpu support, todo: can I deprecate this else?
+        rawTileDataAry= [];
+        if (!BrowserInfo.supportsWebGpu() && !getGpuJsImmediate() ) await getGpuJs(); // make sure the GPU code is loaded up front
+        const cm=getColorModel(colorTableId,nanPixelColor, !BrowserInfo.supportsWebGpu());
```
jaladh-singhal, line 60: I've same concern here and at similar places which rely on version only check for fallback to gpu js.

jaladh-singhal (changes requested):
This works fine - I tested firefly and Euclid. 

I found 2 main issues, rest all are minor comments:
1) the web gpu check is browser version dependent rather than if it was able to intialise/if the API is available? see my code comments below - let me know if I'm missing something here.
2) the MOC look different in web gpu - the tile outline is missing in Auto and Fill modes (see chrome on left and firefox on right)

Image

jaladh-singhal:
Oh and I was just looking at browser compatiblity for web gpu - https://caniuse.com/?search=webgpu says that Firefox, chromium, safari browsers have a flag (see Note number 1 to 3) that can be enabled. Are we not turning this flag on manually until a browser decides to turn it on on a OS by default, because the web gpu support won't be mature enough?

buildScript/webpack.config.js
```diff
@@ -137,7 +147,12 @@ export default function makeWebpackConfig(config) {
                         ['@babel/preset-env',
                             {
                                 targets: {
-                                    browsers: ['safari >= 16', 'chrome >= 130', 'firefox >= 134', 'edge >= 130']
+                                    browsers: [
+                                        'safari >= '+MIN_SAFARI_VERSION,
+                                        'chrome >= '+MIN_CHROME_VERSION,
+                                        'firefox >= '+MIN_FIREFOX_VERSION,
+                                        'edge >= '+MIN_EDGE_VERSION
+                                    ],
```
loitly, line 155: Manually pinning specific versions is fine. However, if we’d prefer Babel to enforce a standard, we can rely on its built-in target policies instead.  e.g.
```
browsers: 'baseline widely available'
browsers: 'last 2 years'
```

src/firefly/js/visualize/rawData/WebGpuJobs.js
```diff
@@ -0,0 +1,292 @@
...
+const modules= {};
+let webGpuDevice= undefined;
+
+export async function initWebGPUDevice() {
+    if (!webGpuDevice) {
+        webGpuDevice= await synchronizeAsyncFunctionById('webGpuDevice', () => initDevice());
+       if (!webGpuDevice) console.log('warning: web gpu cannot initialize webGPU device');
+    }
+    return webGpuDevice;
+}
```
loitly: Probably don't need to be exported.

src/firefly/js/visualize/rawData/WebGpuJobs.js
```diff
@@ -0,0 +1,292 @@
...
+import {toRGB} from '../../util/Color';
+import {synchronizeAsyncFunctionById} from '../../util/SynchronizeAsync';
+import {getGlobalObj} from '../../util/WebUtil';
+import wgslSingleBandSource from './wgsl/imageSingleBand.wgsl?raw';
+import wgsl3ColorSource from './wgsl/image3ColorBand.wgsl?raw';
+import wgslImageRemapColor from './wgsl/imageRemapColor.wgsl?raw';
+import wgslImageMask from './wgsl/imageMask.wgsl?raw';
+
+
+
```
loitly: If you like, you can can add a short description here to explain what WebGPU is, how it’s used in Firefly, and any design considerations developers should be aware of.

src/firefly/js/visualize/rawData/wgsl/imageMask.wgsl
loitly: I wonder if it has to do with the loop is calculating `bit` when it could be using `j`.

loitly (approved):
Tested on a fits cube.  I can see improved performance compared to ops.

src/firefly/js/visualize/rawData/RawDataCommon.js
robyww: This just happens one time so there is not much clutter. I think for now is better to always show it.

src/firefly/js/visualize/rawData/WebGpuJobs.js
robyww: good idea

buildScript/webpack.config.js
robyww: I am not changing anything here. I just added constants that show up in the browser.

src/firefly/js/visualize/rawData/RawImageTilesGPU.js
robyww: supportsWebGpu looks for `navigator.gpu` and the version.

src/firefly/js/visualize/rawData/GpuJsJobs.js
robyww: an implied `undefined` is not always clear in reading the code.

src/firefly/js/util/BrowserInfo.js
robyww: Well there are a couple of reasons.
- using browser and version is more self documenting for such a new technology. This is straight from `canIuse`
- Most browsers only have `navigator.gpu` if it is supported. But not firefox, it defines a gpu even if not supported.
- I actually do what you suggest but only with firefox in the initialization function. Since requestAdapter is async (and probably heavy), it should only be done once. The async is the problem because it has to be done at initialization. I think there is a little bit of a race condition so I don't want to depend on it.
- the fall back is simply. `navigator?.gpu`
- Overall I think it is a little overkill but it is more clear.

robyww:
@jaladh-singhal 

> the MOC look different in web gpu - the tile outline is missing in Auto and Fill modes (see chrome on left and firefox on right)

I am glad the outline is missing! I have tried a couple of time to change that. I don't use the gpu for MOC drawing. I did touch the drawer code but only because I change a name of a `BrowserInfo` function.

I will look into it but it is good news!

After looking at it for 5 minutes... This is a difference between firefox and chrome drawing.  Firefox on mine (updated to use the gpu) is the same as firefox on ops.

src/firefly/js/util/BrowserInfo.js
jaladh-singhal: Ok this makes sense - thanks for the explanation.

jaladh-singhal:
> I am glad the outline is missing! I have tried a couple of time to change that. I don't use the gpu for MOC drawing. I did touch the drawer code but only because I change a name of a BrowserInfo function.
>
> I will look into it but it is good news!
>
>After looking at it for 5 minutes... This is a difference between firefox and chrome drawing. Firefox on mine (updated to use the gpu) is the same as firefox on ops.

@robyww I wasn't aware that's the intended goal (as I had always find outline helpful to realize that this is a HEALPix area). I do notice now that on chrome on ops, outlines never show up whereas on firefox on ops it does. So sorry for false alarm.

jaladh-singhal (approved):
Comments addressed by discussion - good to merge.

src/firefly/js/visualize/rawData/wgsl/imageMask.wgsl
robyww: according to AI it is because each line can be parallelized without dependencies on the previous. I am discovering that the gpu hates loops and does not like functions.

buildScript/webpack.config.js
robyww: one more thought. I actually do like having specific numbers which are more or less the last two years. That way I can get a sense of how our support for version features should be. This was a big deal with this PR. These newer visualization features are all pretty new and knowing if we should have a fallback (based on version) is important.

Another thought: If is very easy to require a more recent chrome and firefox (say last 6 months) because they auto-update. It is much harder with Safari because the version is more controlled by the OSX version.

PR #1892, opened 2025-12-05
Firefly-1903: attempt to fixed north up issue

robyww:
#### [Firefly-1903](https://jira.ipac.caltech.edu/browse/FIREFLY-1903): attempt to fixed north up issue

- fixes: [Firefly-1914](https://jira.ipac.caltech.edu/browse/FIREFLY-1914)

#### Notes
- I can only sort of duplicate this when I first load Spectral Images
- I am attempting to fix what I can duplicate, I don't know if it is fixed in all cases.

#### Testing
- https://firefly-1903-spherex-orient.irsakubedev.ipac.caltech.edu/applications/spherex

### DIFF
--- a/src/firefly/js/visualize/WebPlot.js
+++ b/src/firefly/js/visualize/WebPlot.js
@@ -536,6 +536,7 @@ export const WebPlot= {
             zoomFactor: zf,
             attributes,
             rawData,
+            lastByteRefreshData:Date.now(),
             directFileAccessDataAry,
             dataRequested: false,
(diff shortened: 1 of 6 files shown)


lrebull:
the orientation seems fixed; the size is still wrong. here it's the fourth one that is off (in my original ticket, it was the first one)
Screenshot 2025-12-05 at 11 09 38 AM

lrebull:
amending to note that i said "off" but it may not be "off" just "different" than the others. (i am not sure which is right!)

lrebull:
yes, i think this is fixed!!

PR #1893, opened 2025-12-09
FIREFLY-1916: Per-plane spectral coord display when WCSAXES != NAXIS

tgoldina:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1916
Test URL: https://fireflydev.ipac.caltech.edu/firefly-1916-wcsaxes/firefly

- Fixed bugs in the function that determines if the spectral coordinates are non-separable (should be calculated together).
- Bug fixes for the case when the number of WCS axes is different from the number of pixel axes.

### DIFF
--- a/src/firefly/js/visualize/PlotViewUtil.js
+++ b/src/firefly/js/visualize/PlotViewUtil.js
@@ -992,7 +992,7 @@ export function getPtSpectralCoords(plot, pt, cubeIdx, band= Band.NO_BAND) {
     if (!plot?.wlDataAry?.[band.value] || !hasWLInfo(plot)) return [0,0];
     const ipt= CCUtil.getImageCoords(plot,pt);
     if (!ipt && plot?.wlData?.hasPixelLevelCoordInfo) return [0,0];
-    return getWavelength(ipt, cubeIdx, plot.wlDataAry[band.value]) ?? [0,0];
+    return getWavelength(ipt||makeImagePt(0,0), cubeIdx, plot.wlDataAry[band.value]) ?? [0,0];
 }
 
 //=============================================================
--- a/src/firefly/js/visualize/projection/WavelengthHeaderParser.js
+++ b/src/firefly/js/visualize/projection/WavelengthHeaderParser.js
@@ -4,7 +4,7 @@
 
 import {makeDoubleHeaderParse} from '../FitsHeaderUtil.js';
 import {convertToArraySize} from '../../tables/TableUtil.js';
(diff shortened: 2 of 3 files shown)


src/firefly/js/visualize/PlotViewUtil.js
```diff
@@ -992,7 +992,7 @@ export function getPtSpectralCoords(plot, pt, cubeIdx, band= Band.NO_BAND) {
     if (!plot?.wlDataAry?.[band.value] || !hasWLInfo(plot)) return [0,0];
     const ipt= CCUtil.getImageCoords(plot,pt);
     if (!ipt && plot?.wlData?.hasPixelLevelCoordInfo) return [0,0];
-    return getWavelength(ipt, cubeIdx, plot.wlDataAry[band.value]) ?? [0,0];
+    return getWavelength(ipt||makeImagePt(0,0), cubeIdx, plot.wlDataAry[band.value]) ?? [0,0];
```
robyww, line 995: Just to be clear... an imagePt of 0,0 will only happen if there is no pixel level coordinates. So this parameter is just a place holder that is not really used?

src/firefly/js/visualize/projection/WavelengthHeaderParser.js
```diff
@@ -172,7 +172,7 @@ function isNonSeparableTABGroup(spectralCoords) {
     // all coordinates refer to the same table and output column
     const table = spectralCoords[0].tab.table;
     const ps3_1 = spectralCoords[0].tab.ps3_1.toUpperCase();
-    return spectralCoords.slice(1).every((c) => c.tab.table === table || c.tab.ps3_1.toUpperCase() === ps3_1);
+    return spectralCoords.slice().every((c) => c.tab.table === table && c.tab.ps3_1.toUpperCase() === ps3_1);
```
robyww: I don't think you need the `slice` anymore since it is just doing an array copy for a boolean check.

robyww (approved):
Looks good.

I don't think you need the `slice()`

src/firefly/js/visualize/PlotViewUtil.js
tgoldina: Yes, it's needed to calculate per-plane wavelength for image cubes.
 We need any valid image point to calculate per-plane spectral coord value.

src/firefly/js/visualize/projection/WavelengthHeaderParser.js
tgoldina: `spectralCoords.splice(1)` should remain, I’ve corrected that. We are considering groups of 2 or more spectral coordinates and comparing the tab parameters of the first coordinate with those of the remaining ones.

PR #1894, opened 2025-12-09
FIREFLY-1913: Fix combined chart wiping table sources in cascading mode

jaladh-singhal:
Fixes [FIREFLY-1913](https://jira.ipac.caltech.edu/browse/FIREFLY-1913)

This bug was happening because combining spectra using offsets (cascading), added a "tables::" expression to `data[trace].y` due to which new table sources are generated only for y but not for other fields like x, y_error, etc. We already had code to handle it - generalized it as `addTSMappingsToData` and used it after applying cascading.

## Testing
https://firefly-1913-combined-chart-cascading-bug.irsakubedev.ipac.caltech.edu/applications/euclid/

Follow the instructions in the ticket. Also do regression testing if possible.

### DIFF
--- a/src/firefly/js/charts/ui/CombineChart.jsx
+++ b/src/firefly/js/charts/ui/CombineChart.jsx
@@ -393,6 +393,24 @@ function getParamsFromDialog(props) {
     });
 }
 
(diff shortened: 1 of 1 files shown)


lrebull:
this seems to work!

loitly (approved):
Changes look right.  Good to go.

PR #1896, opened 2025-12-15
FIREFLY-1634: Table toolbar needs to wrap when table is narrow

loitly:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1634

The table toolbar now wraps into multiple sections, flowing from left to right.

**Note:** If the issue is related to the page’s minimum width, I can address that as well. It is currently set to 1000px.

**Test**: https://firefly-1634-table-toolbar-wrap.irsakubedev.ipac.caltech.edu/irsaviewer/dce
- Submit any query
- In the default tri-view layout, the table toolbar does not need to wrap.  It's limited by the page min-size.
- Switch to side-by-side layout mode:  [Table][coverage/images/charts] 
  - Menu(hamburger icon) -> Result Layout -> 5th option
- Use middle slider to shrink the table component and observe the toolbar toolbar wrapping behavior.

### DIFF
--- a/src/firefly/js/tables/ui/TablePanel.jsx
+++ b/src/firefly/js/tables/ui/TablePanel.jsx
@@ -317,13 +317,17 @@ function ToolBar({tbl_id, tbl_ui_id, connector, tblState, slotProps}) {
     if (!showToolbar) return null;
 
     return (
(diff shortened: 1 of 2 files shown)


robyww (approved):
looks good.

PR #1897, opened 2025-12-19
Firefly-1921: support mask cubes

robyww:
#### Firefly-1921: support mask cubes

- mask overlay will support cube mask on cube primary images
- puts the same plane of mask over a plane of the primary images

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1921-mask-grid/firefly
-use the SPHEREx sample mosaic: `sample_mosaic_delivery_v4.fits` 
- upload the fits file, and choose `image` 
- goto the drawing layers popupScreenshot 2025-12-19 at 12 35 22 PM 
- enable the mask Layers with `Enable`
- turn on `bit # 11 - COLD` - you will see the overlay as a wavelength number that matches the cube wavelength number below it.
- iterate through the cube you will see the overlay change with it. 
-  change the color of the overlay and keep interating


lrebull:
it took me a long time to understand what your instructinos were trying to say but i got it finally, and it seems to work!

jaladh-singhal (approved):
LGTM! Tested - works as described!

PR #1900, opened 2026-01-09
Firefly-1926: optimizing tile generation, transfer, and memory usage

robyww:
#### [Firefly-1926](https://jira.ipac.caltech.edu/browse/FIREFLY-1926): optimizing tile generation, transfer, and memory usage

  - Transfer each image tile individually instead of one large data array
  - Server not longer copies tiles into a single byte array from tiles, saving server memory
  - Client not longer parses and copies single byte array into tile array, saving client memory
  - Client makes a single call per tile and processes the image after it is retrieved, while other tiles are being retrieved
     - creates better use of worker thread, using CPU/GPU while retrieving another tile
  - Client `JobRunner.js` acts as a batch queue to control how many calls are made at a time 
     - currently does 3 at a time
     - keeps from overwhelming the server with many concurrent calls

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1926-concurrent-grid/firefly/
- https://firefly-1926-concurrent-grid.irsakubedev.ipac.caltech.edu/applications/euclid
- load images- big or small

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/ServerParams.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/ServerParams.java
@@ -198,6 +198,8 @@ public class ServerParams {
     public static final String BACK_TO_URL= "backToUrl";
     public static final String MASK_DATA= "maskData";
     public static final String MASK_BITS= "maskBits";
+    public static final String TILE_ACTION= "tileAction";
+    public static final String TILE_NUMBER= "tileNumber";
 
     //Workspaces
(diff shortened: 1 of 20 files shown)


loitly (approved):
The developer walked me through the changes in this PR.  It looks like a solid improvement for both the server and client sides.  I suggested testing it with one of our most demanding scenario, such as a grid of large images being stretched at the same time.  If that works as expected, the PR is good to go.

robyww:
Also tested with Euclid with a full grid of images, about 12.5 GB

PR #1903, opened 2026-01-21
FIREFLY-1929: reenabled compression for binary transfers

robyww:
#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1929-mask-performance/firefly/
- load an image and watch chrome devtools network. You will see the image transfers are compress. Mask are compress even more but it does not need to be tested to confirm the fix.


loitly (approved):
I’ve confirmed that HTTP responses are now gzip-compressed in transit. The changes look good.

PR #1905, opened 2026-01-23
Firefly-1931: alpha support for asdf file format

robyww:
#### [Firefly-1931](https://jira.ipac.caltech.edu/browse/FIREFLY-1931): asdf file alpha support 

 - alpha - only data under `roman` node, only uses roman layout
 - supports images and tables created by astropy

#### Reviewer notes.
- All of the work is done in `ASDFUtil.java`, the other file have only have small change to recognize the ASDF files.
- Strategy
   - Images: convert the ASDF image data to a FITS file
      - this happens at the same place we deal with compressed files.
   - Tables: read directly from the ASDF files

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-1931-asdf/firefly
- download an ASDF file and upload into firefly 
- use any ASDF from: https://stsci.app.box.com/s/aarfz7sktz6zrhpvz1763avcbzcs3ch2/folder/356233144630
- I have only tested with the ones in `Box` above, I am not sure others would work but we should try them.


ejoliet:
Exciting to see this and test it.
Before i do, do you know the versions of ASDF, Roman data-models and RAD you are supporting?
https://github.com/spacetelescope/roman_datamodels

Thanks!

robyww:
@ejoliet I can only tell you what is in the asdf file.

Here is the yaml part of one of the asdf files we are using. I see so many different versions in it!

[asdf-header.yaml.gz](https://github.com/user-attachments/files/24839900/asdf-header.yaml.gz)

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbDataIngestor.java
```diff
@@ -105,6 +107,10 @@ static FileInfo ingestFitsTable(TableServerRequest req, DbAdapter dbAdapter, Str
         var table = FITSTableReader.convertFitsToDataGroup(source, req, tableIndex);
         return ingestTable(dbAdapter, table, false);        //logic is already in the reader
     }
+    static FileInfo ingestAsdfTable(TableServerRequest req, DbAdapter dbAdapter, String source, int tableIndex) throws IOException, DataAccessException {
+        var table = ASDFUtil.convertAsdfAstroPyTableToDataGroup(new File(source), req, tableIndex);
+        return ingestTable(dbAdapter, table, false);        //logic is already in the reader
+    }
```
loitly: This works, but it requires loading the entire table into memory during database ingestion, which demands higher deployment memory limit when processing large files.
We have a `TableParseHandler.DbIngest` option that supports inline ingestion and uses minimal memory, regardless of table size.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotBuilder.java
```diff
@@ -234,12 +235,14 @@ private static void validateParams( MultiImageAction multiAction, boolean threeC
...
 
     private static ArrayList<Integer> getFileReadList(String extList, FileReadInfo[] frInfo) {
+        boolean addOne= frInfo[0].imageFileType() == FitsDataEval.ImageFileType.ASDF;
         List<String> idxs = new ArrayList<>(Arrays.asList(extList.split(",")));
         List<Integer> idxsInt = new ArrayList<>();
 
         int maxIdx = -1;
         for (String idxStr : idxs) {
             int idx = Integer.parseInt(idxStr);
+            if (addOne) idx++;
```
loitly: Adding a brief comment explaining why this is needed would be helpful for future reference.

src/firefly/java/edu/caltech/ipac/firefly/server/visualize/ImagePlotBuilder.java
```diff
@@ -352,6 +355,7 @@ private static void initState(PlotState state, FileReadInfo fi, Band band ) {
             bandS.setUploadedFileName(fi.uploadedName());
             bandS.setOriginalImageIdx(fi.originalImageIdx());
             bandS.setImageIdx(fi.originalImageIdx());
+            if (fi.imageFileType()!=FitsDataEval.ImageFileType.FITS) bandS.setFileType(fi.imageFileType().toString());
```
loitly, line 360: I don't think an exception is needed here, since it's just storing the data.

src/firefly/java/edu/caltech/ipac/util/ASDFUtil.java
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    public static DataGroup convertAsdfAstroPyTableToDataGroup(File file, TableServerRequest request, int tableIdx) throws IOException {
+        Path path= file.toPath();
+        try (AsdfFile asdfFile = Asdf.open(path)) {
+            AsdfFileInfo asdfInfo= getAsdfFileInfo(asdfFile);
+            if (asdfInfo == null || asdfInfo.tablesRootNode()==null) throw new IOException("could not read table");
+            int startingTableIdx= asdfInfo.imageNameList.size();
+            if (tableIdx>=startingTableIdx+asdfInfo.tables.size()) throw new IOException("table index exceeds number of tables");
+            String tName= asdfInfo.tableNames().get(tableIdx-startingTableIdx);
```
loitly, line 83: So `tableIdx` refers to the index within the ASDF file itself, not the index of the tables, right?
Also, are images always located at the beginning of the file, with tables at the end?  If not, I don't think this logic will work.

loitly (approved):
I tested this with a sample file and it worked as expected. The code looks clean. Implementing inline table ingestion would be a nice enhancement if we decide to pursue it.  Otherwise, this looks good to go.

src/firefly/java/edu/caltech/ipac/util/ASDFUtil.java
robyww: Yes, (at least for now) the processing make all the images it finds to be first then then all the tables. It is not necessarily about the structure of the file just an arbitrary way to organize it.

This is basically flattening out a very deep structure into something more simple for the user.

loitly: It would be frustrating for users who understand the file structure to have to work with a manufactured version.  I am pretty sure someone will raise the issue.

ejoliet (approved):
I played with one of the file in the box.com shared - those are Level-2 images - i can visualize all the extensions and check tables as well. Congrats!
As i mentioned in a separate discussion, there is much more complex products in ASDF that might requires more logic to extract data that originally image extractor could do. 
The other twist is making sure to highlight the compatibility with datamodel version used, which uniquely identified the content. Typically this is the version of the software, [roman_datamodels](https://github.com/spacetelescope/roman_datamodels/releases) and the [RAD](https://github.com/spacetelescope/rad/releases).

One thing you could always show even if you can't extract data is to show the header, which is very useful and is YAML (ASCII) so you don't need any decoder.

For images based ASDF, i think it's already a win!! I will share the tool once is released so people can start using it.

src/firefly/java/edu/caltech/ipac/util/ASDFUtil.java
robyww: I don't think there is a solution for that right now. For asdf, an image or a table does not have a numbered location. It is just at some arbitrary place. So our numbering scheme it just to help the user get a sence of it.  

In my recent meeting I did talk to the ASDF folks about adding a table of contents to the file format. If that ever happens we will use that.

Also remember this is just an first step at this. I am sure we will iterate on it several times.

src/firefly/java/edu/caltech/ipac/firefly/server/db/DbDataIngestor.java
robyww: ok,  will do that.

PR #1906, opened 2026-02-04
FIREFLY-1833: Fix Python client functions don't switch to Results tab if any other tab is selected

jaladh-singhal:
Fixes [FIREFLY-1833](https://jira.ipac.caltech.edu/browse/FIREFLY-1833)

This was a two part problem:
1. The Menu tabs were not syncing up with the `layoutInfo` in the store esp when results are available. Added an effect in `Menu` to listen to `layoutInfo.dropdown` and update the selected tab when results view.
2. `dropdownHandler` doesn't update `layoutInfo.dropdown` with recent payload correctly because `view: undefined` used for Results view gets ignored when merging objects and the new state has old view. 

## Testing
Point to this firefly server: https://fireflydev.ipac.caltech.edu/firefly-1833-tab-switch/firefly

Follow the instructions in the ticket. Also try other `fc.show_*` methods (can use [reference guide notebook](https://github.com/Caltech-IPAC/firefly_client/blob/master/examples/reference-guide.ipynb) for testing). The menu tab should switch to "Results" if not already there as the method gets executed.

### DIFF
--- a/src/firefly/js/core/LayoutCntlr.js
+++ b/src/firefly/js/core/LayoutCntlr.js
@@ -270,7 +270,7 @@ export function dispatchShowDropDown({view, menuItem, initArgs}) {
  * hide the drop down container
  */
 export function dispatchHideDropDown() {
-    flux.process({type: SHOW_DROPDOWN, payload: {visible: false}});
+    flux.process({type: SHOW_DROPDOWN, payload: resultsViewDropdown});
 }
 
 /**
@@ -415,10 +415,18 @@ export function getResultCounts() {
     return {haveResults,tableCnt,tableLoadingCnt,imageCnt,imageLoadingCnt,pinChartCnt,bgTableCnt};
 }
 
+/* layoutInfo.dropdown for the "Results" panel that shows in standard or expanded layout views */
+export const resultsViewDropdown = {
+    visible: false,
+    view: '' // '' is important to avoid `undefined` getting ignored in state objects merging
+};
+/* check if a layoutInfo.dropdown is for the "Results" view */
+export const isResultsViewDropdown = ({view, visible}) => !visible && view === '';
+
 /**
  * This handles the general use case of the drop-down panel.
- * It will collapse the drop-down panel when new tables or images are added.
- * It will expand the drop-down panel when there is no results to be shown.
+ * It will collapse the drop-down panel when new tables, chart, or images are added, i.e., show the results view (standard/expanded).
+ * It will expand the drop-down panel when there is no results to be shown, i.e., show the non-results panels.
  * @param {LayoutInfo} layoutInfo
  * @param {Action} action
  * @returns {LayoutInfo}  return new LayoutInfo if layout was affected.  Otherwise, return the given layoutInfo.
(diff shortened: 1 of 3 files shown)


src/firefly/js/ui/Menu.jsx
```diff
@@ -95,6 +94,11 @@ export function Menu() {
         }
     }, [selected,ready]);
 
+    useEffect(() => {
+        // if layoutInfo.dropdown in the store updated to results view, set selected menu tab to '' (results)
+        if (isResultsViewDropdown(dropDown)) dispatchSetMenu({...menu, selected: ''});
```
robyww: I think you should check first if it is already in this state before you call the dispatch. Our code layoutInfo a lot.

src/firefly/js/templates/common/MainPanel.jsx
```diff
@@ -14,13 +14,18 @@ import FOOTER_BG from 'images/ipac_bar.jpg';
 
 /**
  * Main panel of the FireflyLayout.  It handles the 3 possible views: drop-down, standard, expanded.
- * It also add a footer to the drop-down if footerComponent is set.
+ * It also adds a footer to the drop-down view if footerComponent is set.
+ *
+ * Typically, these 3 views render the following:
+ * 1) standard:   LandingPage or ResultsPanel.StandardView.DockLayoutPanel with split view of image, table and chart containers
+ * 2) expanded:   ResultsPanel.ExpandedView with one of image, table or chart container expanded to entire main panel
+ * 3) drop-down:  DropDownContainer containing any panel except the above two like search panel, upload panel, etc.
```
robyww, line 22: good!

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -415,10 +415,18 @@ export function getResultCounts() {
     return {haveResults,tableCnt,tableLoadingCnt,imageCnt,imageLoadingCnt,pinChartCnt,bgTableCnt};
 }
 
+/* layoutInfo.dropdown for the "Results" panel that shows in standard or expanded layout views */
+export const resultsViewDropdown = {
+    visible: false,
+    view: '' // '' is important to avoid `undefined` getting ignored in state objects merging
+};
```
robyww, line 422: sometimes if find that importing objects is fragile because of import dependencies. Though slightly less efficient it might be better to use a function here.

src/firefly/js/core/LayoutCntlr.js
```diff
@@ -427,17 +435,18 @@ export function dropDownHandler(layoutInfo, action) {
     // calculate dropDown when new UI elements are added or removed from results
     switch (action.type) {
         case CHART_ADD:
+            return smartMerge(layoutInfo, {dropDown: resultsViewDropdown});
         case TBL_RESULTS_ADDED:
         case TBL_RESULTS_ACTIVE:
         case TABLE_LOADED:
-            const tbl_id= action.type === CHART_ADD ? action.payload.groupId : action.payload.tbl_id;
+            const {tbl_id} = action.payload;
```
robyww, line 442: I don't know why this line was like this before. Are you sure this change is correct?

robyww (approved):
Looks good. 

I am most concerned about checking the menu state before calling the dispatch. I think if you do that it is ready to go.

Always glad to see more documentation.

loitly (approved):
Tested and confirmed it works as described. The code looks good.

src/firefly/js/ui/Menu.jsx
jaladh-singhal: yes, I should do that!

src/firefly/js/core/LayoutCntlr.js
jaladh-singhal: So far it seems to be working fine since this a `*Cntrl.js` file that does more exports than imports. But can change it in future if needed.

src/firefly/js/core/LayoutCntlr.js
jaladh-singhal: Yes. `action.payload.groupId` for `CHART_ADD` action results in group's ID. This will cause `getTblById` or `findGroupByTblId` always return undefined. Based on git blame, it seems this was added just to bypass upload table cases and does nothing for chart.

src/firefly/js/ui/Menu.jsx
jaladh-singhal: Done.

PR #1907, opened 2026-02-04
FIREFLY-1919: Add support for UWS jobs that produce non-table results

loitly:
This PR add the ability to handle UWS job results other than tables (e.g., images). It introduces dynamic loading based on MIME types and add support for multi-result jobs.
It uses the original search request to process the job, and a configured loader to load the results to the UI based on its MIME type.
In cases where a job produces multiple results, a new selection popup allows the user to choose which result to load.

Tests:
https://firefly-1919-jobs-with-mixed-results.irsakubedev.ipac.caltech.edu/irsaviewer/
https://firefly-1919-jobs-with-mixed-results.irsakubedev.ipac.caltech.edu/applications/spherex/
https://firefly-1919-jobs-with-mixed-results.irsakubedev.ipac.caltech.edu/applications/euclid/

Verified that all existing background job worked as before.  Pay extra attention to TAP, IRSA Catalog, Euclid, and SPHEREx SP Tool.

Normally, this operation returns a single table.

To test with a single-image FITS file:
- Go to Upload → Upload from URL
- Enter:
https://dev.irsakubedev.ipac.caltech.edu/api/spherex/spxmosaic/async/sm-6zskz
- Click `Upload`, then `Load`
- Expect to see a UWS job added to the Job Monitor and the image is displayed in Firefly

To test multi results:
- Go to Multi-archive VO TAP
- Select CADC -> any table with any search parameters.
- Click `Search`
- Either wait for monitor in Job Monitor.  
- Show results will popup a list of available results to load.

### DIFF
--- a/src/firefly/js/ui/FileUploadProcessor.jsx
+++ b/src/firefly/js/ui/FileUploadProcessor.jsx
@@ -140,7 +140,7 @@ export function resultSuccess(request,cacheKey) {
             const title= report.fileName ?? 'UWS Job File';
             const jobUrl = report?.parts[0].url;
             const req = makeTblRequest('UwsJob', title, {jobUrl});
-            dispatchTableSearch(req);
+            dispatchTableSearch(req, {backgroundable: true});
             return true;
 
         default: return false;
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/JobManager.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/JobManager.java
@@ -176,6 +176,11 @@ public class JobManager {
             });
             LOG.error(e);
         }
+        if (!job.getWorker().isSelfManaged()) {
+            updateJobInfo(jobId, ji -> {
+                ji.setOwnerId(reqOwner.getUserKey());
+            });
+        }
         return getJobInfo(jobId);
     }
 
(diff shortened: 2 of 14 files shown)


src/firefly/js/ui/FileUploadProcessor.jsx
```diff
@@ -140,7 +140,7 @@ export function resultSuccess(request,cacheKey) {
             const title= report.fileName ?? 'UWS Job File';
             const jobUrl = report?.parts[0].url;
             const req = makeTblRequest('UwsJob', title, {jobUrl});
-            dispatchTableSearch(req);
+            dispatchTableSearch(req, {backgroundable: true});
```
robyww, line 143: good

robyww (commented):
The code looks good and single image results work great.

However, the UI for the multi response is not right. We should talk about what it the best way to proceed. Ultimately we want a TAP search like CADC (and rubin?) to work just like before. So I am not sure how to expose the other results. Let's talk about it in the standup on Monday.

robyww (approved):
Would talked about a couple of things in the stand up. After that I think it is ready to go.

kpuriIpac (changes requested):
Sorry for the late review. All searches (and job monitor) seem to working as expected, but I did notice a UI bug when I tried the Upload from URL as suggested in PR: 

Image


When you expand the `parameters` (or any other UWS job info), it goes "over" the Load button as you can see in the screenshot above. 

All else looks good to me!

loitly:
> Sorry for the late review. All searches (and job monitor) seem to working as expected, but I did notice a UI bug when I tried the Upload from URL as suggested in PR:

This is an existing bug.  If it's okay, I would like to merge this first, then fix the bug at a later time. @kpuriIpac

kpuriIpac (approved):
I am fine with having the existing bug be fixed later. Looks good to me other than that!

PR #1908, opened 2026-02-04
FIREFLY-1933: Add React Compiler

kpuriIpac:
**Ticket**: [FIREFLY-1933](https://jira.ipac.caltech.edu/browse/FIREFLY-1933)  
- This PR adds React Compiler to firefly:
  - I suggest devs take a look at docs: https://react.dev/learn/react-compiler/introduction  
     - whenever you get a chance, not necessarily to review this PR. I think it'll be helpful to have basic understanding of React Compiler moving forward. 
  - React Compiler does not automatically generate a list of components it memoizes, but you can inspect components in `React DevTools` -> `Components` and any components that have `Memo :sparkles:` (sparkles is the emoji with 3 stars) next to them have been memoized by React Compiler.
     - React Compiler is already used in production by several companies, including of course `Meta` and I think even `Github` (inspect & see `components` in React DevTools while on `github.com`) 
   - For our existing components, React docs suggest we leave any `useMemo` and `useCallback` as is (removing them carefully, over time, is optional). 
   - For newer components, we need not memoize them (but we can, _`as an escape hatch to provide control over which values are memoized`_ according to react)
  - Now, as for how React Compiler works: 
    - it does **_attempt_** to, theoretically, memoize every component (trying to optimize re-rendering of components essentially, and only doing so when necessary). It also attempts to optimize JSX subtrees (by caching them). 
    - It will skip memoizing any components that break [Rules of React](https://react.dev/reference/rules). But there's no way to know which ones it skips unless we look at dev tools or get specific warnings/errors from `eslint`. 
   

The bug I encountered in TAP and why React Compiler still chose to memoize this component: 
  - `BasicUI` component (in `TapViewType`) was giving me issues after setting up React Compiler. After it applied memoization, a part of the component was only rendering `Skeleton`. 
    - The issue: Earlier, we had this line in our code: 
      -  `const capabilities = getLoadedCapability(serviceUrl); `
      - `capabilities` is not a react state here. Pre-memoization of this component, we had a number of incidental re-renders that **_eventually_** made `capabilities` have a `truthy` value. 
        - But after memoization, even after `capabilities` was eventually populated with the call to `getLoadedCapability`, this would not cause BasicUI component to re-render! 
     - so turning `capabilities` into a react state (my code change) fixes this bug. 
  - The reason why React Compiler still memoizes this component is because capabilities being `falsy` is not technically breaking any Rules of React (react compiler is only concerned with React props/state/context, and not global values, etc.) 
    - but `capabilities` being `falsy` did lead to this being false: 
      -  `{(capabilities && columnsModel) ?  `
      - and rendering `<Skeleton/>` instead, which is a UI bug, yes but not technically a bug React compiler would catch).

If we find other components with similar issues, they will need to be handled manually. But I am happy to report I didn't come across any more in my testing. If we do happen to find **_several_** issues in testing, we could still keep React Compiler, but adopt incremental adoption instead (https://react.dev/learn/react-compiler/incremental-adoption) 
 - Basically, there's a way to apply React compiler directory by directory instead of to all of firefly at once. So we can fine tune it, keep adding directories one by one, until we've added all and find no issues with React Compiler being in firefly.  

**_Note_**: I may change/remove/add some of the `eslint` rules. 
 

**Testing**: 
- [Firefly](https://fireflydev.ipac.caltech.edu/firefly-1933-react-compiler/firefly), [Irsaviewer](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/irsaviewer), [Finderchart](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/finderchart), [Euclid](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/euclid), [Spherex](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/spherex), [Wise](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/wise), [Sofia](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/sofia), [SHA](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/Spitzer/SHA) and [ZTF](https://firefly-1933-react-compiler.irsakubedev.ipac.caltech.edu/applications/ztf)
- Test the apps as extensively as possible. 
   - I did a few basic searches on all apps, those worked fine. I even tested the download packager on some apps. 
 - Look out for any anomalies, any components not working as expected, and please report them back here.


src/firefly/js/ui/tap/TapViewType.jsx
```diff
@@ -92,18 +92,18 @@ const expandableTapSectionSx = {
...
+        if (!serviceUrl) return;
+        setCapabilities(getLoadedCapability(serviceUrl));
         if (!isCapabilityLoaded(serviceUrl)) {
             loadTapCapabilities(serviceUrl)
-                .then((c) => setCapabilitiesChange(c??{}))
-                .catch( (error) => {
-                    setError(`Fail to retrieve capability for: ${serviceUrl}`);
-                });
+                .then((c) => setCapabilities(c ?? getLoadedCapability(serviceUrl)))
+                .catch(() => setError(`Fail to retrieve capability for: ${serviceUrl}`));
```
loitly: Should it be like this so that it doesn't setState twice when not loaded?
```
if (isCapabilityLoaded(serviceUrl)) {
    setCapabilities(getLoadedCapability(serviceUrl));
} else {
    loadTapCapabilities(serviceUrl)
         .then((c) => setCapabilities(c ?? getLoadedCapability(serviceUrl)))
         .catch(() => setError(`Fail to retrieve capability for: ${serviceUrl}`));
```

loitly (approved):
I didn't test all of the applications, but the ones I tested are working as expected.  I am glad you found and fixed the issue in TAP panel.  It’s reassuring to know this was a real bug rather than a false positive from the React compiler.  I am fine with merging and monitoring it closely.

src/firefly/js/ui/tap/TapViewType.jsx
kpuriIpac: Yes good point, I made this change (slightly edited for better reading flow)

PR #1910, opened 2026-02-18
IRSA-7518, FIREFLY-1948, FIREFLY-1770: SPHEREx mosaic tool, Result Panel's tab layout control, Job Progress UI updates

jaladh-singhal:
See IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/457

See firefly client PR: https://github.com/Caltech-IPAC/firefly_client/pull/78

## Testing
https://fireflydev.ipac.caltech.edu/irsa-7518-spherex-mosaic-tool/firefly

- Check the improved progress bar component as per instructions in https://github.com/Caltech-IPAC/firefly/pull/1912. I used ObsTAP search on MAST images since it remains stuck in Queued phase.
- Do some regression testing for affected components

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/data/ServerRequest.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/data/ServerRequest.java
@@ -3,6 +3,7 @@
  */
 package edu.caltech.ipac.firefly.data;
 
+import edu.caltech.ipac.firefly.util.MathUtil;
 import edu.caltech.ipac.util.StringUtils;
 import edu.caltech.ipac.visualize.plot.ResolvedWorldPt;
 import edu.caltech.ipac.visualize.plot.WorldPt;
@@ -276,6 +277,10 @@ public class ServerRequest implements Serializable, Cloneable {
         return StringUtils.getDouble(getParam(key));
     }
 
+    public double getDoubleParam(String key, double def) {
+        return StringUtils.getDouble(getParam(key), def);
+    }
+
     public float getFloatParam(String key) {
         return StringUtils.getFloat(getParam(key));
     }
@@ -292,6 +297,15 @@ public class ServerRequest implements Serializable, Cloneable {
             wpt= null;
         }
         return wpt;
+    }
+
+    public double getSizeParam(String key, MathUtil.Units units) {
+        double sizeInDeg = getDoubleParam(key); // client sends size in degrees
+        return MathUtil.convert(MathUtil.Units.DEGREE, units, sizeInDeg);
+    }
+
+    public double getSizeParam(String key) {
+        return getSizeParam(key, MathUtil.Units.DEGREE);
     }
 
 //====================================================================
--- a/src/firefly/js/charts/ChartsCntlr.js
+++ b/src/firefly/js/charts/ChartsCntlr.js
@@ -86,6 +86,7 @@ function reducers() {
  *  @param {string} [p.viewerId] – viewer where chart will be displayed
  *  @param {boolean} [p.deletable] - is the chart deletable, if undefined: single chart in a group is not deletable, multiple are deletable
  *  @param {string} [p.help_id] - help id, if undefined, no help icon shows up
+ *  @param {boolean} [p.activateViewer] - activate the chart viewer (tab) when chart is added
  *  @param {Function} [p.dispatcher=flux.process] - only for special dispatching uses such as remote
  *  @public
  *  @function dispatchChartAdd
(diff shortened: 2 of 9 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -120,6 +120,11 @@ public DataGroup fetchDataGroup(TableServerRequest req) throws DataAccessExcepti
             jobUrl = jobUrl == null ? req.getParam(JOB_URL) : jobUrl;
             if (jobUrl == null) {
                 jobUrl = submitJob(req);
+                if (jobUrl.contains("spxmosaic")) {
+                    // workaround for mosaic tool service; otherwise job info gets stuck in PENDING state even when it completes or errors out
+                    // TODO: remove this workaround with a better fix or remove when UWS service no longer has this issue
+                    TimeUnit.SECONDS.sleep(1);
+                }
```
jaladh-singhal: I was having a hard time track down why mosaic tool job gets stuck in “PENDING” phase in the app whereas in reality, it completes almost instantaneously (testable by going to the UWS job link as soon as it’s created and displayed in the app).

Interestingly, when I ran debugger server-side, the phase in app reflects the correct state of UWS job so everything works! So I concluded that there is some race condition that the debugger is resolving somehow by timeout on breakpoints. Hence I put this sleep between submitting and running the job to make it work correctly.

This is ofc a temporary fix. @loitly, do you have any idea why this race condition is happening, given that the HTTP calls made by `submitJob` and `runJob` are synchronous and blocking? I spend more than a day debugging it and I'm out of my depth 😅

loitly: Ah.. thanks for pointing this out.  This is a `backend/uws` bug.  Argo Workflow set a job to `PENDING` when it's waiting to be executed.  The `uws` service should have recognized that `PENDING` has a different meaning in uws and that the translated phase should have been `QUEUED` instead.  Because when it's `PENDING`, UwsJobProcessor will stop watching the job.
https://github.com/Caltech-IPAC/firefly/blob/7fb871495dfa1a058b96caf82a50a2f4edf3dd53/src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java#L167

jaladh-singhal: Ohh. That makes sense because the working jobs were transitioning from `PENDING` -> `QUEUED` -> `EXECUTING` whereas this mosaic tool job goes `PENDING` -> `QUEUED` -> `PENDING` and then the DataAccess exception is thrown (because the execution comes to the `if` clause you linked).

So...is this an easy fix in mosaic tool mock service? If yes, then I'll remove this workaround as you fix that. Otherwise I'll keep it in until we replace mock service with the actual service (whenever that happens).

jaladh-singhal: Removed the workaround since Loi fixed the bug in the mock service.

src/firefly/java/edu/caltech/ipac/firefly/data/ServerRequest.java
```diff
@@ -276,6 +277,11 @@ public double getDoubleParam(String key) {
         return StringUtils.getDouble(getParam(key));
     }
 
+    public double getDoubleParam(String key, double def) {
+        double ret = StringUtils.getDouble(getParam(key));
+        return Double.isNaN(ret) ? def : ret;
+    }
```
loitly, line 282: I would move this to StringUtils as `getFiniteDouble`, since it already handles parsing numbers from strings.  The check should be `isFinite` since there's also (+-)Infinity.

loitly (approved):
I like the new progress look overall. I’m not completely sure about the elapsed time display when the JobInfo dialog is extremely wide, but because it starts narrow, the timer should be noticeable.

src/firefly/java/edu/caltech/ipac/firefly/data/ServerRequest.java
jaladh-singhal: I realized I could have just used `StringUtils.getDouble(str, def)`. I was doing NaN check because `StringUtils.getDouble(str)` returns `NaN` when string is `null`. Fixed it.

loitly: The problem is `StringUtils.getDouble` can still return NaN or (+-)Infinity.  If you need to exclude them, you have to check.

jaladh-singhal: Because of the client-side validation, I cannot imagine anything that will return NaN or +/-Inf. Also, I'll revisit this in the second/third pass when the tool container produces real results over the mock ones right now.

jaladh-singhal:
> I’m not completely sure about the elapsed time display when the JobInfo dialog is extremely wide, but because it starts narrow, the timer should be noticeable.

@loitly I forgot to reply to this. I agree, we can put a max width on the progress bar. I wasn't sure if the message might be too long that we want it to take full width of the dialog so I didn't do that.

PR #1911, opened 2026-02-18
Firefly-1946: certain rubin file not loading

robyww:
#### Testing (_read carefully_, all steps are important)

- Rubin portal: https://rc-2026-1.irsakudev.ipac.caltech.edu/suit
- go to Multi-archive SIAv2 tag
- enter : https://data-int.lsst.cloud/api/sia/prompt/query
   - make sure you hit `enter` after you enter the new url, this is not obvious like with tap, it needs to be looked at.
- turn off spatial
- under `observation type and source` check `Calibrated science-ready data (2)`
- set row limit to 100
- do a search
- After the seach the images should load


kpuriIpac (approved):
Looks good. Tested carefully according to instructions. 
Is that the only search that was broken and that we needed to test?  If so, this is good to go. 

(I did see you mention S3 signed downloads are supported but not yet working - and yes I agree we can look into this).

PR #1912, opened 2026-02-19
FIREFLY-1770: Move pct_complete to progress under uws:jobInfo

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-1770

Firefly currently stores job progress in a custom field within the UWS job. This change updates the implementation to persist progress data under the standard <progress> element inside <uws:jobInfo>, making it the official and standard location for tracking UWS job progress in Firefly.

**UI Behavior**
- A determinate LinearProgress is used when a percentage complete can be computed.
- An indeterminate LinearProgress is used when percentage information is unavailable.
- In both cases, the UI displays:
  - Elapsed time
  - A status message from the <progress> block
- If no progress message is provided, the job phase is displayed instead.

**Testing**
https://firefly-1770-progress-ui.irsakubedev.ipac.caltech.edu/irsaviewer/
https://firefly-1770-progress-ui.irsakubedev.ipac.caltech.edu/applications/spherex/

- Submit any background job to verify that progress appears in the mask panel.
- Click the info **(“i”)** icon to open the Job Info dialog.
- Confirm that:
  - Full job details are displayed
  - Progress information is rendered across the dialog

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/core/background/Job.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/core/background/Job.java
@@ -52,21 +52,19 @@ public interface Job extends Callable<String> {
         try {
             updateJobInfo(getJobId(), ji -> {
                 ji.setStartTime(Instant.now());
-                ji.getMeta().setProgress(0);
             });
             String results = run();
-            // the worker is set in onStart().
+            // the worker is set at onStart().
             getWorker().onComplete();
             if (Thread.currentThread().isInterrupted()) throw new InterruptedException("Job was aborted");
             updateManagedStatus(ji -> {
                 ji.setPhase(JobInfo.Phase.COMPLETED);
-                ji.getMeta().setProgress(100, "");
             });
             return results;
         } catch (InterruptedException | DataAccessException.Aborted e) {
             updateJobInfo(getJobId(), ji -> {
                 ji.setPhase(JobInfo.Phase.ABORTED);
-                ji.getMeta().setProgress(100, "Job was aborted");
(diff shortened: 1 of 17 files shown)


jaladh-singhal (approved):
I tested different jobs, it's working as intended both in BG mask panel and The code looks good too.

About the UI, I think it can be improved. Few things:
- The sheet style creates this boxed layout and we already are in a box of table container so box-in-a-box looks bit awkward. The mask animation in background is great, we can let it bleed to progress sheet too.
- The progressbar is taking too much real estate without conveying meaningful information, it can be made sleeker by adjusting thickness. 
- "Progress:" text should be followed by a text value rather than a progressbar, following the common pattern we have elsewhere
- The font sizes are too small, I tried going one level up and it looks fine. Maybe for dialog we need to keep it bit smaller.
- Elapsed time can be moved to the right corner, common to how progress bars are out there on internet

I did some quick hacking through devtools to implement these (not perfect, can be iterated more):
Image

Let me know if you want me to implement these UI tweaks in my PR (if you like/agree with them) since I'm already touching the BG monitor files. They are not a blocker for the functionality you added.

loitly:
> Let me know if you want me to implement these UI tweaks in my PR (if you like/agree with them) since I'm already touching the BG monitor files. They are not a blocker for the functionality you added.

I like it.  This looks much better here. I wonder how it will look in the popup, especially when you bleed into the background. I’m happy for you to tweak it in your PR, and then we can review it in both contexts. Keep in mind that the progress message can be any length. Remember to keep the ellipsis behavior while still showing the full message in tooltips.

PR #1914, opened 2026-02-24
IRSA-7589: Allow errorSummary to exist even when the phase is not Error

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/IRSA-7589

When Firefly imports a UWS job via URL, it currently treats the presence of an `<errorSummary>` element as an indication of failure and forces the job phase to ERROR.

The presence of an `<errorSummary>` element is not necessarily indicative of a failed job, as it may be included irrespective of the job phase. When importing such jobs, Firefly should respect the phase reported by the service instead of forcing an ERROR state.

Test: https://irsa-7589-handle-errorsummary.irsakubedev.ipac.caltech.edu/irsaviewer/
- Upload -> Upload from URL
- Enter: https://irsa.ipac.caltech.edu/api/spherex/spectrophotometry/async/4c0b58d3-8c8a-40b2-987e-28c0dec0da72
- Click Upload

Verify that `Phase: COMPLETED` and `Error Summary` exists.

Also fixed in this PR:
- Scrollbar appeared when Job information overflow.  The `Load` button correctly rendered on the bottom left of the panel.

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
@@ -99,6 +99,7 @@ public class UwsJobProcessor extends EmbeddedDbProcessor {
         if (status.isError()) {
             logger.warn(status.getErrMsg());
             sendJobUpdate(ji -> {
+                ji.setPhase(Phase.ERROR);
                 ji.setErrorSummary(new ErrorSummary(status.getErrMsg()));
             });
         } else {
@@ -128,6 +129,7 @@ public class UwsJobProcessor extends EmbeddedDbProcessor {
             }
         } catch (Exception e) {
             updateJob(ji -> {
+                ji.setPhase(Phase.ERROR);
                 ji.setErrorSummary(new ErrorSummary(e.getMessage()));
             });
             throw new DataAccessException(e.getMessage());
@@ -144,7 +146,10 @@ public class UwsJobProcessor extends EmbeddedDbProcessor {
                 JobInfo uwsJob = getUwsJobInfo(jobUrl);
                 if (uwsJob == null) {
                     String msg = "Failed to retrieve UWS job info";
-                    sendJobUpdate(ji -> ji.setErrorSummary(new ErrorSummary(msg)));
+                    sendJobUpdate(ji -> {
+                        ji.setPhase(Phase.ERROR);
+                        ji.setErrorSummary(new ErrorSummary(msg));
+                    });
                     throw new DataAccessException(msg);
                 }
                 try {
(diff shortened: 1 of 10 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
```diff
@@ -162,9 +167,12 @@ public DataGroup fetchDataGroup(TableServerRequest req) throws DataAccessExcepti
                         case Phase.ERROR:
                             throw new DataAccessException("Job has failed with the error: " + uwsJob.getErrorSummary().message());
                         case Phase.UNKNOWN: {
-                            if (cnt > 20) {
-                                updateJob(ji -> ji.setErrorSummary(new ErrorSummary("Unknown phase for over 20 seconds")));
-                                throw new DataAccessException("Unknown phase for over 20 seconds");
+                            if (cnt > 70) {
```
jaladh-singhal, line 170: Is use of 70 here instead of 60 (=2 min as in error messages below), intentional?

jaladh-singhal (approved):
Tested, working as expected. Code looks good too! 

Also thanks for fixing UWS upload form overflowing bug. I noticed two semi-related job info UI issues - see next comment since they are out of scope for this PR.

jaladh-singhal:
image

1. In job info results panel: if ID is long, it wraps making the layout look awkward. It should either follow the max-width + ellipsis pattern or take full width without wrapping. Also I feel href is too strict with max width (this becomes pronounced in Upload panel), it should allow all available width until it runs to right edge of the job info results panel (similar to error summary  message)
2. In job info error summary: if `has detail: true`, shouldn't UI show something like `details: <job link>/error`? This will be super helpful in debugging mosaic tool jobs and I remember us talking about it.

src/firefly/java/edu/caltech/ipac/firefly/server/query/UwsJobProcessor.java
loitly: This is because `cnt` uses a linear backoff wait strategy, resulting in around two minutes of elapsed time after 70 iterations.

jaladh-singhal:
@loitly thanks for the updates. They look good.

One thing got left out:
> if ID is long, it wraps making the layout look awkward. It should either follow the max-width + ellipsis pattern or take full width without wrapping.
image

But shouldn't hold this PR from merging, it's good to go!

loitly:
> One thing got left out:
> 
> > if ID is long, it wraps making the layout look awkward. It should either follow the max-width + ellipsis pattern or take

Sorry, you were looking at an old build.  I've fixed it in my latest code changes.  Because you did not response in Slack, I assumed you were not going to look at it again.

PR #1915, opened 2026-02-27
FIREFLY-1951: Create GitHub Action to Build and Publish Firefly Image to GHCR

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1951

Create a GitHub Actions workflow to manually build and publish the Firefly Docker image to GitHub Container Registry (GHCR). The workflow should support multi-platform builds.

- Multi-Platform Support
- Use docker buildx
- Build for:
  - linux/amd64
  - linux/arm64
- Publish a multi-arch manifest to GHCR

This PR is primarily for [documentation](https://github.com/Caltech-IPAC/firefly/blob/FIREFLY-1951-ghaction-build-publish/docs/how-to/github-action.md) + the initial version of the workflow.
The workflow has been created and tested against the lsst/suit.

To test the published image locally:

```bash
docker run  -p 8000:8080 --rm -e PROPS_redis__host=localhost ghcr.io/lsst/suit:release-2025.5.4
```
Then, goto http://localhost:8000/suit/ to confirm it is running correctly.

### DIFF
--- a/.github/workflows/build_publish.yml
+++ b/.github/workflows/build_publish.yml
@@ -3,18 +3,21 @@ name: Build and publish Firefly image
 on:
   workflow_dispatch:
     inputs:
-      tag_name:
-        description: "Docker image tag; git tag without 'release-' prefix (e.g. 'xxxx.x.x')"
+      git_tag:
+        description: "Git tag to use(e.g. 'release-xxxx.x.x' or a full git tag)"
         required: true
+        type: string
+      img_tag:
+        description: "Optional: Docker image tag(e.g. 'xxxx.x.x'); If omitted, `git_tag` will be used."
+        required: false
         type: string
       push_image:
         description: "Push image to GHCR"
         required: false
         default: false
         type: boolean
-# Commenting out release trigger for now since we have not decided on publishing to GHCR.
-#  release:
-#    types: [published]
+  release:
+    types: [published]
 
 permissions:
   contents: read
@@ -26,13 +29,34 @@ jobs:
 
     steps:
       # ------------------------------------------------------------
-      # Checkout firefly repo at configured tag
+      # Resolve checkout ref and image tag
+      # - If triggered by a release event, use the release tag for both ref and image tag
+      # - Otherwise, prefer `inputs.img_tag` (short form) and fallback to `inputs.git_tag`
+      # This ensures `git_tag` is required and independent of `img_tag`.
+      # ------------------------------------------------------------
+      - name: Resolve tags
+        id: resolve_tags
+        run: |
+          if [[ "${{ github.event_name }}" == "release" ]]; then
+            echo "ref=${{ github.event.release.tag_name }}" >> $GITHUB_OUTPUT
+            echo "tag=${{ github.event.release.name }}" >> $GITHUB_OUTPUT
+          else
+            echo "ref=${{ inputs.git_tag }}" >> $GITHUB_OUTPUT
+            if [[ -n "${{ inputs.img_tag }}" ]]; then
+              echo "tag=${{ inputs.img_tag }}" >> $GITHUB_OUTPUT
+            else
+              echo "tag=${{ inputs.git_tag }}" >> $GITHUB_OUTPUT
+            fi
+          fi
+
+      # ------------------------------------------------------------
+      # Checkout firefly repo at the resolved ref
       # ------------------------------------------------------------
       - name: Checkout Firefly
         uses: actions/checkout@v4
         with:
           repository: Caltech-IPAC/firefly
-          ref: release-${{ inputs.tag_name }}
+          ref: ${{ steps.resolve_tags.outputs.ref }}
           path: firefly
 
       # ------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


.github/workflows/build_publish.yml
```diff
@@ -0,0 +1,96 @@
+name: Build and publish Firefly image
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        description: "Docker image tag; git tag without 'release-' prefix (e.g. 'xxxx.x.x')"
```
robyww: I think this should be the full tag. 
We can make a pre-release with a tag like this: `pre-1-2026.1`

.github/workflows/build_publish.yml
```diff
@@ -0,0 +1,96 @@
...
+
+    steps:
+      # ------------------------------------------------------------
+      # Checkout firefly repo at configured tag
+      # ------------------------------------------------------------
+      - name: Checkout Firefly
+        uses: actions/checkout@v4
+        with:
+          repository: Caltech-IPAC/firefly
+          ref: release-${{ inputs.tag_name }}
```
robyww: I don't understand where this comes from for `on: release`, since the tag does start with `release`. I suspect this line might be wrong.

.github/workflows/build_publish.yml
```diff
@@ -0,0 +1,96 @@
...
+          path: firefly
+
+      # ------------------------------------------------------------
+      # Checkout Firefly online help
+      # ------------------------------------------------------------
+      - name: Checkout Firefly Help
+        uses: actions/checkout@v4
```
robyww, line 66: do we have to go through and updates the version of a action everything it has a new version? like this one is `@v4`

docs/how-to/github-action.md
```diff
@@ -0,0 +1,484 @@
...
+
+This results in:
+
+```
+.
+├── my-app/
+└── firefly/
+```
+
+The Docker build `context` should be set appropriately (for example `context: .`) so both directories are visible during the build.
```
robyww: so it is always `context: .` ?

robyww (commented):
Made a pass looking for the PR and included a few questions and comments. It is pretty easy to follow. The documentation you wrong seems very complete. I expect to have more questions next week.

.github/workflows/build_publish.yml
```diff
@@ -0,0 +1,96 @@
...
+      # Build (and optionally push) multi-platform image
+      # ------------------------------------------------------------
+      - name: Build and push image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: firefly/docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name == 'release' || inputs.push_image == 'true' }}
+          tags: ghcr.io/lsst/suit:${{ steps.image_tag.outputs.tag }}
```
jaladh-singhal: Shouldn't it reflect the repository of the workflow?
```suggestion
          tags: ghcr.io/Caltech-IPAC/firefly:${{ steps.image_tag.outputs.tag }}
```

I'd guess, there must be some github workflow variable we can use here instead of hardcoding the `org/repository`

docs/how-to/github-action.md
```diff
@@ -0,0 +1,484 @@
...
+    steps:
+      - run: echo "Workflow created"
+```
+
+### Develop and test the workflow
+
+Work in a separate development branch (for example, `FIREFLY-1234-my-workflow`) while iterating on the workflow.  
+The workflow exists only in this branch, it will **not appear in the GitHub Actions UI** until it is merged into the default branch (e.g., `dev`).  
+
+To test safely before merging, use `act` locally. Refer to the **"How to Safely Develop and Test GitHub Workflows"** section for
```
jaladh-singhal, line 216: I didn't know it is possible - interesting!

jaladh-singhal (approved):
Looks good to me, was easy to follow thanks to all the comments. And thank you for adding a comprehensive how to guide.

.github/workflows/build_publish.yml
jaladh-singhal: I don't think you need to update an action unless it has some security vulnerability or newer version has a feature we need.

Additionally, we can add [Dependabot](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/dependabot-quickstart-guide) to our repo. It will open issue/PRs if some action needs to upgraded if the current version in our workflows is insecure.

.github/workflows/build_publish.yml
loitly: We will resolve this offline.

docs/how-to/github-action.md
loitly: Yes, I should clarify that it has to be the parent directory of firefly, so that all dependent directories are visible.

.github/workflows/build_publish.yml
loitly: I repaced `tag_name` with `git_tag` and `image_tag` to allow them to have different values.

PR #1919, opened 2026-03-03
FIREFLY-1951-ghaction-build-publish

loitly:
Additional follow-up changes to FIREFLY-1951 github actions PR.

- move reusable logic to local action
- fix upper-case tag name not supported error


robyww (approved):
good. now we have examples of how to write actions.

jaladh-singhal:
Oh cool! So we can write local actions too (without needing to publish them - what we talked about in standup)

PR #1922, opened 2026-03-18
FIREFLY-1943: React Compiler bugs

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1943
- Fixes bugs introduced in code after enabling react compiler. See ticket for list of bugs. 
- Reason for most bugs: before react compiler, incidental re-renders would make some of the logic work, but with react compiler enabled, state renders became more obvious (and would happen only when necessary), often leading to stale values and exposing existing bugs (or rather bad code, because everything still worked fine). 
- **Bug 1: SIA Search**:
  - The issue with `constraints` is probably related to `makePanelStatusUpdater`. The way we update constraints here is really complicated, and `constraints` were not updated properly after enabling react compiler and led to stale values. 
    - I simplified it by reading the current field values that we calculate constraints from and depending on those changing to calculate/update `constraintResult`. 
  - if this refactor looks good, I will address the remaining usage of `makePanelStatusUpdater` in the new compiler ticket (see note below for other bugs found for this ticket). It's used in `ObjectIDSearch`, `ObsCore`, and a few other components. 
    - While there are no bugs there yet, it'll just make for better code to repeat the above there and not rely on the `makePanelStatusUpdater` flow. 

- **Bug 2**: **Spherex TAP Search**:
  - This is also related to `constraints`. By the time we called `getAdqlQuery`, the spatial constraint fragment was missing (after enabling react compiler). The above fix (stopped relying on `makePanelStatusUpdater`) and directly computing and updated `constraintResult` from the values it depends on fixed this as well 
    - (BTW, explicit use of `useMemo` helps make sure we use the previous constraintResult value of none of the spatial inputs it relies on change, kind of similar to `makePanelStatusUpdater` logic of checking if new/old values differ)

- **Bug 3**: fixed by Trey in a previous PR

**_Note_**: I found 2 more bugs: The TAP service dropdown and moving between Single and Multi-Object in Spatial Search. The fix for both is likely the same (in `TapSearchRootPanel` and `TapUtil`). I attempted fixing them in this PR, but the changes were a bit complex & out of scope for this PR. I will address those 2 in a new React Compiler ticket. 

I will turn off React Compiler before merging. 

**Testing**:
[Firefly](https://fireflydev.ipac.caltech.edu/firefly-1943-react-compiler-bugs/firefly), [Euclid](https://firefly-1943-react-compiler-bugs.irsakubedev.ipac.caltech.edu/applications/euclid) and [Spherex](https://firefly-1943-react-compiler-bugs.irsakubedev.ipac.caltech.edu/applications/spherex)
- Test bugs found in ticket for Firefly and Spherex
  - TAP search, SIAv2 search, Spectral Image Search on Spherex
- For Euclid, test TAP to do a search and check ADQL in job history to ensure the search was correct
- I tested the apps with react compiler turned off as well, they work as expected with the changes above.

### DIFF
--- a/src/firefly/js/ui/tap/Constraints.js
+++ b/src/firefly/js/ui/tap/Constraints.js
@@ -3,10 +3,27 @@ import React from 'react';
 
 export const ConstraintContext = React.createContext({});
 
+export function hasAdqlConstraint(constraintObj) {
+    return Boolean(
+        constraintObj?.adqlConstraint ||
+        constraintObj?.adqlConstraintsAry?.length
+    );
+}
+
+function getAdqlConstraintsList(constraintObj) {
+    if (constraintObj?.adqlConstraintsAry?.length) return constraintObj.adqlConstraintsAry;
+    if (constraintObj?.adqlConstraint) return [constraintObj.adqlConstraint];
+    return [];
+}
+
+function isConstraintsValid(constraintObj) {
+    return !constraintObj?.constraintErrors?.length;
+}
+
 export function isTapUpload(tapBrowserState) {
     const {constraintFragments}= tapBrowserState;
     return [...constraintFragments.values()]
-        .some( (c) => Boolean(c.uploadFile && c.TAP_UPLOAD && c.adqlConstraint));
+        .some((c) => Boolean(c.uploadFile && c.TAP_UPLOAD && hasAdqlConstraint(c)));
 }
 
 /**
@@ -16,7 +33,8 @@ export function isTapUpload(tapBrowserState) {
  */
 export function getUploadConstraint(tapBrowserState) {
     const {constraintFragments}= tapBrowserState;
-    return [...constraintFragments.values()].find( (c) => Boolean(c.uploadFile && c.TAP_UPLOAD && c.adqlConstraint));
+    return [...constraintFragments.values()]
+        .find((c) => Boolean(c.uploadFile && c.TAP_UPLOAD && hasAdqlConstraint(c)));
 }
 
 export function getTapUploadSchemaEntry(tapBrowserState) {
@@ -38,26 +56,17 @@ export function getHelperConstraints(tapBrowserState) {
     const {constraintFragments}= tapBrowserState;
     const adqlConstraints = [];
     const adqlConstraintErrorsArray = [];
-    const siaConstraints = [];
-    // adqlComponents can apparently be modified during iteration in the forEach...
+
     Array.from(constraintFragments.values()).forEach((constraintObj) => {
-        if (!constraintObj.adqlConstraintErrors?.length) {
-            if (constraintObj.adqlConstraint) {
-                adqlConstraints.push(constraintObj.adqlConstraint);
-            }
+        if (isConstraintsValid(constraintObj)) {
+            adqlConstraints.push(...getAdqlConstraintsList(constraintObj));
         } else {
-            adqlConstraintErrorsArray.push(constraintObj.constraintErrors);
-        }
-        if (!constraintObj.constraintErrors?.length) {
-            if (constraintObj.siaConstraints?.length > 0) {
-                siaConstraints.push(...constraintObj.siaConstraints);
-            }
-        } else {
-            adqlConstraintErrorsArray.push(constraintObj.constraintErrors);
+            adqlConstraintErrorsArray.push(...constraintObj.constraintErrors);
         }
     });
+
     return {
-        valid: adqlConstraintErrorsArray?.length === 0,
+        valid: adqlConstraintErrorsArray.length === 0,
         messages: adqlConstraintErrorsArray,
         where: adqlConstraints.join('\n      AND ')
     };
--- a/src/firefly/js/ui/tap/SpatialSearch.jsx
+++ b/src/firefly/js/ui/tap/SpatialSearch.jsx
@@ -22,19 +22,15 @@ import {calcCornerString, PolygonDataArea} from '../CatalogSearchMethodType.jsx'
 import {FieldGroupCtx, ForceFieldGroupValid} from '../FieldGroup.jsx';
 import {ListBoxInputField} from '../ListBoxInputField.jsx';
 import {RadioGroupInputField} from '../RadioGroupInputField.jsx';
-import {useFieldGroupRerender, useFieldGroupValue, useFieldGroupWatch} from '../SimpleComponent.jsx';
+import {useFieldGroupRerender, useFieldGroupValue, useFieldGroupWatch,
+    useFieldValueOnly} from '../SimpleComponent.jsx';
 import {SizeInputFields} from '../SizeInputField.jsx';
 import {DEF_TARGET_PANEL_KEY} from '../TargetPanel.jsx';
 import {ConstraintContext} from './Constraints.js';
 import {ROW_POSITION, SEARCH_POSITION} from './Cutout';
 import {getDataServiceOption} from './DataServicesOptions';
-import {
-    DebugObsCore,
-    getPanelPrefix,
-    makeCollapsibleCheckHeader,
-    makeFieldErrorList,
-    makePanelStatusUpdater,
-} from './TableSearchHelpers.jsx';
+import {DebugObsCore, getPanelPrefix, makeCollapsibleCheckHeader, makeConstraintEntry, makeEmptyConstraints,
+    makeFieldErrorList} from './TableSearchHelpers.jsx';
 import {showUploadTableChooser} from '../UploadTableChooser.js';
 import {
     getAsEntryForTableName, getColumnAttribute, getTapServiceByURL, makeUploadSchema, maybeQuote,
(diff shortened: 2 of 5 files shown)


src/firefly/js/ui/tap/Constraints.js
```diff
@@ -3,10 +3,23 @@ import React from 'react';
...
+        constraintObj?.adqlConstraint ||
+        constraintObj?.adqlConstraintsAry?.length
+    );
+}
+
+function getAdqlConstraintsList(constraintObj) {
+    if (constraintObj?.adqlConstraintsAry?.length) return constraintObj.adqlConstraintsAry;
+    if (constraintObj?.adqlConstraint) return [constraintObj.adqlConstraint];
+    return [];
+}
```
robyww, line 17: good! `getAdqlConstraintsList` is a real improvement.

robyww: You should add some sort of `hasErrors` to work with `constraintObj.constraintErrors?.length` or call is `isContraintsValid`

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -321,13 +350,11 @@ function getSpacialLayoutMode(spacialType, obsCoreEnabled, canUpload) {
...
-    const spacialType= getVal(SPATIAL_TYPE) ?? SINGLE;
-    const spatialMethod= getVal(SpatialMethod)??CONE_CHOICE_KEY;
-    const closest= getVal(Closest)??'';
-    const cornerCalcTypeValue= getVal(cornerCalcType)??'image';
-    const spatialRegOpValue= getVal(SpatialRegOp) ?? SpatialRegOpType.CONTAINS_POINT;
+    const spacialType = useFieldGroupValue(SPATIAL_TYPE)[0]() ?? SINGLE;
+    const spatialMethod = useFieldGroupValue(SpatialMethod)[0]() ?? CONE_CHOICE_KEY;
+    const closest = useFieldGroupValue(Closest)[0]() ?? '';
+    const cornerCalcTypeValue = useFieldGroupValue(cornerCalcType)[0]() ?? 'image';
+    const spatialRegOpValue = useFieldGroupValue(SpatialRegOp)[0]() ?? SpatialRegOpType.CONTAINS_POINT;
```
robyww: good. I think I went down the wrong path using getVal here.

robyww (changes requested):
This is good work and good  analysis. I like your big solution of `useMemo` I just want you to make it more reusable.

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -236,15 +228,52 @@ export function SpatialSearch({sx, cols, serviceUrl, serviceLabel, serviceId, co
...
+            ...constraints,
+            constraintErrors: [...(constraints.errAry ?? [])],
+            simpleError: constraints.errAry?.[0] ?? '',
+        };
+    }, [isSpatialPanelActive, columnsModel, obsCoreEnabled, uploadInfo, tableName, canUpload,
+        useSIAv2, targetWp, spatialRegOp, spatialType, spatialMethodVal,
+        radiusVal, polygonCornersVal, centerLonVal, centerLatVal,
+        uploadCenterLonVal, uploadCenterLatVal, cornerCalcTypeVal, closestVal,
+    ]);
+
```
robyww, line 244: I _really_ like this approach. It is much clean and easier to follow. However, I think it is to verbose and hard to consistently reuse. So here are a couple of ideas.

this will help a lot
```js
fldListAry.map
```

### Idea 1
define in `TableSearchHelpers.js`
```js
function makeConstrainstHelper(active, make)  {
        if (!active) return { adqlConstraintsAry: [], constraintErrors: [], simpleError: '', };

        const constraints = make();
        return {
            ...constraints,
            constraintErrors: [...(constraints.errAry ?? [])],
            simpleError: constraints.errAry?.[0] ?? '',
        };
 }
```

then your use Memo function
```js
    const constraintResult = React.useMemo(() => {
        return doContrains(
            isSpatialPanelActive,
            () => makeSpatialConstraints( columnsModel, obsCoreEnabled, makeFldObj(fldListAry),
                uploadInfo, tableName, canUpload, useSIAv2)
            );
    }, [...fldListAry.map((v)=> getVal(v)) ]);
```



### Idea 2

define in `TableSearchHelpers.js` the short functions  `makeEmptyConstrains()` and a  `makeConstrainstEntry(constraints)`

then
```js
   const constraintResult1 = React.useMemo(() => {
        if (!isSpatialPanelActive) return makeEmptyConstrains();

        const constraints = makeSpatialConstraints(
            columnsModel, obsCoreEnabled, makeFldObj(fldListAry),
            uploadInfo, tableName, canUpload, useSIAv2
        );

        return makeConstrainstEntry(constraints);
    }, [...fldListAry.map((v)=> getVal(v)) ]);

```

src/firefly/js/ui/tap/SpatialSearch.jsx
```diff
@@ -321,13 +350,11 @@ function getSpacialLayoutMode(spacialType, obsCoreEnabled, canUpload) {
...
                                  hipsUrl, centerWP, fovDeg, capabilities, embeddedInHiPS, slotProps}) => {
 
-    const {getVal}= useContext(FieldGroupCtx);
-
-    const spacialType= getVal(SPATIAL_TYPE) ?? SINGLE;
-    const spatialMethod= getVal(SpatialMethod)??CONE_CHOICE_KEY;
-    const closest= getVal(Closest)??'';
-    const cornerCalcTypeValue= getVal(cornerCalcType)??'image';
-    const spatialRegOpValue= getVal(SpatialRegOp) ?? SpatialRegOpType.CONTAINS_POINT;
+    const spacialType = useFieldGroupValue(SPATIAL_TYPE)[0]() ?? SINGLE;
```
jaladh-singhal: This maybe just preference issue, I find useStoreConnector cleaner for these cases. I consider useFieldGroupValue more for cases where I want a getter function that I'll call later.

```suggestion
    const spacialType = useStoreConnector(() => getFieldVal(groupKey, SPATIAL_TYPE, SINGLE));
```

You can get groupKey by doing `const {groupKey} = useContext(FieldGroupCtx);`

src/firefly/js/ui/tap/SpatialSearch.jsx
jaladh-singhal: I vote for idea 2

jaladh-singhal (changes requested):
Code looks improved, left some minor comments.

Regarding testing, the spatial constraints are working now but other constraints like spectral constraints no longer get added to constraint fragment list.
1. SIA: make a search with location constraints, check job/table info -> job url, you will see `POS` param in URL query string. Now make another search with same location constraints and with some spectral coverage constraints, check job url, you will not see `BAND` param in URL query string. Do this same search again, BAND will now appear.
2. Spectral Image aka SPHEREx ObsTAP: same as above check wavelength params and inspect ADQL query in job info, the constraint on em_min, em_max columns won't appear in the WHERE clause. Same for time constraints

src/firefly/js/ui/tap/SpatialSearch.jsx
robyww: I would prefer to stay consistent with `useFieldGroupValue`. Maybe we should make is support a default

robyww: You could defined this in SimpleComponent. I think it would work. There are many uses.

```js
export function useFieldValueOnly(fieldKey, def, gk) {
    return useFieldGroupValue(fieldKey,gk)[0]() ?? def;
}
```

kpuriIpac:
@robyww  @jaladh-singhal, I addressed the feedback & did lots of testing. I would like you both to take a final look before I merge. You can look at the last commit, and the test builds below. 

@jaladh-singhal: I confirmed the bugs in other constraints (coverage, timing, etc. are pre-existing when you just turn on React Compiler. Tested this on both firefly TAP/SIA and Spherex ObsTAP). I do think the fix for this is very similar to the fix for spatial constraints, but I will address all of the other constraints together in my next React Compiler ticket (already created). 

Builds with React Compiler turned **_on_**: 
- **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1943-react-compiler-bugs/firefly
- **Spherex**: https://firefly-1943-react-compiler-bugs.irsakubedev.ipac.caltech.edu/applications/spherex

Builds with React Compiler turned **_off_** : 
- **Firefly**: https://fireflydev.ipac.caltech.edu/temp-1943-react-compiler/firefly
- **Spherex:** https://temp-1943-react-compiler.irsakubedev.ipac.caltech.edu/applications/spherex
  - (I know we discussed these 2 builds may not be necessary, but I thought better safe than sorry - just do a quick test to make sure all constraints work here. There should be no bugs as react compiler is turned **_off_**)

robyww (approved):

jaladh-singhal (approved):
Thanks @kpuriIpac, I can't reproduce the issues I reported in the builds with react compiler turned off. Good to merge.

PR #1928, opened 2026-04-01
IRSA-7653: Fix precision handling for SizeInputField causing round-trip errors

jaladh-singhal:
See IFE PR: https://github.com/IPAC-SW/irsa-ife/pull/461

- Unlike WavelengthInputField, SizeInputField wasn't calculating precision (number of digits after decimal) based on each unit. Fixed it.
- Also added precision based rounding to min and max bounds to prevent errors in bound validation due to floating point comparison (esp when arcsec is converted to base unit degree and then compared)

### DIFF
--- a/src/firefly/js/util/MathUtil.js
+++ b/src/firefly/js/util/MathUtil.js
@@ -20,8 +20,19 @@ export function getDecimalPlaces(range, numSigDigits) {
     return numDecPlaces;
 }
 
-/*
- * remove trailing zero from toFixed result
+
+/**
+ * Formats a number to at most `digits` decimal places by rounding.
+ * Unlike `toFixed`, the result is a number (not a string) and has no unnecessary trailing zeros.
+ *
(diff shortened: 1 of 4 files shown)


src/firefly/js/util/MathUtil.js
```diff
@@ -20,8 +20,19 @@ export function getDecimalPlaces(range, numSigDigits) {
...
+ * @param {number} digits - Maximum number of decimal places to keep.
+ * @returns {number} The formatted number with at most `digits` decimal places.
+ *
+ * @example
+ * toMaxFixed(3.14159, 4) // 3.1416
+ * toMaxFixed(3.10000, 4) // 3.1
+ * toMaxFixed(3,       2) // 3
  */
 export function toMaxFixed(floatNum, digits) {
     return parseFloat(Number(floatNum).toFixed(digits));
```
robyww, line 38: the function doc is very helpful

robyww (approved):
Looks good.

PR #1933, opened 2026-04-08
Firefly-1151: DCE support for upload

robyww:
#### [Firefly-1151](https://jira.ipac.caltech.edu/browse/FIREFLY-1151):  DCE support for upload

- support upload from DCE
- adde a SIAv2 upload search processor

#### Testing
- https://firefly-1151-dce-upload.irsakubedev.ipac.caltech.edu/irsaviewer/dce
- upload a file
- example file: [3rows.tbl.zip](https://github.com/user-attachments/files/26582148/3rows.tbl.zip)


lrebull:
hm.
There is an upload tab in DCE on ops. Neither ops nor your test build's upload tab can handle the "tbl.zip" file format.
Both ops and your test build can load tbl files.
Your test build has the legend in the coverage map, but I don't think that is what you are asking me to test.
I am not sure exactly what I am supposed to be testing here..?

lrebull:
OHHH this is for searching not just generically the upload tab! Now I understand!

Yes, this works.

In the longer term, we should think about the UX in the context of serving (displaying) results for a list of targets. But this is great, and another step towards decommissioning Atlas. :)

kpuriIpac (approved):
Spent some time looking at the code and testing the UI flow. Looks great!

PR #1936, opened 2026-04-10
FIREFLY-1331: Add Version validation for compatibility check between python client and firefly server

jaladh-singhal:
Fixes [FIREFLY-1331](https://jira.ipac.caltech.edu/browse/FIREFLY-1331)

Exposed the version info as `cmd=CmdVersion` at the commands servelet endpoint (`CmdSrv/sync`). 

Firefly client is not using the `raw=true` param but it maybe helpful in future if validation logic becomes complicated.

See Firefly client PR: https://github.com/Caltech-IPAC/firefly_client/pull/79

## Testing 
1. http://aac8ce8bef8604e7ca9d38b934ab83c5-4855331a01d9fb87.elb.us-east-1.amazonaws.com/firefly-1331/firefly (yes I managed to deploy the PR build on AWS using helm!)

   - Open the version dialog and check if it matches the JSON returned by `/CmdSrv/sync?cmd=CmdVersion` [query params](http://aac8ce8bef8604e7ca9d38b934ab83c5-4855331a01d9fb87.elb.us-east-1.amazonaws.com/firefly-1331/firefly/CmdSrv/sync?cmd=CmdVersion)

   - Check that `/CmdSrv/sync?cmd=CmdVersion?raw=true` [query params](http://aac8ce8bef8604e7ca9d38b934ab83c5-4855331a01d9fb87.elb.us-east-1.amazonaws.com/firefly-1331/firefly/CmdSrv/sync?cmd=CmdVersion&raw=true) returns primitive version object which can also be useful in some cases

2. https://firefly-1331-version-validation.irsakubedev.ipac.caltech.edu/irsaviewer/CmdSrv/sync?cmd=CmdVersion

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
@@ -10,6 +10,9 @@ import edu.caltech.ipac.firefly.data.ServerEvent;
 import edu.caltech.ipac.firefly.data.ServerParams;
 import edu.caltech.ipac.firefly.data.userdata.UserInfo;
 import edu.caltech.ipac.firefly.messaging.JsonHelper;
+import edu.caltech.ipac.firefly.server.util.VersionUtil;
+import edu.caltech.ipac.util.KeyVal;
+import edu.caltech.ipac.util.serialization.Serializer;
 import edu.caltech.ipac.firefly.server.events.FluxAction;
 import edu.caltech.ipac.firefly.server.events.ServerEventManager;
 import edu.caltech.ipac.firefly.server.security.SsoAdapter;
@@ -24,6 +27,7 @@ import org.json.simple.parser.ParseException;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -149,6 +153,23 @@ public class AppServerCommands {
     }
 
 
+
+    public static class GetVersion extends ServCommand {
+        public String doCommand(SrvParam params) throws Exception {
+            boolean raw = params.getOptionalBoolean("raw", false);
+            Object versionData;
+            if (raw) {
+                versionData = VersionUtil.getAppVersion();
+            } else {
+                Map<String, String> versionMap = new LinkedHashMap<>();
+                for (KeyVal<String, String> kv : VersionUtil.getVersionInfo()) {
+                    versionMap.put(kv.getKey(), kv.getValue());
+                }
+                versionData = versionMap;
+            }
+            return Serializer.getJsonMapper().writeValueAsString(versionData);
+        }
+    }
 
     public static class GetAlerts extends ServCommand {
         public String doCommand(SrvParam params) throws Exception {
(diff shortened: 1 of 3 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
```diff
@@ -150,6 +154,15 @@ public String doCommand(SrvParam params) throws Exception {
 
 
 
+    public static class GetVersion extends ServCommand {
+        public String doCommand(SrvParam params) throws Exception {
+            boolean raw = Boolean.parseBoolean(params.getOptional("raw"));
```
robyww: you should use 
```java
parmas.getOptionalBoolean(“raw”,false)
```

src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
```diff
@@ -150,6 +154,15 @@ public String doCommand(SrvParam params) throws Exception {
...
+            Object data = raw ? VersionUtil.getAppVersion()
+                    : VersionUtil.getVersionInfo().stream().collect(Collectors.toMap(KeyVal::getKey, KeyVal::getValue));
+            return Serializer.getJsonMapper().writeValueAsString(data);
```
robyww: it is a little verbose but wrap all out returns in an object.

this is SimpleJson taken from `GetUserInfo` (a few lines below)  just to demonstrate
```
map.put( "success", true);
map.put("data", data);
```

robyww (approved):
I have not looked at the python yet but I think this is fine after the clean up I suggested.

src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
loitly: Using Jackson for JSON conversion is a good approach, and I’d like to gradually migrate SimpleJson to Jackson.

The added complexity here comes from converting a list of <name, value> pairs into a map. This could be simplified if VersionUtil.getVersionInfo() returned a Map directly. From what I’ve seen, all current usages end up converting or streaming the list anyway to access the values.

loitly (approved):
Changes look good.

src/firefly/java/edu/caltech/ipac/firefly/server/AppServerCommands.java
jaladh-singhal: @robyww I was using the imperative approach earlier to create map by looping over list. I reverted functional streaming approach back to it.

@loitly this still keeps the jackson JSON conversion since I am working with map object directly.

robyww: > it is a little verbose but wrap all out returns in an object.

I was reviewing from my ipad last night and the above line got garbled. Let me try that again. 

Our standard return style is a little verbose, but we always wrap the json data returned in an outer json object. You are just returning the data and it should be wrapped in an object `{status: true, data:data}`

jaladh-singhal: Oh I see. I jumped the gun and merged already. Should I revert this merge and make this change in this branch? OR just push this change directly to dev?

PR #1938, opened 2026-04-13
Firefly 1939 tap schema unquoted

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1939

The bug described in the ticket is no longer reproducible, but the changes still make sense. We just need to perform regression testing.
Test: https://fireflydev.ipac.caltech.edu/firefly-1939-tap-schema-unquoted/firefly/

The other update includes adding autoscaling to the Helm chart and revamping the Helm documentation. 
Please review the docs: https://github.com/Caltech-IPAC/firefly/tree/FIREFLY-1939-tap-schema-unquoted/helm#readme

### DIFF
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -305,6 +305,7 @@ def main():
             f"-XX:MaxRAMPercentage={os.getenv('MAX_RAM_PERCENT', '80')}",
             "-XX:+UnlockExperimentalVMOptions",
             "-XX:TrimNativeHeapInterval=30000",
+            "-XX:+UseZGC",
             f"-DADMIN_USER={admin_user}",
             f"-DADMIN_PASSWORD={admin_password}",
             f"-Dhost.name={os.getenv('HOSTNAME', '')}",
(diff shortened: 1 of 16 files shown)


docker/entrypoint.py
```diff
@@ -305,6 +305,7 @@ def main():
             f"-XX:MaxRAMPercentage={os.getenv('MAX_RAM_PERCENT', '80')}",
             "-XX:+UnlockExperimentalVMOptions",
             "-XX:TrimNativeHeapInterval=30000",
+            "-XX:+UseZGC",
```
robyww, line 308: I have not been keeping with gc's. I am glad you noticed this.

helm/templates/hpa.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
```
robyww, line 3: I would like to understand this. I think I would like to discuss a general concern about pod scaling and losing cache.

I am particularly concerned about long term cache like HiPS or big images like are more common now.

If pod are coming and going it defeats our caching. Should we do more caching outside of the pod?

robyww (approved):
It all look good.

I guess it is OK we go to auto-scaling but I have several concerns. We need to dedicate a meeting to it soon.

PR #1939, opened 2026-04-15
Improve AlertViewer UI/UX 

kpuriIpac:
**SUIT PR**: https://github.com/lsst/suit/pull/84
- Remove Job Monitor tab for alert viewer 
- Default tab when you hit `/alertviewer` is now the `Alert` tab (`/alertviewer/upload`) instead of the `Results` tab.
- Simplified `Results` (landing page) text for `AlertViewer` 
  - required a small custom component to modify the bottom section of the `LandingPage`, I could have probably done this through `slotProps`, but it was proving to be trickier (was affecting Euclid/Spherex) and `LandingPage` is used in several places so I thought it was best to create a simple custom component instead in this case 
  - And the changes in RoutedApp are to support being able to call `LandingPage` as well (to have more control over customizing it)

**Testing**: 
- **Firefly's AlertViewer**: [https://fireflydev.ipac.caltech.edu/alert-viewer-changes/firefly/alertviewer](https://fireflydev.ipac.caltech.edu/alert-viewer-changes/firefly/alertviewer)
- **SUIT's AlertViewer**: https://enable-alert-viewer.irsakubedev.ipac.caltech.edu/suit/alertviewer
  - Ensure you're routed to Alert tab directly. Inspect the Results tab (landing page) text. 
  - Checkout this branch locally if you want to test submitting Alert IDs (sample IDs: `170112073844916273`, `170059278837088375`)
  - URL API should still work (if trying on the test builds above, you should get a missing/invalid token error, locally it should take you to the results tab with results loaded).

### DIFF
--- a/src/firefly/js/apps/alertviewer/alertviewer.js
+++ b/src/firefly/js/apps/alertviewer/alertviewer.js
@@ -1,10 +1,15 @@
 import React from 'react';
+import QueryStats from '@mui/icons-material/QueryStats';
+import {Sheet, Stack, Typography} from '@mui/joy';
 
 import {createRouter} from 'firefly/templates/router/RoutedApp.jsx';
 import {FormWatcher} from 'firefly/templates/router/RouteHelper';
+import {LandingPage} from 'firefly/templates/fireflyviewer/LandingPage.jsx';
 import {UploadPanel} from 'firefly/apps/alertviewer/AlertUploadPanel';
 import {AlertResultView} from 'firefly/apps/alertviewer/AlertResultView';
 import {getAlertViewerCommands} from 'firefly/api/webApiCommands/AlertViewerCommands';
+import {dispatchShowDropDown} from 'firefly/core/LayoutCntlr';
+import {dispatchOnAppReady} from 'firefly/core/AppDataCntlr';
 
 function getBasename() {
     const p = window.location.pathname || '';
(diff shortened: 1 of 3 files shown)


src/firefly/js/apps/alertviewer/alertviewer.js
```diff
@@ -27,19 +32,54 @@ const routes = [
...
+        backgroundMonitor: false,
         getRouter: createRouter(basename, routes),
         appIcon: 'n/a',
+        landingPage: <AlertViewerLanding/>,
         slotProps: {
             drawer: { drawerWidth:'24rem'},
-            landing: {
-                desc: 'Some information about the alert viewer here. This is the landing page. This text will be changed to something more useful in the future.',
-                bgImage: null,
-            }
```
loitly: I don’t think this is necessary. slotProps.landing was created specifically for this use case. If you need a custom component, you can define it like this:
```js
landing: {
    component: <AlertViewerLanding/>
}
```

src/firefly/js/apps/alertviewer/alertviewer.js
```diff
@@ -50,4 +90,11 @@ export const alertviewer = {
...
 
 
-function init() {}
+function init() {
+    const path = window.location.pathname || '';
+    const search = window.location.search || '';
+    const hasUrlApi = search.includes('api=');
+    if (!hasUrlApi && (path === basename || path === `${basename}/`)) {
+        dispatchOnAppReady(() => dispatchShowDropDown({view: 'AlertUpload'}));
+    }
```
loitly: Is this logic correct?  Normally, `?api` would show the api help page, but I am not seeing that happen.
Also, using URLSearchParams might help when working with URL parameters.

```js
const { pathname, search } = window.location;
const params = new URLSearchParams(search);
params.has('api');

```

loitly (changes requested):
Please revert `customLandingPage` changes if my suggested solution works.   Thanks.

src/firefly/js/apps/alertviewer/alertviewer.js
kpuriIpac: yes good suggestion, thank you. also fixes the `?api` bug, it should now show the api help page (after I push my commit). 

The **_if_** check here was needed because we want to re-route user to `/alertviewer/upload` path when they try to access `/alertviewer`, but if I didn't add the check for `!hasUrlApi`, it would also re-route you to `/upload` path for the URL API instead of taking you directly to the results (when, for instance, you try to access somethign like: `/alertviewer?api=alert&id=170059278837088375`). 

open to better suggestions for this re-route logic though.

robyww (commented):
UI changes look good. Go ahead and merge when @loitly approves.

loitly (approved):
Looks good.  Thanks for the changes.

PR #1940, opened 2026-04-17
IRSA-7673: Update job image results and table container for SPHEREx Mosaic Tool

jaladh-singhal:
See ife PR for tickets and testing: https://github.com/IPAC-SW/irsa-ife/pull/462

1. `TableContainer` defaults to `SingleTable` when single table which is different from `TabPanel` in 2 ways: 
     1. when table load is completed with error it doesn't render table title with a close icon button so user gets stuck with failed table view and can't clear it
     2. the table toolbar is narrower in height than tab panel which makes layout jumpy vertically when more tables are added
   
     I exposed a prop to force using TabPanel even for single table case to overcome above two issues.

2. `loadImageResult` uses image title from result href's filename instead of using it from job request's title which is implicit in `loadTableResult`. This becomes stark when you have multiple image results plotted and you are trying to identify which image came from which job. Fixed it.


loitly (approved):
Changes look good.

PR #1943, opened 2026-04-22
FIREFLY-1997: Alertviewer use URL API approach for results

kpuriIpac:
**Ticket**: [FIREFLY-1997](https://jira.ipac.caltech.edu/browse/FIREFLY-1997)
- Use URL API approach for all alertviewer searches and open them in a new tab 

**Testing**: 
- **Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1997-alerts-url-api/firefly/alertviewer
  - you should still see the URL API be created and a new tab above, but it'll fail due to token authentication
  - to test results working, checkout my branch and run firefly locally


robyww (changes requested):
Do don't have the logic quite right here.
1. You should return `UploadPanel` is is confusing. Call it something like `AlertIdPanel`
2. The panel need to take a parameter `loadInPlace=true or  false` .
3. if true then load in the application, if false load via the url.
4. This panel parameter is pass in `routes` and `dropdownPanels`

robyww (approved):
Looks right

PR #1944, opened 2026-04-24
IRSA-7687: Make BgMaskPanel's layout responsive and improve the error state

jaladh-singhal:
See ife PR: https://github.com/IPAC-SW/irsa-ife/pull/464

- Refactored `BgMaskPanel`:
   - made button layout responsive via CSS container queries so it adapts to the panel's own width esp when panel becomes small in image grid
   - made it accept a `renderError(errorMsg, infoButton)` render prop for custom error UIs, with the existing default behavior unchanged for all other callers

- I noticed two run-time issues when loading SP results because of accessing prop on undefined, fixed them by using `?.`

Do some regression testing with long running table search and resize the table width to test responsive layout.

### DIFF
--- a/src/firefly/js/visualize/PlotViewUtil.js
+++ b/src/firefly/js/visualize/PlotViewUtil.js
@@ -188,7 +188,7 @@ export const isActivePlotView= (visRoot,plotId) => visRoot.activePlotId===plotId
  * @param {VisRoot} visRoot - root of the visualization object in store
  * @return {PlotView} the active plot view
  */
-export const getActivePlotView= (visRoot) => visRoot?.plotViewAry.find( (pv) => pv.plotId===visRoot.activePlotId);
+export const getActivePlotView= (visRoot) => visRoot?.plotViewAry?.find( (pv) => pv.plotId===visRoot.activePlotId);
 
 /**
  *
--- a/src/firefly/js/visualize/iv/ImageViewer.jsx
+++ b/src/firefly/js/visualize/iv/ImageViewer.jsx
@@ -7,8 +7,8 @@ import {bool, func, string} from 'prop-types';
 import React, {memo, useState, useEffect, useRef, useDeferredValue, use, Suspense} from 'react';
 import {omit} from 'lodash';
 import shallowequal from 'shallowequal';
-import {getPlotViewById,getAllDrawLayersForPlot} from '../PlotViewUtil.js';
-import {ImageViewerView} from './ImageViewerDecorate.jsx';
+import {getPlotViewById, getAllDrawLayersForPlot, getPlotViewProxyById} from '../PlotViewUtil.js';
+import {ImageViewerView, getImageViewerDecorateSx} from './ImageViewerDecorate.jsx';
 import {visRoot, ExpandType} from '../ImagePlotCntlr.js';
 import {extensionRoot} from '../../core/ExternalAccessCntlr.js';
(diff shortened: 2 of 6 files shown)


src/firefly/js/visualize/PlotViewUtil.js
```diff
@@ -188,7 +188,7 @@ export const isActivePlotView= (visRoot,plotId) => visRoot.activePlotId===plotId
  * @param {VisRoot} visRoot - root of the visualization object in store
  * @return {PlotView} the active plot view
  */
-export const getActivePlotView= (visRoot) => visRoot?.plotViewAry.find( (pv) => pv.plotId===visRoot.activePlotId);
+export const getActivePlotView= (visRoot) => visRoot?.plotViewAry?.find( (pv) => pv.plotId===visRoot.activePlotId);
```
robyww, line 191: were you getting an exception? plotViewAry is never suppose to be undefined.

loitly (approved):
Looks good.

src/firefly/js/visualize/PlotViewUtil.js
jaladh-singhal: Yes I was getting error in spectrophotometry search. It is undefined in this caseL https://github.com/Caltech-IPAC/firefly/blob/b78250519b4fae36514425342990376e85bc269c/src/firefly/js/visualize/PlotViewUtil.js#L35

robyww: interesting I did not thing it was ever saved in the store like that, but something in the reducer must be calling it.

src/firefly/js/visualize/iv/ImageViewer.jsx
```diff
@@ -114,24 +114,31 @@ ImageViewer.propTypes= {
 };
 
 
-export function ImageViewerPlaceHolder({fromWorkingTask:task=false, ...rest}){
-    return task ? <WorkingTaskPlaceHolder {...rest}/> : <MessagePlaceHolder {...rest}/>;
+export function ImageViewerPlaceHolder({fromWorkingTask:task=false, plotId, ...rest}){
+    const {pv, vr} = useStoreConnector(() => {
+        const vr= visRoot();
+        return {vr, pv: getPlotViewById(vr, plotId) ?? getPlotViewProxyById(vr, plotId)};
```
robyww, line 120: I you sure it is necessary to call `getPlotViewById(vr, plotId)`? In theory the proxy is removed when the image is created. Maybe there is some sort of intermediate moment

robyww (approved):
Looks good.

One little UI thing. I think you want to set `overflow: auto` and not `scroll` on the error message

Screenshot 2026-04-28 at 10 28 13 AM

src/firefly/js/visualize/iv/ImageViewer.jsx
jaladh-singhal: This is shared with both Working Task and Message placeholder - that's why I'm calling `getPlotViewById(vr, plotId)` and then `getPlotViewProxyById(vr, plotId)` as a fallback for pv proxy cases. `getImageViewerDecorateSx()` mostly cares about pv being an object with plot id inside it.

jaladh-singhal:
> One little UI thing. I think you want to set overflow: auto and not scroll on the error message

@robyww good catch - this was needed for long error message earlier, I'll revert it.

PR #1948, opened 2026-04-28
FIREFLY-1977: Make Select Row Filter button optional 

kpuriIpac:
**Ticket**: https://jira.ipac.caltech.edu/browse/FIREFLY-1977
- Pass through a new prop `showSelectRowFilter` (defaults to true, but false for File Upload panel) 

**Testing**: 
**Firefly**: https://fireflydev.ipac.caltech.edu/firefly-1977-upload-filter-bug/firefly
- Upload a file in the upload panel
- Notice the Filter on Rows button is now gone (compare against https://fireflydev.ipac.caltech.edu/firefly/)
  - other filters should still be available


robyww (approved):
Looks good.

PR #1955, opened 2026-05-04
FIREFLY-1864: Improve application security and configuration hardening

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1864

**Changes:**
- Avoid using Host header when server host is explicitly configured
- Tomcat hardening: strip Server header, disable TRACE method, set autoDeploy=false
- Updated DuckDB to v1.5.2.1
  - Restrict file access via `enable_external_access=false` and `allowed_directories`
  - Query performance improvements

**Test**: https://firefly-1864-security-config.irsakubedev.ipac.caltech.edu/irsaviewer/
- Added 'server.host' properties to defined host instead of relying on http header
- Verify 404-Not Found without server info  (https://firefly-1864-security-config.irsakubedev.ipac.caltech.edu/irsaviewer/something-that-does-not-exists)
- I tested 'enable_external_access=false' from the debugger by sending SQL statement the try to read file from the local filesystem and verified it worked.
- Improved performance based on times from logs.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -200,7 +200,7 @@ RUN sed -i 's/<Connector port="8080" protocol="HTTP\/1.1"/<Connector port="8080"
     && sed -i 's/<Connector /<Connector server="" allowTrace="false" /' \
         $CATALINA_HOME/conf/server.xml \
     # Disable runtime auto-deploy
-    && sed -i 's/autoDeploy="true"/autoDeploy="false">/' \
+    && sed -i 's/autoDeploy="true"/autoDeploy="false"/' \
         $CATALINA_HOME/conf/server.xml  \
     && sed -i '/<\/Host>/i\    <Valve className="org.apache.catalina.valves.ErrorReportValve" showReport="false" showServerInfo="false" />' \
           $CATALINA_HOME/conf/server.xml
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -139,7 +139,7 @@ ARG uid=91
 ARG gid=91
 ARG TARGETARCH=amd64
 ARG TARGETOS=linux
-ARG DUCKDB_VERSION=v1.1.3
+ARG DUCKDB_VERSION=v1.5.2.1
 
 # These are the users replaceable environment variables, basically runtime arguments
 #          - Set the available memory on the command line with --memory="4g"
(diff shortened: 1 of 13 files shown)


docker/Dockerfile
```diff
@@ -195,13 +195,15 @@ COPY --from=builder /opt/work/${build_dir}/build/dist/*.war ${CATALINA_HOME}/web
...
-ENV DUCKDB_TARGET=${DUCKDB_VERSION}/${TARGETOS}_${TARGETARCH}_gcc4
-RUN mkdir -p temp/$DUCKDB_TARGET \
-    && wget -P temp/$DUCKDB_TARGET https://community-extensions.duckdb.org/$DUCKDB_TARGET/magic.duckdb_extension.gz \
-    && gunzip -f temp/$DUCKDB_TARGET/magic.duckdb_extension.gz
+        $CATALINA_HOME/conf/server.xml \
+    # Strip version from Server header and disable TRACE
+    && sed -i 's/<Connector /<Connector server="" allowTrace="false" /' \
+        $CATALINA_HOME/conf/server.xml \
+    # Disable runtime auto-deploy
+    && sed -i 's/autoDeploy="true"/autoDeploy="false">/' \
```
kpuriIpac: I think there's a small type with `>` at the end, replacing `autoDeploy="true"` with `autoDeploy="false">` (so might end up with double `>>` in the XML file.

kpuriIpac: Also related to this, I see in the ticket it was mentioned adding `deployOnStartup="false"` along with setting `autoDeploy="false"`. 

not sure if that was missed or if there's a better reason to not include that here?

kpuriIpac (approved):
Looks good! 

Few comments below. Tested for any new bugs all across `irsaviewer`. Only found `Prepare Download` doesn't work for DCE (but that's becuase I am reviewing this PR late, that bug is now fixed on `dev` so if you were to do a new build, it wouldn't show up as a bug). 

I also tested for `'enable_external_access=false'` . If you'd like, we could also add a small test to `DuckDbAdapteterTest.java`, I got this from Codex: 

```
@Test
	public void testDuckDbExternalAccessIsRestrictedToAllowedDirs() throws Exception {
		DuckDbAdapter db = new DuckDbAdapter((File) null);

		File allowedCsv = new File(ServerContext.getWorkingDir(), "duckdb-allowed.csv");
		Files.writeString(allowedCsv.toPath(), "id,name\n1,allowed\n");
		assertEquals(1, db.execQuery("select * from read_csv('%s')".formatted(allowedCsv.getAbsolutePath()), null).size());

		File deniedCsv = new File(ServerContext.getWorkingDir().getParentFile(), "duckdb-denied.csv");
		Files.writeString(deniedCsv.toPath(), "id,name\n1,denied\n");
		try {
			db.execQuery("select * from read_csv('%s')".formatted(deniedCsv.getAbsolutePath()), null);
			fail("DuckDB should block reads outside allowed_directories");
		} catch (DataAccessException expected) {
			assertNotNull(expected.getMessage());
		} finally {
			allowedCsv.delete();
			deniedCsv.delete();
		}
	}
```

If you do add this, feel free to edit as necessary. But just as is, this test worked for me.

docker/Dockerfile
loitly: Without going into details, this was intentionally not added.

PR #1957, opened 2026-05-08
IRSA-7677: Fix broken URL API for SPHEREx Spectral Image search

jaladh-singhal:
In url API of SPHEREx spectral image search, for some unclear reason, `unit` doesn't get registered in field group state of Spatial search `SizeInputField` (which manifests as `viewProps.unit=undefined|null` in underlying `QuantityInputField`). This was breaking unit conversion and throwing the error. Now ensuring `unit` is always initialized in store from `initialState`.

## Testing
https://irsa-7677-spherex-api-fix.irsakubedev.ipac.caltech.edu/applications/spherex/?api
- the first two examples in above should work instead of throwing errors like before
- rest of examples should be working as they were before

Regression test other TAP queries in [firefly url api](https://fireflydev.ipac.caltech.edu/irsa-7677-spherex-api-fix/firefly/?api=tap). Here's a ObsTAP example (since spherex also uses ObsTAP): https://fireflydev.ipac.caltech.edu/irsa-7677-spherex-api-fix/firefly/?api=tap&service=https://ws.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/argus&schema=ivoa&table=ivoa.ObsCore&WorldPt=202.48417;47.23056;EQ_J2000&sr=20s&spatialRegionOperation=intersects

### DIFF
--- a/src/firefly/js/ui/QuantityInputField.jsx
+++ b/src/firefly/js/ui/QuantityInputField.jsx
@@ -203,18 +203,24 @@ const quantityBoundsValidator = (min, max, nullAllowed, unit, quantityName, form
 /**Uncontrolled component for Quantity Input that wires QuantityInputFieldView's state with the FieldGroup store.**/
 export const QuantityInputField = memo((props) => {
     const { quantityBaseUnit, convertQuantityUnits, quantityName, formatQuantity } = props;
+    const normalizedInitialState = normalizeInitState(props?.initialState, quantityBaseUnit, convertQuantityUnits);
+    const { unit: initUnit } = normalizedInitialState;
     const { viewProps, fireValueChange } = useFieldGroupConnector({
         ...props,
-        initialState: normalizeInitState(props?.initialState, quantityBaseUnit, convertQuantityUnits),
+        initialState: normalizedInitialState,
     });
 
     useEffect(() => {
-        // if value changed in the store (for e.g. by a setFieldValue()) and there's no displayValue, set it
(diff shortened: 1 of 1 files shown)


loitly (approved):
It worked as described. The changes fixed the problem, but the root cause seems to be in `useFieldGroupConnector`. I suggest creating a ticket to fix it there later. It's a problem if we need to apply this logic everywhere it’s used.

lrebull:
This seems to work for me!

jaladh-singhal:
> but the root cause seems to be in useFieldGroupConnector. I suggest creating a ticket to fix it there later. It's a problem if we need to apply this logic everywhere it’s used.

@loitly based on yesterday's long debugging session, I could boil it down that problem is with `doGetInitialState` in `FieldGroupConnector.jsx`: https://github.com/Caltech-IPAC/firefly/blob/0352891e4ff4be9b48b2c1d3bccec6230237f6a7/src/firefly/js/ui/FieldGroupConnector.jsx#L64

In url api, `storeField`  somehow gets unit=undefined (maybe because of ListBox delay) and in the `keepState=true`, because of `||` it never retrieves unit from `initialState`. Maybe it should be:
```
  const initS = !keepState
      ? (initialState || storeField || {})                                   
      : (storeField ? {...initialState, ...storeField} : initialState || {});
```
but this may introduce regression issues so I didn't touch it. Although without urlApi, everything seems to work fine.

Opened an issue for it here: https://jira.ipac.caltech.edu/browse/FIREFLY-2016

PR #1958, opened 2026-05-09
FIREFLY-2009: fix embed panel collapse bug

kpuriIpac:
See IFE PR for details: https://github.com/IPAC-SW/irsa-ife/pull/466

### DIFF
--- a/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
+++ b/src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
@@ -4,7 +4,7 @@ import PropTypes, {oneOf, bool, string, number, arrayOf, object, func, shape, el
 import {defaultsDeep, isString} from 'lodash';
 import {SelectedShape} from '../../drawingLayers/SelectedShape';
 import CoordinateSys from '../../visualize/CoordSys.js';
-import {BOX_CHOICE_KEY, CONE_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY} from '../../visualize/ui/CommonUIKeys.js';
+import {BOX_CHOICE_KEY, CONE_CHOICE_KEY, POINT_CHOICE_KEY, POLY_CHOICE_KEY, UPLOAD_CHOICE_KEY} from '../../visualize/ui/CommonUIKeys.js';
 import {BoxSearchInputFields} from '../BoxSearchInputFields.jsx';
 import {
     DEFAULT_INSTRUCTIONS_CIRCLE, DEFAULT_INSTRUCTIONS_RECT,
@@ -378,7 +378,6 @@ EmbeddedPositionSearchPanel.propTypes= {
         }),
         searchSummary: shape({
             component: elementType,
-            getSummaryInfo: func,
         }),
         spatialSearch: shape({
             component: elementType,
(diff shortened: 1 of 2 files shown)


src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -426,7 +425,7 @@ function SearchSummary({
     request, targetKey, sizeKey, polygonKey, searchTypeKey, searchTypeToggleOptions, getSearchType,
     boxSizeXKey, boxSizeYKey, boxRotationKey
 }) {
-    let {value: searchTypeValue, label: searchTypeLabel} = getSearchType?.(request) ?? {};
+    let {value: searchTypeValue, label: searchTypeLabel, showRadius=true} = getSearchType?.(request) ?? {};
```
jaladh-singhal: I don't think it's right to expect `getSearchType` to return `showRadius` - it just identifies type of search (can be polygon, upload, etc.) not how to generate text for it (for a specific type - cone/point). You can see the signature of this function defined in prop types comment of this component.

src/firefly/js/ui/dynamic/EmbeddedPositionSearchPanel.jsx
```diff
@@ -448,7 +447,7 @@ function SearchSummary({
 
     const summaryFragments = [];
     if (searchTypeValue === CONE_CHOICE_KEY) {
-        if (radius) summaryFragments.push(`${toMaxFixed(radius, 6).toString()}° at `);
+        if (showRadius && radius) summaryFragments.push(`${toMaxFixed(radius, 6).toString()}° at `);
```
jaladh-singhal: I am now recalling that I refactored this when introducing `BOX_CHOICE_KEY` for mosaic tool which broke this case which was relying earlier on not polygon check.

I was planning to introduce a `POINT_CHOICE_KEY` and use that here in it's own conditional block to make it readable.

jaladh-singhal (changes requested):
Thanks for the PR, it works but is a bit sneaky. We should now become explicit about POINT_CHOICE_KEY - can you please define it and just add a conditional branch for it? (sorry this is a regression bug I introduced because I forgot to do that in the first place)

kpuriIpac:
> Thanks for the PR, it works but is a bit sneaky. We should now become explicit about `POINT_CHOICE_KEY` - can you please define it and just add a conditional branch for it? (sorry this is a regression bug I introduced because I forgot to do that in the first place)

Yes fair point. Good suggestion, I'll add `POINT_CHOICE_KEY` as an option. That should be much cleaner.

kpuriIpac:
Ok @jaladh-singhal I implemented the feedback. Can you take a look here and at the IFE PR? It's much more straightforward & concise now. 

In a few minutes, the 3 test builds should be ready as well (Euclid & DCE to make sure nothing broke - just precautionary but the code should not affect those at all)

kpuriIpac:
> Ok @jaladh-singhal I implemented the feedback. Can you take a look here and at the IFE PR? It's much more straightforward & concise now.
> 
> In a few minutes, the 3 test builds should be ready as well (Euclid & DCE to make sure nothing broke - just precautionary but the code should not affect those at all)

@jaladh-singhal wait - the test builds aren't ready. not building on jenkins. I'll let you know when those are ready.

jaladh-singhal (approved):
Thanks for the update - I tested spherex (the rebuild from 10 min ago)

PR #1960, opened 2026-05-13
Firefly 2017 duckdb memory usage

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-2017
- Revert back to open/close connection on each use model
- Close when there's no active connections
- When open, return duplicate connection to avoid multiple connections to the same db

Also:
- Fix intermittently failing tests
- Lazy load ehcache manager.  May resolve failure during redeploy.

Test: https://fireflydev.ipac.caltech.edu/firefly-2017-duckdb-memory-usage/firefly/
It’s hard to test. I checked to ensure that the pod’s memory usage doesn’t increase over time.  I then compare it to nightly builds, as that’s where the bug was introduced.  In nightly builds, the memory usage steadily increased and never decreased.

### DIFF
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -305,6 +305,7 @@ def main():
             f"-XX:MaxRAMPercentage={os.getenv('MAX_RAM_PERCENT', '80')}",
             "-XX:+UnlockExperimentalVMOptions",
             "-XX:TrimNativeHeapInterval=30000",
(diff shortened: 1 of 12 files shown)


kpuriIpac (approved):
Looks good to me for now. I think it's safe to merge and test locally on `dev` to see if we find any new concerns.

PR #1962, opened 2026-05-19
Firefly-1980: firefly standalone installation

robyww:
#### Testing
- install firefly on your machine
- create a install directory
```bash
curl -L https://raw.githubusercontent.com/Caltech-IPAC/firefly/refs/heads/FIREFLY-1980-standalone/bin/get-firefly | bash

```
- start the server `firefly/bin/ff start`
- it should open the web browser
- see more options with the ff command `firefly/bin/ff`

#### Information
- workarea and logs are under `~/.firefly/server`
  - log files `firefly.log` and `application.log`
- installation including webapp dir is where ever you installed firefly
- if you want to change the port edit ~/.firefly/port.txt
- add java start up parameters `~/.firefly/user_ops.sh`
- graphics are not final but I need a placeholder

#### Todo

- [x] End user documentation
- [x] change temporary PR branch endpoints before merge
- [x] gradle: fix dependency
- [x] gradle: put dependencies in local mavin


loitly (commented):
It looks good.  Definitely good enough for phase 1.  Below are my feedbacks.

Instead of a new standalone directory under `firefly` repo root, have you considered moving these scripts into the existing bin directory?

Rather than remoteInstall.sh, maybe call it get-firefly.

Instead of “create an install directory,” install.sh should prompt for an installation location, defaulting to the current directory.

I would move `startFirefly` into the firefly/bin directory. There will likely be additional scripts added later, and we could either add this directory to the user’s PATH automatically or instruct the user to do so during installation.

I’d also think about how to organize the files under ~/.firefly. Maybe place them under a `config` directory so files are not hanging directly off .firefly.

loitly (commented):
Even though this is already in the firefly repo, we still prefix it with “firefly” everywhere. Perhaps instead of “src/firefly_standalone”, it should simply be “standalone” or “application”.

buildScript/dependencies.gradle
```diff
@@ -108,6 +108,11 @@ dependencies {
     implementation 'org.asdf-format:asdf-core:0.1-alpha-10'
     implementation 'edu.stsci:roman-datamodels:0.1-alpha-3'
 
+    // tomcat for standalone
+    implementation 'org.apache.tomcat.embed:tomcat-embed-core:11.0.21'
+    implementation 'org.apache.tomcat.embed:tomcat-embed-websocket:11.0.21'
+    implementation 'org.apache.tomcat.embed:tomcat-embed-jasper:11.0.21'
+
```
loitly, line 112: Is this still necessary when firefly_standalone dependencies are already defined?

robyww: I will remove them

robyww:
@loitly `firefly_standalone` was you idea but I am happy the change it to standalone

robyww:
> Instead of a new standalone directory under firefly repo root, have you considered moving these scripts into the existing bin directory?

sure

> Rather than remoteInstall.sh, maybe call it get-firefly.

good idea

> Instead of “create an install directory,” install.sh should prompt for an installation location, defaulting to the current directory.

many install scripts no longer prompt. they go for simple. For example install java.

> I would move startFirefly into the firefly/bin directory. There will likely be additional scripts added later, and we could either add this directory to the user’s PATH automatically or instruct the user to do so during installation.

yea, I suppose, I was not quite sure where to put it. I wanted a script that the user could copy to other places. If it is in the bin I guess the install script needs to make a clear message to the user.

> I’d also think about how to organize the files under ~/.firefly. Maybe place them under a config directory so files are not hanging directly off .firefly.

I experimented with stuff like that. After that, I purposely tried to make it more shallow I am not sure of the point of making it deeper.

loitly:
One more suggestion:
Instead of using `startFireflyServer.sh`, I recommend using `firefly` with parameters.

Here’s an example of how you can use it:
```bash
firefly start [-d]
firefly stop
firefly status
firefly logs [-f]
```

This script can be extended to include additional functionalities, such as loading a file:
```bash
firefly load <file-name>
```

The purpose of these user-facing suggestions, including the directory organization, is to future-proof the tool.

bin/install.sh
```diff
@@ -0,0 +1,173 @@
...
+fi
+
+
+if isTrue $initialInstall; then
+  echo
+  echo "Firefly successfully installed, to start Firefly use the $startScript command"
+  echo
+  echo ">>>>>>>>>>>>>>>>>>>>>> ${binDir#$PWD/}/ff start"
+  echo
+  echo "You might want to add the bin dir to your path: $binDir"
```
jaladh-singhal: To make it clear that we're talking about `PATH` env var:
```suggestion
  echo "You might want to add the bin dir to your PATH: $binDir"
```

src/standalone/assets/startFireflyServer.sh
```diff
@@ -0,0 +1,184 @@
...
+   echo "------------------------------------------------"
+   echo "---------- Starting Firefly server"
+   echo "---------- $(date)"
+   echo "------------------------------------------------"
+   echo ${JAVA} ${splash} "${nameParam}" ${PROPS} edu.caltech.ipac.app.FireflyApplication
+} >> "$appLog"
+
+if isTrue $inBackground; then
+  (cd "$applicationDir" && ${JAVA} ${splash} "${nameParam}" ${PROPS} edu.caltech.ipac.app.FireflyApplication &> "${fireflyServer}/logs/backgroundStart.log" &)
+  echo "Firefly server starting in background (is takes a few seconds)..."
```
jaladh-singhal:
```suggestion
  echo "Firefly server starting in background (it takes a few seconds)..."
```

src/standalone/assets/startFireflyServer.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
jaladh-singhal, line 7: I think this script should also check if the port is available before starting firefly server.

Before executing `start` command, I had a jupyter server running (it also uses 8888) and the `start` command succeeded with `Firefly ready: use URL: http://localhost:8888/firefly/` but going to this url showed me:

Image

The `start` command should have given port not free error along with instructions to change port.

src/standalone/assets/ff
```diff
@@ -0,0 +1,95 @@
...
+     fi
+}
+
+
+stop() {
+  if [[ "$1" == "help" || "$1" == "-h" || "$1" == "--help" ]]; then
+    echo "stop the firefly server"
+  else
+    pid=$(cat "$fireflyDir/pid.txt");
+    kill "$pid"
```
jaladh-singhal: It might be worth echoing a message (if the process kill was successful) that "Firefly server at \<url\> stopped". It gives no feedback esp if you started Firefly server with `ff start -d`

jaladh-singhal (commented):
This is looking really good. The commands were easy to follow after seeing help. The firefly menu in the topmost tray is helpful.

1) Upload and IRSA image search are working fine but the searches that go through job monitor aren't working for me. In the logs, I see the following two errors after POSTing a TAP, SIA or IRSA catalogs query:
```
java.lang.NullPointerException: Cannot invoke "edu.caltech.ipac.firefly.core.background.JobInfo.getMeta()" because "info" is null

io.lettuce.core.RedisCommandExecutionException: MISCONF Redis is configured to save RDB snapshots, but it is currently not able to persist on disk. Commands that may modify the data set are disabled, because this instance is configured to report errors during writes if RDB snapshotting fails (stop-writes-on-bgsave-error option). Please check the Redis logs for details about the RDB error.
```

2) It will be nice if you can add an option `--port, -p` to start to override port.txt. I find editing file an extra step/

3) The multi-port deployment needs more thought. I started one firefly server at 8888 with `ff start -d` and then changed the port.txt to 8889 and started another one. The state of these two servers are independent - I tired doing different searches in each url - so that's good. Now when I did `ff stop` it successfully shut downned the recent one which was 8999. Then I did `ff stop` again but it fails with `line 32: kill: (84074) - No such process`. These multi-port issues are probably out-of-scope for this pass and we can just prevent starting firefly server at more than one port by checking if it's already running.

(I didn't really do a proper code review but skimmed it and nothing jumps out)

src/standalone/assets/ff
robyww: I will echo a message.

src/standalone/assets/startFireflyServer.sh
robyww: ok, I will put more port analysis in the script

robyww:
> Upload and IRSA image search are working fine but the searches that go through job monitor aren't working for me. In the logs, I see the following two errors after POSTing a TAP, SIA or IRSA catalogs query:

This is because of redis conflicts and database conflicts. I think I will move the default redis port to something different that what we are using in development. That should allow more that one to exist.

lrebull:
lol i actually giggled out loud when it started up.

the "GAIA DR# cone search at M101" yields this:
Screenshot 2026-05-28 at 12 50 35 PM

Also, GAIA isn't an acronym; it should be "Gaia".

The "SDSS HiPS view of m81" yields this:
Screenshot 2026-05-28 at 12 52 25 PM

I just started exercising the regular kinds of searches I would do to IRSA Viewer -- just basic "try things" kinds of testing. Most of it works just fine.
 
A TAP position search, I think, would normally 'inherit' the polygon visible image corners for the images I have loaded, but this incarnation doesn't; i can't make it automatically grab the corners of the images I have loaded. It does, at least, remember the target location.

vandesai1:
I thought we agreed to the most generic tabs rather than the irsa-specific ones?

Copying from slack:

Suggested tabs in downloadable Firefly:

Table (TAP)
Images (SIAv2)
Survey Maps (HiPS)

suggestions for the hidden tabs:

Images --> IRSA Images
IRSA Catalogs (good)
IRSA SIAv2 --> IRSA Images (SIAv2)
VO SCS Search --> Catalogs (SCS)

robyww:
> I thought we agreed to the most generic tabs rather than the irsa-specific ones?

@vandesai1 that is a different PR, this is to make a standalone installation. Were you able to install and run Firefly?

vandesai1:
@robyww , yes I was able to run and install it, and only came up with the same issues that Luisa did!

loitly (approved):
Looks great.  Ready to go.

PR #1964, opened 2026-05-22
FIREFLY-2012: When UWS job fails, display (i) icon to popu job info dialog

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-2012
Show "(i)" icon when UWS job failed

Test:  https://fireflydev.ipac.caltech.edu/firefly-2012-jobinfo-on-table-error/firefly/

- TAP -> Edit ADQL
- ADQL Query: SELECT TOP 1000 * FROM a_dummy_table
- Click `Search`

You should see an error message with the “(i)” icon that will pop up the Job Info Dialog.

### DIFF
--- a/src/firefly/js/tables/TableRequestUtil.js
+++ b/src/firefly/js/tables/TableRequestUtil.js
@@ -2,7 +2,7 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {get, set, unset, cloneDeep, omit, omitBy, isNil, pickBy, uniqueId, merge} from 'lodash';
+import {set, unset, cloneDeep, omit, omitBy, isNil, pickBy, uniqueId, merge} from 'lodash';
 
 import {getTblById, uniqueTblId} from './TableUtil.js';
 import {SelectInfo} from './SelectInfo.js';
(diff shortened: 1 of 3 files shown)


jaladh-singhal (approved):
Looks good, works as expected.

I noticed a related issue (likely out of scope for this PR): when a job result (table) fails, there's no way to dismiss it. This is specifically annoying when this job was the first search you did because single table doesn't have table container toolbar to close the tab and you can't go back to empty results screen. I ran into same problem for spherex mosaic results and added two ways to overcome it:
1. Use `forceSingleTableAsTab`: https://github.com/Caltech-IPAC/firefly/blob/ceef8ece501936fa8ce68bf16cde94f5dc379e3e/src/firefly/js/tables/ui/TablesContainer.jsx#L89
2. Let the BGMaskPanel `showError` but give it the `renderError`: https://github.com/Caltech-IPAC/firefly/blob/7eaa8a17100dc33f4ed285ecc2241ee2e93b6824/src/firefly/js/core/background/BgMaskPanel.jsx#L100
 
I'm not sure if option 2 is possibe for general table case but if it is, it will save you from creating the job info button again in `TablePanel`.

loitly:
@jaladh-singhal I've addressed your concern.  I've instead added a 'Dismiss' button when there is no tabs and the table cannot be 'Reload'.  This would remove the bad table.  Please review if it's acceptable.

jaladh-singhal:
@loitly it looks good, thanks!

PR #1966, opened 2026-05-28
FIREFLY-2007: Properly initialize jobs, distinguishing between UWS and  Firefly-managed jobs.

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-2007

Also:
- Continue polling(with timeout) for the UWS status during bad phases to ensure its survival even in transient errors like network issues.

Test: https://fireflydev.ipac.caltech.edu/firefly-2007-uws-job-init/firefly/
- Ensure that background jobs continue to function as intended.
- Verify that the UWS job is properly initialized and accurately reflects information from the UWS service, without introducing any default values.

An example:  IRSA TAP do not have Creation Time.


robyww (approved):
Looks fine

PR #1968, opened 2026-06-10
Firefly-2037: gpu disable for linux, better gpu testing with async test

robyww:
#### [Firefly-2037](https://jira.ipac.caltech.edu/browse/FIREFLY-2037): gpu disable for linux, better gpu testing with async test

- also fixes a bug with firefox, it can now use webgpu

#### Testing 
- https://fireflydev.ipac.caltech.edu/firefly-2037-gpu/firefly
- hard to test the linux without a desktop, image should work the same with firefox, chrome, safari

### DIFF
--- a/src/firefly/js/util/BrowserInfo.js
+++ b/src/firefly/js/util/BrowserInfo.js
@@ -3,6 +3,7 @@
  */
 import Enum from 'enum';
 import {getProp} from 'firefly/util/WebUtil.js';
+import {synchronizeAsyncFunctionById} from './SynchronizeAsync';
 
 /**
  * @typedef {Object} Browser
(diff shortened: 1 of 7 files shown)


src/firefly/js/util/BrowserInfo.js
```diff
@@ -91,18 +92,29 @@ export default BrowserInfo;
     BrowserInfo.majorVersion=majorVersion;
     BrowserInfo.minorVersion=minorVersion;
     BrowserInfo.browser=browser;
+    BrowserInfo.webGpuVerifiedPromise= verifyWebGpuDevice();
```
loitly, line 95: What is the purpose of saving this promise?

src/firefly/js/util/BrowserInfo.js
```diff
@@ -91,18 +92,29 @@ export default BrowserInfo;
...
-        navigator?.gpu?.requestAdapter()
-            .then( () => BrowserInfo.firefoxWebGpuEnabled= true)
-            .catch(() => BrowserInfo.firefoxWebGpuEnabled= false);
-    }
     BrowserInfo.platformMajorVerison= platformMajorVersion(ua,BrowserInfo.platform);
     logIfVersionNotSupported();
 })();
 
+
+async function verifyWebGpuDevice() {
```
loitly, line 104: I see.  You want to make sure this function only get called once.  You could have used `_.once` here.

src/firefly/js/util/BrowserInfo.js
```diff
@@ -230,20 +242,24 @@ function supportsCssColorMix() {
...
+    else if (isBrowser(Browser.FIREFOX)) {
+        if (linux) return false;
+        if (!isVersionAtLeast(145)) return false;
+        if (BrowserInfo.isPlatform(Platform.WINDOWS)) return verified;
+        if (BrowserInfo.isPlatform(Platform.MACOS)) return verified;
         return false;
     }
-    return foundGpu;
+    else if (isBrowser(Browser.SAFARI)) return isVersionAtLeast(26) && verified;
+    return verified;
```
loitly, line 264: What other browser types have you tested that you expect `verified` to be true at this line? If none, can you simply return `false` for `Platform.LINUX`?  Also, you should return `false` at the end to rule out false positives.

loitly (approved):
Based on the code, the changes look right. However, I haven’t tested it.

src/firefly/js/util/BrowserInfo.js
robyww: I purposely did not do it that way because it is easier to follow by breaking them down by browsers. Others are not tested so they just use the fallback.

src/firefly/js/util/BrowserInfo.js
robyww: it is already being called in a immediate function so once is not necessary.

src/firefly/js/util/BrowserInfo.js
robyww: They promised is saved because the check at the bottom is async and uses it. Since an await on a given promise can be called multiple times, the promise also functions as a container. That is much cleaner then trying to resolve it in the initialization step.

PR #1972, opened 2026-06-17
Firefly 2045: Fix WebSocket connection leak on client and server

loitly:
https://jira.ipac.caltech.edu/browse/FIREFLY-2045

**Server:**
- Cap WS connections at 10 per channel (`max.ws.per.channel`), evicting oldest
  under a synchronized block to prevent race conditions.
- Fix `ArrayIndexOutOfBoundsException` in `ReplicatedQueueList` by passing a
  snapshot copy to MessagePack serialization.
- Delete ehcache `.lock` on shutdown to allow hot redeployment without Tomcat restart.

**Client:**
- Wrap `getOrCreateWsConn()` with `synchronizeAsyncFunctionById()` so concurrent
  callers share a single in-flight connection attempt instead of each opening their own.
- Remove redundant `onClose()` call from `onerror` handler — browser always fires
  `onclose` after `onerror`, so cleanup now happens exactly once.
  
#### Test
https://fireflydev.ipac.caltech.edu/firefly-2045-runaway-websocket/firefly/
- [ ] Verify normal WebSocket connection and event flow in the browser
- [ ] Verify multiple tabs each receive their own events and closing one doesn't affect others
- [ ] Verify that flooding connections to the same channel results in no more than 10 active connections
- [ ] Verify that a fresh connection can always be established after the limit is reached
- [ ] Verify client reconnects automatically after a network interruption with no duplicate connections

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
@@ -382,15 +382,16 @@ public class ServerStatus extends BaseHttpServlet {
         int qCnt= ServerEventManager.getActiveQueueCnt();
         w.println("  - Total active queues:" + qCnt);
         if(qCnt>0) {
-            w.println("  - "+ (qCnt>10? "10 Most recently used channels:" :  "Channel list, ordered by last use:"));
-            w.println(makeQueueList());
+            w.println("  - "+ (qCnt>20? "Top 20 channels by connection count:" :  "Channel list, ordered by connection count:"));
+            w.println(makeQueueList(20));
         }
     }
 
-    private static String makeQueueList() {
-        return ServerEventManager.getQueueDescriptionList(10).stream()
-                .map( d -> String.format("     - %s, %s\n",d.channel(), new Date(d.lastPutTime())))
-                .reduce("", (all, entry) -> all+entry);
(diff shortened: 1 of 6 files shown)


src/firefly/java/edu/caltech/ipac/firefly/server/servlets/ServerStatus.java
```diff
@@ -375,22 +377,40 @@ private static boolean isK8S(String hostname) {
...
-            w.println("  - "+ (qCnt>10? "10 Most recently used channels:" :  "Channel list, ordered by last use:"));
-            w.println(makeQueueList());
+        if (qCnt > 0) {
+            List<ServerEventQueue.QueueDescription> conns = ServerEventManager.getQueueList();
+            w.println("  - Top 10 channels by connection count:");
+            w.println(formatSummaries(getChannelSummaries(conns).stream()
+                    .sorted(Comparator.comparingInt(ChannelSummary::count).reversed())
+                    .limit(10).toList()));
+            w.println("  - Top 10 most recently active connections:");
+            conns.stream().limit(10).forEach(d ->
```
robyww, line 404: why don't you make this larger like 40. I don't think it will hurt.

robyww (approved):
All looks good, I test 1,2 and 5. I seems to work fine.

- Showing information in the admin/status is good I am glad you added that.

PR #1973, opened 2026-06-23
Firefly-2026: url api: added SIA, clean up of hipsPanel

robyww:
### [Firefly-2026](https://jira.ipac.caltech.edu/browse/FIREFLY-2026): url api: added SIA, clean up of hipsPanel

- Updated todo item in` StandaloneFireflyLanding.jsx` `DEFAULT_CHIPS`
- also a little refactor of PositionParser, there was overlap with CoordSys.js

#### Testing
- https://fireflydev.ipac.caltech.edu/firefly-2026-urlapi/firefly
- try https://fireflydev.ipac.caltech.edu/firefly-2026-urlapi/firefly?api=sia
- try https://fireflydev.ipac.caltech.edu/firefly-2026-urlapi/firefly?api=HipsPanel
- Landing page examples all go to search panels, no direct searches

### DIFF
--- a/src/firefly/js/api/webApiCommands/ImageCommands.js
+++ b/src/firefly/js/api/webApiCommands/ImageCommands.js
@@ -113,15 +113,15 @@ const hipsExamples= [
 
 const hipsPanelOverview= {
     overview: [
-        'Show HiPS Panel, configure HiPS panel, show HiPS: you may also use any parameter from the hips command'
+        'Show HiPS Panel and/or load, configure HiPS panel, or just display HiPS: you may also use any parameter from the hips command'
     ],
     parameters: {
-        showPanel: {desc:'show HiPS panel'},
+        showPanel: {desc:'Show HiPS search panel. If not included or false, loads the HiPS (executes the HiPS search'},
         [ReservedParams.POSITION.name]: ['coordinates to center HiPS',...ReservedParams.POSITION.desc],
         [ReservedParams.SR.name]: ['Radius of the field of view of the HiPS',...ReservedParams.SR.desc],
         hipsListName: {desc:'a HiPS list server name'},
         hipsListUrl: {desc:'a HiPS list server url'},
-        uri: {desc:'if included, the HiPS panel will not initially show: the HiPS will load'},
+        uri: {desc: 'URI of the HiPS to load, same as in cmd=hips. If showPanel is true, it selects the HiPS from the HiPS list else it loads the HiPS.' },
     },
 };
 
@@ -214,11 +214,22 @@ function showHiPs(cmd,inParams) {
 }
 
 function validateHiPSPanel(params) {
+    const {hipsListUrl,uri,showPanel}= params;
+    if (!hipsListUrl && !uri && !showPanel) {
+        return {valid:false, msg:'hipsListUrl or showPanel or uri is required'};
+    }
     return {valid:true};
 }
 
 function showHiPSPanel(cmd,inParams) {
     setTimeout( async () => void showHiPSPanelAsync(cmd,inParams) );
+}
+
+function executeHiPSUri(cmd, plotParams) {
+    setTimeout(() => {
+        dispatchHideDropDown();
+        showHiPs(cmd, plotParams);
+    });
 }
 
 async function showHiPSPanelAsync(cmd, inParams) {
@@ -228,7 +239,7 @@ async function showHiPSPanelAsync(cmd, inParams) {
         urlApi.radius= inParams[ReservedParams.SR.name];
     }
 
-    const {hipsListName,hipsListUrl,showPanel, ...plotParams}= inParams;
+    const {hipsListName,hipsListUrl,showPanel, execute, ...plotParams}= inParams;
     if (hipsListUrl) {
         const name= hipsListName || hipsListUrl;
         dispatchAppOptions({extraHiPSListName:name});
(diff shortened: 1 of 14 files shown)


src/firefly/js/templates/fireflyviewer/StandaloneFireflyLanding.jsx
```diff
@@ -69,19 +69,16 @@ const DEFAULT_CHIPS = [
...
     },
     {
         label: '🔭 Gaia DR3 cone search at M101 →',
         url: '?api=tap&service=https://gea.esac.esa.int/tap-server/tap&schema=gaiadr3&table=gaiadr3.gaia_source&WorldPt=210.80227;54.34895;EQ_J2000&sr=500s',
     },
     {
-        // TODO: api=hipsPanel doesn't allowing selecting a HiPS server without executing the search, need to make "uri" not execute
         label: '🌐 SDSS HiPS view of M81 →',
-        uri: '?api=hipsPanel&showPanel=true&ra=148.88822&dec=69.06529&sr=40m&uri=ivo://CDS/P/SDSS9/color',
+        url: '?api=hipsPanel&showPanel=true&ra=148.88822&dec=69.06529&sr=4d&uri=ivo://CDS/P/SDSS9/color',
```
jaladh-singhal:
```suggestion
        url: '?api=hipsPanel&showPanel=true&ra=148.88822&dec=69.06529&sr=2d&uri=ivo://CDS/P/SDSS9/color',
```
Can you zoom more? otherwise I see incomplete patches of HiPS at 4deg FOV and scientist wanted to focus on M81 only.

src/firefly/js/api/webApiCommands/ImageCommands.js
```diff
@@ -113,15 +113,16 @@ const hipsExamples= [
...
     ],
     parameters: {
         showPanel: {desc:'show HiPS panel'},
         [ReservedParams.POSITION.name]: ['coordinates to center HiPS',...ReservedParams.POSITION.desc],
         [ReservedParams.SR.name]: ['Radius of the field of view of the HiPS',...ReservedParams.SR.desc],
         hipsListName: {desc:'a HiPS list server name'},
         hipsListUrl: {desc:'a HiPS list server url'},
-        uri: {desc:'if included, the HiPS panel will not initially show: the HiPS will load'},
+        uri: {desc:'not using the showPanel parameter, the HiPS will load, this is the same as cmd=hips'},
+        execute: 'use with showPanel, true or false - if true execute the HiPS load'
```
jaladh-singhal: As API user, I found this pretty confusing since `execute=false` overlaps with `showPanel`. I gave it more thought... we can **remove `showPanel`, and fold it into the `execute` pattern**

| `?api=hipsPanel` Behaviour | Current params | Proposed params |
|---|---|---|
| Open panel | `showPanel=true` | N/A |
| Open panel with HiPS list pre-added | `showPanel=true` + `hipsListUrl` | `hipsListUrl` alone |
| Open panel with uri pre-selected | `showPanel=true` + `uri` | `uri` alone |
| Load HiPS directly | `uri` alone **or** `showPanel=true` + `uri` + `execute=true` | `uri` + `execute=true` |

Since the rest of the URL API already uses `execute=false` → open panel, `execute=true` → submit, `hipsPanel` can follow the same pattern. `showPanel` can be dropped entirely - the command is already called `hipsPanel`, so opening the panel is the default. 

About the "N/A" in above table, it's consistent with `tap`, `sia`, etc. which also require at least one param since a bare panel can always be opened via `?api=menu&tab=`.

Besides, this will need changes in:
- `validateHiPSPanel` to enforce that `hipsListUrl` or `uri` must be present (since bare `?api=hipsPanel` correctly falls through to the API command help page). 
- "add the toolbar option" branch to run when `execute=false` (i.e. whenever the panel is shown) since it currently only runs inside the showPanel=true block.

jaladh-singhal (changes requested):
Thanks for doing this @robyww. The landing page examples of SIA and HiPS are consistent now and `?api=sia` is working correctly.

But I'm finding `?api=hipsPanel` confusing and I think we can improve its design.

src/firefly/js/api/webApiCommands/ImageCommands.js
robyww: HiPS panel was added for rubin, it has three uses. add a new hips service (hipsListUrl), showing the panel, or searching. It was added for rubin so they could bring firefly up with a extra hips configurations for experimenting. These configuration can show up in other places beside the HiPS panel.

This is different than most search panels. 

i agree that we needed to add `execute` and set the uri correctly. However, I don't agree we should remove `showPanel`. Having `showPanel` is not an accident it is necessary to completely implement all the use cases. It could be refactored but I don't thing there is a better way that how it is. Besides, I don't want to break the backward compatibility.

jaladh-singhal: To preserve the backward compatibility, we can safely remove `execute` param since `showPanel=falsy` is equivalent to `execute=true`.

Then the param descriptions (from User's POV) become:
- `showPanel: {desc:'show HiPS panel'},` -> `showPanel: {desc: 'Show HiPS search panel. If undefined or false, loads the HiPS (executes the HiPS search'},`
- `uri: {desc:'not using the showPanel parameter, the HiPS will load, this is the same as cmd=hips'}` -> `uri: {desc:'URI of the HiPS to load, same as in cmd=hips. If showPanel is true, it selects the HiPS from the HiPS list else it loads the HiPS.'},`
- `execute: 'use with showPanel, true or false - if true execute the HiPS load'` -> delete

Then you will just need couple of lines of changes in code to remove the execute branch (L267-L268). And `else if` you added in L271 is already ensuring `showPanel=false + uri` executes the HiPS search.

jaladh-singhal (approved):
Looks good now, thanks!

PR #1975, opened 2026-06-26
Firefly-2039: Fixed: Incorrect flag mask shown for a pixel with no flags

robyww:
#### Test
- https://firefly-2039-mask-corner.irsakubedev.ipac.caltech.edu/firefly

### DIFF
--- a/src/firefly/js/visualize/rawData/ManageRawDataThread.js
+++ b/src/firefly/js/visualize/rawData/ManageRawDataThread.js
@@ -99,11 +99,17 @@ async function doMaskColorChange(payload) {
 }
 
 function convertToBits(ary) {
-    const retAry= new Uint8ClampedArray(Math.trunc(ary.length/8)+1);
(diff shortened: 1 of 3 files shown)


tgoldina (changes requested):
Previously, the lower-right pixel was always shaded for every mask plane; now it is never shaded for any mask plane.

This suggests that the bug lies elsewhere in the shader. I have attached `lrset_5x5.fits` to the ticket; it now has flags set for every pixel. The lower-right pixel, which the mouse pointer is hovering over, is not shaded.

Image

tgoldina (approved):
Thank you, Trey. The latest version produces correct mask shading.

I am wondering if we have a similar bug in image shading. In my 5x5 sample, the corner pixel with value 6.536 is not shaded while the adjacent pixel (to the left) with value 6.373 is shaded light gray. But this is a separate issue, unrelated to this PR.

Image

PR #1977, opened 2026-07-06
FIREFLY-1954: React Split Pane version upgrade

kpuriIpac:
**Ticket**: [FIREFLY-1954](https://jira.ipac.caltech.edu/browse/FIREFLY-1954)
**[IFE PR](url)**: https://github.com/IPAC-SW/irsa-ife/pull/472
- This PR upgrades Firefly from react-split-pane `v0.1.92` to` v3.2.0`
   - Using the [migration doc](https://github.com/tomkp/react-split-pane/blob/master/MIGRATION.md) and [readme](https://github.com/tomkp/react-split-pane/blob/master/README.md) for reference 
- The core changes are in `DockLayoutPanel`. v0 allowed sizing props to be applied at the parent `SplitPane` level, but v3 moves that to the individual `Pane` components. 
  - collapsible split panel logic is also slightly altered now to handle v3's pane sizing requirement.
- Changed several consumers of `SplitPane` (using `react-split-pane` directly) to use our `SplitPanel` wrapper instead
- `AdvancedADQL` and `TabPanel` include timing fixes for react-split-pane's v3 onwards ResizeObserver-based layout
  - In `AdvancedADQL` in particular, earlier the code assumed DOM was present when the useEffect for `Prism.highlighAll` ran), but with v3.2.0 of react-split-pane using ResizeObserver for layout measurements, I noticed the Pane content can become available after the useEffect where we try to initialize `Prism`
    - This led to crashes and/or the SQL examples not being highlighted 
    - So the new retry logic aims to fix this race issue. Could possibly be polished further by extracting the logic into a small separate hook. Open to other feedback as well. 
- The `dockerfile` change to Node 20 from Node 18 was needed because `react-split-pane@3.2.0` declares `node >= 20` and my Jenkins builds were failing without this change. I will talk to @loitly about this before deciding if we need to keep this. 

**Testing**: 
[Firefly](https://firefly-1954-react-split-pane.irsakubedev.ipac.caltech.edu/firefly), [Euclid](https://firefly-1954-react-split-pane.irsakubedev.ipac.caltech.edu/applications/euclid), [Spherex](https://firefly-1954-react-split-pane.irsakubedev.ipac.caltech.edu/applications/spherex), [Timeseries](https://firefly-1954-react-split-pane.irsakubedev.ipac.caltech.edu/irsaviewer/timeseries)
- Confirm visually in TAP, Advanced ADQL (Edit ADQL), tri view and bi view results, collapsible panels (spherex spectrophotometry) and in Timeseries that the results look good. 
  - Split Pane is used in several places, like the TAP screen as well. Test visually and compare against dev/ops if needed. Move around the panes to make sure there's no bugs there. 
- Dev links, to compare the test builds against: [Firefly Dev,](https://fireflydev.ipac.caltech.edu/firefly/) [Euclid Dev,](https://irsadev.ipac.caltech.edu/applications/euclid) [Spherex Dev,](https://irsadev.ipac.caltech.edu/applications/spherex/) [Timeseries Dev](https://irsadev.ipac.caltech.edu/irsaviewer/timeseries)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "react-grid-layout": "~1.5",
     "react-router-dom": "^6.14.0",
     "react-slick": "~0.31",
-    "react-split-pane": "~0.1",
+    "react-split-pane": "~3.2.0",
     "redux": "~3.7",
     "redux-logger": "~3.0",
     "redux-saga": "~1.1",
--- a/src/firefly/js/charts/ui/PinnedChartContainer.jsx
+++ b/src/firefly/js/charts/ui/PinnedChartContainer.jsx
@@ -32,7 +32,7 @@ import {ActiveChartsPanel} from './ChartsContainer';
 import {StatefulTabs, switchTab, Tab} from '../../ui/panel/TabPanel';
 import {HelpIcon} from '../../ui/HelpIcon';
 import {getComponentState, dispatchComponentStateChange} from '../../core/ComponentCntlr';
-import {SplitPanel} from '../../ui/panel/DockLayoutPanel';
+import {SplitPanel, Pane} from '../../ui/panel/DockLayoutPanel';
 import {hideInfoPopup, showInfoPopup, showPinMessage} from '../../ui/PopupUtil.jsx';
 import {dispatchAddActionWatcher} from 'firefly/core/MasterSaga';
 import {PinButton, ShowTableButton} from 'firefly/visualize/ui/Buttons.jsx';
@@ -61,19 +61,23 @@ export const PinnedChartContainer = (props) => {
         return (
             <Stack id='chart-pinned-sideBySide' overflow='hidden' flexGrow={1}>
                 <Stack flexGrow={1} position='relative'>
-                    <SplitPanel split='vertical' defaultSize={400} style={{display: 'inline-flex'}} pKey='chart-sideBySide'>
-                        <Stack>
-                            <Typography level='title-md'>
-                                {activeLabel}
-                            </Typography>
-                            <ActiveChartsPanel {...props}/>
-                        </Stack>
-                        <Stack>
-                            <Typography level='title-md'>
-                                {pinnedLabel}
-                            </Typography>
-                            <PinnedChartPanel {...props}/>
-                        </Stack>
+                    <SplitPanel direction='horizontal' defaultSize={400} style={{display: 'inline-flex'}} pKey='chart-sideBySide'>
+                        <Pane>
+                            <Stack>
+                                <Typography level='title-md'>
+                                    {activeLabel}
+                                </Typography>
+                                <ActiveChartsPanel {...props}/>
+                            </Stack>
+                        </Pane>
+                        <Pane>
+                            <Stack>
+                                <Typography level='title-md'>
+                                    {pinnedLabel}
+                                </Typography>
+                                <PinnedChartPanel {...props}/>
+                            </Stack>
+                        </Pane>
                     </SplitPanel>
                 </Stack>
             </Stack>
--- a/src/firefly/js/tables/ui/FilterEditor.jsx
+++ b/src/firefly/js/tables/ui/FilterEditor.jsx
@@ -6,13 +6,12 @@ import React, {useEffect, useRef} from 'react';
 import PropTypes from 'prop-types';
 import {Box, Button, Link, Sheet, Stack, Typography} from '@mui/joy';
 import {cloneDeep, get} from 'lodash';
-import SplitPane from 'react-split-pane';
 import Tree from 'rc-tree';
 import 'rc-tree/assets/index.css';
 
 import {FilterInfo, getFiltersAsSql} from '../FilterInfo.js';
 import {getTableUiById, getSqlFilter, DOC_FUNCTIONS_URL} from '../TableUtil.js';
-import {SplitContent} from '../../ui/panel/DockLayoutPanel.jsx';
+import {SplitPanel, Pane, SplitContent} from '../../ui/panel/DockLayoutPanel.jsx';
 import {InputAreaFieldConnected} from '../../ui/InputAreaField.jsx';
 import {RadioGroupInputField} from '../../ui/RadioGroupInputField.jsx';
 import {insertAtCursor} from '../../ui/tap/AdvancedADQL.jsx';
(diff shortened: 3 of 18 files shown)


src/firefly/js/charts/ui/PinnedChartContainer.jsx
```diff
@@ -61,7 +61,7 @@ export const PinnedChartContainer = (props) => {
         return (
             <Stack id='chart-pinned-sideBySide' overflow='hidden' flexGrow={1}>
                 <Stack flexGrow={1} position='relative'>
-                    <SplitPanel split='vertical' defaultSize={400} style={{display: 'inline-flex'}} pKey='chart-sideBySide'>
+                    <SplitPanel direction='horizontal' defaultSize={400} style={{display: 'inline-flex'}} pKey='chart-sideBySide'>
```
robyww, line 64: This strikes me as unexpected.  did the layout change?

src/firefly/js/tables/ui/FilterEditor.jsx
```diff
@@ -105,10 +104,10 @@ export function SqlTableFilter({tbl_ui_id, tbl_id, onChange, style={}, samples,
     placeholder = placeholder || 'e.g., "ra" > 180 AND "ra" < 185';
 
     return (
-        <SplitPane split='vertical' defaultSize={200} style={{display: 'inline-flex', ...style}}>
+        <SplitPanel direction='horizontal' defaultSize={200} style={{display: 'inline-flex', ...style}} pKey='filter-editor'>
```
robyww, line 107: again, when did `vertical` become `horizontal`

src/firefly/js/ui/tap/AdvancedADQL.jsx
```diff
@@ -201,7 +235,7 @@ export function AdvancedADQL({adqlKey, defAdqlKey, serviceUrl, capabilities, sty
                     if (uploadTableAlias) insertTname+= ' AS '+uploadTableAlias;
                 }
                 setVal(adqlKey, `SELECT TOP 1000 * FROM ${maybeQuote(insertTname,true)}`);
-                window.setTimeout( () => prismLiveRef.current.syncStyles?.(), 10);
+                window.setTimeout( () => prismLiveRef.current?.syncStyles?.(), 10);
```
robyww: You should define a function at the top of the component `syncStyles` that is
```js
const syncStyles= window.setTimeout( () => prismLiveRef.current?.syncStyles?.(), 10);
```

robyww (approved):
This looks good.  It works in my testing. There are some style things that I don't really understand that I assume is because of the component.

my big question is: why `vertical` to `horizontal` ?

src/firefly/js/tables/ui/FilterEditor.jsx
kpuriIpac: Yes so this is a big (but easy) change from v0.x to v3.x, they swapped `vertical` and `horizontal`  to match CSS flex direction. Hence all of these changes.

src/firefly/js/charts/ui/PinnedChartContainer.jsx
kpuriIpac: See below, layout remains the same. Just the meaning of vertical/horizontal is swapped.

src/firefly/js/ui/panel/DockLayoutPanel.jsx
```diff
@@ -25,28 +27,83 @@ const SplitContext = createContext({});
...
+                {childArray.map((child, idx) => {
+                    const paneStyle = {overflow: 'hidden', minWidth: 0, minHeight: 0, ...paneProps[idx]?.style};
+                    return (
+                        <Pane
+                            key={idx}
+                            {...paneProps[idx]}
+                            style={paneStyle}
+                        >
+                            {child}
+                        </Pane>
```
loitly: I would expect `Pane` to be the wrapper used in `SplitContent` and not injected here. This container should translate the out-dated 'maxSize/minSize' props into `minSize` `defaultSize`, than pass-along to the children, which should be `Pane` since we use SplitContent in all of our split panes.

loitly (approved):
I’ve tested it, and it worked fine. However, I have a suggestion that you should consider.

package.json
```diff
@@ -45,7 +45,7 @@
     "react-grid-layout": "~1.5",
     "react-router-dom": "^6.14.0",
     "react-slick": "~0.31",
-    "react-split-pane": "~0.1",
+    "react-split-pane": "^3.2.0",
```
jaladh-singhal: We generally pin the minor version too since some packages introduce breaking change in minor version too (not sure about react-split-pane, looks stable enough) and hence use `~` instead of `^`. We had a long discussion over this last time I had to update package.json 😅  and I added that to our [dev documentation](https://github.com/IPAC-SW/irsa-ife/blob/dev/docs/developer-guide.md#updating-packagejson).

src/firefly/js/ui/tap/AdvancedADQL.jsx
```diff
@@ -173,15 +173,49 @@ export function AdvancedADQL({adqlKey, defAdqlKey, serviceUrl, capabilities, sty
...
+            window.clearTimeout(timeoutId);
+        };
     },  [serviceUrl]);
 
     useEffect(() => {
-        // We need to get prism-live to adopt to the textarea
-        const textArea = adqlEl.current?.querySelector('textarea');
-        // adopt textArea
-        prismLiveRef.current = new Prism.Live(textArea);
+        let cancelled = false;
```
jaladh-singhal: Might be worth abstracting this out as a hook since above useEffect follows same pattern and it obscures the two-line business logic.

src/firefly/js/ui/panel/DockLayoutPanel.jsx
jaladh-singhal: +1. This way you won't need to define and maintain `paneProps=[{}, {}]` dedicatedly here and pass-through props will allow consumers of `SplitContent` to leverage new/better API of `Pane` if they decide to.

src/firefly/js/ui/panel/DockLayoutPanel.jsx
```diff
@@ -92,12 +151,12 @@ export function SplitContent({sx={}, style={}, children, panelTitle,
  */
 function CollapsibleSplitContent({sx={}, panelTitle, children, ...props}) {
     const {isOpen, onToggle} = useContext(SplitContext);
-    
-    // TODO: dynamically calculate styles based on the position of toggle button and split direction (passed as props)
```
jaladh-singhal, line 96: This is an important TODO that I'd advise against removing since this component isn't mature enough to handle both vertical and horizontal splits. If you want to remove it, we should at least add this note to docstring of this component and mark it `@experimental`

src/firefly/js/ui/panel/DockLayoutPanel.jsx
```diff
@@ -139,32 +199,50 @@ function CollapsibleSplitContent({sx={}, panelTitle, children, ...props}) {
...
-            // to let SplitPane manage sizing of the collapsible content panel, and let the other content panel grow/shrink
-            primary: collapseSecondContent ? 'second' : 'first',
-            // control the sizing of collapsible content panel
-            size: isCollapsibleOpen ? openSize : collapsedSize,
-            minSize: collapsedSize,
-            // to create collapsing animation effect only when not dragging otherwise it will be jerky
-            onDragStarted: () => setIsSplitterDragging(true),
-            onDragFinished: (currentSize) => {
-                setIsSplitterDragging(false);
-                setIsCollapsibleOpen(currentSize > collapsedSize); // update the open state after dragging
```
jaladh-singhal, line 161: again these code comments (and below) were helpful to understand the complex layout manipulation code.

jaladh-singhal (commented):
@kpuriIpac this is thorough! I tested Firefly and spherex primarily and they are working great (and better than before esp since there's no flicker on dragging the split pane resizer).

One minor issue in collapsible panel that appears in SP results: when you collapse it, it's still 2-4px in height as opposed to 0px height on dev. But it's not a dealbreaker since collapsible panel needs work in future anyways to make it handle both vertical and horizontal cases.

kpuriIpac:
@jaladh-singhal @loitly I addressed your feedback in the latest commit, please have a look when you get a chance. 

I did not move `Pane` inside `SplitContent` because this version of `react-split-pane` reads `pane` sizing props from immediate children passed to `SplitPane`. Even if `SplitContent` returned `Pane` internally, `SplitContent` would be treated as the immediate child so `SplitPanel` would still need to wrap/inject a `Pane` around each `SplitContent`. 

But making `Pane` explicit at consumer sites (making the flow `SplitPanel` -> `Pane` -> `SplitContent`) is still better code than what we had before (though it may look a bit verbose), and as @jaladh-singhal mentioned too consumers can pass additional props via `Pane`. And `SplitPanel` still supports the old sizing API props (at some point we can clean that up, but for this migration I think it's better to leave this as is) by translating those props and merging them on the child `Panes`. 

@jaladh-singhal I also re-added those comments. Good point about keeping those in, at least for now. 

I rebuilt all test builds. Please test all apps (including the IRSA apps in`irsa-ife`) for a couple of searches each or so, and double check `Edit ADQL` on the TAP screen as well.

jaladh-singhal (approved):
Thanks for addressing my comments.

I'm not sure about double use of `<Pane> <SplitContnet> consumer code </SplitContent> </Pane>` in several consumers. But approving it since we can tackle removing that redundancy later on.

PR #1978, opened 2026-07-07
Firefly-2055: Fixed: WCS Match: east-right images are not matching correctly

robyww:
#### Testing
- https://firefly-2055.irsakubedev.ipac.caltech.edu/firefly
- use panstairs hips and image


tgoldina (approved):
Thank you. The WCS alignment now works as expected for my test case with Pan-STARR DR1 HiPS and one of the tiles, e.g. `https://alasky.cds.unistra.fr/Pan-STARRS/DR1/g/Norder11/Dir14780000/Npix14783802.fits`:

Image

PR #1979, opened 2026-07-08
Firefly-2056: fix gpu array overflow and optimize color reading

robyww:
#### Testing
- https://firefly-2056-gpu-overflow.irsakubedev.ipac.caltech.edu/firefly
- test with a 25x25 image

### DIFF
--- a/src/firefly/js/util/Color.js
+++ b/src/firefly/js/util/Color.js
@@ -1,7 +1,7 @@
 import chroma from 'chroma-js';
 import {isArray} from 'lodash';
 
-export default { makeSimpleColorMap, getBWBackground, shadeColor};
+export default { getBWBackground, shadeColor};
 
 export const toRGBA= (color, alpha=1) => chroma.valid(color) ? chroma(color).alpha(alpha).rgba() : [0,0,0,alpha];
 export const toRGB= (color) => chroma.valid(color) ? chroma(color).rgb() : [0,0,0];
--- a/src/firefly/js/util/WebUtil.js
+++ b/src/firefly/js/util/WebUtil.js
@@ -148,8 +148,8 @@ export function loadScript(scriptName) {
 
 /**
  * Create a function that caches last N (up to 20) function call results
- * This is better for functions that take immutable objects. it compare every argument using ===
- * This cache size has to be smaller (<29=0) since if must iterate though all the last results.
+ * This is better for functions that take immutable objects. it compares every argument using ===
+ * This cache size has to be smaller since it must iterate though all the last results.
  * @param fn - the function to wrap
  * @param {number} [cacheSize=1] the number of saved calls with a maximum of 20
  * @return {function}
--- a/src/firefly/js/visualize/rawData/ColorTable.js
+++ b/src/firefly/js/visualize/rawData/ColorTable.js
@@ -10,6 +10,9 @@ import {getColorTableDefinitionInfo} from './ColorTableDefinitions';
 
 export const REVERSED_END_CHAR='R';
 export const NO_COLOR_TABLE='NO_COLOR_TABLE';
+export const PERCENT= 'PERCENT';
+export const PACKED= 'PACKED';
+export const RGB= 'RBG';
 
 
 export const getColorTableMap = () => getColorTableDefinitionInfo().map;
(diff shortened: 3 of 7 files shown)


tgoldina (approved):
The bug affecting the shading of the corner pixel in the single-color image has been fixed.

I don't have enough expertise in this area to review the implementation myself, so I asked Claude's code-review to take a look. Hopefully, its feedback is useful. I'll leave it to your judgment.

## Scope

```
git diff origin/dev...HEAD
```

One commit — `6bb06d204 Firefly-256: fix gpu array overflow and more optimize color` — touching 4 files (~287 lines):

- `src/firefly/js/visualize/rawData/ColorTable.js`
- `src/firefly/js/visualize/rawData/WebGpuJobs.js`
- `src/firefly/js/visualize/rawData/wgsl/imageRemapColor.wgsl`
- `src/firefly/js/visualize/rawData/wgsl/imageSingleBand.wgsl`

## Summary

The diff optimizes GPU color-table lookups by pre-packing the color palette into a single `Uint32` per entry (a new `PACKED` model form) instead of computing RGBA packing per-pixel inside the WGSL shaders. `getColorModel` /
`getColorModelBase` gain a 3-way `modelForm` parameter (`PERCENT` / `PACKED` / `RGB`) replacing the old boolean `asPercent`. Both WGSL shaders were rewritten to read directly from the packed 256-entry array. `imageSingleBand.wgsl` also adds tail-handling logic (`outPixelCnt`) so a per-invocation batch of 4 pixels no longer overwrites past the real output buffer when the image's pixel count isn't a multiple of 4 — this is the "array overflow" fix referenced in the commit message.

Review method: 8 finder angles (3 correctness, 3 cleanup, 1 altitude, 1 conventions) via independent agents, followed by targeted verification of the most consequential candidates.

## Findings (most severe first)

### 1. image3ColorBand.wgsl still has the identical overflow bug this PR fixes elsewhere
**File:** `src/firefly/js/visualize/rawData/wgsl/image3ColorBand.wgsl:37`
**Verdict:** CONFIRMED

`process3CTileViaWebGPU` dispatches `ceil(width*height/256)` workgroups of 256 threads, but this shader's `main()` has **no** `if (idx >= numPixels) return` guard at all (unlike the guard kept/added in `imageSingleBand.wgsl`), and unconditionally writes 4 `u32` values to `outBuf` at `idx*4` and reads `redAry[idx]`/`greenAry[idx]`/`blueAry[idx]` for every invocation. For any tile whose `width*height` isn't a multiple of 256 (routine for edge/corner tiles), the tail threads read past the end of `redAry`/`greenAry`/`blueAry` and write past the end of `outBuf` — the exact "gpu array overflow" class of bug this commit claims to fix, still live for 3-color images.

### 2. Cache-thrashing regression between PACKED and RGB model forms
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:90`
**Verdict:** CONFIRMED (independently verified)

`getColorModelByGPUType` now returns `PACKED`-format models on WebGPU-capable browsers, while `HipsColor.js`'s non-GPU-path `getColorModel(ct)` call (used for irregular HiPS tile sizes) still defaults to `RGB`. Pre-diff, both call sites resolved to `asPercent=false`, so the size-1 `memorizeLastCall` cache in `getColorModelBase` (`ColorTable.js:122`) was always a hit across both call sites for the same `colorTableId`. Post-diff these are different cache keys, so every call to `changeHiPSColor` for an irregular-size tile now recomputes the full 256-step color interpolation from scratch instead of hitting the cache — a real, if minor, performance regression introduced by this "optimize color" diff.

### 3. Typo in new exported constant
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:15`
**Verdict:** CONFIRMED

```js
export const RGB= 'RBG';
```

Letters transposed. Harmless today because every comparison in this diff goes through the imported constant rather than a literal string, but any future code that logs/serializes `modelForm`, or compares against the literal `'RGB'` (matching the JSDoc on lines 87/119, which correctly says "RGB"), will silently fail to match and fall through unexpectedly.

### 4. Bit-packing formula duplicated 3×
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:104` (also `:197`, and pre-existing `WebGpuJobs.js:256`)

The formula `(255<<24) | (b<<16) | (g<<8) | r` appears three times instead of being extracted into one helper. A future change to the packing layout (e.g. premultiplied alpha or a byte-order swap) has to be found and edited in three separate places; missing one silently produces a color/alpha mismatch between the NaN-pixel override path and the normal palette-building path.

### 5. getColorModelBase double-allocates/double-loops to build the PACKED form
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:189-199`

Builds a full 768-entry `Uint32Array` of r,g,b triples, then loops over all 256 entries a second time just to bit-pack them into a new `Uint32Array(256)`, discarding the first array. Every cache-miss for the PACKED form allocates an extra ~3KB scratch array and runs a full second 256-iteration pass, avoidable by packing inline during the original interpolation loop.

### 6. Percent-normalization block duplicated verbatim
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:168-172` and `:182-186`

The three-line `paletteData[idx] /= 255` (r/g/b) block is copy-pasted between the main interpolation loop and the post-loop tail-color handling. A future tweak to normalization needs to be applied in two spots; easy to update one and miss the other.

### 7. getColorModel's PERCENT/RGB branches are near-duplicates
**File:** `src/firefly/js/visualize/rawData/ColorTable.js:96-110`

PERCENT and the RGB (`else`) branch are structurally identical 3-line assignment blocks differing only by a `/255` divisor. Adding another representation or changing how `nanPix` maps into the palette means editing multiple near-identical blocks by hand; the two paths can silently drift apart even though they represent the same value at two scales.

### 8. Overflow fix implemented as a per-shader bandaid instead of at the root
**File:** `src/firefly/js/visualize/rawData/wgsl/imageSingleBand.wgsl:44-55` (root cause: `WebGpuJobs.js:202` / `WebGpuJobs.js:108-117`)

The tail-handling fix (`outPixelCnt` min/if-else) patches this one shader, while the actual mismatch is that `outBuf` is sized to the exact `width*height` but `pixelAry` is padded up to a multiple of 4 by `makePixelBuf`. Padding `outBuf`'s allocation the same way would let every invocation keep the simpler unconditional 4-write path safely (the copy-back already reads only the exact `width*height*4` bytes), and would have also covered `image3ColorBand.wgsl` (finding 1) for free instead of leaving it with the identical unfixed hazard.

## Investigated and refuted

**Candidate:** `imageRemapColor.wgsl`'s clamp bound was narrowed from `clamp(i32(floor(avg)),0,255)` to `clamp(avg,0,254)`, seemingly making palette index 255 unreachable and causing an off-by-one color error at the bright end of the HiPS color ramp.

**Verdict: REFUTED.** `HipsColor.js`'s own CPU fallback (`findPixelIdx`, untouched by this diff) already explicitly caps at index 254 (`if (pixel===255) pixel=254`; `newPixel>254 ? 762 : ...`). The pre-existing, intentional design for HiPS color remap across both CPU and GPU paths already treats 254 as the ceiling — the new shader brings the WebGPU path in line with that established convention rather than introducing a regression.

PR #1980, opened 2026-07-08
FIREFLY-1746: Firefly is slow at loading very wide tables

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1746
- Replace the per-column SELECT DISTINCT loop with a single array_agg(DISTINCT...) query over all qualifying columns.
- Dispatch the check to a background task when the column count exceeds 100.

Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-1746

Upload the wide tables attached to the ticket. Compare the performance of this version with the nightly build version (https://firefly.irsakubedev.ipac.caltech.edu/firefly/).

In the logs, you’ll also notice that the nightly version issues 1000 distinct queries and takes several seconds.


robyww (commented):
looks good.

> Dispatch the check to a background task when the column count exceeds 100.

I don't see where you do this. I see the check (` vals.length <= MAX_COL_ENUM_COUNT`) but not how a background task is dispatched.

robyww (approved):

PR #1982, opened 2026-07-17
FIREFLY-2059: Table query optimization

loitly:
Ticket: https://jira.ipac.caltech.edu/browse/FIREFLY-2059
- Root problem: every sort/filter fully copied every column of every matching row into a new table
- Thin index instead of full copy: store just (ROW_IDX, ROW_NUM) for the resultset, cheap regardless of table width.
- Resultset as a view joining that index to DATA, projecting only requested columns, so every existing consumer keeps working unchanged.
- Healpix index kept separate: its own small (ROW_IDX, pixel) table, computed once and joined at query time, instead of a persisted column on the shared resultset

Test: https://firefly-2059-table-query-optimization.irsakubedev.ipac.caltech.edu/firefly/
Compare this with the nightly build: https://firefly.irsakubedev.ipac.caltech.edu/firefly/

Try this: 10 million rows, 150 columns. The same FITS table converted into parquet to avoid high memory usage during loading.  https://irsawebdev1.ipac.caltech.edu:9201/test-data/10m-150col.parquet
You’ll notice a significant performance improvement during sorting and filtering.

These smaller tables work smoothly.
https://irsawebdev1.ipac.caltech.edu:9201/test-data/table_1mil.parquet
https://irsawebdev1.ipac.caltech.edu:9201/test-data/wise-1_6-million.tbl

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/db/BaseDbAdapter.java
@@ -164,41 +164,15 @@ abstract public class BaseDbAdapter implements DbAdapter {
     }
     
 
-    public void createTempResults(TableServerRequest treq, String resultSetID) {
+    /**
+     * Builds resultSetID table including _DD/_META/_AUX related tables.
+     */
+    public final void createTempResults(TableServerRequest treq, String resultSetID) {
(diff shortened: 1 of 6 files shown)


robyww (approved):
This looks fine but I I don’t really understand much. It feels lowlevel enough that I think it is best we decided not to put it in the RC. I did not test, I am too loaded with rubin issues.

PR #1985, opened 2026-07-21
FIREFLY-1971: Add spectral lines plotting in the spectral charts

jaladh-singhal:
Fixes [FIREFLY-1971](https://jira.ipac.caltech.edu/browse/FIREFLY-1971)

**This pass focuses on plotting; the next pass will cover loading (separate dialog and configurable line groups on app level).**

Added a spectral lines overlay for spectrum charts — users can toggle a recommended line list and see the selected lines plotted as labeled, hoverable vertical lines, correctly unit- and redshift-corrected. 

- Recommended line list is served from a new `SpectralLinesProcessor` endpoint backed by a bundled CSV, loaded headlessly into a client-side table so it can be selected/filtered without appearing in the results UI
- Lines list uses a checkbox group (not single-select) so multiple sources can combine later — only "Recommended" is wired up, "Upload mine" is disabled pending the loading pass
- Bumped `plotly.js-dist-min` to `~3.6.0` for native shape labels, used to render each line's species text directly on the chart
- Added a `makeShapeHoverTrace` helper (`ChartUtil.js`) for spectral lines — a companion scatter trace per shape's `hovertext` since plotly doesn't support hover on shapes. I tried Plotly annotations for hover but they caused pan/zoom lags at realistic line counts. 
- "Redshift correction" options (previously Rest-Frame-only) are now always visible: in Rest Frame they still correct the spectrum (unchanged); in Observed Frame they now correct the lines' rest wavelengths to match the spectrum. A "Redshift Correction: Spectrum / Spectral Lines" label clarifies which

## Testing
https://firefly-1971-spectral-lines.irsakubedev.ipac.caltech.edu/firefly/

- Load a spectra from the ticket, toggle "Spectral lines" on, confirm lines plot with correct hover info, check if selection in table reflects correctly in the plot.
- Toggle Rest vs Observed Frame: spectrum shifts in Rest Frame, lines shift to match in Observed Frame; "Redshift correction" label switches accordingly
- Pan/zoom with several lines plotted without any lags
- Regression test:
    - redshift correction of spectrum (in rest frame) working as before (spitzer TOPOCENTER spectra, spitzer CUSTOM spectra)
    - other chart types, since `PlotlyChartArea.jsx`/`ChartUtil.js` and the `plotly.js` bump touch shared rendering code

### DIFF
--- a/src/firefly/js/charts/ChartUtil.js
+++ b/src/firefly/js/charts/ChartUtil.js
@@ -377,6 +377,61 @@ export function flattenAnnotations(annotations) {
     return [];
 }
 
+const SHAPE_HOVER_STEPS = 10; // points per line stacked along the shape's height, for a hover target that tracks the cursor vertically
+
+/**
+ * Builds companion hover trace(s) for Plotly shapes that carry a `hovertext`.
+ *
+ * Shapes themselves can't be hovered (no hovertext/hoverinfo support on shapes as of now unlike annotations; see
+ * plotly.js#4780: https://github.com/plotly/plotly.js/issues/4780), so this fakes it with a real scatter
+ * trace: a fully transparent, null-separated line per shape, on a fixed 'y2' axis (paired with the returned
+ * `layout.yaxis2`) so it stays aligned with the shape's fixed paper-relative span regardless of the real
+ * y-axis' zoom/pan.
+ *
+ * Not specific to any one feature — any shape with `hovertext` set gets a hover target this way. One trace is
+ * built per distinct `legendgroup` found among the hoverable shapes (and sharing that legendgroup + showlegend:false
+ * means toggling a shape's legend entry hides its hover trace along with it).
+ *
+ * NOTE: Only handles vertical, paper-height line shape geometry (`yref:'paper', y0:0, y1:1, x0=x1`);
+ * shapes of other types/positions are ignored.
+ *
+ * @param {Array<object>} shapes - candidate shapes, e.g. from layout.shapes
+ * @returns {{traces: Array<object>, layout: object}} traces to add to `data`, and the `layout` fragment
+ * (the 'y2' axis they're plotted on) to merge in alongside them — empty/{} when there's nothing hoverable
+ */
+export function makeShapeHoverTrace(shapes=[]) {
+    const hoverable = shapes.filter((s) => s?.type === 'line' && s.yref === 'paper' && s.hovertext);
+    if (hoverable.length === 0) return {traces: [], layout: {}};
+
+    const byGroup = new Map();
+    hoverable.forEach((s) => {
+        const group = s.legendgroup;
+        if (!byGroup.has(group)) byGroup.set(group, []);
+        byGroup.get(group).push(s);
+    });
+
+    const traces = Array.from(byGroup.entries()).map(([group, groupShapes]) => {
+        const x = [], y = [], hovertext = [];
+        groupShapes.forEach((s) => {
+            for (let i = 0; i < SHAPE_HOVER_STEPS; i++) {
+                x.push(s.x0);
+                y.push(i / (SHAPE_HOVER_STEPS - 1));
+                hovertext.push(s.hovertext);
+            }
+            x.push(null); y.push(null); hovertext.push(null);
+        });
+        return {
+            x, y, hovertext,
+            yaxis: 'y2',
+            type: 'scatter', mode: 'lines', hoverinfo: 'text',
+            line: {width: 10, color: 'rgba(0,0,0,0)'},
+            legendgroup: group, showlegend: false, name: '',
+        };
+    });
+
+    return {traces, layout: {yaxis2: {overlaying: 'y', range: [0, 1], visible: false, fixedrange: true}}};
+}
+
 export function updateSelection(chartId, selectInfo) {
     const {data, activeTrace=0} = getChartData(chartId);
 
--- a/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
+++ b/src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
@@ -47,6 +47,32 @@ export function getUnitConvExpr({cname, from, to, alias, args=[]}) {
 }
 
 /**
+ * Converts a single numeric value from one unit to another, using the same conversion table as {@link getUnitConvExpr}.
+ * Unlike getUnitConvExpr, this returns a number rather than a SQL-like expression string.
+ * @param {number} value    the value to convert
+ * @param {string} from     from unit
+ * @param {string} to       to unit
+ * @returns {number|undefined} the converted value, or undefined if no conversion path exists between the units
+ */
+export function convertUnitValue(value, from, to) {
+    const {unit: fromKey, factor: fromFactor} = normalizeUnit(from);
+    const {unit: toKey, factor: toFactor} = normalizeUnit(to);
+
+    const formula = UnitXref?.[fromKey]?.[toKey]; // SQL-like expression string
+    if (!formula) return undefined;
+
+    // every UnitXref formula is one of: '%s', '%s * n', or '%s / n', apply the operator on number values
+    const [, op, n] = formula.match(/^%s(?:\s*([*/])\s*([\d.eE+-]+))?$/) ?? [];
+    let result = op === '*' ? value * Number(n) : op === '/' ? value / Number(n) : value;
+
+    // handle factors if present in the units
+    if (fromFactor) result *= Number(fromFactor);
+    if (toFactor) result /= Number(toFactor);
+
+    return result;
+}
+
+/**
  * returns an array of options for unit conversion.
  * @param {string} unit         the unit to get the conversion options for
  * @returns {Array<{value: string, label: LatexDelimited}>}
--- a/src/firefly/js/charts/ui/PlotlyChartArea.jsx
+++ b/src/firefly/js/charts/ui/PlotlyChartArea.jsx
@@ -5,7 +5,8 @@ import {PlotlyWrapper} from './PlotlyWrapper.jsx';
 import {showInfoPopup} from '../../ui/PopupUtil.jsx';
 
 import {dispatchChartHighlighted, dispatchChartUpdate, dispatchSetActiveTrace, getAnnotations, getChartData, usePlotlyReact} from '../ChartsCntlr.js';
-import {clearChartConn, flattenAnnotations, handleTableSourceConnections, isSpectralOrder, isScatter2d} from '../ChartUtil.js';
+import {clearChartConn, flattenAnnotations, handleTableSourceConnections, isSpectralOrder, isScatter2d,
+    makeShapeHoverTrace} from '../ChartUtil.js';
 import {useStoreConnector} from 'firefly/ui/SimpleComponent.jsx';
 import {Skeleton, useTheme} from '@mui/joy';
 
@@ -68,10 +69,16 @@ export function PlotlyChartArea({chartId, widthPx, heightPx, thumbnail}) {
             }
         }
     }
+
+    // hoverable shapes (e.g. spectral lines) can't show hovertext directly
+    // so build their companion hover trace(s) plus the axis layout they need
+    const {traces: hoverTraces, layout: hoverLayout} = makeShapeHoverTrace(layout.shapes);
+    pdata = pdata.concat(hoverTraces);
+
     const {chartWidth, chartHeight} = calculateChartSize(widthPx, heightPx, xyratio, stretch);
 
     const showlegend = data.length > 1;
-    const playout = cloneDeep(Object.assign({showlegend}, adjustLayout(layout, theme), {width: chartWidth, height: chartHeight, annotations}));
+    const playout = cloneDeep({showlegend, ...adjustLayout(layout, theme), width: chartWidth, height: chartHeight, annotations, ...hoverLayout});
 
     const style = {float: 'left'};
     if (chartWidth > widthPx || chartHeight > heightPx) {
--- a/src/firefly/js/charts/ui/options/SpectralLines.jsx
+++ b/src/firefly/js/charts/ui/options/SpectralLines.jsx
@@ -1,0 +1,181 @@
+import React, {useEffect} from 'react';
+import {get} from 'lodash';
+import {Stack} from '@mui/joy';
+import {SwitchInputField} from 'firefly/ui/SwitchInputField';
+import {CheckboxGroupInputField} from 'firefly/ui/CheckboxGroupInputField';
+import {CollapsibleGroup, CollapsibleItem} from 'firefly/ui/panel/CollapsiblePanel';
+import {useStoreConnector, useFieldValueOnly} from 'firefly/ui/SimpleComponent';
+import {getChartData} from '../../ChartsCntlr.js';
+import {isKnownRefPos} from 'firefly/voAnalyzer/SpectrumDM';
+import {canUnitConv, convertUnitValue} from '../../dataTypes/SpectrumUnitConversion.js';
+import {makeTblRequest} from 'firefly/tables/TableRequestUtil';
+import {dispatchTableFetch, dispatchTableUiUpdate, dispatchTableSelect} from 'firefly/tables/TablesCntlr';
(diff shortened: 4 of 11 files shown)


src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
```diff
@@ -46,6 +46,32 @@ export function getUnitConvExpr({cname, from, to, alias, args=[]}) {
...
+export function convertUnitValue(value, from, to) {
+    const {unit: fromKey, factor: fromFactor} = normalizeUnit(from);
+    const {unit: toKey, factor: toFactor} = normalizeUnit(to);
+
+    const formula = UnitXref?.[fromKey]?.[toKey]; // SQL-like expression string
+    if (!formula) return undefined;
+
+    // every UnitXref formula is one of: '%s', '%s * n', or '%s / n', apply the operator on number values
+    const [, op, n] = formula.match(/^%s(?:\s*([*/])\s*([\d.eE+-]+))?$/) ?? [];
+    let result = op === '*' ? value * Number(n) : op === '/' ? value / Number(n) : value;
```
loitly, line 66: I like your clever use of the existing UnitXref. Regarding the expression evaluation, this regex currently works but is restricted to the current set of formulas. Have you considered using JavaScript to evaluate the expression, as shown below?

```js
const expr = formula.replace(/%s/g, ’s’);
let result = new Function(’s’, `return ${expr};`)(Number(value));
```
This approach will work for any arithmetic-based formula, such as `(%s - 32) * 5/9`.

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }, []);
+
+    const {enabled: initialEnabled = false, source: initialTblId = RECOMMENDED_LINES_TBL_ID} =
+        useStoreConnector(() => get(getChartData(chartId), 'fireflyLayout.spectralLines')) ?? {};
+    const initialSourceOptions = tblIdToSourceOption(initialTblId);
+
+    // spectral lines need a redshift to correct against, which only exists when Spectral Frame options are shown (as opposed to read-only value)
+    const hasSpectralFrame = useStoreConnector(() =>
+        isKnownRefPos(getChartData(chartId)?.fireflyData?.[activeTrace]?.spectralFrame?.refPos));
```
loitly: Can `hasSpectralFrame` toggle between true and false for the same chartId and activeTrace? If not, you don’t need useStoreConnector. However, if useStoreConnector is required and chartId or activeTrace can change, you must include them as dependencies.

```js
useStoreConnector( () => ‘blah’, [chartId, activeTrace]);
```

src/firefly/js/charts/ChartUtil.js
```diff
@@ -377,6 +377,61 @@ export function flattenAnnotations(annotations) {
...
+ * means toggling a shape's legend entry hides its hover trace along with it).
+ *
+ * NOTE: Only handles vertical, paper-height line shape geometry (`yref:'paper', y0:0, y1:1, x0=x1`);
+ * shapes of other types/positions are ignored.
+ *
+ * @param {Array<object>} shapes - candidate shapes, e.g. from layout.shapes
+ * @returns {{traces: Array<object>, layout: object}} traces to add to `data`, and the `layout` fragment
+ * (the 'y2' axis they're plotted on) to merge in alongside them — empty/{} when there's nothing hoverable
+ */
+export function makeShapeHoverTrace(shapes=[]) {
```
loitly, line 402: Wow, I didn’t realize so much effort went into these lines overText.

loitly (approved):
Tested and it worked as described.  Good to go.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -164,19 +168,18 @@ const getRedshiftInfo = (inFields, path, fireflyData, activeTrace) => {
...
-            ?.label.replace(/ with confidence.*$/, ''); //don't need to show confidence in axis label
-
-        if(redshiftOption==='userSpecified') {
-            redshift = userSpecifiedRedshift;
-            redshiftLabel = `Redshift = ${userSpecifiedRedshift}`;
-        }
+        redshiftLabel = redshiftOption==='userSpecified'
+            ? `Redshift = ${userSpecifiedRedshift}`
+            : getRedshiftOptions(fireflyData[activeTrace]).find((opt)=> opt.value===redshiftOption)
+                ?.label.replace(/ with confidence.*$/, ''); //don't need to show confidence in axis label
```
robyww: this is a lot in a tertiary operator, why not make the second option a small function?  That way it would be easier to understand when the find and replace are doing.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -303,6 +306,29 @@ export function submitChangesSpectrum({chartId, activeTrace, fields, tbl_id, ren
...
+    const spectralLinesEnabled = toBoolean(fields['spectralLines.enabled']);
+    const spectralLinesTblId = sourceOptionToTblId(fields['spectralLines.sourceOptions']);
+    fields = omit(fields, ['spectralLines.enabled', 'spectralLines.sourceOptions']);
+
+    // persisted only so UI controls can seed their initial state from the chart data next time this dialog opens
+    fields = updateSet(fields, ['fireflyLayout.spectralLines.enabled'], spectralLinesEnabled);
+    fields = updateSet(fields, ['fireflyLayout.spectralLines.source'], spectralLinesTblId);
+
+    // replace only this feature's shapes, preserve others
+    const otherShapes = (getChartData(chartId)?.layout?.shapes ?? []).filter((s) => s.legendgroup !== SPECTRAL_LINES_GROUP);
```
robyww: I was trying to understand the goal of this line in an `undefined` case. it is `[]`.  It would be easier to follow like this.

```js
const otherShapes = getChartData(chartId)?.layout?.shapes?.filter((s) => s.legendgroup !== SPECTRAL_LINES_GROUP) ?? [];
```

src/firefly/js/charts/ui/PlotlyChartArea.jsx
```diff
@@ -68,10 +69,16 @@ export function PlotlyChartArea({chartId, widthPx, heightPx, thumbnail}) {
...
+    // hoverable shapes (e.g. spectral lines) can't show hovertext directly
+    // so build their companion hover trace(s) plus the axis layout they need
+    const {traces: hoverTraces, layout: hoverLayout} = makeShapeHoverTrace(layout.shapes);
+    pdata = pdata.concat(hoverTraces);
+
     const {chartWidth, chartHeight} = calculateChartSize(widthPx, heightPx, xyratio, stretch);
 
     const showlegend = data.length > 1;
-    const playout = cloneDeep(Object.assign({showlegend}, adjustLayout(layout, theme), {width: chartWidth, height: chartHeight, annotations}));
+    const playout = cloneDeep(Object.assign({showlegend}, adjustLayout(layout, theme), {width: chartWidth, height: chartHeight, annotations}, hoverLayout));
```
robyww: we should take out old Object.assign lines when possible.
```js
 const playout = cloneDeep({showlegend, ...adjustLayout(layout, theme), width: chartWidth, height: chartHeight, annotations, hoverLayout});


```

src/firefly/js/charts/ChartUtil.js
```diff
@@ -377,6 +377,61 @@ export function flattenAnnotations(annotations) {
...
+    if (hoverable.length === 0) return {traces: [], layout: {}};
+
+    const byGroup = new Map();
+    hoverable.forEach((s) => {
+        const group = s.legendgroup;
+        if (!byGroup.has(group)) byGroup.set(group, []);
+        byGroup.get(group).push(s);
+    });
+
+    const traces = Array.from(byGroup.entries()).map(([group, groupShapes]) => {
```
robyww, line 413: I think you can now use `map` directly from `map.entries()` working as a iterator and the end the whole thing with `.toArray()`.

src/firefly/js/ui/CheckboxGroupInputField.jsx
```diff
@@ -24,10 +24,10 @@ export function CheckboxGroupInputFieldView({fieldKey, onChange, label, tooltip:
                 {label && <FormLabel {...slotProps?.label}>{label}</FormLabel>}
                 {/*following should be nested in a <FormGroup/> once joy-ui exposes it: https://mui.com/material-ui/react-checkbox/#formgroup*/}
                 <Stack className='ff-Checkbox-container' spacing={orientation==='vertical'?1:2} direction={orientation==='vertical' ? 'column' : 'row'}>
-                    {options.map( ({value,label,tooltip}) => {
+                    {options.map( ({value,label,tooltip,...checkboxProps}) => {
                         const cb= (
                             <Checkbox {...{ className:'ff-Checkbox-item', size:'sm', name:fieldKey, key:value, value,
```
robyww, line 29: good idea. I had never considered doing this before.

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+        type: 'line',
+        x0: x, x1: x,
+        y0: 0, y1: 1,
+        xref: 'x', yref: 'paper',
+        line: {color: SPECTRAL_LINE_COLOR, width: 1, dash: 'dot'},
+        label: {
+            text: label,
+            textposition: 'end',
+            yanchor: 'bottom',
+            font: {size: 9.5, color: SPECTRAL_LINE_COLOR, family: "'SF Mono', ui-monospace, monospace"},
```
robyww: `"'SF Mono', ui-monospace, monospace"` should probably be a constant with your other ones.

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+            dispatchTableFetch(request); // headless: doesn't render in results UI
+
+            // all lines are checked by default; set post-load since request.META_INFO.selectInfo can silently get lost in transit
+            onTableLoaded(RECOMMENDED_LINES_TBL_ID).then((tableModel) => {
+                dispatchTableSelect(RECOMMENDED_LINES_TBL_ID, {
+                    selectAll: true, exceptions: new Set(), rowCount: tableModel.totalRows
+                });
+            });
+        }
+    }, []);
```
robyww: I think you should break this part out in an async function so you can use await to make it easier to follow.
something like `ensureRecommandedLines()`

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+        },
+        hovertext: `<b>${label}</b>  λ ${x} ${xUnit}` +
+            (transition ? `<br>${transition}` : '') + (phase ? `<br>${phase}` : ''),
+        legendgroup: SPECTRAL_LINES_GROUP,
+        showlegend: i === 0, // legend entry goes only on the first shape item
+        name: 'Lines',
+    }));
+}
+
+export function SpectralLinesOptions({activeTrace, groupKey, chartId}) {
```
robyww: not necessary to pass groupKey if you use `useFieldGroupValue` or `useFieldValueOnly`

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+    }, []);
+
+    const {enabled: initialEnabled = false, source: initialTblId = RECOMMENDED_LINES_TBL_ID} =
+        useStoreConnector(() => get(getChartData(chartId), 'fireflyLayout.spectralLines')) ?? {};
+    const initialSourceOptions = tblIdToSourceOption(initialTblId);
+
+    // spectral lines need a redshift to correct against, which only exists when Spectral Frame options are shown (as opposed to read-only value)
+    const hasSpectralFrame = useStoreConnector(() =>
+        isKnownRefPos(getChartData(chartId)?.fireflyData?.[activeTrace]?.spectralFrame?.refPos));
+    const isEnabledField = useStoreConnector(() => toBoolean(getFieldVal(groupKey, ENABLED_KEY)));
```
robyww: better to use
```js
const isEnabledField= useFieldValueOnly(ENABLED_KEY, false);  // (second param is the default, I was guessing false)
```

same with line 132, then passing groupKey is not necessary.

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+    const isEnabled = hasSpectralFrame && isEnabledField;
+
+    // TODO: combine different source tables to a client-side table
+    const sourceOptions = useStoreConnector(() => getFieldVal(groupKey, SOURCE_OPTIONS_KEY, initialSourceOptions));
```
robyww: here as well

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -0,0 +1,180 @@
...
+                                  initialState={{value: initialEnabled}}/>
+            </Stack>
+            {isEnabled && (
+                <Stack spacing={1}>
+                    <CheckboxGroupInputField fieldKey={SOURCE_OPTIONS_KEY}
+                                             label='Lines list:'
+                                             initialState={{value: initialSourceOptions}}
+                                             options={[
+                                                 {label: 'Recommended', value: SOURCE_RECOMMENDED},
+                                                 {label: 'Upload mine', value: SOURCE_UPLOAD, disabled: true}
```
robyww, line 153: I do like the improvement to CheckboxGroupInputField

robyww (approved):
It looks good. I made a few clean up comments, after that I think it is ready to merge.

src/firefly/js/charts/ChartUtil.js
robyww: I just check https://caniuse.com/ on this suggestion. It is not supported until safari 18.4. So you should hold off.

All the other browsers support it but we claim to support safari 17. It is possible that babel shims it but that is hard to test.

src/firefly/js/charts/ChartUtil.js
jaladh-singhal: Yes, I tried different options because shapes don't support hover yet.
- Annotations cause lag since each of it attaches its own DOM instead of making same SVG
- Transparent scatter on y1 axis conflict with our regular traces because we need yref paper like in shapes

src/firefly/js/charts/dataTypes/SpectrumUnitConversion.js
jaladh-singhal: Yes, I didn't think of Function constructor but I thought of using `eval()` (which is same as this) but I didn't go with it because it triggers CSP warnings. I tried your syntax and I see same error as I saw with `eval` when typing an example in console:
```Uncaught EvalError: Evaluating a string as JavaScript violates the following Content Security Policy directive because 'unsafe-eval' is not an allowed source of script: script-src chrome://resources chrome://webui-test 'self'"```

I know since we're not sourcing it externally but from a local dict `UnitXRef` it should be fine but I didn't risk it thinking if it changes in future or if gets flagged in the vulnerability fixing work you had to do for cloud deployments.

Also our formulas are really simple `/` and `*` so we don't need full arithmetic capabilites yet.

src/firefly/js/charts/ui/options/SpectralLines.jsx
jaladh-singhal: Yes, `hasSpectralFrame` can toggle for the same `chartId`/`activeTrace`: `fireflyData[activeTrace].spectralFrame` is populated asynchronously as the chart's data loads, so `refPos` can go from unknown to known after mount even without switching traces so `useStoreConnector` is needed. And `activeTrace` can also change without unmounting this component (the trace selector just dispatches a store update, no remount).

I've added `[chartId, activeTrace]` as deps - good catch!

src/firefly/js/charts/ui/options/SpectralLines.jsx
jaladh-singhal: Good idea, I keep forgetting about `useFieldValueOnly` since we added it recently.

src/firefly/js/charts/ui/options/SpectralLines.jsx
jaladh-singhal: Done!

PR #1986, opened 2026-07-23
FIREFLY-2021: Improve job monitor UI

loitly:
Ticket https://jira.ipac.caltech.edu/browse/FIREFLY-2021
- Preserve sort and filter state across automatic refreshes
- Add a clear filter button when filters are active
- Show a loading indicator while data is being refreshed
- Highlight newly added job after sort/filter is applied instead of always placing it at the top

Test: 
https://firefly-2021-job-monitor-update.irsakubedev.ipac.caltech.edu/applications/spherex/
https://firefly-2021-job-monitor-update.irsakubedev.ipac.caltech.edu/firefly/

Addressed the issues reported in the ticket:
- I was able to stop a job and then delete it successfully. Could you please test this again?
- Added a Clear Filters button next to the filter inputs. Let me know if you think it would be better placed elsewhere.
- Newly created jobs no longer automatically appear at the top of the table now that sort and filter state is preserved across refreshes. Instead, the new job is highlighted after the current sort and filter are applied. This behavior change should be documented if the previous behavior was documented.

Additional Changes after PR:

- Moved the **Job Summary** into the table toolbar, enabling use of the built-in **Filter** and **Clear Filter** actions.
- Removed data types from the table headers.
- Cleaned up the table header layout so header heights render consistently regardless of the show/hide options.

Test URL instances updated with the latest change.  Please test again.

### DIFF
--- a/src/firefly/js/core/background/JobMonitor.jsx
+++ b/src/firefly/js/core/background/JobMonitor.jsx
@@ -1,11 +1,13 @@
 import React, {useContext, useEffect, useState} from 'react';
 import {object, string, shape} from 'prop-types';
-import {IconButton, Button, Sheet, Stack, Typography, ListItemDecorator, Tab} from '@mui/joy';
+import {IconButton, Button, Sheet, Skeleton, Stack, Typography, ListItemDecorator, Tab} from '@mui/joy';
 import moment from 'moment';
 
 import {Slot, useStoreConnector} from '../../ui/SimpleComponent';
 import {getBackgroundInfo, getJobInfo, getJobTitle, getMetadata, getPhaseTips, isActive, isArchived, isDone, isExecuting, isFail, isSearchJob, isSuccess, loadAllJobs, Phase, loadJobResult, fixTapResults, getProgressMsg} from './BackgroundUtil';
 import {TablePanel} from '../../tables/ui/TablePanel';
+import {OptionsFilterStats} from '../../tables/ui/TablePanelOptions';
+import {HeaderText} from '../../tables/ui/TableRenderer';
 import {getAppOptions} from '../AppDataCntlr';
 import {dispatchBgJobInfo, dispatchBgSetInfo, dispatchJobCancel, dispatchJobRemove, dispatchSetJobNotif} from './BackgroundCntlr';
 import {InputField} from '../../ui/InputField';
@@ -40,7 +42,9 @@ export const jobMonitorPath = '/jobMonitor';
 export const jobMonitorGroupKey = 'jobMonitor';
 export const useLocalTimeKey = 'useLocalTime';
 
-const jobIdColIdx = 7;  // the index of the jobId column in the table
+const jobIdColIdx = 7;   // the index of the jobId column in the table
+const jobHistoryTblId = 'JobHistoryTable';
+let seenJobIds = null;      // map of jobIds seen in the previous poll, used to detect newly submitted jobs
 
 export function JobMonitor({initArgs, help_id, slotProps, ...props}) {
     const pollInterval = getAppOptions()?.background?.historyPollInterval || 30000;       // every 30 seconds
@@ -60,7 +64,7 @@ export function JobMonitor({initArgs, help_id, slotProps, ...props}) {
     const titleProps = {...slotProps?.title, ...title};
     const tableProps = {...slotProps?.table, ...table};
     const width = useStoreConnector(() => {
-        const {columnWidths=[]} = getTableUiById('JobHistoryTable-ui') || {};
+        const {columnWidths=[]} = getTableUiById(`${jobHistoryTblId}-ui`) || {};
         return columnWidths.reduce((acc, val, idx) => idx < 8 ? acc + val : acc, 3);
 
     });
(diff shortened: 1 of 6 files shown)


lrebull:
Screenshot 2026-07-24 at 10 59 01 AM
Having trouble running sp jobs in your build. They fail so quickly I had to work to get this shot with one in progress. I can't test some of the other things while jobs are running or queued.
I appear to be able to delete failed jobs but that was never the problem on ops.
I see why you put the cancel filters where you did, and I agree that that's the best place for it, given the layout, but I bet others will... Have Opinions. 

Can you fix it so that I can run sp jobs? Then I can test the rest of the issues.

src/firefly/js/tables/ui/TablePanelOptions.jsx
```diff
@@ -142,16 +142,16 @@ TablePanelOptions.propTypes = {
...
     const filterCnt = useStoreConnector(() => getFilterCount(getTblById(tbl_id)));
     const clearFilters = () => dispatchTableFilter({tbl_id, filters: ''});;
 
     if (filterCnt === 0) return null;
     return (
         // wrapper stack added because ToolbarButton did not expose root
-        <Stack sx = {{[`& .${badgeClasses.root}`]: {margin: '1px 3px 0 0'}}}>
-            <ClearFilterButton iconButtonSize='34px' onClick={clearFilters}
+        <Stack sx={{[`& .${badgeClasses.root}`]: {margin: '1px 3px 0 0'}}} {...props}>
+            <ClearFilterButton iconButtonSize={iconButtonSize} onClick={clearFilters}
```
jaladh-singhal: I'd rather pass `props` to `ClearFilterButton` and maybe a `slotProp.root` to the `Stack` - this way you can override `tip` too. I don't think the default tooltip makes sense for job monitor but  @lrebull can verify the wording.

src/firefly/js/core/background/JobMonitor.jsx
```diff
@@ -270,6 +289,15 @@ function PhaseRenderer({cellInfo}) {
     );
 }
 
+function ControlHeadRenderer({col}) {
+    return (
+        <Stack alignItems='center' justifyContent='space-between' height={1} mt='1px'>
+            <HeaderText val={col?.label || col?.name} level='title-sm'/>
+            <OptionsFilterStats tbl_id={jobHistoryTblId}  alignSelf='start'/>
```
jaladh-singhal: I liked it better as center aligned (or similar to how Control Rendered places its cell contents - with same margin/padding)

Image

jaladh-singhal: Another placement of this button could be:
2. in the job summary area below/right of time zone
3. bring the table toolbar back and let it show filter controls how we do elsewhere?

I'm honestly not sure which one is better since this is not a typical table.

jaladh-singhal (changes requested):
@loitly this is working as expected. Clear filter looks correct as in other tables.

However I noticed a bug: Apply any filter to enum type columns -> click on clear filter button -> the enum selection (▼) disappears. Sometimes they come back after a delay though (can't reliably reproduce the delay duration)

src/firefly/js/core/background/JobMonitor.jsx
loitly: I don't like the center align too much because it seems awkward to have clear filter there.
Another option:
Move the entire ‘Job Summary’ row to the Table’s header and enable it with filter. It’s a bigger change, but worth doing.  I will give it a try, but will require scientist to approve new look.

src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -219,13 +258,11 @@ function EnumSelect({col, tbl_id, filterInfoCls, onFilter}) {
 export function SelectableHeader ({checked, onSelectAll, showUnits, showTypes, showFilters, showSelectRowFilter=true,
                                       onFilterSelected, sx}) {
     return (
-        <Stack alignItems='center' height={1} justifyContent='space-between' py='2px' sx={sx}>
+        <Stack alignItems='center' height={1} justifyContent='space-between' pt='4px' pb='2px' sx={sx}>
```
jaladh-singhal, line 261:
```suggestion
        <Stack alignItems='center' height={1} justifyContent='space-between' pt={0.5} pb={0.25} sx={sx}>
```

src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -36,6 +36,29 @@ import {FilterButton} from 'firefly/visualize/ui/Buttons.jsx';
 
 export const headerStyle = {fontSize:'var(--joy-fontSize-sm)', fontWeight:'var(--joy-fontWeight-md)'};  // maybe faulty becuase it's translated from Typography title-sm, which is dynamic.
 
+/**
+ * Pixel heights of the rows making up a column header.  fixed-data-table needs the header's
+ * total height up front, so these must match what HeaderCell renders.  HeaderText's font size is
+ * pinned to headerFontSize so the text rows stay in sync; headers do not scale with the font .
+ */
+export const headerFontSize = 12.25;                                // Joy's fontSize.sm at firefly's root size
```
jaladh-singhal, line 44: I wonder if we can pull it from joy ui theme instead of hardcoding this. What happens if Rubin or someone else configures Firefly to render different than our default font at a different size?

src/firefly/js/tables/ui/TableRenderer.js
```diff
@@ -36,6 +36,29 @@ import {FilterButton} from 'firefly/visualize/ui/Buttons.jsx';
...
+export const headerLineHeight = Math.ceil(headerFontSize * 1.2);    // HeaderText uses lineHeight 1.2
+
+export const headerRowHeight = {
+    label:  22,
+    units:  headerLineHeight,
+    types:  headerLineHeight,
+    filter: 25,
```
jaladh-singhal, line 51: same here for label and filter.

src/firefly/js/core/background/JobMonitor.jsx
```diff
@@ -250,11 +265,30 @@ function JobMonitorTable({help_id, ...props}) {
...
-                    {...{showToolbar: false, selectable:false, showFilters:true, showOptionButton:false}}
-                    slotProps={{...props}}
-        />
+        <Stack sx={{position: 'relative', height: 1, ...sx}}>
+            <Skeleton loading={loading} sx={sx}>
+                <TablePanel rowHeight={32} tbl_id={tbl_id} help_id={help_id} renderers={renderers}
+                            sx={{'& .fixedDataTableCellGroupLayout_cellGroup > :last-child': {borderRight: 'none'}, ...sx}}
+                            showToolbar={true}
+                            showTitle={false}
+                            leftButtons={[() => <JobSummary jobs={jobs} overflow={overflow}/>]}
```
jaladh-singhal, line 274: [No fix required] Just noticed that JobSummary doesn't wrap in smaller than default width of this table. Although this is how ops behaves too.

jaladh-singhal (approved):
@loitly I like your updated layout utilizing table toolbar, also the conditional notification part looks clean and separate now.

I've left some stylistic/CSS improvement comments - none of them are dealbreaker for merging this as-is.

jaladh-singhal:
> However I noticed a bug: Apply any filter to enum type columns -> click on clear filter button -> the enum selection (▼) disappears. Sometimes they come back after a delay though (can't reliably reproduce the delay duration)

@loitly not sure if you noticed this issue - it happens intermittently for me.

src/firefly/js/tables/ui/TableRenderer.js
loitly: Our table library requires a fixed pixel height for headers, rows, and footers to enhance rendering performance. To simplify the logic of calculating the precise pixel height for the user’s specified font and size combination, I’ve defined the exact height for each component of the header, irrespective of their settings. However, if there’s a need to support dynamic height at the expense of re-rendering the table, we’ll add that hook (which may require an update to the library).

src/firefly/js/tables/ui/TableRenderer.js
loitly: To enforce a predefined pixel height, as mentioned in the previous comment.

loitly:
> @loitly not sure if you noticed this issue - it happens intermittently for me.

I’ve fixed it. Thanks for pointing it out; I missed it before.

PR #1992, opened 2026-08-11
FIREFLY-2006: Charts Grouping functionality 

kpuriIpac:
**Ticket**: [FIREFLY-2006](https://jira.ipac.caltech.edu/browse/FIREFLY-2006) 
- Added grouping functionality to scatter plots (to allow user to **_group by_** any enum like columns) 
- Generalizing grouping to all scatter plots (we previously already supported grouping by **_order_** column for spectra)  exposed several existing chart bugs (some related to grouping, some to filtering/selection, etc.).   
  - This PR fixes those bugs as well. 
- `makeScatterGroupByChanges` is the main function handling grouping logic (in `ChartUtil`). Some other supporting functions are also in `ChartUtil`. My reasoning behind keeping these here was that even though we only offer grouping for scatter plots, it's general enough to belong in `ChartUtil`. 
  - open to feedback and moving this elsewhere, perhaps `ScatterOptions`. 
- **_Final note_**: I tried to keep code separated (following separation of concern logic) & generic, as much as possible. And tightened up some areas prone to frequent errors. But an overall refactor of Firefly's Charts (in the near future, perhaps last cycle of 2026) will make working with charts much easier moving forward. 

**Testing**: 
- [IRSAViewer](https://firefly-2006-charts-groups.irsakubedev.ipac.caltech.edu/irsaviewer), [Spherex](https://firefly-2006-charts-groups.irsakubedev.ipac.caltech.edu/applications/spherex), [Euclid](https://firefly-2006-charts-groups.irsakubedev.ipac.caltech.edu/applications/euclid), [Firefly](https://firefly-2006-charts-groups.irsakubedev.ipac.caltech.edu/firefly)  
- Do searches on any of the apps for regular scatter plots or spectra. Play around with the chart dialog. 
   - In the chart dialog, Spectra should load with the chart dialog already set to `Group by: Order`, and all other scatter plots should load with `Group by: None`
  - Apply `Group By` with any column. Select any trace and try changing trace style/symbol, etc. Set back `Group By` to `None`. 
  - While trying these, make a selection on a portion of the chart, and from the toolbar,  try both `select the enclosed points`,  and `Filter on the selected points`.  And then remove the selection/filter. Do this after you set `Group By` to a column and also after you set it back to `None`. 
- The above workflow led to several bugs I encountered and fixed, ensure you don't run into any more. 
  - After selection/filtering - also take a look at the `Trace Options` and `Chart Options` in the charts dialog. To make sure none of your selections changes (filtering would sometimes remove **_reverse_** from X/Y axes, even on ops - under `Chart Options`, as an example of a bug). 
- Also, for spectra, try turning on/off **Spectral Lines** from the chart dialog as well. And change traces while Spectral Lines are on. This should all work without any bugs.

### DIFF
--- a/src/firefly/js/charts/ChartsCntlr.js
+++ b/src/firefly/js/charts/ChartsCntlr.js
@@ -2,27 +2,27 @@
  * License information at https://github.com/Caltech-IPAC/firefly/blob/master/License.txt
  */
 
-import {cloneDeep, has, get, isArray, isEmpty, isString, isUndefined, omit, omitBy, set, range, unset} from 'lodash';
+import {cloneDeep, has, get, isArray, isEmpty, isString, isUndefined, omit, omitBy, set, range, unset, uniqueId} from 'lodash';
 import shallowequal from 'shallowequal';
 
 import {flux} from '../core/ReduxFlux.js';
 import {updateSet, updateObject, toBoolean} from '../util/WebUtil.js';
 import {Logger} from '../util/Logger.js';
-import {getTblById, getColumns, isFullyLoaded, COL_TYPE} from '../tables/TableUtil.js';
+import {getTblById, getColumn, getColumns, isFullyLoaded, COL_TYPE} from '../tables/TableUtil.js';
 import {dispatchAddActionWatcher} from '../core/MasterSaga.js';
 import * as TablesCntlr from '../tables/TablesCntlr.js';
 import {dispatchAddViewerItems, dispatchUpdateCustom, dispatchRemoveViewerItems,
     getMultiViewRoot, getViewer} from '../visualize/MultiViewCntlr.js';
 import {DEFAULT_PLOT2D_VIEWER_ID} from '../visualize/VisConst';
 import {applyDefaults, flattenAnnotations, formatColExpr, getPointIdx, getRowIdx, handleTableSourceConnections,
-        clearChartConn, newTraceFrom, setupTableWatcher, HIGHLIGHTED_PROPS, SELECTED_PROPS, TBL_SRC_PATTERN,
-        getSelIndexes, isSpectralOrder, combineAllTraceFrom} from './ChartUtil.js';
+        clearChartConn, newTraceFrom, setupTableWatcher, updateHighlighted, updateSelection, HIGHLIGHTED_PROPS, SELECTED_PROPS, TBL_SRC_PATTERN,
+        getSelIndexes, getTblIdFromChart, isGroupedChart, combineAllTraceFrom, getResetAxesChanges} from './ChartUtil.js';
 import {FilterInfo} from '../tables/FilterInfo.js';
 import {SelectInfo} from '../tables/SelectInfo.js';
 import {REINIT_APP} from '../core/CoreConst';
 import {getAppOptions} from '../core/AppDataCntlr.js';
 import {showInfoPopup} from '../ui/PopupUtil.jsx';
-import {makeHistogramParams, makeXYPlotParams, getDefaultChartProps} from './ChartUtil.js';
+import {makeHistogramParams, makeXYPlotParams, getDefaultChartProps, getChartProps} from './ChartUtil.js';
 import {adjustColorbars, hasFireflyColorbar} from './dataTypes/FireflyHeatmap.js';
 
 export const CHART_SPACE_PATH = 'charts';
@@ -130,13 +130,14 @@ export function dispatchChartTraceRemove(chartId, traceNum, dispatcher= flux.pro
  *  @param {Object} p - dispatch parameetrs
  *  @param {string} p.chartId - chart id
  *  @param {Object} p.changes - object with the path-string keys and values of the changed props
+ *  @param {boolean} [p.replaceTableSources=false] - replace and reload the complete table-source set
  *  @param {Function} [p.dispatcher=flux.process] - only for special dispatching uses such as remote
  *  @public
  *  @function dispatchChartUpdate
  *  @memberof firefly.action
  */
-export function dispatchChartUpdate({chartId, changes, dispatcher=flux.process}) {
-    dispatcher({type: CHART_UPDATE, payload: {chartId, changes}});
+export function dispatchChartUpdate({chartId, changes, replaceTableSources=false, dispatcher=flux.process}) {
+    dispatcher({type: CHART_UPDATE, payload: {chartId, changes, replaceTableSources}});
 }
 
 /**
@@ -312,19 +313,52 @@ function chartTraceRemove(action) {
         tablesources.forEach((ts,i) => {
             if (i>=traceNum && ts && ts._cancel) {
                 ts._cancel();
-                if (i !== traceNum) {
-                    const newIdx = i-1;
-                    ts._cancel = setupTableWatcher(chartId, ts, newIdx);
-                }
+                // mark the cancelled watcher as inactive so it is recreated with the
+                // correct trace index after the trace arrays are reduced.
+                ts._cancel = undefined;
             }
         });
         dispatch(action);
+
+        // Reconnect surviving table sources using their new trace indices;
+        // preserve their metadata, and assign new IDs to sources that moved
+        const {tablesources:remainingSources=[], mounted} = getChartData(chartId) || {};
+        if (mounted > 0 && remainingSources.length > 0) {
+            const reconnectedSources = remainingSources.map((tablesource, idx) => {
+                if (!tablesource?.tbl_id) return tablesource;
+                //give moved sources a new ID so requests from their old index are ignored.
+                const reconnectedSource = idx >= traceNum
+                    ? {...tablesource, _sourceId: uniqueId('chart-source-')}
+                    : tablesource;
+                return {...reconnectedSource, _cancel: setupTableWatcher(chartId, reconnectedSource, idx)};
+            });
+            dispatchChartUpdate({chartId, changes: {tablesources: reconnectedSources}});
+
+            const activeSource = reconnectedSources[getChartData(chartId)?.activeTrace];
+            const activeTable = activeSource?.tbl_id && getTblById(activeSource.tbl_id);
+            if (activeSource && activeTable) {
+                const currentChart = getChartData(chartId);
+                const activeTraceNow = currentChart.activeTrace;
+                const axisChanges = {};
+                const xMapping = activeSource.mappings?.x;
+                const currentXTitle = currentChart.layout?.xaxis?.title?.text;
+                //restore the shared X-axis label if trace removal left it empty (historgam removal did this, even on ops)
+                if (xMapping && !currentXTitle) {
+                    const xColumn = getColumn(activeTable, xMapping);
+                    const xUnit = get(xColumn, 'units', '');
+                    axisChanges['layout.xaxis.title.text'] = xMapping + (xUnit ? ` (${xUnit})` : '');
+                }
+                if (!isEmpty(axisChanges)) dispatchChartUpdate({chartId, changes: axisChanges});
+                updateHighlighted(chartId, activeTraceNow, activeTable.highlightedRow);
+                updateSelection(chartId, activeTable.selectInfo || {});
+            }
+        }
     };
 }
 
 function chartUpdate(action) {
     return (dispatch) => {
-        const {chartId, changes} = action.payload;
+        const {chartId, changes, replaceTableSources=false} = action.payload;
         // when selection  is undefined, selections layer must be removed
         if (changes.hasOwnProperty('selection') && !changes.selection) changes['layout.selections'] = [];
         // remove any table's mappings from changes because it will be applied by the connectors.
@@ -339,8 +373,8 @@ function chartUpdate(action) {
 
         // lazy table connection
         const {mounted} = getChartData(chartId);
-        if (mounted > 0) {
-            handleTableSourceConnections({chartId, data, fireflyData});
+        if (mounted > 0 && (!isEmpty(data) || !isEmpty(fireflyData))) {
+            handleTableSourceConnections({chartId, data, fireflyData, replaceTableSources});
         }
     };
 }
@@ -349,7 +383,7 @@ function chartHighlight(action) {
     return (dispatch) => {
         const {chartId, highlighted=0, chartTrigger=false} = action.payload;
         // TODO: activeTrace is not implemented.  switch to trace.. then highlight(?)
-        const {data, fireflyData, tablesources, activeTrace:activeDataTrace=0, selected} = getChartData(chartId);
+        const {data, fireflyData, activeTrace:activeDataTrace=0, selected} = getChartData(chartId);
 
         const {traceNum=activeDataTrace, traceName} = action.payload; // highlighted trace can be selected or highlighted trace of the data trace
 
@@ -359,14 +393,19 @@ function chartHighlight(action) {
 
         const ttype = get(data, [traceNum, 'type'], 'scatter');
 
-        if (!isEmpty(tablesources) && ttype.includes('scatter')) {
+        if (ttype.includes('scatter')) {
             // activeTrace is different from activeDataTrace if a selected point highlighted, for example
-            const {tbl_id} = tablesources[activeDataTrace] || {};
-            if (!tbl_id) return;
+            const tbl_id = getTblIdFromChart(chartId, activeDataTrace);
+            if (!getTblById(tbl_id)) return;
             // avoid updating chart twice
             // update only as a response to table highlight change
             if (!chartTrigger) {
                 const traceAnnotations = get(fireflyData, `${traceNum}.annotations`);
+                const hlTrace = newTraceFrom(data[traceNum], [highlighted], HIGHLIGHTED_PROPS, traceAnnotations);
+                dispatchChartUpdate({chartId, changes: {highlighted: hlTrace}});
+            } else if (data[traceNum]) {
+                //update the visual highlight immediately
+                const traceAnnotations = fireflyData?.[traceNum]?.annotations;
                 const hlTrace = newTraceFrom(data[traceNum], [highlighted], HIGHLIGHTED_PROPS, traceAnnotations);
                 dispatchChartUpdate({chartId, changes: {highlighted: hlTrace}});
             }
@@ -392,7 +431,7 @@ function chartHighlight(action) {
 function chartSelect(action) {
     return (dispatch) => {
         const {chartId, selIndexes=[], chartTrigger=false} = action.payload;
-        const {activeTrace=0, data, fireflyData, tablesources} = getChartData(chartId);
+        const {activeTrace=0, data, fireflyData} = getChartData(chartId);
 
         // when skipping hover, selecting chart points does not work
         // disable chart select in this case
@@ -401,18 +440,21 @@ function chartSelect(action) {
         // avoid updating chart twice
         // don't update before table select
         if (chartTrigger) {
-            if (!isEmpty(tablesources)) {
-                const {tbl_id} = tablesources[activeTrace] || {};
-                const {totalRows} = getTblById(tbl_id);
+            const tbl_id = getTblIdFromChart(chartId, activeTrace);
+            const tableModel =  getTblById(tbl_id);
+            if (tableModel) {
+                const {totalRows} = tableModel;
                 const selectInfoCls = SelectInfo.newInstance({rowCount: totalRows});
 
                 selIndexes.forEach(([ptIdx, traceIdx]) => selectInfoCls.setRowSelect(getRowIdx(data[traceIdx], ptIdx), true));
                 TablesCntlr.dispatchTableSelect(tbl_id, selectInfoCls.data);
+                // Keep the chart overlay in sync even while table-source watchers rebuild.
+                dispatchChartSelect({chartId, selIndexes});
             }
         } else {
             let selected = undefined;
             const hasSelected = !isEmpty(selIndexes);
-            if (isSpectralOrder(chartId)) {
+            if (isGroupedChart(chartId)) {
                 selected = combineAllTraceFrom(chartId, selIndexes, SELECTED_PROPS);
                 dispatchChartUpdate({chartId, changes: {hasSelected, selected, selection: undefined}});
             } else {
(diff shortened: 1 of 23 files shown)


kpuriIpac:
@robyww @jaladh-singhal 

**Updates**, PR updated as of 8/14, bug free from all my tesing:

- Lots of changes initially came from bugs discovered, but the code attempts to tighten the logic and improve existing code as well (as much as possible, before our big chats refactor in the near future)
- Feel free to ask me about any of the logic - I’ve tried describing some of it via in code comments 
- Test builds now updated
  - Test chart dialog as you please. Overplot all kinds of charts (or **_add_** different kinds of charts). Remove active traces (of the original chart or the one you overplotted)
     - Filter/select points where applicable 
     - `Group by` should only show for scatter plots, including spectrum plots
     -  Ungroup (`Group by: None`) / `Group By` and play around with chart toolbar as well


Please report any new bugs, style fix suggestions, etc.

src/firefly/js/charts/ui/options/FireflyHistogramOptions.jsx
```diff
@@ -47,15 +47,22 @@ export function FireflyHistogramOptions({activeTrace:pActiveTrace, tbl_id:ptbl_i
...
     const changes = {};
     changes[`fireflyData.${activeTrace}.dataType`] = 'fireflyHistogram';
     changes[`fireflyData.${activeTrace}.tbl_id`] = tbl_id;
+
+    //keep the trace and Firefly metadata linked to the same table during overploting
+    changes[`data.${activeTrace}.tbl_id`] = tbl_id;
     fields && Object.entries(fields).forEach( ([k,v]) => {
+        // do not let an empty histogram label erase the shared X-axis label while overplotting
+        if (k === 'layout.xaxis.title.text' &&
+            (!v || v === 'undefined') &&
```
robyww:
```
 (!v || v === 'undefined')
```
I don't understand this line. can it just be `!v`. Am I missing something?

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -126,11 +129,14 @@ export function SpectralLinesOptions({activeTrace, chartId}) {
     const hasSpectralFrame = useStoreConnector(() =>
         isKnownRefPos(getChartData(chartId)?.fireflyData?.[activeTrace]?.spectralFrame?.refPos),
         [chartId, activeTrace]);
-    const isEnabledField = useFieldValueOnly(ENABLED_KEY, false);
-    const isEnabled = hasSpectralFrame && isEnabledField;
+
+    // The field can be undefined for one render while the dialog mounts
+    // Use live form state when present, otherwise fall back to chart state
+    const isEnabledField = useStoreConnector(() => getFieldVal(groupKey, ENABLED_KEY), [groupKey]);
```
robyww: just use, you don't need `groupKey` it is picked up from the context.
```jsx
const isEnabledField- useFieldValueOnly(ENABLED_KEY)
```

src/firefly/js/charts/ui/options/SpectralLines.jsx
```diff
@@ -126,11 +129,14 @@ export function SpectralLinesOptions({activeTrace, chartId}) {
...
+    const isEnabled = hasSpectralFrame && (isEnabledField ?? initialEnabled);
 
     // TODO: combine different source tables to a client-side table
-    const sourceOptions = useFieldValueOnly(SOURCE_OPTIONS_KEY, initialSourceOptions);
+    const sourceOptions = useStoreConnector(() => getFieldVal(groupKey, SOURCE_OPTIONS_KEY), [groupKey]) ?? initialSourceOptions;
```
robyww: same here but also takes default value
```js
 const sourceOptions = useFieldValueOnly(SOURCE_OPTIONS_KEY, initialSourceOptions)
```

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -257,12 +259,38 @@ export const applyUnitConversion = ({fireflyData, data, inFields, axisType, newU
...
-    const sfOptionFields = Object.fromEntries(Object.entries(SFOptionFieldKeys(activeTrace))
-        .map(([subKey, fieldKey])=>[subKey, fields[fieldKey]]));
+    const currentSFOptionFields = getEffectiveSFOptionFields(fireflyData?.[activeTrace]);
+    const sfFieldKeys = SFOptionFieldKeys(activeTrace);
+    const sfOptionFields = {
+        value: fields[sfFieldKeys.value] ?? currentSFOptionFields.value,
+        redshift: fields[sfFieldKeys.redshift] ?? currentSFOptionFields.redshift,
+        userSpecified: fields[sfFieldKeys.userSpecified] ?? currentSFOptionFields.userSpecified
+    };
+    const xUnitChanged = xUnit !== undefined && xUnit !== fireflyData?.[activeTrace]?.xUnit;
```
robyww: use `isDefined(xUnit)` (in WebUtil.js) it is easier to read

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -359,16 +393,21 @@ function chartHighlight(action) {
 
         const ttype = get(data, [traceNum, 'type'], 'scatter');
 
-        if (!isEmpty(tablesources) && ttype.includes('scatter')) {
+        if (ttype.includes('scatter')) {
             // activeTrace is different from activeDataTrace if a selected point highlighted, for example
-            const {tbl_id} = tablesources[activeDataTrace] || {};
-            if (!tbl_id) return;
+            const tbl_id = getTblIdFromChart(chartId, activeDataTrace);
+            if (!tbl_id || !getTblById(tbl_id)) return;
```
robyww: I think you can just say `getTblById(tbl_id)` if not modify `getTblById` to have a 
```
if (!tbl_id) return;
```

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -359,16 +393,21 @@ function chartHighlight(action) {
...
             // avoid updating chart twice
             // update only as a response to table highlight change
             if (!chartTrigger) {
                 const traceAnnotations = get(fireflyData, `${traceNum}.annotations`);
                 const hlTrace = newTraceFrom(data[traceNum], [highlighted], HIGHLIGHTED_PROPS, traceAnnotations);
                 dispatchChartUpdate({chartId, changes: {highlighted: hlTrace}});
+            } else if (data[traceNum]) {
+                //update the visual highlight immediately
+                const traceAnnotations = get(fireflyData, `${traceNum}.annotations`);
```
robyww: can you take out the `get`?
```js
const traceAnnotations = fireflyData?.[`${traceNum}.annotations`];
 
```

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -427,10 +469,19 @@ function chartSelect(action) {
...
+        const {activeTrace=0, selection, data=[], fireflyData=[]} = getChartData(chartId) || {};
+        const activeTraceType = data?.[activeTrace]?.type;
+        const activeDataType = fireflyData?.[activeTrace]?.dataType;
+
+        //Histograms use bins and counts rather than table-backed X/Y points,so chart filtering does not apply
+        //ignore the action even if a transient selection state or toolbar entry is still present while the trace is loading
+        if (activeDataType === 'fireflyHistogram' || activeTraceType === 'bar') return;
+
+        const tbl_id = getTblIdFromChart(chartId, activeTrace);
+        const tableModel = tbl_id && getTblById(tbl_id);
```
robyww: another example of `getTblById` needing to return an undefined.

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -427,10 +469,19 @@ function chartSelect(action) {
...
+        const {mappings={}} = tableModel ? getChartProps(chartId, tbl_id, activeTrace) : {};
+        if (tableModel && selection && mappings.x && mappings.y) {
```
robyww: this is a huge `if` and even bigger after the PR. I think it would be better to
```js
 if (!tableModel || !election || !mappings.x || mappings.y) return;
```

src/firefly/js/charts/ui/PinnedChartContainer.jsx
```diff
@@ -180,8 +180,12 @@ export function ChartBadgeLabel({labelStr}) {
 
 function doPinChart({chartId, autoLayout=true, displayPinMessage=true }) {
 
-    const chartData = cloneDeep(omit(getChartData(chartId), ['_original', 'mounted']));
-    chartData?.tablesources?.forEach((ts) => Reflect.deleteProperty(ts, '_cancel'));
+    const chartData = cloneDeep(omit(getChartData(chartId), ['_original', 'mounted', 'selection']));
+    if (chartData?.layout) delete chartData.layout.selections;
```
robyww: if you are going to use `Reflect.deleteProperty` you should do it here as well.

src/firefly/js/charts/ui/CombineChart.jsx
```diff
@@ -74,7 +74,11 @@ export const CombineChart = ({chartIds, selectedChartId, showChartSelectionTable
         onCombineComplete?.(); // post-combination callback
     };
 
-    return (chartIds?.length > 1)
+    const unresolvedChartIds = chartIds?.[0] instanceof Promise;
```
robyww: I have read doing an `instanceof Promise` was not very reliable. I look it up on AI and it pointed out some issues.

Why don't you make a isPromsie in WebUtil.js that looks something like
```js
const isPromise = (p) => isFunction(p?.then);
```
line 245 is the other place it is use.
In our code base we only use `instanceof Promise` in this file.

robyww: I have read doing an `instanceof Promise` was not very reliable. I looked it up on AI and it pointed out some issues.

Why don't you make a `isPromsie` in `WebUtil.js` that looks something like

```js
const isPromise = (p) => isFunction(p?.then);
```
line 245 is the other place it is use.
In our code base we only use instanceof Promise in this file.

src/firefly/js/charts/dataTypes/FireflyHistogram.js
```diff
@@ -22,15 +22,20 @@ import Color from '../../util/Color.js';
...
 
     const options = {};
 
     // server call parameters
     const {fireflyData, layout} = getChartData(chartId) || {};
-    const histogramParams = get(fireflyData, `${traceNum}.options`);
+    //preserve histogram options when chart state updates are partial
+    const histogramParams = {
+        ...(traceTS.options || {}),
+        ...(fireflyData?.[traceNum]?.options || {})
```
robyww: using `...` is safe on undefined object so you don't need `|| {}`
example
```js
const a={};
const b=undefined;
const c= {...a,...b};
```

src/firefly/js/charts/dataTypes/FireflyGenericData.js
```diff
@@ -212,7 +230,7 @@ export function addScatterChanges({changes, chartId, traceNum, tablesource, tabl
 
     const colors = get(changes, [`data.${traceNum}.marker.color`]);
     let cTipLabel = isArray(colors) ? get(mappings, 'marker.color') : '';
```
robyww: take out the `get` in the above two lines.

robyww (approved):
This is looking really good!  I think it will be a very nice feature.

#### Code
- Several small things to clean up.
- Glad you added stuff to ChartUtil, I have a concern there is a lot of redundancy in the chart code.

#### Testing and UI.
- Test when very well. I did not run into any issues.
- I looks and works very nice! Easy to use!
- UI change? Maybe in the grouping case we do not need the trace selection on the toolbar. I don't really see what it does.

kpuriIpac:
@robyww Thanks! I will address your feedback. Meanwhile, I found that when opening the chart dialog, the `X` and `Y`  input lines would sometimes appear a split second late (moving in the UI, making for bad UX). Could be even more delayed for bigger tables/charts (since I changed how we rely on `colValStats`, and these inputs wait for `colValStats` to render). I fixed this in the latest commit. Feel free to do some basic testing again.

src/firefly/js/charts/ui/options/ScatterOptions.jsx
```diff
@@ -101,19 +106,29 @@ export function ScatterCommonOptions({activeTrace:pActiveTrace, tbl_id:ptbl_id,
...
+    }, [chartId, isGrouped]);
 
     return (
-        <CollapsibleItem header='Trace Options' componentKey='chart-scatter-options'
+        <CollapsibleItem header='Trace Options' componentKey={TRACE_OPTIONS_COMPONENT_KEY} isOpen={isGrouped}
                                    {...props /* to allow AccordionGroup styling props inserted by joy UI, propagate to Accordion */}>
             <Stack spacing={2} sx={{'.MuiFormLabel-root': {width: '6rem'}}}>
+                {isGrouped && (
+                    <Stack spacing={1}>
+                        <Typography level='body-md'>You can modify the selected trace:</Typography>
```
jaladh-singhal:
```suggestion
                        <Typography level='body-sm'>You can modify the selected trace:</Typography>
```
to stay consistent with the style of other such text preceeding input fields in scatter options.

src/firefly/js/charts/ui/options/SpectralLines.jsx
jaladh-singhal: I updated these in my spectral lines (first pass) PR before merging - @kpuriIpac have you verified if this file is up-to-date with dev?

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -92,6 +92,7 @@ export function SpectrumOptions ({activeTrace:pActiveTrace, tbl_id:ptbl_id, char
                         {yMin && <Ymin label = 'Flux axis lower limit column:'/>}
                         <Yunit/>
                     </GroupedStack>
+                    <GroupBy/>
```
jaladh-singhal: Can you pass appropriate props to GroupBy here to make the "label" `fontSize: 'md'` (instead of default 'sm')?

Right now in spectrum options case, GroupBy is hard to distinguish from Y-axis sub-fields (despite de-indentation). Matching its fontSize with X-axis and Y-axis labels makes it immediately stand out on the same level as them (from devtools css hacking):

Image

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -145,13 +146,14 @@ export function spectrumReducer({chartId, activeTrace, tbl_id}) {
 
 /* ------------------------------ Redshift correction handling ------------------------------ */
 const getRedshiftCorrectedExpr = ({cname, spectralFrame, sfOption, redshift=undefined}) => {
-    const {refPos, redshift: customRedshift} = spectralFrame;
-    const multiplyBy = refPos.toUpperCase() === REF_POS.CUSTOM ? ` * (1 + ${customRedshift ?? '0'})` : '';
+    const {refPos, redshift: customRedshift} = spectralFrame || {};
+    const normalizedRefPos = refPos?.toUpperCase?.() || REF_POS.TOPOCENTER;
```
jaladh-singhal: normalisation really doesn't matter since multiplyBy is empty for anything not REF_POS.CUSTOM

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -145,13 +146,14 @@ export function spectrumReducer({chartId, activeTrace, tbl_id}) {
...
+    const multiplyBy = normalizedRefPos === REF_POS.CUSTOM ? ` * (1 + ${customRedshift ?? '0'})` : '';
     const divideBy = sfOption === 'rest' && redshift ? ` / (1 + ${redshift})` : '';
-    let expr = `"%s"${multiplyBy}${divideBy}`;
+    let expr = `${quoteNonAlphanumeric(cname)}${multiplyBy}${divideBy}`;
```
jaladh-singhal, line 154: Is there a reason for this - just curious?

It's working fine in my testing.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -365,24 +410,24 @@ function ReadOnlyField({value, label, ...props}) {
...
             const allProps = {label: 'Spectral frame:', ...props};
             const sfRefPos = fireflyData[activeTrace].spectralFrame.refPos.toUpperCase();
             return isKnownRefPos(sfRefPos) //only show options when TOPOCENTER or CUSTOM
                 ? <SpectralFrameOptions groupKey={groupKey} activeTrace={activeTrace} refPos={sfRefPos} fireflyData={fireflyData} {...allProps}/>
                 : <ReadOnlyField value={sfRefPos} {...allProps}/>;
-        }, deps),
+        }, [activeTrace, fireflyData, groupKey]),
         SpectralLines: useCallback((props={}) =>
-            <SpectralLinesOptions activeTrace={activeTrace} chartId={chartId} {...props}/>, deps),
+            <SpectralLinesOptions activeTrace={activeTrace} chartId={chartId} {...props}/>, [activeTrace, chartId]),
```
jaladh-singhal, line 431: thanks for adding these deps, it was long due and I remember I didn't do it because it required other changes in code above which I think you covered.

src/firefly/js/charts/dataTypes/FireflySpectrum.js
```diff
@@ -61,10 +63,17 @@ async function fetchData(chartId, traceNum, tablesource) {
         // extra changes based on trace type
         addOtherChanges({changes, chartId, traceNum, tablesource, tableModel: originalTableModel});
 
+        if (resetAxes) {
+            const {layout} = getChartData(chartId) ?? {};
+            changes['layout.xaxis.autorange'] = getFilterResetAutorange(layout?.xaxis);
+            changes['layout.xaxis.range'] = undefined;
+            changes['layout.yaxis.autorange'] = getFilterResetAutorange(layout?.yaxis);
+            changes['layout.yaxis.range'] = undefined;
+        }
```
jaladh-singhal, line 68: I see this repeated across all chart types - can be moved to a function (but I do see other untouched code is a repeat too across these files so also ok to leave it for next refactor)

src/firefly/js/charts/dataTypes/FireflySpectrum.js
```diff
@@ -156,7 +166,7 @@ export function getSpectrumProps(tbl_id, spectrumDM) {
 
     // get default spectral frame and labels, needed to initialize spectrum
     // (in future, move redshift processing functions & defaults from SpectrumOptions to a separate file where they can be exported to avoid redundancy)
-    const refPos = spectralFrame.refPos.toUpperCase();
+    const refPos = spectralFrame?.refPos?.toUpperCase?.() || REF_POS.TOPOCENTER; //add a fallback to TOPOCENTER if refPos is undefined
```
jaladh-singhal, line 165: this was a bug I remember - thanks for fixing it!

src/firefly/js/charts/ui/options/HeatmapOptions.jsx
```diff
@@ -154,14 +157,31 @@ export function submitChangesHeatmap({chartId, activeTrace, fields, tbl_id, rend
...
 
+    //store heatmap axes in both formats required by table sources and heatmap fetches
+    ['x', 'y'].forEach((axis) => {
+        const fieldValue = fields?.[`_tables.data.${activeTrace}.${axis}`];
+
+        if (fieldValue) {
+            const expression = String(fieldValue).replace(/^tables::/, '');
+
+            changes[`data.${activeTrace}.${axis}`] = `tables::${expression}`;
+            changes[`fireflyData.${activeTrace}.options.${axis === 'x' ? 'xColOrExpr' : 'yColOrExpr'}`] =
```
jaladh-singhal:
```suggestion
            changes[`fireflyData.${activeTrace}.options.${axis}ColOrExpr`] =
```

src/firefly/js/charts/ui/options/ScatterOptions.jsx
```diff
@@ -360,4 +411,23 @@ export const useScatterInputs = ({activeTrace:pActiveTrace, tbl_id:ptbl_id, char
     };
 };
 
+/**
+ * Keep chart option fields in sync while table column statistics are loaded.
+ * ColumnOrExpression cannot render its input until these statistics are ready.
+ * @param tbl_id table identifier used to load and retrieve column statistics
+ */
+function useTableColValStats(tbl_id) {
```
jaladh-singhal: This seems independent enough from charts stuff. If so, you can:
- update docstring
- move it to TableStatsCntlr.js or TableUtils.js

src/firefly/js/charts/ui/options/ScatterOptions.jsx
```diff
@@ -233,6 +248,19 @@ export function submitChangesScatter({chartId, activeTrace, fields, tbl_id, rend
...
+    const groupByKey = groupByFieldKey(activeTrace);
+    const groupByColumn = fields[groupByKey] || '';
+    const currentGroupBy = getGroupByColumn(chartId) || '';
+    const groupByChanged = groupByColumn !== currentGroupBy;
+    if (chartId && has(fields, groupByKey) && groupByChanged && isGroupByAllowed(chartId, activeTrace)) {
+        const groupedFields = omit(changes, groupByKey);
+        const evalChanges = evalChangesFromFields(chartId, tbl_id, groupedFields);
+        const groupChanges = makeScatterGroupByChanges({chartId, tbl_id, activeTrace, groupByColumn, changes: evalChanges});
+        dispatchChartUpdate({chartId, changes: groupChanges, replaceTableSources: true});
+        dispatchComponentStateChange(TRACE_OPTIONS_COMPONENT_KEY, {isOpen: Boolean(groupByColumn)});
```
jaladh-singhal: Are you sure that you need to fire these dispatch functions here? I know they are guarded by your `if` block but I don't see pattern of state changing function within a function that submits payload for chart state changes. It might be worth:
- putting them in a useEffect inside the <SpectrumOptions/> instead of here OR
- remove dispatch functions but batch their changes with other changes payload and let submitChanges update the chart state

You know the flow better to make right call.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
```diff
@@ -337,6 +371,17 @@ export function submitChangesSpectrum({chartId, activeTrace, fields, tbl_id, ren
     submitChangesScatter({chartId, activeTrace, fields, tbl_id, renderTreeId});
 }
 
+function getEffectiveSFOptionFields(trace={}) {
+    const spectralFrame = trace.spectralFrame || {};
+    const spectralFrameOption = trace.spectralFrameOption || {};
+    const refPos = spectralFrame.refPos?.toUpperCase?.();
+    return {
+        value: spectralFrameOption.value ?? (refPos === REF_POS.TOPOCENTER ? 'observed' : 'rest'),
```
jaladh-singhal, line 380: You should double check it from code but IIRC, `spectralFrameOption.value` becomes same as refPos when neither TOPOCENTER nor CUSTOM. 

There were 3 cases I developed redshift correction for, so test BARYCENTER in this: [z-correction-data.zip](https://github.com/user-attachments/files/31284748/z-correction-data.zip)

src/firefly/js/charts/ui/PlotlyWrapper.jsx
```diff
@@ -344,6 +346,9 @@ export class PlotlyWrapper extends Component {
         const {layout} = getChartData(chartId);
         if (layout && !this.props.thumbnail) {
             Object.entries(changes).forEach( ([k, v]) => {
+                //ignore Plotly's transient selection-box layout edits, e.g. selections[0].yref.
+                //the supported selection state is captured from plotly_selected in PlotlyChartArea
+                if (k.startsWith('selections[')) return;
```
jaladh-singhal: If the goal is to prevent syncLayout to do anything when `k.startsWith('selections[')`, shouldn't this be moved outside this loop in a `changes.find()` call? Because otherwise layout will still be changed for keys in `changes` object that appear before `selections` key. 

And can an object key really contain `[`??

src/firefly/js/charts/ChartsCntlr.js
```diff
@@ -462,27 +513,34 @@ function chartFilterSelection(action) {
...
             TablesCntlr.dispatchTableFilter(newRequest);
-            dispatchChartUpdate({chartId, changes:{selection: undefined}});
+            const {layout} = getChartData(chartId) ?? {};
+            dispatchChartUpdate({chartId, changes:{
+                selection: undefined,
+                'layout.xaxis.autorange': getFilterResetAutorange(layout?.xaxis),
+                'layout.xaxis.range': undefined,
+                'layout.yaxis.autorange': getFilterResetAutorange(layout?.yaxis),
+                'layout.yaxis.range': undefined
+            }});
```
jaladh-singhal: I think I saw this in a couple more files. If so, can move it to a util function returning this change payload.

jaladh-singhal (approved):
@kpuriIpac this is a lot of effort. I tested the firefly/irsaviewer - it's working really well.

There are two visual style suggestions, other than that most of my comments are questions or cautionary. I should note that I only skimmed histogram, heatmap code since i'm not super familiar with it.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
kpuriIpac: good point! i'll make this work

src/firefly/js/charts/ui/options/SpectralLines.jsx
kpuriIpac: Yes, rebased against dev. Is it not supposed to be : 

`const isEnabledField = useStoreConnector(() => getFieldVal(groupKey, ENABLED_KEY), [groupKey]);`

from your spectral lines work? (or did you mean the current unmerged PR you have - actually I don't see a PR, so i think your local spectral lines work?)

src/firefly/js/charts/ui/options/FireflyHistogramOptions.jsx
kpuriIpac: I think this is left over from when I trying to debug something with AI. Good catch, thanks! Found another check in `ChartsCntlr` too. Fixed in both places.

src/firefly/js/charts/ChartsCntlr.js
kpuriIpac: Yes `getTblById` should return `undefined` when `tbl_id` is missing since it already uses `get`. So I can safeyl change this to `getTblById` check only.

src/firefly/js/charts/ChartsCntlr.js
kpuriIpac: I think the way `getTblById` is currently defined (using `get`), we should get an undefined for tbl_id being undefined. so this can also simplify to: 

`const tableModel = tbl_id && getTblById(tbl_id);`

but if you want me to add an explicit check in `getTblById`, i can do that.

src/firefly/js/charts/ChartsCntlr.js
kpuriIpac: good point. I did: 


```
if (!tableModel || !selection) return;

const {mappings = {}} = getChartProps(chartId, tbl_id, activeTrace);
if (!mappings.x || !mappings.y) return;
```

and then removed the if condition entirely from the big code block.

src/firefly/js/charts/ui/options/SpectralLines.jsx
kpuriIpac: Sorry @jaladh-singhal I see your comment below properly now. You're saying it should have picked up a change you made from your first PR, right? Do you not see that change? that'd be odd if so, since I did rebase against `dev` before starting and have been doing so regularly..

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
kpuriIpac: got it, makes sense. removed the fallback. I think we may still need to check `refPos?.toUpperCase?.()` when comparing to `REF_POS.CUSTOM` so i made it: 

```
const multiplyBy = refPos?.toUpperCase?.() === REF_POS.CUSTOM
        ? ` * (1 + ${customRedshift ?? '0'})`
        : '';
```

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
kpuriIpac: Yes, I think this is where I was seeing a bug with the Spectral axis column names showing up sometimes as, for example, `"lamda"` instead of just `lamda` after applying some `Group by` changes / canceling `Group by`. 

this (`quoteNonAlphanumeric(cname)`) should still show quotes when the column name requires them

src/firefly/js/charts/ui/PlotlyWrapper.jsx
kpuriIpac: Good point. Moved the check outside the loop. And yes, object keys I think can contain `[` and as in Plotly's case, it does have keys like `selections[0].yref`.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
kpuriIpac: @jaladh-singhal I am unsure about the correct change to be made here, but I did test `BARYCENTER` in the file and it did set a fixed spectral frame to `BARYCENTER`. Would that be the expected behavior then? Can you test it and let me know if it looks correct?

kpuriIpac:
@jaladh-singhal @robyww I addressed your PR feedback, and updated the test builds. Can you please take a quick look at the code changes, and repeat some basic testing on the test builds above? Let me know if anything looks problematic (in code or UI).
   - Roughly, `first round of PR feedback` commit addresses @robyww changes and `second round of PR feedback` commit addresses @jaladh-singhal 's. 

The final commit is a small change that changes our enum-like column detection logic to look for more than 1 distinct value in a column to designate this column as enum-like. 
   - The result is that for most searches, you should see fewer columns in the `Group By` dropdown now. (Spherex `Spectral Image Search` -> `Active Chart`-> `Group By` is a good test for this, previously this would show several options, most of them meaningless. Now it shows only 3-4 options).

robyww:
It is a little to hard to evaluate the code now without completely starting over. I am fine with you merging.

src/firefly/js/charts/ui/options/SpectrumOptions.jsx
jaladh-singhal: I tested it - it's working fine for now!

src/firefly/js/charts/ui/options/SpectralLines.jsx
jaladh-singhal: Nvm, I think I got confused too. I'll take care of it when rebasing my upcoming PR branch over your merged changes

jaladh-singhal:
Thanks for making the changes @kpuriIpac. I think you covered most of them. I did a quick round of testing, it's still working fine - good to merge!

PR #1994, opened 2026-08-12
Firefly-2048: fix bug in reading tables with TZERO

robyww:
### [Firefly-2048](https://jira.ipac.caltech.edu/browse/FIREFLY-2084): fix bug in reading tables with TZERO

- nom-tam-fits bug: hduTable.getColumnMeta(icol, "TZERO") will return a null. The problem seems to be with the nom-tam-fits function
- We do not handle longs correctly when we need to scale them. We are converting the long to a double
- Fixed the two bugs, implement a work around convenience function for getColumnMeta and do the right thing with longs.
- A little clean up on HiPSRetrieve.java

#### Testing
- build the rubin portal locally with this branch
- choose `DP2 & DP2 Images`
- under `Observation Type and Source` -> `Collection` -> LSST.DP2
- target: 53.25, -28.089 EQ_J2000
- `Search`
- for the first image use the `File Contents` pull down to go to `HDU #9` and show the table.
- The first 4 columns should be  `int`, `int`, `long`, `int`
- compare with `data-int` to see the wrong data

### DIFF
--- a/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileUpload.java
+++ b/src/firefly/java/edu/caltech/ipac/firefly/server/servlets/AnyFileUpload.java
@@ -256,7 +256,9 @@ public class AnyFileUpload extends BaseHttpServlet {
                 File dir= getSessUploadDir(sp.convertToServerRequest());
                 statusFileInfo = LockingRetrieve.downloadWithCacheMsg(fromUrl, dir);
             }
-            if (isUrlFail(statusFileInfo)) throw new Exception(codeFailMsg(statusFileInfo));
+            if (isUrlFail(statusFileInfo))  {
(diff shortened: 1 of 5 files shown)


timj:
If we think this is also a nom-tam-fits bug, can we also file a ticket with them?

robyww:
I am going to make a nom-tam-fits issue soon.

robyww:
nom-tam-fits issue: https://github.com/nom-tam-fits/nom-tam-fits/issues/882

kpuriIpac (approved):
UI Test looks good, as compared to changes on data-int. ~~My only concern is that from the reading I did on FITS and `TSCAL`/`TZERO`, it seems `TSCAL` and `TZERO` can be floating points. Do we need to handle that case too?~~ Never mind, I see a check for this later. All good.

PR #1995, opened 2026-08-18
Firefly-2024: New MOC features in DCE and code clean up

robyww:
### [Firefly-2024](https://jira.ipac.caltech.edu/browse/FIREFLY-2024): new moc features in DCE and code clean up

- refactor and some clean up of code cleanup, created new file `CisxSerDescUtils.js` and moved some functions to it.
- support:
```xml
  <PARAM name="moc_style" datatype="char" arraysize="*" value="outline"/>
  <PARAM name="moc_short_description" datatype="char" arraysize="*" value="herschel_pep"/>
```

#### Testing

- https://firefly-2024-moc-outline.irsakubedev.ipac.caltech.edu/irsaviewer
- drop test file into upload, and load, you will see the following
  - color red (just to help you see it is different)
  - moc is outlined
  - the legend will say `dustpedia`
- compare to the normal DCE load of this file: https://firefly-2024-moc-outline.irsakubedev.ipac.caltech.edu/irsaviewer/dce
  - bring up dce
  - choose dustpedia about nine rows down or you can search for it.
  - notice that it come up in auto mode and the legend is different

#### Test File
[dustpedia.xml](https://github.com/user-attachments/files/31188003/dustpedia.xml)

### DIFF
--- a/src/firefly/js/ui/dynamic/CisxSerDescUtil.js
+++ b/src/firefly/js/ui/dynamic/CisxSerDescUtil.js
@@ -1,0 +1,134 @@
+import {isValidFullUrl} from '../../util/WebUtil';
+import CoordinateSys from '../../visualize/CoordSys';
+import {makeWorldPt} from '../../visualize/Point';
+
+
+/*
+ * -------------------------------------
+ * Supports the processing of IRSA extension of service descriptor: <RESOURCE type="meta" utype="CISX:adhoc:service"
+ * -------------------------------------
+ */
+
+
+
+/**
+ * Pull the service descriptor (xml) moc related keys out and convert them to a form that can be later processed.
+ * The service descriptor might define multiple MOCs
+ * @param cisxUI
+ * @return {*} and object the defines multiple mocs that will be process by other parts of the application
+ */
+
+function findAndTranslateMocParameters(cisxUI) {
+    const translateMocXml = {
+        moc_color: 'mocColor',
+        moc_style: 'mocStyle',
+        moc_short_description: 'shortTitle',
+    };
+    const itemsList = Object.keys(translateMocXml);
+
+    const allMocsObj= cisxUI.filter(({name}) => {
+            if (!name) return false;
+            const key = name.toLowerCase();
+            return key === 'moc' || key.match(/moc\d+$/i) || itemsList.some((s) => key.startsWith(s));
+        })
+        .reduce((allMocs, {name, value, desc}) => {
+            const key = itemsList.find((k) => name.startsWith(k));
+            const id= name.substring(key?.length ?? 'moc'.length);
+            if (!allMocs[id]) allMocs[id] = {};
+            if (key) {
+                allMocs[id][translateMocXml[key]] = value;
+            } else {
+                allMocs[id].mocUrl= value;
+                allMocs[id].title= desc;
+            }
+            return allMocs;
+        }, {});
+    return Object.values(allMocsObj);
+}
+
+/**
+ *
+ * @param {Object} cisxUI
+ * @param {number} defaultMaxMOCFetchDepth
+ * @return {SearchAreaInfo}
+ */
+export function makeSearchAreaInfo(cisxUI, defaultMaxMOCFetchDepth) {
+    if (!cisxUI) return;
+    const tmpObj = cisxUI.reduce((obj, {name, value, UCD}) => {
+        switch (name) {
+            case 'hips_initial_fov':
+                obj[name] = Number(value);
+                break;
+            case 'hips_initial_dec':
+            case 'hips_initial_ra':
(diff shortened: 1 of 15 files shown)


src/firefly/js/visualize/HiPSMocUtil.js
```diff
@@ -1,4 +1,6 @@
 import {get, set, isEmpty, flatten, isArray} from 'lodash';
+import React from 'react';
```
jaladh-singhal: I don't see the use of React import in this file. Also it's a `.js` not `.jsx` so either we should remove it or change file extension?

src/firefly/js/voAnalyzer/VoDataLinkServDef.js
```diff
@@ -326,3 +325,32 @@ export function findWorldPtInServiceDef(serDef,sourceRow) {
...
+        const isDatalinkSerDesc = isDataLinkServiceDesc(sd);
+        if (isDatalinkSerDesc) {
+            const productUrl = {
+                accessURL: sd?.accessURL || '',
+                inputParams: {}
+            };
+
+            if (sd?.serDefParams) {
+                for (const param of sd.serDefParams) {
+                    productUrl.inputParams[param?.name] = {value: param.value, ref: param.ref};
```
jaladh-singhal: You may wanna check `param?.name` before doing assignment otherwise there will be a "undefined" key  in `inputParams`. I've noticed this issue in some of our other code as well.

src/firefly/js/ui/dynamic/CisxSerDescUtil.js
```diff
@@ -0,0 +1,150 @@
...
+    if (!inExample) return {targetPanelExampleRow1: undefined, targetPanelExampleRow2: undefined};
+    const examples = inExample.split('|');
+    if (examples?.length > 1) {
+        const cnt = examples.length;
+        return {
+            targetPanelExampleRow1: examples.slice(0, Math.trunc(cnt / 2)),
+            targetPanelExampleRow2: examples.slice(Math.trunc(cnt / 2))
+        };
+    } else {
+        return {targetPanelExampleRow1: [inExample], targetPanelExampleRow2: []};
```
jaladh-singhal, line 100: This probably out of scope since it will change data structure everywhere else but instead of breaking examples to two rows we can let CSS handle that during rendering based on line wrapping.

src/firefly/js/ui/dynamic/CisxSerDescUtil.js
```diff
@@ -0,0 +1,150 @@
...
+        .map( (k) => {
+            const cnt= k.substring(3,k.length);
+            return {
+                mocUrl : searchAreaInfo[k],
+                title : searchAreaInfo['mocTitle'+cnt] ?? 'MOC'+cnt,
+                [SHORT_TITLE]: searchAreaInfo[SHORT_TITLE+cnt],
+                [MOC_COLOR]: searchAreaInfo[MOC_COLOR+cnt],
+                maxFetchDepth: searchAreaInfo['maxFetchDepth'+cnt],
+                [MOC_STYLE]: searchAreaInfo[MOC_STYLE+cnt],
+            };
```
jaladh-singhal: I think this code can be improved:
- You can omit second argument in [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) - it will default to string.length instead of you specifying it again and again
- There seems to be some string parsing and reverse parsing association between these two functions for object keys e.g. `cnt= k.substring(3,k.length); searchAreaInfo['mocTitle'+cnt]` and `obj['mocTitle' + name.substring(3, name.length)] = desc;`. It might be worth abstracting those associations out to helper functions so that they don't go out of sync and are more readable. 

(I recommend feeding it to AI, it's usually good at this kind of things)

jaladh-singhal (approved):
I tested the builds, working as described. @lrebull can test in depth when she's back.

I left some minor code improvement comments, otherwise good to go.

src/firefly/js/voAnalyzer/VoDataLinkServDef.js
robyww: I will put an if around it.

src/firefly/js/ui/dynamic/CisxSerDescUtil.js
robyww: That could be really complex and often they want more control. It is something we could look at in the future.

src/firefly/js/ui/dynamic/CisxSerDescUtil.js
robyww: This processing was happening at two different places in the code. I realize it could be combined and happen very early. 
I will look at cleaning it up some more.

PR #1996, opened 2026-08-19
Firefly-2016: fixed FieldGroupConnector to properly initialize in keepState case

robyww:
### [Firefly-2016:](https://jira.ipac.caltech.edu/browse/FIREFLY-2016) fixed FieldGroupConnector to properly initialize in keepState case
 
- also, unrelated: added outputFormats to the tap capabilities retrieve

#### Testing
- https://firefly-2016-fg-connector.irsakubedev.ipac.caltech.edu/applications/spherex?api
- use SPHEREx url api page, try first to examples
- The `VosiCapabilityRetrieve.java` does not need to be tested now. It is a small change that is a required for another ticket.


jaladh-singhal (approved):
Code looks good.

Had to test it locally as you said, size params are working for both spherex and TAP url api

PR #1997, opened 2026-08-19
Firefly-2079: Fixed: Statistics dialog fails to come up when selected region contains only NaNs

robyww:
### [Firefly-2079](https://jira.ipac.caltech.edu/browse/FIREFLY-2079): Fixed: Statistics dialog fails to come up when selected region contains only NaNs

 - including a little code cleanup

#### Testing
- https://firefly-2079-stats.irsakubedev.ipac.caltech.edu/firefly
   - Go to upload tab
   - choose upload and enter the url below
      - `http://web.ipac.caltech.edu/staff/roby/many-images/_data_SOFIA_FIFI-LS_L4_2016-07-05_FI_F316_p3993_g10_F0316_FI_IFS_04015210_BLU_WXY_600284-700065.fits`
    - take the defaults and load the image
    - on the imag,e NaN pixels show up redish, the first image will be all NaNs
    - goto plane 3 which will have a mix of NaNs and non-NaNs
    - using the selection tool choose an area with a mix of both, choose the statistics button Screenshot 2026-08-19 at 4 19 16 PM
    - you should see stats
    - now choose an area with only NaNs and click the the statistics button again. It should show an error.

### DIFF
--- a/src/firefly/js/visualize/ui/CtxToolbarFunctions.js
+++ b/src/firefly/js/visualize/ui/CtxToolbarFunctions.js
@@ -121,6 +121,7 @@ function tabulateStatics(wpResult, cc) {
     const bandData = {};
 
     function getOneStatSet(b) {
+        if (!b.MEAN || !b.STDEV || !b.INTEGRATED_FLUX) return;
 
         const tblData = {};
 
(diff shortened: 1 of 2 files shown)


aventura121 (approved):
Hi Trey, looks good to me. One non-blocking comment pertaining to something I'm not sure is an actual use case in practice.

Testing: 

- I performed the specified steps on the test build and validated the expected dialogs displayed.

Comment:

Since the code allows more than one band being submitted, then it seems one band can be valid but not others. If so, it might be more accurate to have the string say something like: "Could not compute statistics: one or more bands have no valid (non-NaN) pixels"

robyww:
I that is a good message. The normal case is one band, so I can check and have two messages depending on normal (one band) or the 3-color (multi-band) case.

PR #2001, opened 2026-08-26
FIREFLY-2057: disable contextual font ligatures

aventura121:
[Firefly-2057](https://jira.ipac.caltech.edu/browse/FIREFLY-2057): Minor display bug in TAP list of 2MASS catalogs

Description:

The font styling we're using is including contextual ligatures (e.g. it will read 2x3 and redraw it as arithmetic). We disable this on the root element and also list the form inputs separately, because the browser can undo it on those (i.e. reapply the treatment).

Testing:

- Build: https://firefly-2057-disable-font-contextual-alternates.irsakubedev.ipac.caltech.edu/firefly

1. Navigate to the Tables (TAP) view.
2. Select Project: 2mass.
3. Expand the Tables selection menu. 
4. We should see the x as a plain character in the string "pt_src_6x2"

Before:

before

After:

after

### DIFF
--- a/src/firefly/html/css/global.css
+++ b/src/firefly/html/css/global.css
@@ -22,6 +22,14 @@
 @def toolbarGroupColor    rgba(0,0,0,.51);
 */
 
+/*
+ * Disable contextual ligatures. The form controls are listed to account for the browser's font shorthand reset.
+ */
+:root,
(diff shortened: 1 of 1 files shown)


kpuriIpac (approved):
Looks good to me. Thank you for fixing this!

PR #2003, opened 2026-08-28
FIREFLY-2094: Support pre-assigned colors and order for grouped charts 

kpuriIpac:
**Ticket**: [FIREFLY-2094](https://jira.ipac.caltech.edu/browse/FIREFLY-2094)
- Added the ability to set colors/order for grouped traces via `dataServiceOptions`
- The changes in `Firefly.js` are strictly for testing purposes, added hardcoded colors and orders in the `enumConfig` in Firefly.js's `dataServiceOptions` (for the 2 test files). This will be removed before this PR is merged 
    - I also locally tested this in `euclid.js` and uploaded files on `Euclid` to get similar results - so this should work similarly for Rubin as well.
- The reason for 2 different test files below, and similar code changes in 2 different areas of code as well are because: 
   - we need to handle the path where a spectra is grouped by a column by default when you load it
   - and also the 2nd case when a chart has the `Group By` functionality manually turned on after the chart loads with `Group By` initially set to `None` (this is in `makeScatterGroupByChanges` function)

**Testing**: 
- **Firefly**: https://firefly-2094-preassigned-group-colors.irsakubedev.ipac.caltech.edu/firefly
- Use the 2 files in the ticket for testing
  - when uploading both files, select the option to `Attempt to interpret tables as spectra`
  - The `DO_Tau_SH_SPECTRA.tbl` file will result in the spectra being automatically grouped by `order` column
     - the traces should follow the `'u', 'g', 'r', 'i', 'z', 'y'`  specified 
     - and the same for colors specified as well.
  - For the `table_ForcedSource-time-series.vot` file, enable `Group By` by `band` column, also change Y-axis to `psfDiffFlux` for a better view
     - the traces here should also follow the order specified in Firefly.js file .. `'1819', '15', '12', '1112'`,  and so on...
     - again, colors will also follow the list of specified colors.
- feel free to use any other chart/spectra files to test this functionality (and for any columns not hardcoded in `firefly.js`, it should all just follow default order/color mapping) 
- Also, load a chart (either via TAP search or file upload) and test the `Group By` functionality from the previous PR as well, to make sure it works as expected.

### DIFF
--- a/src/firefly/js/ui/tap/DataServicesOptions.js
+++ b/src/firefly/js/ui/tap/DataServicesOptions.js
@@ -2,7 +2,6 @@ import {isObject} from 'lodash';
 import {getAppOptions} from '../../core/AppDataCntlr';
 import {MetaConst} from '../../data/MetaConst';
 import {getMetaEntry, getObjectMetaEntry} from '../../tables/TableUtil';
-import {getServiceMetaOptions} from './SiaUtil';
 
 const dsOps= () => getAppOptions().dataServiceOptions ?? {};
 
@@ -35,6 +34,18 @@ export function getDataServiceOptionByTable(key, tableOrId, defVal=undefined) {
     const entry= getObjectMetaEntry(tableOrId, MetaConst.DATA_SERVICE_OPTIONS)?.[key];
     if (entry) return entry;
     return getDataServiceOption(key, getMetaEntry(tableOrId,MetaConst.DATA_SERVICE_ID), defVal);
+}
+
+/**
+ * Get the `enumConfig` (preassigned order/color palette for an enum-like column, e.g. a photometric
+ * band) for a table, if one is configured via dataServiceOptions and applies to the given column.
+ * @param {TableModel|String} tableOrId - the table model or table id
+ * @param {String} columnName - the column being grouped/enumerated
+ * @return {{columnNames: String[], order: String[], palette: Object}|undefined}
+ */
+export function getEnumConfigForColumn(tableOrId, columnName) {
+    const enumConfig = getDataServiceOptionByTable('enumConfig', tableOrId);
+    return enumConfig?.columnNames?.includes(columnName) ? enumConfig : undefined;
 }
 
 export function getDataServiceOptionsFallback(dataServiceId, hostname) {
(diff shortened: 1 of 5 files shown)


jaladh-singhal (approved):
Code looks good - remember to clean up firefly.js before merging.

Tested the app - working well. I notice a weirdness though, not a bug:

When I set Group by to "order" in forced photometry table and apply, the `ugrizy` order is preserved in traces dropdown on chat but not exactly in legend - first trace `u` is rendered last in the legend.

Image


As soon as I choose a different trace (see "g" selected), it also jumps out of the order come to the last:

Image

I remember this is by design in our chart code to push the selected/active trace to last (I forgot the reasoning what it helps with). A chart initializes active trace as per the highlighted row in its linked table so whatever is the value of group-by column in the first row of table, renders last in the legend breaking the order. Something you can discuss with Gregory or scientists.

kpuriIpac:
> When I set Group by to "order" in forced photometry table and apply, the `ugrizy` order is preserved in traces dropdown on chat but not exactly in legend - first trace `u` is rendered last in the legend.
> 

Yes @jaladh-singhal this seems intentional as we do this in `PlotlyChartArea` (pushing the active trace down). So maybe there's a good reason for that. Worth discussing with scientists like you pointed out, because if this is not needed anymore, it's relatively simple to set `legendrank` and have the columns in the legend not move around like this.

kpuriIpac:
@robyww @jaladh-singhal 

I made an update to have the table's enum-like columns also be sorted according to the given order (if one exists). For testing, use both of the same files above. Then in table, notice for `band` and `order` columns (in each file respectively), look at the filter dropdown, these should be sorted according to the same order as seen in the charts trace select dropdown. 

Example of what to expect: 

Screenshot 2026-09-01 at 4 22 33 PM

@robyww this is what we were going for, right? Let me know, else I can undo the commit.

src/firefly/js/ui/tap/DataServicesOptions.js
```diff
@@ -37,6 +37,33 @@ export function getDataServiceOptionByTable(key, tableOrId, defVal=undefined) {
...
+ * @param {String[]} order - the desired order of those values
+ * @return {String[]}
+ */
+export function orderByConfig(values, order) {
+    const rank = (v) => {
+        const i = order.indexOf(v);
+        return i === -1 ? Number.MAX_SAFE_INTEGER : i;
+    };
+    return [...values].sort((a, b) => rank(a) - rank(b));
+}
```
robyww: put this in `WebUtil.js`. It probably should something like `sortUsing`  or `sortBy` ?

robyww (approved):
PR looks good, Just move that one function to WebUtil.js

PR #2012, opened 2026-09-11
Firefly-2101: Bug Fix for Expanded Spectrum Plot

aventura121:
[Firefly-2101](https://jira.ipac.caltech.edu/browse/FIREFLY-2101): "DCE: no way to expand spectrum plot"

**Bug description**

- Expanding a chart lost the reference to the viewer from which it was expanded. RightSide owned the expanded slot and rendered either the default active chart viewer or the pinned one. A chart expanded from another viewer, e.g. the spectrum in DCE's data-product pane, fell through to the default and showed the active chart.

**Changes**

- The expanded chart is now built independently. makeExpandedChart() picks the panel matching expandedViewerId, and renders any other viewer directly by id.
- Remove RightSide's expanded-mode branch and expanded prop.
- Drop chartExpandedMode from tab call sites (always false).

**Builds**

Irsa-viewer: https://firefly-2101-dce-expand-spectrum-plot.irsakubedev.ipac.caltech.edu/irsaviewer

Firefly: https://firefly-2101-dce-expand-spectrum-plot.irsakubedev.ipac.caltech.edu/firefly

**Testing**

Using the irsa-viewer build, we can reproduce following the ticket's steps:

1. Go to the DCE application (Data Collections).
2. Select the SASS program in the Choose Data Collection table.
3. Search using 20h11m52.82s +36d11m52.1s Equ J2000, e.g.
4. In the upper right of the data product window (the top-left panel), click on the double-headed arrow in the top right corner.
5. We expect the spectrum graph to be expanded (i.e., not the "Active Chart" from the right-hand panel as in the error scenario).
6. Closing it should work and return to the three panel layout.
7. The other panels also expand and close as expected.


lrebull:
i confirm the original ticket is fixed!
was FIREFLY-2102 also supposed to be fixed here, or no?

aventura121:
> i confirm the original ticket is fixed! was FIREFLY-2102 also supposed to be fixed here, or no?

No, Trey is currently tasked with FIREFLY-2102 separately.

robyww (approved):
Looks Good!

PR #2013, opened 2026-09-11
Firefly-2099: Harden and improve Firefly standalone install

robyww:
### [Firefly-2099](https://jira.ipac.caltech.edu/browse/FIREFLY-2099): Harden and improve Firefly standalone install

  - Add to `build_publish` an option make only standalone.zip 
  - `get_firefly` now has a development mode to pull from a branch build
  - More checking after each download
  - Java install less fragile
  - Documentation has better trouble shooting
  - fallback clean up of Redis
  - some bug fixes

#### Testing
- install firefly using `get_firefly` in branch testing mode. This command should be from a bash prompt
```bash
curl -L https://raw.githubusercontent.com/Caltech-IPAC/firefly/refs/heads/FIREFLY-2099-stand-hard/bin/get-firefly | FIREFLY_BRANCH=FIREFLY-2099-stand-hard bash
```
- Try
   - `ff start`
   - use firefly
   - `ff status`
   -  `ff logs -f`
   - `ff stop`
   - `ff status` (again)
   - `ff uninstall`

#### Alternate which to do test install
- for testing install `get_firefly` requires `jq` be installed. If you don't have it (`which jq`) then you will need to us this command which defines the github run id directly
```bash
curl -L https://raw.githubusercontent.com/Caltech-IPAC/firefly/refs/heads/FIREFLY-2099-stand-hard/bin/get-firefly | FIREFLY_BRANCH=FIREFLY-2099-stand-hard FIREFLY_RUN_ID=35272093406 bash
```


src/standalone/java/edu/caltech/ipac/app/FireflyApplication.java
```diff
@@ -193,7 +193,7 @@ public static boolean doAutoUpdateCheck() {
     }
 
     public static boolean isNewVersionAvailable(String currVer, String availableVer) {
-        if (currVer==null) currVer= "0,0.0";
+        if (currVer==null) currVer= "0.0.0";
         if (availableVer==null) availableVer= "0,0.0";
```
loitly: The typo `0,0.0` is fixed in the first instance, but not in the second.

loitly (changes requested):
I tested the instructions and they worked great.  I found a bug, and have a few comments:

**ff status**
- Report on the wrong port
```sh
$ ff start --port 9999
$ ff status
The Firefly server is not running
```
- Additional information would be nice, e.g.
```sh
Firefly server:  running and healthy  (pid 48213)
URL:             http://localhost:10233/firefly/
Ports:           firefly 10233, redis 10234, debug 5005 (suspend=y)
Version:         2026.2.1
Work dir:        ~/.firefly/server/workarea
Logs dir:        ~/.firefly/server/logs 
```
- Or, keep status simple but add an `info` command to show full dump

**Install**
There a long pause(> 30s) after I enter the suggested command.  Why?
```
curl -L https://raw.githubusercontent......
```
- It allow me to install on top of an existing install_dir.  Does it cleanly write over it?  Will it properly update a later version of JRE, for example?

Please fix typo and `ff status` bug.

robyww:
I had fixed the status bug. Maybe I forgot to rebuild. I will do it again.

robyww:
I can add more status as well

robyww:
@loitly I did your suggested updates on the the status and the about.  Try it again, the build is ready.

loitly:
I also mentioned what appeared to be 'stuck' when running the suggested install command.  I traced it down to 
```sh
+ curl -fsSL https://nightly.link/Caltech-IPAC/firefly/actions/runs/35140976592/standalone-zip.zip -o /var/folders/c0/1t8k80r13b3dw3n6y7l966000000gp/T/firefly-build-artifact.XXXXXX.CymFcfR5fL
```
I think it should use some progress printouts so that it does not appear to be stalled.

Maybe add `-s` to the command so that it does not print the status since it's not very useful.
As is, I am stuck here for 30+ seconds.

```sh
curl -L https://raw.githubusercontent.com/Caltech-IPAC/firefly/refs/heads/FIREFLY-2099-stand-hard/bin/get-firefly | FIREFLY_BRANCH=FIREFLY-2099-stand-hard bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3495  100  3495    0     0   257k      0 --:--:-- --:--:-- --:--:--  262k

```

robyww:
I did not do that because this is only for the testing branch. I can in the future but the end-user will not see it, but I will take out the `-s`

did it work for you?

loitly:
> did it work for you?

Yes, the update status output looks great.  However, it's still not picking up the right pid when using a different port.  But, you said it should be fixed already.

```sh
$ ff start --port 9999
Firefly server starting in background (it takes a few seconds)...

---------------------------------
Firefly URL: http://localhost:9999/firefly/
---------------------------------
Firefly server waiting for init to complete...Ready
to stop server: firefly/bin/ff stop
$ ff status
The Firefly server is not running
```

robyww:
i missed what you were doing (--port 9999). There had been another bug I fixed before it was out for review. I thought is was a regression issue. 

I don't think I actually save the port when it is overridden like that. Let me look at that.

robyww:
@loitly - ok, I think I have it fixed now.  If the first install line does not work the use the second. The first only works if `buld_publish` was the last one done. With the PR work going on it might not be. The second is more specific and should always work.  Remember they are both only for testing. 

in ops we pull from from the release asset.

loitly (approved):
Looks good.  Thanks for addressing my concerns.

PR #2014, opened 2026-09-11
FIREFLY-2077: Refactor cleanup.sh to improve configuration and maintainability

loitly:
Ticket:  https://jira.ipac.caltech.edu/browse/FIREFLY-2077
Additional changes here: https://github.com/IPAC-SW/irsa-ife/pull/478

- Fix entrypoint.py to prevent duplicate props added to the command line

Many changes to the cleanup process.  Best way to understand it is to read the [script's docs](https://github.com/Caltech-IPAC/firefly/blob/0c48d5e1f2fc07e274af2f42848cb70ecca735a5/docker/cleanup.sh#L3).

Test: https://firefly-2077-cleanup-script.irsakubedev.ipac.caltech.edu/firefly/

The best way to test this is to remote shell into the pod and observe how cleanup.sh behaves.
Cleanup logs are under `/firefly/logs/cleanup/firefly/`
```
kubectl -n firefly-dev exec -ti firefly-2077-cleanup-script-6cd4d6787d-mqrps -- bash
```

Built with envVars:
```bash
CLEANUP_INTERVAL=1m
CLEANUP_AGE_HIPS=30m
CLEANUP_AGE_STAGE=1h

CLEANUP_PCT_FULL=80
CLEANUP_PCT_RECLAIM=10
CLEANUP_SHARED_PCT_FULL=20
```

So, cleanup check every minute:
```
- Removes temp files older than 1 minute.
- Removes HiPS files older than 30 minutes.
- Removes stage files older than 1h.
- Leaves upload files at the default retention period of 7 days.
```
But, because `CLEANUP_SHARED_PCT_FULL=20` is set and the `shared-workarea` is above 20% used,
The script will remove all files under `shared-workarea` including `stage` and `upload`.


robyww:
I will look at this PR on Monday. However, a quick comment. Why not use both age mode and percent full mode together?  I think that is what Frossie wanted.

The age mode runs and the percent full is a fallback if the disk is still getting too full after the age mode run.

loitly:
> The age mode runs and the percent full is a fallback if the disk is still getting too full after the age mode run.

Ah, I misunderstood. Let me make the change, and I’ll let you know when to review it.

robyww (approved):
Looks good.

- Lots of use of functions make it more readable, one you get use to bash function.
- never seen `readonly` before. I did not know you could do that in bash. I probably should explore other place to use it.
- I like the documentation of the script. It will be easy to point others to it.

